PR #2, opened 2025-06-09
Add pointers to install lp_solve

asm582:
github repo pointer to lp_solve and for mac users brew command to install lp_solve


atantawi:
Handled in PR #3

PR #16, opened 2025-07-17
Actuator Emit metrics to promethues

mamy-CS:
- Emit metrics back to Prometheus (via the actuator), enabling external systems like HPA to react to scaling recommendations.
- Expose replica target metrics


mamy-CS:
Moved to https://github.com/llm-d-incubation/inferno-autoscaler/pull/39

PR #24, opened 2025-07-25
Vendor vllme (new-metric branch) into inferno-autoscaler

mamy-CS:
- vllme inside inferno repo for ease of maintainability. 
- Readme is updated in consecutive standing pr (https://github.com/llm-d-incubation/inferno-autoscaler/pull/18)


asm582:
@mamy-CS, the vllme repo has portions that we don't need. I recommend creating a slim image and using loadgen—a script that modifies the deployment, taking RPM as input.

vishakha-ramani (approved):

PR #41, opened 2025-07-30
Support configured maximum batch size

atantawi:
Allow setting the maximum batch size of a variant and not letting the optimizer to set it.


mamy-CS (approved):
/lgtm

PR #117, opened 2025-08-15
hpa integration configuration

mamy-CS:
```
kubectl get hpa vllme-deployment-hpa -n llm-d-sim -o wide --watch
NAME                   REFERENCE                     TARGETS     MINPODS   MAXPODS   REPLICAS   AGE
vllme-deployment-hpa   Deployment/vllme-deployment   3/1 (avg)   1         10        1          46s
vllme-deployment-hpa   Deployment/vllme-deployment   1/1 (avg)   1         10        3          60s
```

Time for HPA to scale deployment starting from inferno optimization = 20-40 secs


README.md
```diff
@@ -455,6 +497,16 @@ For detailed information about the custom metrics, see [Custom Metrics Documenta
...
+## Documentation
+
+### HPA Integration
+- [HPA Integration Quick Start](docs/hpa-integration.md) - Setup HPA with Inferno optimization signals
+
+### Additional Resources
+- [Custom Metrics Documentation](docs/custom-metrics.md) - Understanding Inferno's metrics system
+- [CRD Documentation](docs/crd-docs.md) - VariantAutoscaling custom resource details
+- [Modeling & Optimization](docs/modeling-optimization.md) - Deep dive into optimization algorithms
+
```
asm582, line 509: I think you should remove this section from readme but please keep the design docs in the docs folder.

config/samples/hpa-integration.yaml
```diff
@@ -0,0 +1,93 @@
...
+        periodSeconds: 30
+      - type: Percent
+        value: 50                        # Or 50% of current replicas
+        periodSeconds: 30
+      selectPolicy: Max                  # Use the higher of the two policies
+    scaleDown:
+      stabilizationWindowSeconds: 180    # Slower scale-down for stability
+      policies:
+      - type: Pods
+        value: 1                         # Scale down by 1 pod max
```
asm582, line 46: Same as above, scale down should be a significant value and let Inferno handle it.

internal/actuator/actuator.go
```diff
@@ -25,65 +25,62 @@ func NewActuator(k8sClient client.Client) *Actuator {
...
-			// Don't fail the deployment patch for metric emission errors
-		}
+	// Fallback to spec if status not ready
+	if deploy.Spec.Replicas != nil {
+		return *deploy.Spec.Replicas, nil
 	}
 
-	return nil
+	// Final fallback
+	return 1, nil
```
asm582, line 50: Final fallback should not be 1, it looks like we don't need this function anymore.

internal/actuator/actuator.go
```diff
@@ -25,65 +25,62 @@ func NewActuator(k8sClient client.Client) *Actuator {
...
-			// Metrics are critical for HPA, but emission failures shouldn't break core functionality
+			// Metrics are critical for external autoscalers, but emission failures shouldn't break core functionality
 		} else {
-			logger.Log.Debug("EmitReplicaMetrics completed")
+			logger.Log.Debug("EmitReplicaMetrics completed successfully", "variant", VariantAutoscaling.Name)
 		}
 	} else {
-		logger.Log.Debug("Skipping EmitReplicaMetrics - NumReplicas is 0")
+		logger.Log.Debug("Skipping EmitReplicaMetrics - NumReplicas is 0", "variant", VariantAutoscaling.Name)
 	}
```
asm582, line 81: Is this just for adding a logger message? we should remove this else branch

internal/actuator/actuator.go
asm582: > That fallback is an edge case where deployment exists but has no `spec.replicas` set

spec.replicas defaults to 1 as per the deployment controller design.

asm582:
Thanks for this PR. Closing this draft PR, as it has helped with HPA integration that has been working on KinD and OpenShift.

PR #135, opened 2025-08-22
Changing E2E to check emitted Inferno metrics

WheelyMcBones:
This PR aims to change the E2E tests to check the replica metrics emitted by Inferno, to verify scale up and scale down. The previous check regarding the deployment patch is removed, to be compliant with the introduction of HPA.

It also includes the changes made by @mamy-CS in #117 to have Prometheus fetch metrics from the controller using HTTPS.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -752,6 +752,8 @@ var _ = Describe("Test Inferno-autoscaler with vllme deployment - single VA - cr
 		serviceMonName = "vllme-servicemonitor"
 		appLabel = "vllme"
 		port = 8000
+		// Consider adjusting the load rate based on test environment capabilities
+		// TODO: lowering the load rate to 20 makes checks pass quicker, but causes instabilities in the `constant load` test case
 		loadRate = 25
 
 		By("ensuring unique app label for deployment and service")


internal/actuator/actuator.go
```diff
@@ -78,7 +78,7 @@ func (a *Actuator) EmitMetrics(ctx context.Context, VariantAutoscaling *llmdOptv
 			// Don't fail the reconciliation for metric emission errors
 			// Metrics are critical for HPA, but emission failures shouldn't break core functionality
 		} else {
-			logger.Log.Debug("EmitReplicaMetrics completed")
+			logger.Log.Debug("EmitReplicaMetrics completed for ", "variantAutoscaling-name: ", VariantAutoscaling.Name, ", current-replicas: ", VariantAutoscaling.Status.CurrentAlloc.NumReplicas, ", desired-replicas: ", VariantAutoscaling.Status.DesiredOptimizedAlloc.NumReplicas, ", accelerator: ", VariantAutoscaling.Status.DesiredOptimizedAlloc.Accelerator)
```
asm582, line 81: I agree this code is mainly not part of this PR but still we should not have else branches just to perfrom logging, can we try the below simplified version:

```
if VariantAutoscaling.Status.DesiredOptimizedAlloc.NumReplicas > 0 {
	if err := a.MetricsEmitter.EmitReplicaMetrics(
		ctx,
		VariantAutoscaling,
		int32(VariantAutoscaling.Status.CurrentAlloc.NumReplicas),
		int32(VariantAutoscaling.Status.DesiredOptimizedAlloc.NumReplicas),
		VariantAutoscaling.Status.DesiredOptimizedAlloc.Accelerator,
	); err != nil {
		logger.Log.Error(err, "Failed to emit replica metrics for variantAutoscaling - ",
			"variantAutoscaling-name: ", VariantAutoscaling.Name)
		// Don't fail the reconciliation for metric emission errors
		// Metrics are critical for HPA, but emission failures shouldn't break core functionality
	}
}
```

asm582 (changes requested):
Address review

internal/actuator/actuator.go
WheelyMcBones: Addressed, changed the flow a bit to avoid `else` branches.

asm582 (approved):
lgtm

PR #143, opened 2025-08-27
Computing scaling decision based on ratio metric for HPA

WheelyMcBones:
This PR aims to add the new `inferno_desired_ratio` metric, that the Inferno-autoscaler will emit for each variant, computed as  the optimized desired number of replicas for the same VA divided by the current number of replicas for the VA:

`inferno_desired_ratio` = $\frac{desiredReplicas}{currentReplicas}$ 

This is needed to work with HPA on a `per variant` logic, rather than a `per pod` logic, aiming to possibly target other `scale` resources in the future (e.g. `LeaderWorkerSets`).
It also includes changes to the HPA documentation and samples.

Minor fixes to the `deploy-llm-d-inferno-emulate-on-kind` target were done to be coherent with the other Make targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,8 +104,8 @@ undeploy-inferno-on-kind:
 .PHONY: deploy-llm-d-inferno-emulated-on-kind
 deploy-llm-d-inferno-emulated-on-kind:
 	@echo ">>> Deploying integrated llm-d and Inferno-autoscaler (cluster args: $(KIND_ARGS), image: $(IMG))"
-	export KIND=$(KIND) KUBECTL=$(KUBECTL) IMG=$(IMG) && \
-		hack/deploy-llm-d-inferno-emulated-on-kind.sh $(KIND_ARGS)
+	export KIND=$(KIND) KUBECTL=$(KUBECTL) && \
+		hack/deploy-llm-d-inferno-emulated-on-kind.sh $(KIND_ARGS) -i $(IMG)
 
 ## Deploy Inferno Autoscaler to OpenShift cluster with specified image.
 .PHONY: deploy-inferno-on-openshift
--- a/docs/hpa-integration.md
+++ b/docs/hpa-integration.md
@@ -5,10 +5,10 @@ This guide shows how to integrate Kubernetes HorizontalPodAutoscaler (HPA) with 
 ## Overview
 
 After deploying the Inferno-autoscaler following the provided guides, this guide allows the integration of the following components:
-1. **Inferno Controller** processes VariantAutoscaling objects and emits the `inferno_desired_replicas` metrics
+1. **Inferno Controller** processes VariantAutoscaling objects and emits the `inferno_current_replicas`, the `inferno_desired_replicas` and the `inferno_desired_ratio` metrics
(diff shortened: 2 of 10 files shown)


WheelyMcBones:
The current example configuration setup has been reverted, for simplicity of understanding, to the `AverageValue` setup of HPA consuming the `inferno_desired_replicas` metric.

This was also done to avoid possible timing conflicts between the HPA control loop and the Inferno-Autoscaler reconciliation, due to the HPA poll mechanism.

Makefile
```diff
@@ -104,8 +104,8 @@ undeploy-inferno-on-kind:
 .PHONY: deploy-llm-d-inferno-emulated-on-kind
 deploy-llm-d-inferno-emulated-on-kind:
 	@echo ">>> Deploying integrated llm-d and Inferno-autoscaler (cluster args: $(KIND_ARGS), image: $(IMG))"
-	export KIND=$(KIND) KUBECTL=$(KUBECTL) IMG=$(IMG) && \
-		hack/deploy-llm-d-inferno-emulated-on-kind.sh $(KIND_ARGS)
+	export KIND=$(KIND) KUBECTL=$(KUBECTL) && \
+		hack/deploy-llm-d-inferno-emulated-on-kind.sh $(KIND_ARGS) -i $(IMG)
```
asm582, line 108: why do we have change in this make file?

docs/hpa-integration.md
```diff
@@ -245,9 +265,29 @@ kubectl -n kube-system get pod -l component=kube-controller-manager -o yaml | gr
...
+# config/samples/hpa-integration.yaml
+# ...
+      target:
+        type: Value
+        value: "1"
+```
+
+This issue could come up if HPA operates with default behavior.
+
+This can be strongly mitigated by adding an appropriate `StabilizationWindow` (e.g. `35s`) in both scaling up and down, to help avoid *flapping* due to timing issues between the Inferno-Autoscaler and the HPA loops.
```
asm582, line 285: remove 35 sec value

internal/metrics/metrics.go
```diff
@@ -96,11 +107,19 @@ func (m *MetricsEmitter) EmitReplicaMetrics(ctx context.Context, va *llmdOptv1al
 	}
 
 	// These operations are local and should never fail, but we handle errors for debugging
-	if currentReplicas == nil || desiredReplicas == nil {
+	if currentReplicas == nil || desiredReplicas == nil || desiredRatio == nil {
```
asm582, line 110: Add a comment on the reason behind this check

Makefile
WheelyMcBones: In the previous state, we were not using the `IMG` env variable correctly and coherently with other Make targets, e.g.:
`make my-target IMG=...`

With this change, we can

asm582: ok, i would have expected it as a different PR. It 's fine for now, but let's try to keep PRs to solve just one issue

docs/hpa-integration.md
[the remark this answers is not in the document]
asm582: community doc is a public doc

WheelyMcBones: Moved along with the rest to the community doc

PR #164, opened 2025-09-12
added README for KEDA integration with WVA on Kind

WheelyMcBones:
This PR aims to show the integration of the WVA with the **Kubernetes Event-Driven Autoscaler** (**KEDA**), used to apply the scaling decisions emitted via Prometheus metrics by the WVA. 
It will support scale to zero, thanks to the native support implemented by KEDA.
The guide includes the integration steps on **Kind**, as well as a basic configuration example to:

- install KEDA on the Kind cluster.
- deploy a `ScaledObject` resource tied to a specific Deployment to be scaled, querying for Prometheus metrics emitted by the WVA.

Internally, KEDA deploys an associated HPA that will scale the Deployment according to the decision computed by the WVA Optimizer.


asm582 (approved):
lgtm

PR #177, opened 2025-09-18
Adding unit test coverage for the Actuator and the Optimizer interface

WheelyMcBones:
This PR increments the test coverage, adding unit tests for the `Actuator` as well as the code interfacing with the `Optimizer`.


internal/actuator/suite_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	llmdv1alpha1 "github.com/llm-d-incubation/inferno-autoscaler/api/v1alpha1"
+	"github.com/llm-d-incubation/inferno-autoscaler/internal/logger"
+	"github.com/llm-d-incubation/inferno-autoscaler/internal/metrics"
+	// +kubebuilder:scaffold:imports
```
asm582, line 39: why do we need this marker?

asm582 (changes requested):
Can you please tell what is pending on this PR?

internal/actuator/suite_test.go
WheelyMcBones: It is not necessary and can be removed, it is just a default marker on every `suite_test.go` that we had.

PR #182, opened 2025-09-24
Update ci-release.yaml to do multi arch

clubanderson:
updating the ci-release.yaml to create multi-arch builds.


asm582:
@clubanderson, can we please add a description to the PR?

PR #183, opened 2025-09-24
Update README.md to include ref to new home for container image

clubanderson:
this PR updates the README to point to the new home of the container image for wva - ghcr.io


clubanderson:
done

PR #189, opened 2025-09-26
Setting minNumReplicas to 1 by default

WheelyMcBones:
This PR sets the minimum number of replicas to 1 (disabling scale to zero) by default.

### DIFF
--- a/internal/controller/variantautoscaling_controller.go
+++ b/internal/controller/variantautoscaling_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


internal/controller/variantautoscaling_controller.go
```diff
@@ -99,7 +99,7 @@ func (r *VariantAutoscalingReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 	}
 
-	if os.Getenv("WVA_SCALE_TO_ZERO") != "false" {
+	if os.Getenv("WVA_SCALE_TO_ZERO") != "false" && os.Getenv("WVA_SCALE_TO_ZERO") != "" {
 		logger.Log.Info("Scaling to zero is enabled!")
 	}
```
asm582, line 105: Can we simplify this with something like below:

```
if strings.EqualFold(os.Getenv("WVA_SCALE_TO_ZERO"), "true") {
    logger.Log.Info("Scaling to zero is enabled!")
}
```

WheelyMcBones: Done

eero-t:
Something like this should IMHO be specified in per-model profiles, like e.g. KubeAI does: https://github.com/substratusai/kubeai/blob/main/docs/how-to/configure-autoscaling.md

asm582:
@eero-t Thanks for your suggestion. Can you create a GitHub issue and describe the use case that you have? We have a CR called VA that helps us in scaling a variant.

PR #200, opened 2025-10-07
add helm chart for ocp, supporting files, and instructions to install wva, prom adaptor, and llm-d in simple form

clubanderson:
adding helm chart for autoscaler


clubanderson:
@WheelyMcBones @asm582

mamy-CS:
Looks like llmd-infra is submoduled twice, one in the root folder and again in the `/hack` folder, I will test this on another cluster as well. Thanks

clubanderson:
> Looks like llmd-infra is submoduled twice, one in the root folder and again in the `/hack` folder, I will test this on another cluster as well. Thanks

got rid of both - thanks for pointing it out

PR #220, opened 2025-10-16
Enhancements to OC E2E Testing

Vezio:
# Summary

This PR includes the following:

## Test Enhancement
- Fixes a bug in the `e2e openshift test` where `light loads` would error out even if they were expected to _not scale_ - the resolution was to add the ability to recognize loads that may not actually scale up.
- Enhances the `e2e openshift test` to allow parametrized options for things such as `namespaces`, `load size`, and many others as documented in the respective readme.
- Parametrized options reduce the need for the users to understand the test code base and more so understand and drive workload specific tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,15 @@ KIND_ARGS ?= -t mix -n 3 -g 2   # Default: 3 nodes, 2 GPUs per node, mixed vendo
 KUBECONFIG ?= $(HOME)/.kube/config
 K8S_VERSION ?= v1.32.0
 
+CONTROLLER_NAMESPACE ?= workload-variant-autoscaler-system
+MONITORING_NAMESPACE ?= openshift-user-workload-monitoring
+LLMD_NAMESPACE       ?= llm-d-inference-scheduling
+GATEWAY_NAME         ?= infra-inference-scheduling-inference-gateway-istio
+MODEL_ID             ?= unsloth/Meta-Llama-3.1-8B
+DEPLOYMENT           ?= ms-inference-scheduling-llm-d-modelservice-decode
+REQUEST_RATE         ?= 20
+NUM_PROMPTS          ?= 3000
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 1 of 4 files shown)


test/e2e-openshift/sharegpt_scaleup_test.go
```diff
@@ -33,6 +33,8 @@ import (
 	v1alpha1 "github.com/llm-d-incubation/workload-variant-autoscaler/api/v1alpha1"
 )
 
+var lowLoad = numPrompts <= 2000 && requestRate <= 8
```
WheelyMcBones: This is definitely an improvement, although maybe a bit hard-coded, could it be possible to have `lowLoad` even with e.g. 10 prompts and 80 req/s?
I don't really have a _good_ solution, but maybe it could be changed to something similar to:
```go
const (
    lowLoadRateThreshold = 8  // req/s
    promptThreshold = 500  // prompts
    minScalingDuration = 300 
)

var testDuration = float64(numPrompts) / float64(requestRate)

var lowLoad = requestRate < lowLoadRateThreshold ||
              testDuration < minScalingDuration ||
              numPrompts < promptThreshold
```
Not sure about the threshold values though.

Makefile
```diff
@@ -6,6 +6,15 @@ KIND_ARGS ?= -t mix -n 3 -g 2   # Default: 3 nodes, 2 GPUs per node, mixed vendo
 KUBECONFIG ?= $(HOME)/.kube/config
 K8S_VERSION ?= v1.32.0
 
+CONTROLLER_NAMESPACE ?= workload-variant-autoscaler-system
+MONITORING_NAMESPACE ?= openshift-user-workload-monitoring
+LLMD_NAMESPACE       ?= llm-d-inference-scheduling
+GATEWAY_NAME         ?= infra-inference-scheduling-inference-gateway
```
WheelyMcBones: This should change to default as:
```bash
GATEWAY_NAME         ?= infra-inference-scheduling-inference-gateway-istio
```
As we are now using Istio as default gateway provider.

WheelyMcBones (changes requested):
Except for the minor change in `GATEWAY_NAME` default value, /lgtm
Tested with that change and works:
```bash
Ran 7 of 7 Specs in 228.859 seconds
SUCCESS! -- 7 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestE2EOpenShift (228.86s)
PASS
ok      github.com/llm-d-incubation/workload-variant-autoscaler/test/e2e-openshift      229.396s
```
@Vezio Please sign all your commits and address the change and we can merge, thanks a lot.

PR #248, opened 2025-11-04
[DO NOT MERGE] Add Dynamic Parameter Tuning for Model-based Optimization

vishakha-ramani:
Currently, WVA uses static performance model parameters defined in the VariantAutoscaling CRD to derive prefill and decode latencies (TTFT and ITL):

**Decode params** (alpha, beta): Model per-token latency as `ITL = alpha + beta * batchSize`
**Prefill params** (gamma, delta): Model prefill latency as `TTFT = gamma + delta * inputTokens * batchSize`

These latencies drive the autoscaler decisions.

**Problems with static parameters:**
- Suboptimal allocations - Optimizer makes decisions based on inaccurate parameters
- Manual tuning required - Operators must update CRD specs when performance drifts

Solution: 
This PR integrates an Extended Kalman filter-based parameter tuner into the WVA to dynamically adjust queueing model parameters (alpha, beta, gamma, delta) based on observed metrics (TTFT, ITL, arrival rate, average input/output tokens), thereby improving optimization accuracy.

cc @atantawi @WheelyMcBones

### DIFF
--- a/internal/controller/variantautoscaling_controller.go
+++ b/internal/controller/variantautoscaling_controller.go
@@ -374,7 +374,7 @@ func (r *VariantAutoscalingReconciler) Reconcile(ctx context.Context, req ctrl.R
 			}
 
 			if tunerEnabled {
-				logger.Log.Debug("Experimental model tuner is enabled globally (EXPERIMENTAL_MODEL_TUNER_ENABLED=true) tuning model performance parameters for active VAs")
+				logger.Log.Debug("Experimental model tuner is enabled globally: tuning model performance parameters for active VAs")
 
 				// Check if auto-guess initial state is enabled globally
 				autoGuessInitStateEnabled, err := r.isAutoGuessInitialStateEnabled(ctx)
@@ -385,15 +385,6 @@ func (r *VariantAutoscalingReconciler) Reconcile(ctx context.Context, req ctrl.R
 				// Tune queueing model parameters for all servers using the system data and all active VAs
 				if err := tuner.TuneModelPerfParams(updateList.Items, systemData, autoGuessInitStateEnabled); err != nil {
 					logger.Log.Warn(err, "failed to tune system data")
-				}
-			} else {
-				logger.Log.Debug("Model tuner is disabled globally (EXPERIMENTAL_MODEL_TUNER_ENABLED=false), using spec parameters for all VAs")
-				// Populate TunerPerfData with spec parameters for all VAs
-				for i := range updateList.Items {
-					va := &updateList.Items[i]
-					if err := tuner.SetFallbackTunedParamsInVAStatus(va); err != nil {
-						logger.Log.Warnf("Failed to set fallback tuned parameters for variant %s/%s: %v", va.Name, va.Namespace, err)
-					}
 				}
 			}
 
--- a/internal/tuner/tuner.go
+++ b/internal/tuner/tuner.go
@@ -1,6 +1,7 @@
 package tuner
 
 import (
+	"errors"
 	"fmt"
 
 	llmdVariantAutoscalingV1alpha1 "github.com/llm-d-incubation/workload-variant-autoscaler/api/v1alpha1"
@@ -122,10 +123,17 @@ func runTuningWithFallback(
 	// Attempt to tune the server
 	tunedResults, err := tuneServer(va, systemData, server, autoGuessInitialState)
 
-	// Handle failure - complete tuner failure
+	// Handle failure
 	if err != nil {
-		logger.Log.Warnf("Tuner failed completely for variant %s/%s: %v. Using fallback parameters.",
-			va.Name, va.Namespace, err)
+		// Check if failure is due to insufficient metrics for tuning
+		if errors.Is(err, ErrorInsufficientMetrics) {
+			logger.Log.Debugf("Skipping tuning for variant %s/%s due to insufficient load. Using fallback parameters.",
+				va.Name, va.Namespace)
+		} else {
+			// Actual tuner failure
+			logger.Log.Warnf("Tuner failed completely for variant %s/%s: %v. Using fallback parameters.",
+				va.Name, va.Namespace, err)
+		}
 		return setFallbackParamsInSystemData(va, systemData, server, autoGuessInitialState)
 	}
 
@@ -185,18 +193,22 @@ func tuneServer(
 	// Create tuner for this server
 	tuner, err := createTuner(va, systemData, server, autoGuessInitialState)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get/create tuner: %w", err)
+		// Check if failure is due to insufficient metrics for tuning
+		if errors.Is(err, ErrorInsufficientMetrics) {
+			logger.Log.Debugf("Skipping tuning for variant %s/%s due to insufficient load. Using fallback parameters.",
+				va.Name, va.Namespace)
+		} else {
+			// Actual tuner failure
+			logger.Log.Warnf("Tuner failed completely for variant %s/%s: %v. Using fallback parameters.",
+				va.Name, va.Namespace, err)
+		}
+		return nil, err
 	}
 
 	// Convert server's CurrentAlloc to Environment
 	env, err := convertAllocToEnvironment(server.CurrentAlloc)
 	if err != nil {
 		return nil, fmt.Errorf("failed to convert allocation to environment: %w", err)
-	}
-
-	// Validate environment has meaningful data
-	if !env.Valid() {
-		return nil, fmt.Errorf("invalid environment for server %s", server.Name)
 	}
 
 	// Update environment with latest metrics
@@ -294,6 +306,11 @@ func createTuner(
 	env, err := convertAllocToEnvironment(server.CurrentAlloc)
 	if err != nil {
 		return nil, fmt.Errorf("failed to convert allocation to environment: %w", err)
+	}
+
+	if !env.Valid() {
+		// Invalid environment is typically due to no traffic observed
+		return nil, ErrorInsufficientMetrics
 	}
 
 	// create tuner
--- a/internal/tuner/utils.go
+++ b/internal/tuner/utils.go
@@ -1,8 +1,8 @@
 package tuner
 
 import (
+	"errors"
 	"fmt"
-	"maps"
 	"strconv"
 	"time"
 
@@ -14,6 +14,9 @@ import (
 	"gonum.org/v1/gonum/mat"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// ErrorInsufficientMetrics indicates that tuning was skipped due to insufficient metrics (e.g., no traffic).
+var ErrorInsufficientMetrics = errors.New("insufficient metrics for tuning")
 
 // build config data from defaults, init state and slos
 func BuildTunerConfig(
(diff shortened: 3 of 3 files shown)


go.mod
```diff
@@ -1,36 +1,43 @@
 module github.com/llm-d-incubation/workload-variant-autoscaler
 
-go 1.23.0
+go 1.24.0
```
asm582: Use 1.23

internal/controller/variantautoscaling_controller.go
```diff
@@ -451,6 +463,11 @@ func (r *VariantAutoscalingReconciler) SetupWithManager(mgr ctrl.Manager) error
 	}
 	logger.Log.Info("Prometheus client and API wrapper initialized and validated successfully")
 
+	// Create Tuner Manager
+	r.TunerMgr = tuner.NewTunerManager()
```
asm582: tunermanager should only be created when flag is enabled

internal/tuner/manager.go
```diff
@@ -0,0 +1,225 @@
...
+	return nil
+}
+
+// tuneServer tunes parameters for a single server and updates SystemData.
+func (tm *TunerManager) tuneServer(systemData *infernoConfig.SystemData, server *infernoConfig.ServerSpec) error {
+	// Check if we should skip tuning due to transient delay
+	skip, err := tm.skipIfInTransition(server)
+	if err != nil {
+		logger.Log.Info("Failed to check transition state: %w", err)
+	}
```
asm582: what is this transient delay? why is timing important?

pkg/tuner/configurator.go
```diff
@@ -0,0 +1,152 @@
...
+
+// Model configuration data
+type TunerModelData struct {
+	InitState            []float64 `json:"initState"`            // initial state of model parameters
+	PercentChange        []float64 `json:"percentChange"`        // percent change in state
+	BoundedState         bool      `json:"boundedState"`         // are the state values bounded
+	MinState             []float64 `json:"minState"`             // lower bound on state
+	MaxState             []float64 `json:"maxState"`             // upper bound on state
+	ExpectedObservations []float64 `json:"expectedObservations"` // expected values of observations
+}
```
asm582: why are json tags needed, do we perform marshall and unmarshall?

pkg/tuner/configurator.go
```diff
@@ -0,0 +1,152 @@
...
+	// dimensions
+	nX int
+	nZ int
+
+	// matrices
+	X0 *mat.VecDense
+	P  *mat.Dense
+	Q  *mat.Dense
+	R  *mat.Dense
+
```
asm582: can we add more meaningful names to this struct

asm582 (changes requested):
partial review

mamy-CS:
pr is going to be reworked with `status` persistence.

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -92,29 +92,31 @@ type VariantAutoscalingStatus struct {
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 
-	// TunerPerfData specifies the tuned prefill and decode parameters for ttft and itl models
+	// TunerPerfData specifies the tuned prefill and decode parameters of the model used by the queue analyzer.
 	TunerPerfData TunerPerfData `json:"tunerPerfData,omitempty"`
```
asm582: Why is tunerPerfData not optional when VAs are only tuned when opted in?

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -92,29 +92,31 @@ type VariantAutoscalingStatus struct {
...
+	// Model specifies the unique identifier of the model used in tuning.
 	Model string `json:"model,omitempty"`
 
-	// Accelerator is the type of accelerator currently allocated.
+	// Accelerator is the type of accelerator used in tuning.
 	Accelerator string `json:"accelerator,omitempty"`
 
-	// UpdateAt specifies the time last succesfull tuner update was performed
+	// UpdateAt specifies the time last succesfull tuner update was performed.
 	UpdatedAt metav1.Time `json:"updatedAt,omitempty"`
```
asm582: Status also includes a timestamp at the top level. Can we reuse the same field?

go.mod
```diff
@@ -1,15 +1,17 @@
 module github.com/llm-d-incubation/workload-variant-autoscaler
 
-go 1.23.0
+go 1.23.2
 
 godebug default=go1.23
 
-toolchain go1.23.10
+toolchain go1.23.12
```
asm582: Why do we need this toolchain?

pkg/tuner/configurator.go
```diff
@@ -40,10 +38,10 @@ type Configurator struct {
...
 
 	// matrices
-	X0 *mat.VecDense // initial values of state parameters
-	P  *mat.Dense    // covariance matrix of estimation error
-	Q  *mat.Dense    // covariance matrix of noise on state
-	R  *mat.Dense    // covariance matrix of noise on observation
+	X *mat.VecDense // (initial or prior) values of state parameters
+	P *mat.Dense    // covariance matrix of estimation error
+	Q *mat.Dense    // covariance matrix of noise on state
+	R *mat.Dense    // covariance matrix of noise on observation
```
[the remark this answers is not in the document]
asm582: Thanks. If such is the case, I would like to see a weblink added to the Kalman filter formulae that represents our implementation. For instance, I found this Wikipedia link: https://en.wikipedia.org/wiki/Kalman_filter. I would like a specific link to be added to the code base that represents a formula.

internal/tuner/tuner.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// Get previous NIS from VA status if it exists
+	var prevNIS float64
+	if hasTunedResults(va) {
+		prevNISStr := va.Status.TunerPerfData.NIS
+		if prevNISStr != "" {
+			parsedNIS, err := strconv.ParseFloat(prevNISStr, 64)
+			if err == nil {
+				prevNIS = parsedNIS
+			}
```
asm582: flip this if loop. we should handle cases when err != nil only.

internal/tuner/tuner.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		// Complete failure
+		return nil, fmt.Errorf("tuner run failed: %w", err)
+	}
+	// Valid update
+	logger.Log.Info("Tuner validation succeeded",
+		"variant", va.Name,
+		"server", server.Name,
+		"NIS", tunedResults.NIS)
+	return tunedResults, nil
```
asm582: Remove business logic inside err != nil loop.

internal/tuner/tuner.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// build tuner config from initial state and slo
+	configData, err := BuildTunerConfig(state, covMatrix, slos)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build config: %w", err)
+	}
+
+	// get environment
+	env := convertAllocToEnvironment(server.CurrentAlloc)
```
asm582: its worth retuning err from func `convertAllocToEnvironment`

pkg/tuner/utils.go
```diff
@@ -0,0 +1,51 @@
+package tuner
+
+import (
+	"math"
+
+	"gonum.org/v1/gonum/mat"
+)
+
+// infinitesimal value to check float equality
+const DefaultEpsilon = 1e-6
```
asm582: Can this be part of constants/tuner.go?

internal/controller/variantautoscaling_controller.go
```diff
@@ -139,6 +139,13 @@ func (r *VariantAutoscalingReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, err
 	}
 
+	// TODO: Whether we need a global switch EXPERIMENTAL_MODEL_TUNER_ENABLED to enable/disable autotuner for VAs.
+
+	// tune queueing model parameters for all servers using the system data and all active VAs
+	if err := tuner.TuneModelPerfParams(activeVAs, systemData); err != nil {
```
mamy-CS: Are we missing status updates in the controller? Checking if I missed anything:
The controller calls `tuner.TuneModelPerfParams()` which updates the VA status via `updateVAStatusWithTunedParams()` (called in `internal/tuner/tuner.go:56`), but the tuner status is never persisted to the API server is it?

1. Line 136: `prepareVariantAutoscalings()` creates `updateList` by fetching fresh VAs from API server
2. Line 145: `TuneModelPerfParams(activeVAs, ...)` is called with `activeVAs` (slice of values, not the `updateList`)
3. Inside `TuneModelPerfParams`, `updateVAStatusWithTunedParams(&va, ...)` modifies the local copy in `activeVAs`
4. Line 201: `applyOptimizedAllocations(ctx, updateList, ...)` uses `updateList` (not `activeVAs`)
5. `applyOptimizedAllocations` copies `CurrentAlloc`, `DesiredOptimizedAlloc`, and `Conditions` from `updateList`, but never copies `TunerPerfData`

The tuner status updates are made on `activeVAs` but `applyOptimizedAllocations` persists `updateList`, which doesn't have the tuner updates.

Don’t we need to either pass `updateList.Items` to tuning and modify them directly or copy tuner status from `activeVAs` to `updateList` before calling `applyOptimizedAllocations`?

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -87,6 +91,33 @@ type VariantAutoscalingStatus struct {
...
+// TunerPerfData captures data related to the status of the performance (queueing) model tuner for a variant.
+// It is used as a persistent store of the model tuner state, keeping the model tuner stateless.
+type TunerPerfData struct {
+	// Model specifies the unique identifier of the model used in tuning.
+	Model string `json:"model,omitempty"`
+
+	// Accelerator is the type of accelerator used in tuning.
+	Accelerator string `json:"accelerator,omitempty"`
+
+	// UpdateAt specifies the time last succesfull tuner update was performed.
```
mamy-CS:
```suggestion
	// UpdateAt specifies the time last successful tuner update was performed.
```

internal/controller/variantautoscaling_controller.go
```diff
@@ -139,6 +139,13 @@ func (r *VariantAutoscalingReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		logger.Log.Warn(err, "failed to tune system data")
```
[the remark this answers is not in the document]
mamy-CS: Maybe add them in conditions as TunerReady, TunerSucceeded, TunerFailed and TunerSkipped (with reason why it was skipped?)

pkg/tuner/tuner.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+		requestData := &analyzer.RequestSize{
+			AvgInputTokens:  t.env.AvgInputToks,
+			AvgOutputTokens: t.env.AvgOutputToks,
+		}
+
+		qa, err := analyzer.NewQueueAnalyzer(qConfig, requestData)
+		if err != nil {
+			logger.Log.Warn(err, "model tuner observation function: failed to create queue analyzer")
+			return nil
```
mamy-CS: Since `makeObservationFunc()` returns `nil` on error, this could cause panics. Maybe return a zero vector or handle error differently here.

mamy-CS (changes requested):
Address reviews, lets also add controller integration tests that verify status updates are persisted.

internal/tuner/utils_test.go
```diff
@@ -0,0 +1,1177 @@
+package controller
```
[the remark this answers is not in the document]
WheelyMcBones: Changed to `package tuner`

internal/tuner/tuner.go
WheelyMcBones: Addressed

atantawi:
PR #791 implements this on the newer controller design.

PR #255, opened 2025-11-06
updates to helm templates to avoid 404 on prom adapter calls

clubanderson:
I changed some things based on feedback from customers and research.

1) I added the label 
```
openshift.io/user-monitoring: "true"
```
to the servicemonitor and the release namespace (workload-variant-autoscaler-system is default ns in values.yaml)
2) I changed the bearerTokenFile to a bearerTokenSecret in the WVA servicemonitor. When doing a describe of the WVA servicemonitor you would see errors re: the use of a bearerTokenFile. So I switched based on feedback from customer and research.
3) the servicemonitor for WVA is now in the WVA namespace and not the openshift-user-workload-monitoring namespace. This keeps the monitor with the other artifacts (cleaner) and it works with the new bearerTokenSecret that is needed
4) I created a secret that creates a token. The secret is called 'workload-variant-autoscaler-controller-manager-token-manual' and is referenced by the WVA servicemonitor
5) the 'workload-variant-autoscaler-metrics-reader' clusterrole was being created, but it was never bound in previous releases of the chart. It now has a clusterrolebinding. This missing CRB was also preventing proper reading of the metrics endpoint exposed by WVA.

### DIFF
--- a/charts/workload-variant-autoscaler/values.yaml
+++ b/charts/workload-variant-autoscaler/values.yaml
@@ -26,7 +26,7 @@ wva:
     #   YOUR_CA_CERTIFICATE_HERE
     #   -----END CERTIFICATE-----
 llmd:
-  namespace: llm-d-inference-scheduling
+  namespace: llm-d-inference-scheduler
   modelName: ms-inference-scheduling-llm-d-modelservice
   modelID: "unsloth/Meta-Llama-3.1-8B"
 


asm582:
Thanks, did we run some sample workload and see if the WVA works?

desimonemike123:
I pulled this PR and the installation was successful (cleared up the prometheus adapter 404 issues)

asm582:
lgtm

PR #256, opened 2025-11-11
add cafile trust stanza

mamy-CS:
- When **wva.prometheus.caCert** is set on openshift and TLS verification is enabled, Prometheus relies on that caFile to trust WVA’s certificate. Without it the scrape fails, leading to missing metrics. This stanza keeps dev/test behavior unchanged (because the block is gated behind the existing condition) while ensuring production clusters keep working


asm582:
Did we test this on OpenShift?

PR #261, opened 2025-11-14
missing cm in mon ns

clubanderson:
in my last PR I moved one of the 2 svc monitors to the wva ns and left the vllm monitor in the monitoring ns. I also move the prometheus-ca to the wva ns. That left the prom adapter in mon ns without the cm. It is not clear to me whether the cm needs to be in wva and mon ns. But it absolutely needs to be in mon ns!


WheelyMcBones (approved):
/lgtm

PR #266, opened 2025-11-18
add service monitor watch

mamy-CS:
- Added ServiceMonitor watch functionality for the controller, watch detects for ServiceMonitor deletion
- CRITICAL log emitted when the service monitor is deleted
- Kubernetes event emitted
- The implementation is verified on both Kind and OpenShift clusters.
- Also updated a test in e2e that I noticed failed a couple of times before as well in the CI
```
LAST SEEN   TYPE      REASON                  OBJECT                                                                          MESSAGE
13m         Warning   ServiceMonitorDeleted   servicemonitor/workload-variant-autoscaler-controller-manager-metrics-monitor   ServiceMonitor workload-variant-autoscaler-system/workload-variant-autoscaler-controller-manager-metrics-monitor is being deleted. Prometheus will not scrape controller metrics. Actuator will not be able to access optimized replicas metrics. Please recreate the ServiceMonitor.
```

### DIFF
--- a/internal/controller/variantautoscaling_controller.go
+++ b/internal/controller/variantautoscaling_controller.go
@@ -511,10 +511,13 @@ func (r *VariantAutoscalingReconciler) SetupWithManager(mgr ctrl.Manager) error 
 				return true
 			},
 			UpdateFunc: func(e event.UpdateEvent) bool {
+				gvk := e.ObjectNew.GetObjectKind().GroupVersionKind()
+				// Allow Update events for ConfigMap (needed to trigger reconcile on config changes)
+				if gvk.Kind == "ConfigMap" && gvk.Group == "" {
+					return true
+				}
 				// Allow Update events for ServiceMonitor when deletionTimestamp is set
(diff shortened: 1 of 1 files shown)


internal/controller/variantautoscaling_controller.go
```diff
@@ -592,3 +648,45 @@ func (r *VariantAutoscalingReconciler) readOptimizationConfig(ctx context.Contex
...
+	if !ok {
+		return nil
+	}
+
+	name := serviceMonitor.GetName()
+	namespace := serviceMonitor.GetNamespace()
+
+	// Check if ServiceMonitor is being deleted
+	if !serviceMonitor.GetDeletionTimestamp().IsZero() {
+		logger.Log.Errorw("CRITICAL: ServiceMonitor being deleted - Prometheus will not scrape controller metrics",
```
asm582: we are already logging with level error, I dont think we need word `CRITICAL` in logger

asm582 (changes requested):
Address review

PR #304, opened 2025-11-21
turn ci e2e off

mamy-CS:
Turning the e2es off in CI temporarily, until the e2es are updated with new capacity work.


mamy-CS:
> CI failed

should pass now

PR #309, opened 2025-11-21
Discover namespace post install

asm582:
This PR uses the downward API to discover the namespace where the pod is running

```
{"level":"INFO","ts":"2025-11-21T22:49:35.676Z","msg":"Capacity scaling ConfigMap changed, reloading cache"}
{"level":"INFO","ts":"2025-11-21T22:49:35.678Z","msg":"Capacity scaling config cache updated: entries=1, has_default=true"}
```

### DIFF
--- a/internal/controller/variantautoscaling_controller.go
+++ b/internal/controller/variantautoscaling_controller.go
@@ -102,6 +102,13 @@ const (
 	EnvExperimentalHybridOptimization = "EXPERIMENTAL_HYBRID_OPTIMIZATION"
 )
 
+func getNamespace() string {
+	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
+		return ns
+	}
+	return "workload-variant-autoscaler-system"
+}
+
 var (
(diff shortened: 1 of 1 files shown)


mamy-CS (approved):
/lgtm

internal/controller/variantautoscaling_controller.go
```diff
@@ -110,6 +109,7 @@ var (
 		Version: "v1",
 		Kind:    "ServiceMonitor",
 	}
+	configMapNamespace = os.Getenv("POD_NAMESPACE")
```
lionelvillard: if value is empty, default to "workload-variant-autoscaler-system"

asm582: Yeah, thought about this. I wanted to fail. If you need a default, I can add.

mamy-CS: This will help the units to pass as well. If not has to be set in the test file

PR #315, opened 2025-11-24
fix: load capacity config after cache initialization

ev-shindin:
## Summary

  Fixes controller startup errors by deferring capacity scaling config initialization until after the controller-runtime
   cache has started.

  ## Problem

Issue: [314](https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/314) 
The controller was attempting to read the `capacity-scaling-config` ConfigMap before the controller-runtime cache was
  initialized, causing repeated errors during startup:

  {"level":"ERROR","ts":"2025-11-24T21:36:29.809Z","msg":"the cache is not started, can not read objects..."}
  {"level":"WARN","ts":"2025-11-24T21:36:31.413Z","msg":"Failed to load initial capacity scaling config, will use
  defaults..."}

  ## Root Cause

  In `cmd/main.go`, `InitializeCapacityConfigCache()` was called directly before `mgr.Start()`, so the cached Kubernetes
   client couldn't access resources yet. The cache is only ready after the manager starts.

  ## Solution

  Moved the capacity config initialization to a manager runnable that executes after the cache has started:

  - Added `manager.RunnableFunc` that calls `InitializeCapacityConfigCache()` after cache sync
  - Added import for `sigs.k8s.io/controller-runtime/pkg/manager`
  - Initialization now happens at the right time in the controller lifecycle

  ## Changes

  - **cmd/main.go**: Moved config initialization to manager runnable
  - Added proper cache-ready check before ConfigMap access

  ## Testing

  - [x] Code compiles successfully
  - [x] golangci-lint passes with no issues
  - [x] Eliminates "cache is not started" errors during controller startup
  - [x] ConfigMap is loaded successfully after cache initialization

  ## Impact

  - **Low risk**: Only changes initialization timing, no logic changes
  - **Backward compatible**: Still uses default config if ConfigMap loading fails
  - **Production safe**: Proper error handling maintained

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -240,6 +240,9 @@ func main() {
 	// +kubebuilder:scaffold:builder
 
 	// Add runnable to initialize capacity scaling config cache after cache has started
+    // This must be a runnable because the cached client is not ready until after mgr.Start()
+    // begins and the cache syncs. Using a runnable ensures initialization happens at the
+    // correct point in the controller lifecycle.
 	if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 		setupLog.Info("Loading initial capacity scaling configuration (after cache start)")
 		if err := reconciler.InitializeCapacityConfigCache(ctx); err != nil {


mamy-CS (approved):
/lgtm
/approve

asm582 (approved):
This code is a temporary patch, it is using main.go to add a runnable. This code will be simplified and moved to the reconciler.

PR #319, opened 2025-11-25
concurrency pattern

Gregory-Pereira:
This PR should enable multiple deployments of the WVA both in a namespace and across separate namespaces


mamy-CS:
What's the status of this pr @Gregory-Pereira @clubanderson?

Issue #172, opened 2025-09-16
Add capped back-off logic around prometheus query


asm582:
This call https://github.com/llm-d-incubation/inferno-autoscaler/blob/f24dcad7b2dea45ee623e30c7b327ae2668bb181/internal/collector/collector.go#L87 should have capped back-off. 
The Prom client makes a fresh HTTP call each time; transient blips, such as Prom restart, network hiccup, 5xx, and timeouts, will surface as errors.

PR #341, opened 2025-11-27
collector: add capped backoff for Prometheus queries

Phil-OSophy-42:
### Summary
Add exponential backoff (capped at 5 s) to every Prometheus query executed by internal/collector/collector.go

### Problem
Issue: #172


internal/utils/utils.go
```diff
@@ -44,9 +46,18 @@ var (
 		Steps:    5,
 	}
 
+	// Lightweight backoff for individual Prometheus queries (collector, etc.)
+	PrometheusQueryBackoff = wait.Backoff{
+		Duration: 500 * time.Millisecond,
+		Factor:   2.0,
+		Jitter:   0.1,
+		Steps:    4, // 500ms, 1s, 2s, 4s = ~7.5s total
```
mamy-CS: Minor comment here, comment can be updated to reflect actual backoff sequence with 4 steps

WheelyMcBones:
@Phil-OSophy-42 thank you for your contribution!
Tests look good now, once you add a verified signature to all commits we should be good to go, thanks!

PR #355, opened 2025-12-02
Saturation-based end-to-end tests

### LINKED ISSUE #346: Minimal E2Es for saturation based autoscaling
Focus on saturation-based E2Es, scenario to test using llmd-sim:
- For WVA deployment, check the configmap spawn before submitting VAs
- Submit a single VA check if a VA exists in the system for deployment
- For submitted VA check replicas before load (scale-up)
- For submitted VA check replicas (scale-down) after load spike
- For submitted VA, check if the replica count is maintained for several 3 mins
- For the two submitted VAs (same base model but different cost), could you check if the cheapest VA scales up

WheelyMcBones:
This PR aims at adding first minimal E2E tests to verify the behavior of the `saturation-based` mode.
Closes #346.


lionelvillard:
@WheelyMcBones @mamy-CS the vast majority of e2e tests should be environment agnostic. #295 can definitely run on kind.

asm582:
> @WheelyMcBones setting up the environment and e2e tests are two different concerns. All e2e tests, in general and there are exceptions, should assume the env is already setup. So I encourage you to decouple these 2 concerns and reuse what #295 already implement. Copy and pasting code is fine.

Agree, e2e should be independent. Make command calls two things: kind cluster creation with the installation of LLMD components on top of which E2Es are run. 
These e2e runs are peculiar in that they focus on running locally, hence they consume llmd sim, which has a queue metrics but no KVcache metrics. Test cases in this PR are modeled after #295, with an emphasis on using llmd-sim.

Issue #323, opened 2025-11-26
Remove tools/vllm_emulator


asm582:
vllm_emulator was a stopgap solution. All models based on e2es now consume vllm-sim. This folder should be deleted from the repository.

PR #369, opened 2025-12-03
Remove vllm emulator in favor of llmd sim

asm582:
This PR removes vllm emulator, as it is not used in the project and we now consume llm-d sim

solves #323


WheelyMcBones (approved):
/lgtm

PR #370, opened 2025-12-03
Remove unused slo class ref from crd

asm582:
The slo class ref was used in the legacy Inferno autoscaler and was present in the CRD to support the legacy implementation. The changes beyond version 0.0.5 of WVA do not intend to support the legacy Inferno implementation.


asm582:
> @asm582 the field should probably be removed from the Helm chart files and templates too

Thanks, removed.

mamy-CS:
Do we need ConfigMapKeyRef now? Appears unused.

PR #397, opened 2025-12-05
Reconciler refactoring part 1: target resolution condition and engine placeholders

lionelvillard:
Part of https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/379

This is the first step to extract optimizers into their own loop. 

This PR includes:

- Partial refactoring of the `VariantAutoscalingReconciler` to 
   - process one VA per reconciliation loop. (**Previous behavior, ie. global reconciliation, is still preserved in this PR**)
   - to include explicit logic for resolving the target deployment and updating the VA status condition (`TypeTargetResolved`)
- The beginning of an optimization engine management framework with clearer separation between per VA reconciliation and flexible global optimization logic.  
- Added 3 optimization engine placeholders. 
- A new executor package to allow various optimization execution strategies such as polling (periodic optimization) and reactive (not implemented) for scale-from-zero and eventually faster 1-N scaling. 

Remaining tasks (follow-on PRs):
- executor unit tests
- add reactive executor
- move existing engines as-is to `engines/...`

### DIFF
--- a/api/v1alpha1/variantautoscaling_types.go
+++ b/api/v1alpha1/variantautoscaling_types.go
@@ -186,6 +186,8 @@ func init() {
 
 // Condition Types for VariantAutoscaling
 const (
+	// TypeTargetResolved indicates whether the target model variant has been resolved successfully
+	TypeTargetResolved = "TargetResolved"
 	// TypeMetricsAvailable indicates whether vLLM metrics are available from Prometheus
 	TypeMetricsAvailable = "MetricsAvailable"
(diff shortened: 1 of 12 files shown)


mamy-CS:
This is a work in progress in multiple steps but added some questions on the big issue here for some clarification - https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/379#issuecomment-3618436484

asm582 (approved):
lgtm

PR #404, opened 2025-12-05
update capacity references to saturation based

mamy-CS:
per title


asm582 (approved):
This is a doc only PR, let's keep associated issue open and not close with this PR.

lgtm

PR #417, opened 2025-12-08
[Do Not Merge] Parallelize Prometheus queries and add circuit breaker

mamy-CS:
This is part of Phase 1 - P0 implementation here [WVA Architecture Refactoring](https://docs.google.com/document/d/17avCH_ZYE7YtUPVoF3CF0meGs5EjEV_I3tbSIQu_rxg/edit?tab=t.0#heading=h.bsro1gpbn2zt)
- Parallelize the Prometheus queries per VA to improve metric collection time
- Added a circuit breaker pattern for Prometheus to prevent cascade failures and catch failures early

General design here: 
Prometheus (may be down) ->
WVA Controller (circuit breaker protects here if prom is down) ->
Collects metrics, makes decisions ->
Emits metrics to Prometheus ->
Prometheus Adapter reads metrics ->
HPA reads from Adapter ->
HPA scales deployments

lint, unit and e2es pass. 

```
[AfterSuite] PASSED [17.331 seconds]

Ran 11 of 12 Specs in 917.375 seconds
SUCCESS! -- 11 Passed | 0 Failed | 1 Pending | 0 Skipped
--- PASS: TestSaturationE2E (917.38s)
PASS
ok      github.com/llm-d-incubation/workload-variant-autoscaler/test/e2e-saturation-based       918.308s
```

### DIFF
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -291,16 +291,17 @@ func AddMetricsToOptStatusWithCircuitBreaker(ctx context.Context,
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
+		var val float64
+		var err error
 		// Check context cancellation before starting query
 		if ctx.Err() != nil {
-			mu.Lock()
-			results.arrivalErr = fmt.Errorf("context cancelled: %w", ctx.Err())
-			mu.Unlock()
-			return
+			err = fmt.Errorf("context cancelled: %w", ctx.Err())
+		} else {
+			val, err = queryAndExtractMetricWithCircuitBreaker(ctx, promAPI, arrivalQuery, "ArrivalRate", circuitBreaker)
+			val = val * 60 // convert from req/sec to req/min
 		}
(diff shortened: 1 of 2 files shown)


internal/collector/collector.go
```diff
@@ -209,35 +267,142 @@ func AddMetricsToOptStatus(ctx context.Context,
...
+		ttftErr         error
+		itlErr          error
 	}
-	arrivalVal *= 60 // convert from req/sec to req/min
 
-	avgInputTokens, err := queryAndExtractMetric(ctx, promAPI, avgPromptToksQuery, "AvgInputTokens")
-	if err != nil {
-		return llmdVariantAutoscalingV1alpha1.Allocation{}, err
+	results := &queryResults{}
+	var wg sync.WaitGroup
```
ev-shindin: Acquiring mutex twice per goroutine (once for error, once for result) is fine but could be simplified  to single lock per goroutine.

PR #428, opened 2025-12-15
Move prom init and with metrics collector to main

asm582:
This PR enables splitting the current monolith reconciler. To enable splitting, we need to move prom init and metrics collector to main so that other threads can consume them. The PR has commented out the copy-pasted code for the reconciler saturation mechanism in engine.go, preserving the comments and order, and lists the following Go methods that need to be exported for external consumption.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -45,6 +45,7 @@ import (
 	llmdVariantAutoscalingV1alpha1 "github.com/llm-d-incubation/workload-variant-autoscaler/api/v1alpha1"
 	"github.com/llm-d-incubation/workload-variant-autoscaler/internal/collector"
 	"github.com/llm-d-incubation/workload-variant-autoscaler/internal/collector/prometheus"
+	"github.com/llm-d-incubation/workload-variant-autoscaler/internal/config"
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -299,11 +304,101 @@ func main() {
...
+	if err != nil {
+		setupLog.Error("failed to create prometheus client", zap.Error(err))
+		os.Exit(1)
+	}
+
+	promAPI := promv1.NewAPI(promClient)
+
+	// Validate that the API is working by testing a simple query with retry logic
+	if err := utils.ValidatePrometheusAPI(context.Background(), promAPI); err != nil {
+		logger.Log.Errorf("CRITICAL: Failed to connect to Prometheus - Inferno requires Prometheus connectivity for autoscaling decisions: error=%v", err)
```
lionelvillard: what's Inferno?

asm582: Ok, that's what we called WVA in the early days, inferno-autoscaler. The greedy optimizer is still called Inferno. A few log files need to be changed. Thanks, I will change it.

PR #432, opened 2025-12-16
update to support multi-ns deployment

### LINKED ISSUE #429: Installation and Reproducability Feedback
# Summary

This documents user-experience from an installation perspective on an OpenShift cluster with several Nvidia H100s. I will share issues regarding reproducing successful installations as well just general comments, or issues experienced.

## Seemingly Random Inconsistent Installation

There were **many** times where I would find myself having to uninstall and reinstall WVA to finally achieve a working installation. And almost the entire time this was without even having to change the configuration, which was what I found most striking. Here's a run down of ultimately what I found:

- (A) Installing the `WVA` controller within the same namespace as the `model` and `llm-d` stack was most successful but was unreliable in creating the `VA` resource in a timely manner - or just not at all. What I mean by this that there would be times where the `VA` resource that is created for the given model would sit completely unpopulated, containing missing values for `Accelerator`, `Current Replicas`, and `Optimized`. _Sometimes_ these values were filled in after a sufficient amount of time passed, others, never, and on the best case, they would be immediately populated.

- (B) Installing `WVA` controller in a separate namespace from the `model` and `llm-d` stack was the most unsuccessful. Out of _many_ (~10-15) tries I only saw this work twice.  It too struggled from the above illustrated `VA` problem, but worse so. There was no other notable issues,

Given these two cases - (A) is the most successful, but obviously this offers zero flexibility. Since WVA is a cluster wide controller - meant to handle multiple models - having to deploy _all_ of the models in the _very same_ namespace is **not** convenient. In a previous release, (B) was possible, and on a consistent manner.

## Seemingly Random TLS Issues and Strange Logging

During the installation period, I noticed _many_ times that there would be TLS issues in the _controller_ logs of WVA that I couldn't figure out where they were coming from or why they were occurring, for example:

```
{"level":"error","ts":"2025-12-15T16:14:51Z","logger":"controller-runtime.metrics","msg":"Authentication failed","path":"/metrics","error":"[invalid bearer token, Token does not match server's copy, token lookup failed]","errorCauses":[{"error":"[invalid bearer token, Token does not match server's copy, token lookup failed]"}],"stacktrace":"sigs.k8s.io/controller-runtime/pkg/metrics/filters.WithAuthenticationAndAuthorization.func1.1\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/metrics/filters/filters.go:89\nnet/http.HandlerFunc.ServeHTTP\n\t/usr/local/go/src/net/http/server.go:2220\nnet/http.(*ServeMux).ServeHTTP\n\t/usr/local/go/src/net/http/server.go:2747\nnet/http.serverHandler.ServeHTTP\n\t/usr/local/go/src/net/http/server.go:3210\nnet/http.(*conn).serve\n\t/usr/local/go/src/net/http/server.go:2092"}
```

Eventually these would stop cluttering the logs and resolve automatically - and other times they would continue - and other...other times...they would occur at all. It is important to know that this issue is completely disjoint from the issues experienced in the first section - I know this from the installation experience...

The other TLS issue is related to what appears in the logs of the controller as well, but at the beginning:

```
{"level":"WARN","ts":"2025-12-15T21:15:39.381Z","msg":"TLS certificate verification is disabled - this is not recommended for production"}
{"level":"INFO","ts":"2025-12-15T21:15:39.381Z","msg":"Initializing Prometheus client -> address: https://thanos-querier.openshift-monitoring.svc.cluster.local:9091/, tls_enabled: true"}
{"level":"INFO","ts":"2025-12-15T21:15:39.382Z","msg":"CA certificate loaded successfullypath/etc/ssl/certs/prometheus-ca.crt"}
{"level":"INFO","ts":"2025-12-15T21:15:39.382Z","msg":"TLS configuration applied to Prometheus HTTPS transport"}
{"level":"INFO","ts":"2025-12-15T21:15:39.382Z","msg":"Bearer token loaded from filepath/var/run/secrets/kubernetes.io/serviceaccount/token"}
```

Above is a case where I had the below:

```
   tls:
      caCertPath: "/etc/ssl/certs/prometheus-ca.crt"
      insecureSkipVerify: true
```

But still `TLS configuration` was applied? Unless this has nothing to do with WVA...

## Resolution Steps Taken

I didn't take much unique or novel steps, other than merely uninstalling and reinstalling, observing the results and retrying. The kinks eventually worked itself out, after a **while**.

Although - I would be interested in seeing how this could play out in a different, OpenShift, environment. I will try to work with @clubanderson on this offline to experiment with a few things.

clubanderson:
when the models are running in a diff ns than wva is installed in, the svc monitor needs to be in the workload ns, not the wva ns. fixes #429

### DIFF
--- a/charts/workload-variant-autoscaler/templates/vllm-servicemonitor.yaml
+++ b/charts/workload-variant-autoscaler/templates/vllm-servicemonitor.yaml
@@ -3,7 +3,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: vllm-servicemonitor
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.llmd.namespace }}
   labels:
     llm-d.ai/model: {{ .Values.llmd.modelName }}
     openshift.io/user-monitoring: "true"


clubanderson:
@Vezio please have a look

asm582 (approved):
lgtm

lionelvillard:
Do we need to backport this PR?

clubanderson:
I would edit for next release

PR #436, opened 2025-12-17
Fix: update svc monitor to use the ns associated with the model workload

clubanderson:
need to help benchmark team release - back porting this on top of v0.4.1 release


asm582 (approved):
lgtm

PR #437, opened 2025-12-17
Allow WVA to watch a select namespace

asm582:
This PR adds a new controller flag to enable WVA and its cache to monitor a specified namespace. If the API remains the same, users can now run multiple versions of WVA.


clubanderson (approved):
lgtm

PR #443, opened 2025-12-18
add OpenShift e2e CI workflow for self-hosted runner

clubanderson:
- Add ci-e2e-openshift.yaml workflow that runs on self-hosted OpenShift runner
- Remove KUBECONFIG requirement from Makefile and test suite


asm582 (approved):
lgtm

PR #468, opened 2025-12-23
Remove VA API calls from core methods consumed by engine

asm582:
engine.go code remains unchanged, but the underlying method that performs VA API server calls does not read VAs from cache, which is populated by the reconciler.


clubanderson:
this should help with timing issues in the scale up and down. good improvement

clubanderson (approved):

PR #469, opened 2025-12-24
fix for openshift test cleaning up resources outside of the PR's 3 namespaces

clubanderson:
the openshift test was cleaning up resources outside of the PR's 3 namespaces. Each test is clobbering another test. Need this PR merged to main to test the fix to only cleanup the PR's own resources.


asm582 (approved):
lgtm

PR #472, opened 2025-12-24
fix(e2e): only clean up resources in current PR's namespaces

clubanderson:
## Summary
- Fixes race condition where concurrent e2e test runs delete each other's VAs
- Previously: `kubectl delete variantautoscaling -A -l ...` deleted ALL VAs cluster-wide
- Now: Only cleans up resources in the 3 namespaces for the current PR

## Root Cause
When multiple e2e tests ran concurrently, the cleanup step from one run would delete VAs being used by another run, causing test failures with 'VA not found'.

## Test plan
- [ ] Trigger e2e test with `/ok-to-test`
- [ ] Verify cleanup only affects current PR's namespaces in logs


clubanderson:
Duplicate of #469 - changes have been added there

Issue #422, opened 2025-12-11
Roadmap - 2026


lionelvillard:
This is a living document! Roadmap discussions happen in the [llm-d sigs-autoscaling slack channel](https://llm-d.slack.com/archives/C08T899332A).

## Optimization Engines

- Add support for scale from zero (new engine).
- Add support for scale to zero.
- Solidify the model tuner.
- Define reasonable alpha/beta tuning parameters.
- Add support for constraints environment. 
- Add support for P/D.
- Machine-learning-based optimization.
- Multi-node support.

## Benchmarking

- Define benchmarking performance metrics and scenarios.
- Evaluate saturation-based optimizer.

## Performance and UX Enhancement

- Decouple optimization engine loops from reconciler loop.
- Make optimization loop pipeline extensible and configurable.
- Add support for collecting metrics from multiple source.
- Asynchronous (push and pull) and synchronous metrics collection.
- VA CRD changes to support optimization-specific parameters.
- VA CRD changes to support HPA features (min/max replicas, behaviors, etc..).
- Grouping and policy-driven autoscaling.

## Testing

- Validate support for multiple AI base models.

Issue #408, opened 2025-12-06
Pluggable Analyzer Architecture for Optimization Loop


ev-shindin:
## Context: Architecture Refactoring

This document details the **Pluggable Analyzer Architecture** for the optimization system, which is part of **Phase 3** of the architecture refactoring epic.

See [WVA Architecture Refactoring](https://docs.google.com/document/d/17avCH_ZYE7YtUPVoF3CF0meGs5EjEV_I3tbSIQu_rxg/edit?tab=t.0#heading=h.bsro1gpbn2zt) for the complete 4-phase plan:

1. **Phase 1**: Fix immediate issues without architectural changes
2. **Phase 2**: Extract MetricsCollectorLoop into separate goroutine
3. **Phase 3**: Extract OptimizationLoop into separate goroutine ← **THIS DOCUMENT**
4. **Phase 4**: Simplify Reconciler to read from caches

### Prerequisites

Phase 3 requires completion of:
- **Phase 1**: Parallel Prometheus queries, circuit breaker
- **Phase 2**: MetricsCache and MetricsCollectorLoop (provides metrics for analyzers)

### Integration with Phase 2: MetricsCollector

Phase 3 Analyzers consume data from the Phase 2 MetricsCollector via **MetricsCache**:

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                     Data Flow: Collector → Analyzer                         │
├─────────────────────────────────────────────────────────────────────────────┤
│                                                                             │
│  Phase 2: MetricsCollector                Phase 3: Analyzers                │
│  ─────────────────────────                ──────────────────                │
│                                                                             │
│  ┌──────────────────────┐                 ┌──────────────────────┐          │
│  │ PrometheusSource     │                 │ CapacityAnalyzer     │          │
│  │ KubernetesSource     │                 │ ModelBasedAnalyzer   │          │
│  └──────────┬───────────┘                 └──────────┬───────────┘          │
│             │                                        │                      │
│             ▼                                        │                      │
│  ┌──────────────────────┐                            │                      │
│  │    MetricsCache      │◄───────────────────────────┘                      │
│  │    (shared)          │   input.Cache.GetAggregated(...)                  │
│  │                      │   input.Cache.GetTimeSeries(...)                  │
│  │  • TimeSeries        │                                                   │
│  │  • Aggregations      │                                                   │
│  │  • Snapshots         │                                                   │
│  └──────────────────────┘                                                   │
│                                                                             │
[...]

PR #498, opened 2026-01-05
Implement limiting the number of accelerators allocated to VAs based on inventory 

atantawi:
This is related to feature "Limited number of accelerators optimizer" of the scaling engine in the roadmap for v0.5 #422 
It addresses issue #536 

Current implementation capitalizes on the (inferno) optimizer (pkg/solver/optimizer) and its related data structures.

`TODO`: Once the Limiter objectives, policies, and algorithms stabilize, consider implementing the limited capacity allocation logic directly using workload variants.

### DIFF
--- a/internal/engines/limiter/targetlimiter.go
+++ b/internal/engines/limiter/targetlimiter.go
@@ -15,16 +15,12 @@ import (
 const (
 	// DefaultTargetLimiterSaturationPolicy is the default policy used when the limited capacity is saturated
 	DefaultTargetLimiterSaturationPolicy = "PriorityRoundRobin"
-
-	// DefaultTargetLimiterServiceClassName is the default name for the service class used in limited capacity allocation
-	DefaultTargetLimiterServiceClassName = "Default"
 )
 
 // TargetLimiterConfig holds configuration for the TargetLimiter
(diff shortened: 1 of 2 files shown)


atantawi:
> @atantawi where this fails w.r.t. architecture described in #408? Lets discuss this.

The PR is meant for the current implementation with target v0.5, not the future planned architectural change.

atantawi:
We could potentially have multiple types of `Limiters` based on a limiting policy. For now, we will implement two Limiters:

- `SaturationLimiter`: Assign the limited available accelerators to variants according to the calculated saturation scores, with the objective of minimizing the largest score while keeping cost low. This is done using a greedy algorithm.
- `TargetLimiter`: Assign the limited available accelerators to variants according to a target number of replicas in a round-robin fashion.

atantawi:
> @atantawi To be ownest I don't think we need complex implementation This PR provide a lot of dependencies to old inferno code. I don't think we need concepts of SeviceClass, Server, System, etc. I would propose to remove all unnecessary dependencies and implement this on top of current infrastarcture with models and variants as entities.

True. I mentioned this in a TODO. The PR builds on current code, but a more efficient, direct implementation is in plan.

PR #506, opened 2026-01-06
remove contextWithRespectedDeadline

lionelvillard:
`context.WithTimeout`  already respects the parent's deadline. From Go docs:

"The derived context is canceled when the timeout elapses, when the returned cancel function is called, or when the parent context's Done channel is closed, whichever happens first."


asm582 (approved):
/lgtm
/approve

PR #509, opened 2026-01-06
fix: apply controller-instance filter to VA list operations

clubanderson:
## Summary
- Fixes multi-controller isolation bug introduced in PR #495
- The saturation engine's polling loop was listing ALL VAs instead of only those assigned to the controller instance
- Adds label selector filtering to `readyVariantAutoscalings()` when `CONTROLLER_INSTANCE` is configured

## Problem
PR #495 added `VariantAutoscalingPredicate()` to filter watch events, but the saturation engine bypasses the watch:
- It runs on a 30-second polling loop
- Calls `ActiveVariantAutoscaling()` → `readyVariantAutoscalings()` → `client.List()`
- The List operation had no label selector, so it returned ALL VAs in the cluster
- This caused controllers to process VAs belonging to other controller instances

## Fix
When `CONTROLLER_INSTANCE` env var is set, the List operation now uses a `MatchingLabels` selector to filter by `wva.llmd.ai/controller-instance` label, matching the predicate behavior.

## Test plan
- [ ] Verify unit tests pass
- [ ] Verify lint passes
- [ ] E2E test with multiple concurrent PRs to confirm isolation works

### DIFF
--- a/internal/constants/metrics.go
+++ b/internal/constants/metrics.go
@@ -86,3 +86,11 @@ const (
 	LabelAcceleratorType    = "accelerator_type"
 	LabelControllerInstance = "controller_instance"
 )
+
+// Kubernetes Label Keys
+// Label keys used on Kubernetes resources for filtering and identification.
+const (
+	// ControllerInstanceLabelKey is the label key used to associate VAs with specific controller instances.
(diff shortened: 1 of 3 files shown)


clubanderson:
@asm582 missed this in 495

asm582 (approved):

PR #514, opened 2026-01-07
Remove variantCost from status

asm582:
Removed variantCost from status; the field is still part of the spec, which is used to select a cost-effective variant.


asm582:
/hold

Emulated mode tests passed, but OpenShift E2E failed. @clubanderson thoughts on OpenShift E2E failures?

clubanderson:
@asm582 are you rebased after 509 merged? If not, thats why your having failure

clubanderson:
@asm582 I re-ran the test and it passed. I don't know why it failed for you initially. will keep an eye on this behavior in other PRs

lionelvillard (approved):

PR #535, opened 2026-01-07
feat: Add collector v2 package with per-source query registry

lionelvillard:
Make the collector package more modular. 

Instead of changing the v1 collector package, I moved it to v2 and refactored it there, for better reading. 

The V2 collector is enabled by default. The v1 collector will be removed when this PR is merged.

Key Changes
---

**New files in v2:**

- prometheus_source.go - PrometheusSource with built-in caching and per-instance QueryList
- query_template.go - QueryTemplate definition and registration 
- cache.go - CachedValue with timestamp-based staleness detection (similar to v1)
- pod_va_mapper.go - PodVAMapper for pod-to-VariantAutoscaling resolution
- registry.go - Single point of entry for all registered metric sources.

**Integration with saturation engine:**

- replica_metrics.go - ReplicaMetricsCollector using v2 API
- engine.go - Conditional v2 usage via COLLECTOR_V2=true env var and UseCollectorV2() method

**Missing in v2:**

- No automatic metrics refresh. Not needed for saturation engine. Can we easily added by reusing common executor package. 
-  Configuration. Will do in a separate PR.

Testing
---
- Enable with COLLECTOR_V2=true environment variable
- E2E tests in Makefile use COLLECTOR_V2=1 by default

Code partially written by Claude.


test/utils/e2eutils.go
```diff
@@ -497,7 +497,7 @@ func CreateLoadGeneratorJob(namespace, targetURL, modelName string, rate, maxSec
 
 	// Always use a standard python image and install guidellm at runtime
 	// using python:3.10 and installing cpu-only torch (~200MB) to be lightweight and fast
```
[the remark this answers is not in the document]
lionelvillard: done in other PR.

internal/engines/saturation/engine.go
```diff
@@ -72,9 +78,23 @@ func NewEngine(client client.Client, scheme *runtime.Scheme, recorder record.Eve
 		RetryBackoff: 100 * time.Millisecond,
 	})
 
+	if strings.EqualFold(os.Getenv("COLLECTOR_V2"), "true") {
+
+		promSource := metricsRegistry.Get("prometheus").(*collectorv2.PrometheusSource)
+		engine.ReplicaMetricsCollectorV2 = saturationmetrics.NewReplicaMetricsCollector(promSource, client)
+
+		saturationmetrics.RegisterSaturationQueries(metricsRegistry)
```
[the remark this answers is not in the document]
lionelvillard: this code will change when removing collector v1.

internal/collector/v2/cache.go
```diff
@@ -0,0 +1,127 @@
...
+	// Start background cleanup if interval is set
+	if cleanupInterval > 0 {
+		go c.startCleanup(ctx)
+	}
+
+	return c
+}
+
+// Get retrieves cached metrics by key
+func (c *Cache) Get(key CacheKey) (*CachedValue, bool) {
```
asm582: Should we return a deep copy?

internal/collector/v2/pod_va_mapper.go
```diff
@@ -0,0 +1,98 @@
...
+	return &PodVAMapper{
+		k8sClient: k8sClient,
+	}
+}
+
+// FindVAForPod finds the VariantAutoscaling object for a pod by first finding
+// its deployment and then finding the VA that targets that deployment.
+//
+// Returns the VA name if found, empty string otherwise.
+func (m *PodVAMapper) FindVAForPod(
```
asm582: I thought this was discussed to be part of the reconciler

internal/engines/saturation/metrics/replica_metrics.go
```diff
@@ -0,0 +1,303 @@
...
+				"queueLength", int(value.Value),
+				"metricAge", metricAge.String())
+		}
+	}
+
+	// Build replica metrics from non-stale pod data
+	replicaMetrics := make([]interfaces.ReplicaMetrics, 0, len(podData))
+	collectedAt := now
+
+	for podName, data := range podData {
```
asm582: Is there a scale issue here, O(Podmap x num deployments)?

asm582 (changes requested):
Address review

internal/collector/v2/pod_va_mapper.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		for _, pod := range podList.Items {
+			if pod.Name == podName {
+				return deploymentName
+			}
+		}
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
lionelvillard: will do a follow-up PR

internal/engines/saturation/metrics/replica_metrics.go
lionelvillard: this code is the same as the one that existed before in collector v1. That said I suggest we try to simplify it in a separate PR?

asm582 (approved):
Please add TODOs where needed.

PR #553, opened 2026-01-09
Removed allocation struct used in modelbased scaling

asm582:
allocation struct has ITL, TTFT, and load average fields. This was handled by model-based scaling, which is not required for saturation-based scaling and therefore is not part of the API.


internal/interfaces/allocation.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// Load describes the workload characteristics for the current allocation.
+	Load LoadProfile `json:"load"`
+}
+
+// LoadProfile represents the configuration for workload characteristics,
+// including the rate of incoming requests (ArrivalRate) and the average
+// length of each request (AvgLength). Both fields are specified as strings
+// to allow flexible input formats.
+type LoadProfile struct {
```
lionelvillard: Why do we still need this?

internal/interfaces/allocation.go
```diff
@@ -0,0 +1,37 @@
...
+	Accelerator string `json:"accelerator"`
+
+	// NumReplicas is the number of replicas currently allocated.
+	NumReplicas int `json:"numReplicas"`
+
+	// MaxBatch is the maximum batch size currently allocated.
+	MaxBatch int `json:"maxBatch"`
+
+	// ITLAverage is the average inter token latency for the current allocation.
+	ITLAverage string `json:"itlAverage"`
```
lionelvillard: Is this still needed?

lionelvillard (approved):

PR #569, opened 2026-01-12
Add model-level transition protection to saturation analyzer

ev-shindin:
## Summary

This PR adds model-level transition protection to the saturation analyzer to prevent incorrect scaling decisions when the model's capacity picture is incomplete.

**Problem:**
Previously, the saturation analyzer only protected individual variants that were transitioning (desired ≠ current). Other variants of the same model could still receive scaling decisions, potentially based on incomplete capacity data. This could lead to:
- Cascading scale-ups when pods aren't yet reporting metrics
- Incorrect decisions when one variant is mid-scaling

**Solution:**
Implement model-level transition blocking:
- If ANY variant of a model is transitioning, ALL variants are blocked from receiving new scaling decisions
- Two transition conditions are now checked:
  1. `desired ≠ current` (scaling operation in progress)
  2. `metrics ≠ current` (pods not yet reporting metrics)
- When model is in transition, variants preserve their desired replicas (if set) or maintain current replicas
- Scaling decisions (scale-up/down) only proceed when all variants of the model are stable

This change is compatible with the recently added `PendingReplicas` checking (#549) - when the model is stable, the pending replicas check is still performed for individual variants during scale-up.

## Changes

- `internal/saturation/analyzer.go`:
  - Added model-level transition detection in `CalculateSaturationTargets`
  - Early return when model is transitioning with detailed logging
  - Preserved `PendingReplicas` check for stable models during scale-up

- `internal/saturation/analyzer_test.go`:
  - Added `TestCalculatesaturationTargets_ModelLevelTransitionBlocking`
  - Added `TestCalculatesaturationTargets_MetricsMismatchBlocksScaling`

## Behavior Change

| Scenario | Old Behavior | New Behavior |
|----------|--------------|--------------|
| v1: desired=4, current=2<br>v2: stable | v1 preserved, v2 could scale | Both blocked until v1 completes |
| v1: metrics=2, current=3<br>v2: stable | v2 could scale | Both blocked until v1 has all metrics |

### DIFF
--- a/internal/saturation/analyzer.go
+++ b/internal/saturation/analyzer.go
@@ -297,8 +297,8 @@ func (a *Analyzer) isScaleDownSafe(
 // Step 1: Pure saturation-based target calculation
 // Uses replica count from Saturation metrics (ready replicas) to avoid excessive scale-up.
 // Rules:
-// - If desired ≠ 0 and desired ≠ current: target = desired (preserve previous optimizer decision)
-// - Else if Saturation needs scale-up: cheapest variant gets readyReplicas+1
+// - If ANY variant is transitioning (desired ≠ current OR metrics ≠ current): block all scaling for the model
+// - Else if Saturation needs scale-up: cheapest variant (without pending replicas) gets readyReplicas+1
 // - Else if Saturation allows scale-down: most expensive variant gets readyReplicas-1
 // - Else: target = readyReplicas (replicas with metrics)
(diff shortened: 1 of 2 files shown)


clubanderson:
@ev-shindin how does this complement #PR 495. I isolated the variants from affecting each other using the vaMap and other places that inadvertently scaled same-name variants or same-cluster variants, which caused interference between them

PR #586, opened 2026-01-14
Implement generic podscaraping source 

mamy-CS:
This pr implements a podscraping metrics source focusing on epp pods for scale from zero use case for now. Below is the design summary based on the use case. 
  * One source instance per InferencePool (not per pod)
  * Service discovery: {inferencePoolName}-epp pattern
  * Pod selection: Ready pods only via Service selector
  * Authentication: Bearer token from secret inference-gateway-sa-metrics-reader-secret
  * Metrics endpoint: Port, path, and scheme provided by client (engine) — not hardcoded
  * On-demand scraping: Refresh() called by engine (no background goroutine, no periodic refresh)
  * Concurrent scraping: All pods in parallel (max concurrent configurable)
  * Result aggregation: All pod metrics combined into single result
  * No metric filtering: Return all metrics, client handles filtering
  * Caching: Results cached with TTL to avoid duplicate scrapes within window
  * Error handling: Partial results on pod failures, continue with available pods
  * Implements MetricsSource interface: Compatible with existing collector infrastructure
  * Configurable: Secret name, timeout, TLS, cache TTL, max concurrent
  * Client-provided: Port, path, scheme (passed via RefreshSpec params or config)
  * Not needed: Refresh interval (engine handles periodic refresh)
  * Client decides: Engine selects source and provides endpoint configuration
  * Engine responsibility: Periodic refresh timing, endpoint configuration (port/path/scheme), and source registration
  * There is a fall back for service name from InferencePool name discovery, this should be provided by the scale-from-zero engine directly in PodScrapingSourceConfig. This pattern-based discovery is only used if ServiceName is not explicitly set by the engine. The discovery and fall back can be removed if not needed. 

- PR also updates `colllector/v2` -> `collector/source` and related files for more flexibility 
- unit tests and e2e tests (both emulated kind and openshift related tests consuming actual gpus) are added testing epp pod scraping. Integration tests with scale from zero engine will be added in future phase with scale from zero work. 

-> the scale from zero engine should create and register a PodScrapingSource for each EPP (InferencePool) it needs to monitor.

### DIFF
--- a/docs/developer-guide/pod-scraping-source.md
+++ b/docs/developer-guide/pod-scraping-source.md
@@ -1,0 +1,373 @@
+# PodScrapingSource Usage Guide
+
+## Overview
+
+`PodScrapingSource` is a generic metrics source implementation that scrapes Prometheus-format metrics directly from Kubernetes pods via HTTP requests. It implements the `MetricsSource` interface and can be used as an alternative to Prometheus queries when you need to collect metrics directly from pods.
+
+### Key Features
+
+- **Direct Pod Scraping**: Scrapes metrics from pod IPs via HTTP (no Prometheus required)
+- **Service-Based Discovery**: Automatically discovers pods using Kubernetes Service selectors
+- **Concurrent Scraping**: Scrapes multiple pods concurrently with configurable limits
+- **Caching**: Built-in caching with configurable TTL
+- **Authentication Support**: Optional Bearer token authentication
+- **Prometheus Format**: Parses standard Prometheus text format metrics
+
+## Prerequisites
+
+1. **Kubernetes Cluster Access**: Valid kubeconfig and cluster connectivity
+2. **Service with Pod Selector**: A Kubernetes Service that selects the target pods
+3. **Metrics Endpoint**: Pods must expose a `/metrics` endpoint (or custom path) in Prometheus format
+   - For EPP pods: Port `9090` (default)
+4. **RBAC Permissions**: The client needs:
+   - `get`, `list`, `watch` on `services`
+   - `get`, `list`, `watch` on `pods`
+   - `get` on `secrets` (if using authentication)
+   - For EPP metrics: `get` on non-resource URLs `/metrics` and `/debug/pprof/*` (see [official RBAC guide](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/site-src/guides/metrics-and-observability.md#scrape-metrics--pprof-profiles))
+5. **Authentication Secret** (optional): If metrics endpoint requires authentication
+   - For EPP: Typically `inference-gateway-sa-metrics-reader-secret` (see [official authentication guide](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/site-src/guides/metrics-and-observability.md#scrape-metrics--pprof-profiles))
+
+## Basic Usage
+
+### Import the Package
+
+```go
+import (
+    "context"
+    "time"
+    
+    "sigs.k8s.io/controller-runtime/pkg/client"
+    
+    "github.com/llm-d-incubation/workload-variant-autoscaler/internal/collector/source/pod"
+)
+```
+
+### Create a PodScrapingSource Instance
+
+```go
+ctx := context.Background()
+k8sClient := // your Kubernetes client
+
+config := pod.PodScrapingSourceConfig{
+    // Required: Service identification
+    ServiceName:      "my-service",
+    ServiceNamespace: "my-namespace",
+    
+    // Required: Metrics endpoint
+    MetricsPort: 9090,
+    
+    // Optional: Customize defaults
+    MetricsPath:   "/metrics",  // default: "/metrics"
+    MetricsScheme: "http",      // default: "http"
+}
+
+source, err := pod.NewPodScrapingSource(ctx, k8sClient, config)
+if err != nil {
+    // handle error
+}
+```
+
+### Use Default Configuration
+
+```go
+config := pod.DefaultPodScrapingSourceConfig()
+config.ServiceName = "my-service"
+config.ServiceNamespace = "my-namespace"
+config.MetricsPort = 9090
+
+source, err := pod.NewPodScrapingSource(ctx, k8sClient, config)
+```
+
+## Configuration Reference
+
+### PodScrapingSourceConfig
+
+| Field | Type | Required | Default | Description |
+|-------|------|----------|---------|-------------|
+| `ServiceName` | `string` | Yes | - | Kubernetes Service name that selects target pods |
+| `ServiceNamespace` | `string` | Yes | - | Namespace of the service |
+| `MetricsPort` | `int32` | Yes | - | Port number where pods expose metrics |
+| `MetricsPath` | `string` | No | `"/metrics"` | HTTP path for metrics endpoint |
+| `MetricsScheme` | `string` | No | `"http"` | URL scheme (`"http"` or `"https"`) |
+| `MetricsReaderSecretName` | `string` | No | `""` | Secret name containing Bearer token (optional) |
+| `MetricsReaderSecretKey` | `string` | No | `"token"` | Key in secret containing the token |
+| `BearerToken` | `string` | No | `""` | Direct Bearer token (overrides secret) |
+| `ScrapeTimeout` | `time.Duration` | No | `5s` | HTTP request timeout per pod |
+| `MaxConcurrentScrapes` | `int` | No | `10` | Maximum concurrent pod scrapes |
+| `DefaultTTL` | `time.Duration` | No | `30s` | Cache TTL for scraped metrics |
+
+### Configuration Examples
+
+#### Minimal Configuration (Required Fields Only)
+
+```go
+config := pod.PodScrapingSourceConfig{
+    ServiceName:      "epp-service",
+    ServiceNamespace: "default",
+    MetricsPort:      9090,
+}
+```
+
+#### With Authentication
+
+```go
+config := pod.PodScrapingSourceConfig{
+    ServiceName:             "epp-service",
+    ServiceNamespace:        "default",
+    MetricsPort:             9090,
+    MetricsReaderSecretName: "metrics-reader-secret",
+    MetricsReaderSecretKey:  "token",
+}
+```
+
+#### With Custom Timeouts and Concurrency
+
+```go
+config := pod.PodScrapingSourceConfig{
+    ServiceName:          "epp-service",
+    ServiceNamespace:     "default",
+    MetricsPort:          9090,
+    ScrapeTimeout:        10 * time.Second,
+    MaxConcurrentScrapes: 5,
+    DefaultTTL:           60 * time.Second,
+}
+```
+
+#### With Direct Bearer Token
+
+```go
+config := pod.PodScrapingSourceConfig{
+    ServiceName:      "epp-service",
+    ServiceNamespace: "default",
+    MetricsPort:      9090,
+    BearerToken:      "your-token-here",
+}
+```
+
+## Using PodScrapingSource
+
+### Refresh Metrics
+
+`Refresh()` discovers pods, scrapes metrics from all ready pods, and updates the cache.
+
+```go
+results, err := source.Refresh(ctx, source.RefreshSpec{
+    Queries: []string{"all_metrics"}, // Empty = refresh all registered queries
+    Params:  map[string]string{},     // Query parameters (not used for pod scraping)
+})
+if err != nil {
+    // handle error
+}
+
+// Access results
+result := results["all_metrics"]
+if result != nil {
+    for _, value := range result.Values {
+        fmt.Printf("Metric: %s, Value: %f, Pod: %s\n",
+            value.Labels["__name__"],
+            value.Value,
+            value.Labels["pod"],
+        )
+    }
+}
+```
+
+### Get Cached Metrics
+
+`Get()` retrieves cached metrics without scraping. Returns `nil` if not cached or expired.
+
+```go
+cached := source.Get("all_metrics", nil)
+if cached != nil && !cached.IsExpired() {
+    result := cached.Result
+    // Use cached metrics
+} else {
+    // Cache miss or expired - call Refresh()
+    results, err := source.Refresh(ctx, source.RefreshSpec{})
+    // ...
+}
+```
+
+### Query List
+
+`QueryList()` returns the query registry. PodScrapingSource registers a default `"all_metrics"` query.
+
+```go
+registry := source.QueryList()
+queries := registry.List()
+// queries = ["all_metrics"]
+```
+
+## Understanding Results
+
+### MetricResult Structure
+
+```go
+type MetricResult struct {
+    QueryName   string        // Query name (e.g., "all_metrics")
+    Values      []MetricValue // All metric values
+    CollectedAt time.Time     // When metrics were scraped
+    Error       error         // Error if query failed
+}
+```
+
+### MetricValue Structure
+
+```go
+type MetricValue struct {
+    Value     float64            // Metric value
+    Timestamp time.Time          // When metric was sampled
+    Labels    map[string]string  // Metric labels
+}
+```
+
+### Label Structure
+
+Each `MetricValue` includes labels:
+
+- `__name__`: The metric name (e.g., `"inference_pool_average_queue_size"`)
+- `pod`: The pod name that provided this metric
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -274,7 +274,7 @@ CONTROLLER_TOOLS_VERSION ?= v0.17.2
 ENVTEST_VERSION ?= $(shell go list -m -f "{{ .Version }}" sigs.k8s.io/controller-runtime | awk -F'[v.]' '{printf "release-%d.%d", $$2, $$3}')
 #ENVTEST_K8S_VERSION is the version of Kubernetes to use for setting up ENVTEST binaries (i.e. 1.31)
 ENVTEST_K8S_VERSION ?= $(shell go list -m -f "{{ .Version }}" k8s.io/api | awk -F'[v.]' '{printf "1.%d", $$3}')
-GOLANGCI_LINT_VERSION ?= v1.63.4
```
asm582: Did we change the linter version? Is this in sync with CI?

test/utils/pod_scraping_test_helpers.go
```diff
@@ -0,0 +1,768 @@
...
+
+	// In Kind, pod IPs are not routable from outside the cluster, so scraping will fail.
+	// This is expected behavior. The actual scraping works when the controller runs
+	// inside the cluster. For e2e tests, we verify:
+	// 1. Source can be created
+	// 2. Infrastructure (pods, service, secret) exists
+	// 3. Controller can access pods (verified through controller logs or status)
+	//
+	// The unit tests verify the actual scraping logic with mock HTTP servers.
+	// Here we just verify the source is functional and infrastructure is ready.
```
asm582: Is this comment needed?

test/utils/pod_scraping_test_helpers.go
```diff
@@ -0,0 +1,768 @@
...
+
+	if config.Environment == "kind" {
+		// In Kind, we expect scraping to fail from outside the cluster
+		// This is a known limitation - pod IPs are only accessible from within the cluster
+		// The controller (running inside the cluster) can successfully scrape metrics
```
asm582: do we need this comment?

test/utils/pod_scraping_test_helpers.go
```diff
@@ -0,0 +1,768 @@
...
+	if err == nil && len(controllerPods.Items) > 0 {
+		// Use controller pod to test scraping
+		controllerPod := controllerPods.Items[0]
+
+		// Verify that the controller pod can resolve the EPP pod IP
+		// This confirms network connectivity within the cluster
+		// The metrics endpoint URL would be:
+		// fmt.Sprintf("%s://%s:%d%s", config.MetricsScheme, testPod.Status.PodIP, config.MetricsPort, config.MetricsPath)
+		// But we can't test it directly from outside the cluster
+		// The actual scraping is verified through unit tests and controller logs
```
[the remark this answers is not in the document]
mamy-CS: oh yes, need to consolidate these. Added them as I was working along. Consolidated. Thanks.

test/utils/pod_scraping_test_helpers.go
```diff
@@ -0,0 +1,768 @@
...
+        expected_token = '%s'
+        if not auth_header.startswith('Bearer ') or auth_header[7:] != expected_token:
+            self.send_response(401)
+            self.end_headers()
+            self.wfile.write(b'Unauthorized')
+            return
+        
+        # Serve metrics
+        metrics = base64.b64decode('%s').decode('utf-8')
+        self.send_response(200)
```
asm582: Nit: Do we have a better way to handle this? I would at least add a TODO, as it seems odd to have a whole class file as a command-line argument.

test/utils/pod_scraping_test_helpers.go
```diff
@@ -0,0 +1,768 @@
...
+			// Cache might be empty if Refresh failed, which is expected from outside cluster
+			_ = cached // Just verify Get doesn't panic
+		} else if results != nil {
+			// If scraping succeeded (unlikely from outside cluster), verify results
+			g.Expect(results).To(gom.HaveKey("all_metrics"), "Should have all_metrics result")
+		}
+	} else {
+		// For OpenShift, pod IPs may or may not be accessible from outside the cluster
+		// depending on network configuration. We verify infrastructure is correct
+		// and that scraping can be attempted. Actual scraping from pod IPs works
```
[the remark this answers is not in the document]
mamy-CS: Consolidated

test/utils/pod_scraping_test_helpers.go
mamy-CS: Consolidated

test/e2e-openshift/e2e_suite_test.go
```diff
@@ -45,14 +46,98 @@ var (
 	llmDNamespace       = getEnvString("LLMD_NAMESPACE", "llm-d-inference-scheduler")
 	// Secondary llm-d namespace for Model B (multi-model testing)
 	llmDNamespaceB = getEnvString("LLMD_NAMESPACE_B", "")
-	gatewayName    = getEnvString("GATEWAY_NAME", "infra-inference-scheduling-inference-gateway")
+	gatewayName    = getEnvString("GATEWAY_NAME", "") // Auto-derived from namespace if empty
 	modelID        = getEnvString("MODEL_ID", "unsloth/Meta-Llama-3.1-8B")
-	deployment     = getEnvString("DEPLOYMENT", "ms-inference-scheduling-llm-d-modelservice-decode")
+	deployment     = getEnvString("DEPLOYMENT", "") // Auto-derived from namespace if empty
```
asm582: Can we remove auto derivation logic? If nothing is specified, we should use defaults. I think this should be more stable.

internal/collector/replica_metrics.go
```diff
@@ -42,30 +45,30 @@ const (
...
 // NewReplicaMetricsCollector creates a new replica metrics collector.
-func NewReplicaMetricsCollector(source collector.MetricsSource, k8sClient client.Client) *ReplicaMetricsCollector {
+func NewReplicaMetricsCollector(metricsSource source.MetricsSource, k8sClient client.Client) *ReplicaMetricsCollector {
 	return &ReplicaMetricsCollector{
-		source:             source,
+		source:             metricsSource,
 		k8sClient:          k8sClient,
-		podVAMapper:        collector.NewPodVAMapper(k8sClient),
+		podVAMapper:        source.NewPodVAMapper(k8sClient),
 		stalenessThreshold: DefaultMetricStalenessThreshold,
```
asm582: I know this could be from the previous implementation, but do we need this?

config/rbac/role.yaml
```diff
@@ -39,6 +39,20 @@ rules:
...
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
```
asm582: Are these part of the VA reconciler via kubebuilder markers?

test/e2e-saturation-based/e2e_saturation_test.go
```diff
@@ -1047,6 +1047,121 @@ var _ = Describe("Test workload-variant-autoscaler - Saturation Mode - Multiple
 
 		_, _ = fmt.Fprintf(GinkgoWriter, "Cleanup completed for multiple VAs Saturation-based E2E tests\n")
 	})
+
+	// PodScrapingSource tests using mock EPP pods for Kind cluster
+	var _ = Describe("PodScrapingSource - Mock EPP Pods", Ordered, func() {
```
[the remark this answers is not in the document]
mamy-CS: we don't, initially used them for unit tests and used it for quick verification in e2es. Currently working on updating them, full e2e and integration tests will be added together with scale from zero work.

config/rbac/role.yaml
mamy-CS: Yes, scale from zero engine/reconciler usage, yet to be implemented. But also added it here for consistency.

test/e2e-openshift/e2e_suite_test.go
mamy-CS: As I mentioned above, I came across an issue. That's why I added it. From past learnings, I am testing using the upstream deployment method, and the current setup is hard-coded. This just derives from the ns the llmd stack is running on. No confusion about it, I vote we keep it.

internal/collector/source/pod/config.go
```diff
@@ -0,0 +1,48 @@
...
+//
+// This file contains configuration types and defaults for PodScrapingSource.
+package pod
+
+import "time"
+
+// PodScrapingSourceConfig contains configuration for pod scraping.
+type PodScrapingSourceConfig struct {
+	// InferencePool identification
+	InferencePoolName      string
```
lionelvillard: Since PodScrapingSource is a generic pod scraper this field is not needed.

internal/collector/source/pod/config.go
```diff
@@ -0,0 +1,48 @@
...
+	InferencePoolNamespace string
+
+	// Service discovery
+	// The scale-from-zero engine should provide ServiceName explicitly.
```
lionelvillard: I would remove this comment as this is a generic component

Makefile
```diff
@@ -274,7 +274,7 @@ CONTROLLER_TOOLS_VERSION ?= v0.17.2
...
+GOLANGCI_LINT_VERSION ?= v2.8.0
```
lionelvillard: this change seems unrelated to the generic podscraping source. It should be part of a different PR.

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
+// Package pod provides the Pod scraping metrics source implementation.
+//
+// This package implements the PodScrapingSource that scrapes metrics directly
+// from EPP pods via HTTP requests to their /metrics endpoints.
```
lionelvillard: EPP is just an example. Please remove the comment.

internal/collector/source/pod/config.go
```diff
@@ -0,0 +1,48 @@
...
+	// If empty, falls back to pattern-based discovery: {poolName}-epp
+	ServiceName string
+
+	// Metrics endpoint (provided by client/engine)
+	MetricsPort   int32  // provided by client
+	MetricsPath   string // provided by client, default: "/metrics"
+	MetricsScheme string // provided by client, default: "http"
+
+	// Authentication
+	MetricsReaderSecretName string // default: "inference-gateway-sa-metrics-reader-secret"
```
lionelvillard: this is not a reasonable default so I would remove it.

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
...
+
+// NewPodScrapingSource creates a new PodScrapingSource for an InferencePool.
+func NewPodScrapingSource(
+	ctx context.Context,
+	k8sClient client.Client,
+	config PodScrapingSourceConfig,
+) (*PodScrapingSource, error) {
+	// Auto-discover service name if not provided
+	if config.ServiceName == "" {
+		config.ServiceName = discoverServiceName(config.InferencePoolName)
```
lionelvillard: Since this source is EPP-agnostic it can't discover the service name. Either the service name is provided or not. It is not discovered.

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
...
+
+	// Set defaults
+	if config.MetricsPath == "" {
+		config.MetricsPath = "/metrics"
+	}
+	if config.MetricsScheme == "" {
+		config.MetricsScheme = "http"
+	}
+	if config.MetricsReaderSecretName == "" {
+		config.MetricsReaderSecretName = "inference-gateway-sa-metrics-reader-secret"
```
lionelvillard: I would remove as it's not a reasonable default (IMO)

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
...
+	cacheKey := source.BuildCacheKey(queryName, params)
+	cached, ok := p.cache.Get(cacheKey)
+	if !ok || cached.IsExpired() {
+		return nil
+	}
+
+	return cached
+}
+
+// discoverPods finds all Ready pods for the EPP service.
```
lionelvillard: remove EPP

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
...
+		Name:      p.config.ServiceName,
+		Namespace: p.config.InferencePoolNamespace,
+	}
+	if err := p.k8sClient.Get(ctx, svcKey, svc); err != nil {
+		return nil, fmt.Errorf("failed to get service %s: %w", svcKey, err)
+	}
+
+	// List pods using Service selector
+	podList := &corev1.PodList{}
+	selector := labels.SelectorFromSet(svc.Spec.Selector)
```
lionelvillard: Service selector may be empty (headless service). In that case, no pods should be selected. I'm not sure this code does the right thing when there is no selector. Can you test it?

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
...
+	if err := p.k8sClient.List(ctx, podList, &client.ListOptions{
+		Namespace:     p.config.InferencePoolNamespace,
+		LabelSelector: selector,
+	}); err != nil {
+		return nil, fmt.Errorf("failed to list pods: %w", err)
+	}
+
+	// Filter to Ready pods only
+	readyPods := []corev1.Pod{}
```
lionelvillard: I would suggest to use a pointer (`[]*corev1.Pod{}`) to avoid excessive shallow copies.

lionelvillard (commented):
thanks @mamy-CS, overall this PR looks good to me. Added some minor comments.

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
...
+	return podSource, nil
+}
+
+// discoverServiceName derives service name from InferencePool name.
+// NOTE: This is a fallback mechanism. The scale-from-zero engine should provide
+// the ServiceName directly in PodScrapingSourceConfig. This pattern-based
+// discovery is only used if ServiceName is not explicitly set by the engine.
+func discoverServiceName(inferencePoolName string) string {
+	return fmt.Sprintf("%s-epp", inferencePoolName)
+}
```
asm582: Why do we need a fallback? If this is required in the config, can we raise an explicit error?

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
...
+		"successCount", len(results),
+		"metricCount", len(aggregated.Values))
+
+	return map[string]*source.MetricResult{
+		"all_metrics": aggregated,
+	}, nil
+}
+
+// Get retrieves cached metrics.
+func (p *PodScrapingSource) Get(queryName string, params map[string]string) *source.CachedValue {
```
asm582: Do we need a lock? How many clients call?

asm582 (changes requested):
Address comments

internal/collector/source/pod/pod_scraping_source.go
mamy-CS: ok, PodScrapingSource is now EPP-agnostic and generic, requiring the caller (for example scale-from-zero engine) to provide ServiceName and ServiceNamespace explicitly

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -0,0 +1,404 @@
...
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == corev1.PodReady {
+			return condition.Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
+
+// scrapeAllPods scrapes metrics from all pods concurrently.
+func (p *PodScrapingSource) scrapeAllPods(ctx context.Context, pods []corev1.Pod) map[string]*source.MetricResult {
```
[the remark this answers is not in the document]
lionelvillard: as @mamy-CS said, this is a generic pod scraper. This code is strictly not needed for EPP but critical in the case of scraping metrics directly from vLLM (there is an issue open with this requirement). 

Also the worker queue pattern is quite common, and used in several places in the codebase. It's probably a good idea to factor it out.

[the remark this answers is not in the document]
asm582: Some comments on the threading and concurrency control sections from the author would also be great to refactor or address later.

PR #587, opened 2026-01-14
Add core limiter interfaces and types

ev-shindin:
## Summary
- Introduce foundational interfaces for resource limiting: Limiter, AllocationAlgorithm, ResourceAllocator, Inventory
- Add core data types: ScalingProposal, ScalingDecision, AllocationResult, ScalingAction
- ScaleTargetRef reuses existing K8s CrossVersionObjectReference for compatibility with VariantAutoscaling

### New Files
- `internal/limiter/interfaces.go` - Core interfaces
- `internal/limiter/types.go` - Data types

### Interfaces

| Interface | Purpose |
|-----------|---------|
| `Limiter` | Combines inventory with allocation algorithm to constrain scaling decisions |
| `AllocationAlgorithm` | Defines how to distribute limited resources across proposals |
| `ResourceAllocator` | Handles granularity-specific resource reservation |
| `Inventory` | Provides resource availability at different levels (cluster, type, node) |

### Data Types

| Type | Purpose |
|------|---------|
| `ScalingProposal` | Proposed scaling action from optimizer |
| `ScalingDecision` | Final scaling decision after limiting |
| `AllocationResult` | Outcome for a single proposal allocation |
| `ScalingAction` | Enum: ScaleUp, ScaleDown, NoChange |

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -99,6 +99,6 @@ jobs:
           ENABLE_SCALE_TO_ZERO: ${{ matrix.scale_to_zero }}
           CLUSTER_GPU_TYPE: ${{ matrix.gpu_type }}
           E2E_GPU_TYPE: ${{ matrix.e2e_gpu_type }}
-          MULTI_MODEL_TESTING: "true"
+          MULTI_MODEL_TESTING: "false"
         run: |
           make test-e2e
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -518,6 +518,16 @@ spec:
           name: grpc-health
         - containerPort: 9090
           name: metrics
+        readinessProbe:
+          grpc:
+            port: 9003
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        livenessProbe:
+          grpc:
+            port: 9003
+          initialDelaySeconds: 15
+          periodSeconds: 20
 ---
 apiVersion: v1
 kind: Service
@@ -537,6 +547,11 @@ spec:
     port: 9090
     targetPort: 9090
 EOF
+
+    # Wait for second EPP to be ready
+    log_info "Waiting for second EPP deployment to be ready..."
+    kubectl wait --for=condition=Available deployment/${POOL_NAME_2}-epp -n $LLMD_NS --timeout=120s || \
+        log_warning "Second EPP deployment not ready yet - check 'kubectl get pods -n $LLMD_NS -l app=${POOL_NAME_2}-epp'"
 
     # Create second modelservice deployment (using llm-d-inference-sim)
     log_info "Creating second modelservice deployment: $MS_NAME_2"
--- a/test/e2e-saturation-based/e2e_limiter_test.go
+++ b/test/e2e-saturation-based/e2e_limiter_test.go
@@ -19,91 +19,52 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
-// Limiter test constants
+// Limiter test constants - matching saturation test patterns
 const (
-	// Test model IDs - different models for each variant to ensure metrics separation
-	// These must match the models configured in install.sh (MODEL_ID and MODEL_ID_2)
-	limiterModel1 = "unsloth/Meta-Llama-3.1-8B"
-	limiterModel2 = "unsloth/Llama-3.2-1B"
-
-	// Gateway name for routing load through InferencePools
-	// This is set up by install.sh and matches the saturation tests
-	limiterGatewayName = "infra-sim-inference-gateway-istio"
-
 	// GPU configurations - each node has 4 GPUs
-	gpusPerReplicaVariant1 = 2 // Variant 1: 2 GPUs/replica, max 2 replicas on 4-GPU node
-	gpusPerReplicaVariant2 = 1 // Variant 2: 1 GPU/replica, max 4 replicas on 4-GPU node
+	gpusPerReplicaLimiter = 2 // 2 GPUs/replica, max 2 replicas on 4-GPU node
 
 	// Min GPUs per specific GPU type (one node with 4 GPUs)
 	minRequiredGPUsPerType = 4
-
-	// Load generation parameters - must match saturation tests to ensure enough load
-	limiterLoadRate     = 8   // requests per second (same as saturation tests)
-	limiterMaxExecTime  = 300 // seconds (5 minutes to ensure saturation is detected)
-	limiterInputTokens  = 128 // same as saturation tests
-	limiterOutputTokens = 128 // same as saturation tests
 )
 
 // getGPUResourceName returns the GPU resource name based on E2E_GPU_TYPE env var.
-// Defaults to "nvidia" if not set.
 func getGPUResourceName() corev1.ResourceName {
 	gpuType := os.Getenv("E2E_GPU_TYPE")
 	if gpuType == "" {
-		gpuType = "nvidia" // default
+		gpuType = "nvidia"
 	}
 	return corev1.ResourceName(gpuType + ".com/gpu")
 }
 
-// getGPUNodeSelectors returns node selectors for the given GPU configuration.
-// For nvidia (nvidia-mix cluster): variant1 targets H100, variant2 targets A100
-// For amd (amd-mix cluster): variant1 targets MI300X, variant2 targets MI250
-func getGPUNodeSelectors() (map[string]string, map[string]string, string, string) {
+// getGPUNodeSelector returns node selector for targeting specific GPU type.
+// For nvidia (nvidia-mix cluster): targets H100 nodes
+// For amd (amd-mix cluster): targets MI300X nodes
+func getGPUNodeSelector() (map[string]string, string) {
 	gpuType := os.Getenv("E2E_GPU_TYPE")
 	if gpuType == "" {
 		gpuType = "nvidia"
 	}
 
-	var variant1Selector, variant2Selector map[string]string
-	var variant1Acc, variant2Acc string
-
 	if gpuType == "nvidia" {
-		// nvidia-mix cluster: H100, A100, MI300X
-		variant1Selector = map[string]string{"gpu-config": "4H100"}
-		variant2Selector = map[string]string{"gpu-config": "4A100"}
-		variant1Acc = "H100"
-		variant2Acc = "A100"
-	} else {
-		// amd-mix cluster: MI300X, MI250, A100
-		variant1Selector = map[string]string{"gpu-config": "4MI300X"}
-		variant2Selector = map[string]string{"gpu-config": "4MI250"}
-		variant1Acc = "MI300X"
-		variant2Acc = "MI250"
+		return map[string]string{"gpu-config": "4H100"}, "H100"
 	}
-
-	return variant1Selector, variant2Selector, variant1Acc, variant2Acc
+	return map[string]string{"gpu-config": "4MI300X"}, "MI300X"
 }
 
 var _ = Describe("Test workload-variant-autoscaler - GPU Limiter Feature", Ordered, func() {
 	var (
 		ctx context.Context
 
-		// Variant 1 resources
-		variant1Name           string
-		variant1DeployName     string
-		variant1ServiceName    string
-		variant1ServiceMonName string
-		variant1HPAName        string
-		variant1AppLabel       string
+		// Resource names - following saturation test pattern
+		name           string
+		deployName     string
+		serviceName    string
+		serviceMonName string
+		hpaName        string
+		appLabel       string
+		namespace      string
 
-		// Variant 2 resources
-		variant2Name           string
-		variant2DeployName     string
-		variant2ServiceName    string
-		variant2ServiceMonName string
-		variant2HPAName        string
-		variant2AppLabel       string
-
-		namespace       string
 		initialReplicas int32
 	)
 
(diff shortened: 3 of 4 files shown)


internal/limiter/interfaces.go
```diff
@@ -0,0 +1,57 @@
...
+	Name() string
+
+	// Refresh updates inventory from the cluster.
+	Refresh(ctx context.Context) error
+
+	// CreateAllocator returns a ResourceAllocator for this inventory.
+	CreateAllocator(ctx context.Context) ResourceAllocator
+
+	// TotalAvailable returns total available GPUs (for metrics/logging).
+	TotalAvailable() int
```
atantawi: - We may need to distinguish between available (meaning existing) and un-allocated resources. An algorithm may need both.
- This design assumes a single type of GPUs. I understand that that is the current use case. However, should we define the interfaces accordingly, or the more general heterogeneous case, given that the current implementation of engines work with heterogeneous resources?

internal/limiter/interfaces.go
```diff
@@ -0,0 +1,57 @@
...
+		allocator ResourceAllocator,
+	) ([]AllocationResult, error)
+}
+
+// ResourceAllocator abstracts resource reservation at different granularities.
+// Created by Inventory to handle node-level vs cluster-level allocation logic.
+type ResourceAllocator interface {
+	// TryAllocate attempts to allocate GPUs for a proposal.
+	// Returns actual GPUs allocated (may be less than requested if constrained).
+	TryAllocate(proposal ScalingProposal, gpusRequested int) (gpusAllocated int, err error)
```
atantawi: - Not clear what is the implication of "`Try`"? Is there a commit after a trial?
- `gpusRequested` could be calculated from the `ScalingProposal`. Do we need a separate argument?

internal/limiter/interfaces.go
```diff
@@ -0,0 +1,57 @@
...
+// Implementations define the granularity (cluster, type, node).
+type Inventory interface {
+	// Name returns inventory identifier.
+	Name() string
+
+	// Refresh updates inventory from the cluster.
+	Refresh(ctx context.Context) error
+
+	// CreateAllocator returns a ResourceAllocator for this inventory.
+	CreateAllocator(ctx context.Context) ResourceAllocator
```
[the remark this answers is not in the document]
ev-shindin: The factory pattern allows Inventory to encapsulate granularity-specific allocation logic. Each allocator is created fresh per `Limit()` call - it holds ephemeral state for that single allocation round (tracking remaining GPUs as `TryAllocate` is called repeatedly). The Inventory itself is read-only after `Refresh()`; the allocator serves as mutable "scratch space" for one allocation pass. This avoids thread-safety issues since each allocation pass gets its own allocator instance - there's no shared mutable state between concurrent operations.

internal/limiter/interfaces.go
ev-shindin: Good point on available vs unallocated distinction. The meaning of Availible is Unallocated. We could rename `TotalAvailable()` to `TotalUnallocated()` or add both methods for clarity.

  For heterogeneity: the interfaces are GPU-type-agnostic at the algorithm level. `ResourceAllocator.TryAllocate` receives a proposal containing `AcceleratorType` - implementations can use this for type-aware allocation. We kept the initial design simple but the interfaces don't preclude heterogeneous implementations. Here is example of implementation:
 ```go
// TypeInventory provides per-accelerator-type GPU availability.
type TypeInventory struct {
    client client.Client
    limits map[string]int // acceleratorType -> max GPUs (0 = unlimited)
    // Cached state
    available map[string]int // acceleratorType -> available GPUs
}
func NewTypeInventory(client client.Client, limits map[string]int) *TypeInventory {
    return &TypeInventory{
        client: client,
        limits: limits,
    }
}
func (t *TypeInventory) Name() string { return "type" }
func (t *TypeInventory) Refresh(ctx context.Context) error {
    var nodes corev1.NodeList
    if err := t.client.List(ctx, &nodes); err != nil {
        return err
    }
    t.available = make(map[string]int)
    for _, node := range nodes.Items {
        // Get GPU type from node labels
        gpuType := node.Labels[gpuTypeLabelKey]
        if gpuType == "" {
            continue
        }
        allocatable := node.Status.Allocatable[gpuResourceName]
        allocated, _ := t.calculateAllocatedOnNode(ctx, node.Name)
        spare := int(allocatable.Value()) - allocated
        t.available[gpuType] += spare
    }
    // Apply limits
    for gpuType, limit := range t.limits {
        if limit > 0 && t.available[gpuType] > limit {
            t.available[gpuType] = limit
        }
    }
    return nil
}
func (t *TypeInventory) TotalAvailable() int {
    total := 0
    for _, avail := range t.available {
        total += avail
    }
    return total
}
func (t *TypeInventory) CreateAllocator(ctx context.Context) ResourceAllocator {
    // Copy available map for allocator
    remaining := make(map[string]int)
    for k, v := range t.available {
        remaining[k] = v
    }
    return &typeAllocator{
        remaining: remaining,
    }
}
// typeAllocator tracks remaining GPUs per accelerator type.
type typeAllocator struct {
    remaining map[string]int // acceleratorType -> remaining GPUs
}
func (a *typeAllocator) TryAllocate(p ScalingProposal, gpusRequested int) (int, error) {
    gpuType := p.AcceleratorType
    avail := a.remaining[gpuType]
    toAllocate := min(gpusRequested, avail)
    toAllocate = (toAllocate / p.GPUsPerReplica) * p.GPUsPerReplica
    if toAllocate > 0 {
        a.remaining[gpuType] -= toAllocate
    }
    return toAllocate, nil
}
func (a *typeAllocator) Remaining() int {
    total := 0
    for _, r := range a.remaining {
        total += r
    }
    return total
}
```

File: internal/limiter/interfaces.go [hunk 4]
```diff
@@ -0,0 +1,57 @@
...
+
+	// Remaining returns total remaining allocatable GPUs.
+	Remaining() int
+}
+
+// Inventory provides resource availability information and creates allocators.
+// Implementations define the granularity (cluster, type, node).
+type Inventory interface {
+	// Name returns inventory identifier.
+	Name() string
```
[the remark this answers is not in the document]
ev-shindin: The primary use case is different granularity **implementations** (ClusterInventory, TypeInventory, NodeInventory) - not multiple concurrent instances. The interface enables swapping implementations via configuration (strategy pattern). At runtime, typically one inventory implementation is active based on the configured limiting granularity.

internal/limiter/types.go
```diff
@@ -0,0 +1,69 @@
...
+	// Priority information (for algorithm ordering)
+	SpareCapacity float64
+	Priority      int
+	Weight        float64
+
+	// Cost information
+	Cost float64
+
+	// Reference to scale target (uses existing K8s type)
+	ScaleTargetRef *autoscalingv1.CrossVersionObjectReference
```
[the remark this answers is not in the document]
ev-shindin: Agreed this adds overhead if engines already have the deployment, but what if we scale object other then deployment? `ScaleTargetRef` was included for generality (supporting other scalable resources).

  Options:
  1. Remove it - assume caller provides deployment reference separately
  2. Make it optional (`*autoscalingv1.CrossVersionObjectReference`)
  3. Keep for flexibility but document it's optional

internal/limiter/types.go
```diff
@@ -0,0 +1,69 @@
...
+	Proposal      ScalingProposal
+	GPUsAllocated int
+	ReplicasAdded int
+	Partial       bool
+	Reason        string
+}
+
+// ScalingDecision represents the final scaling decision after limiting.
+type ScalingDecision struct {
+	ScalingProposal
```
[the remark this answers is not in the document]
ev-shindin: The embedding ensures each decision is self-contained with all proposal data - no external lookups needed. Key-based lookup would require maintaining a proposals map and doing lookups, adding complexity.

  If memory is a concern for large proposal counts, we could switch to:
 ```go
 type ScalingDecision struct {
      ProposalKey    string  // lookup key
      TargetReplicas int32
      GPUsAllocated  int
      // ... limiting metadata
  }
```

  But this adds indirection. Current approach trades some memory for simpler consumption. What's your preference?

**Comment by atantawi** [comment]
Regarding the (4) proposed interfaces, the main ones relevant to this PR are: `Limiter` and `Inventory`.

It seems to me that the other two: `ResourceAllocator` and `AllocationAlgorithm` are implementation-specific. A `Limiter` implementation may not use them necessarily, so why define them here? A user of a`Limiter` cares only about the interface to the `Limiter` and is not concerned with `ResourceAllocator` and `AllocationAlgorithm`. A user needs a factory to create a `Limiter` based on the available choices of `Limiter` implementations.

As for the `Inventory`, there is already one defined and implemented in the collector package. Granted that we need to provide a different/modified implementation. The question is: do we need to define a new one here, or re-use (modify) the existing one? Really, the inventory is an input to the `Limiter` and not a task on the `Limiter` to get the inventory. IMO, the inventory should be collected and passed on to the `Limiter` in the `Limit()` call.

Finally, in the `Limit()` call, there are two new defined types: `ScalingProposal` and `ScalingDecision`. The question is: do we need a new definition here given that there is already a `VariantDecision` in the types package that we could re-use?

In sum, what I am proposing is to simplify the interface to just a `Limiter` interface which uses `Inventory` and `VariantDecision`. Is that possible?

**Comment by ev-shindin** [comment]
> Regarding the (4) proposed interfaces, the main ones relevant to this PR are: `Limiter` and `Inventory`.
> 
> It seems to me that the other two: `ResourceAllocator` and `AllocationAlgorithm` are implementation-specific. A `Limiter` implementation may not use them necessarily, so why define them here? A user of a`Limiter` cares only about the interface to the `Limiter` and is not concerned with `ResourceAllocator` and `AllocationAlgorithm`. A user needs a factory to create a `Limiter` based on the available choices of `Limiter` implementations.

Why `ResourceAllocator` and `AllocationAlgorithm` as core interfaces?
  These enable the key design goal - algorithms work independently of inventory implementation:
  - `AllocationAlgorithm` doesn't know if resources are tracked at cluster, type, or node level
  - `ResourceAllocator` is the bridge created by` Inventory` and used by `Algorithm`
  - You can swap `GreedyBySaturation` → `RoundRobin` without changing `NodeInventory`
  - You can swap `ClusterInventory` → `TypeInventory` without changing any algorithm
 
The interfaces are exposed for extensibility - custom implementations can be added without modifying the package.

> As for the `Inventory`, there is already one defined and implemented in the collector package. Granted that we need to provide a different/modified implementation. The question is: do we need to define a new one here, or re-use (modify) the existing one? Really, the inventory is an input to the `Limiter` and not a task on the `Limiter` to get the inventory. IMO, the inventory should be collected and passed on to the `Limiter` in the `Limit()` call.

Inventory reuse vs new definition
   The limiter's Inventory is intentionally different from collector's:
  - Collector's inventory: metrics collection, storage, staleness tracking
  - Limiter's inventory: resource availability for allocation
  - Different responsibilities, different lifecycles
  - Limiter inventory is passed availability data **but doesn't depend on how it was collected**

> Finally, in the `Limit()` call, there are two new defined types: `ScalingProposal` and `ScalingDecision`. The question is: do we need a new definition here given that there is already a `VariantDecision` in the types package that we could re-use?

` ScalingProposal/ScalingDecision` vs `VariantDecision`
  The existing `VariantDecision` is saturation-specific (has `SaturationBased`, `SafetyOverride`, `SaturationOnly` fields). The limiter types represent clear input/output boundaries:
  - `ScalingProposal` is generic optimizer input - works with any optimizer
  - `ScalingDecision` is limiter output with constraint metadata
  -  Clear input/output boundary enables testing and composition
 
  However, we could simplify by using `VariantDecision` if we make fields optional. This is a reasonable change.

File: internal/limiter/types.go [hunk 7]
```diff
@@ -0,0 +1,90 @@
...
+	NoChange  ScalingAction = "NoChange"
+)
+
+// ScalingProposal represents a proposed scaling action from any optimizer.
+//
+// This type is intentionally generic - it captures WHAT resources are needed,
+// not implementation-specific details like saturation metrics. This allows the
+// limiter to work with any optimizer (saturation-based, model-based, hybrid).
+//
+// Compare to interfaces.VariantDecision which is saturation-analyzer specific
```
lionelvillard: What's the relationship between `interfaces.VariantDecision` and `ScalingProposal`

atantawi:
> > Regarding the (4) proposed interfaces, the main ones relevant to this PR are: `Limiter` and `Inventory`.
> > It seems to me that the other two: `ResourceAllocator` and `AllocationAlgorithm` are implementation-specific. A `Limiter` implementation may not use them necessarily, so why define them here? A user of a`Limiter` cares only about the interface to the `Limiter` and is not concerned with `ResourceAllocator` and `AllocationAlgorithm`. A user needs a factory to create a `Limiter` based on the available choices of `Limiter` implementations.
> 
> Why `ResourceAllocator` and `AllocationAlgorithm` as core interfaces? These enable the key design goal - algorithms work independently of inventory implementation:
> 
>     * `AllocationAlgorithm` doesn't know if resources are tracked at cluster, type, or node level
> 
>     * `ResourceAllocator` is the bridge created by` Inventory` and used by `Algorithm`
> 
>     * You can swap `GreedyBySaturation` → `RoundRobin` without changing `NodeInventory`
> 
>     * You can swap `ClusterInventory` → `TypeInventory` without changing any algorithm
> 
> 
> The interfaces are exposed for extensibility - custom implementations can be added without modifying the package.
> 
> > As for the `Inventory`, there is already one defined and implemented in the collector package. Granted that we need to provide a different/modified implementation. The question is: do we need to define a new one here, or re-use (modify) the existing one? Really, the inventory is an input to the `Limiter` and not a task on the `Limiter` to get the inventory. IMO, the inventory should be collected and passed on to the `Limiter` in the `Limit()` call.
> 
> Inventory reuse vs new definition The limiter's Inventory is intentionally different from collector's:
> 
>     * Collector's inventory: metrics collection, storage, staleness tracking
> 
>     * Limiter's inventory: resource availability for allocation
> 
>     * Different responsibilities, different lifecycles
> 
>     * Limiter inventory is passed availability data **but doesn't depend on how it was collected**
> 
> 
> > Finally, in the `Limit()` call, there are two new defined types: `ScalingProposal` and `ScalingDecision`. The question is: do we need a new definition here given that there is already a `VariantDecision` in the types package that we could re-use?
> 
> ` ScalingProposal/ScalingDecision` vs `VariantDecision` The existing `VariantDecision` is saturation-specific (has `SaturationBased`, `SafetyOverride`, `SaturationOnly` fields). The limiter types represent clear input/output boundaries:
> 
>     * `ScalingProposal` is generic optimizer input - works with any optimizer
> 
>     * `ScalingDecision` is limiter output with constraint metadata
> 
>     * Clear input/output boundary enables testing and composition
> 
> 
> However, we could simplify by using `VariantDecision` if we make fields optional. This is a reasonable change.

OK, with me. Let's see the implementation. We may revisit the interfaces.

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -91,3 +91,115 @@ func (d *K8sWithGpuOperator) Discover(ctx context.Context) (map[string]map[strin
...
+		}
+	}
+	// Also check init containers
+	for _, container := range pod.Spec.InitContainers {
+		for _, vendor := range vendors {
+			resName := corev1.ResourceName(vendor + "/gpu")
+			if qty, ok := container.Resources.Requests[resName]; ok {
+				// Init containers run sequentially, so we take the max, not sum
+				// But for simplicity, we can sum here as they don't run concurrently with main containers
+				total += int(qty.Value())
```
atantawi: `Init` containers execute in series, so they do not exist simultaneously. Thus, it seems to me that a `max` instead of a `plus` would be more appropriate here.
Further, the `max` over `init` containers should be maxed with the sum over regular containers, as a pod is either be in `init` or regular phase, running its regular containers simultaneously.

atantawi (changes requested):

internal/discovery/k8s_with_gpu_operator.go
ev-shindin: Resolved in [470394b](https://github.com/llm-d-incubation/workload-variant-autoscaler/pull/587/commits/470394b4bf936710a3d4dbe9cede61d4d4a86517)

PR #424, opened 2025-12-11
Initial drop of the scale-from-zero implementation

dumb0002:
This PR starts the implementation of the logic to support scale from zero outlined in the design doc [here](https://docs.google.com/document/d/1jEhbz3UcDNJcYRpaIdhtnAPwBXvDP-YwI4yY6V3ZVqc/edit?tab=t.0) 

**This PR is responsible for the following:**

- Create a reconciler to watch for inferencePool objects
- Create a datastore to store relevant metadata for each inferencePool
- Create helper functions for parsing inferecePool objects to extract relevant metadata. 
- Adds the logic to process VA objects, then process the target variant object  (e.g., deployment, statefulSet, replicaSet, etc)
- Extract the pod labels associated the target inferencePool
- Find the target EPP endpoint using information associated with inferencePools from the datastore
- Create Unit Tests for all components 

**Next step includes:**
- Create the EPP source & query Prometheus metrics
- Implement direct actuation: https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/421
- Build E2E test scripts

**Related Issue:** https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/372

PR #588, opened 2026-01-14
🌱 Update golangci-lint to v2.8.0

clubanderson:
## Summary
- Updates golangci-lint from v2.1.6 to v2.8.0
- Fixes false positive "possible nil pointer dereference" warnings in test files
- The older staticcheck version didn't recognize that `t.Fatal()` terminates execution

## Context
PR #424 is blocked by these false positives in `pkg/core/system_test.go` and `pkg/core/allocation_test.go`. The code correctly checks for nil and calls `t.Fatal()` which stops execution, but the older linter version doesn't understand this control flow.

## Test plan
- [ ] CI passes with the updated linter version
- [ ] No new legitimate lint warnings introduced

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -31,7 +31,7 @@ jobs:
       - name: Install golangci-lint
         uses: golangci/golangci-lint-action@v8
         with:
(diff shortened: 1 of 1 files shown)


mamy-CS (approved):
/lgtm

PR #603, opened 2026-01-18
feat: using indexes to optimize Pod-VariantAutoscaling mapping

### LINKED ISSUE #546: Optimize computing mapping between pod and VA
`FindVAForPod` is expensive. One solution is to leverage k8s go client supports  custom indexes supports for fast lookup (see https://buraksekili.github.io/articles/controller-runtime-1/#indexer).

WheelyMcBones:
This PR introduces a module that registers a custom field index to enable for faster look-ups for `VariantAutoscaling`s.
**Note**: to avoid breaking changes, in case of missing `spec.scaleTargetRef.apiVersion`, a default to `apps/v1` is added for Deployments or unsupported Kinds.

Closes #546.

### DIFF
--- a/internal/collector/replica_metrics.go
+++ b/internal/collector/replica_metrics.go
@@ -207,7 +207,7 @@ func (c *ReplicaMetricsCollector) CollectReplicaMetrics(
 				"deployments", getDeploymentNames(deployments))
 			continue
 		}
-		variantKey := utils.GetVariantKey(namespace, vaName)
+		variantKey := utils.GetNamespacedKey(namespace, vaName)
 
 		// Get accelerator name from VariantAutoscaling label
 		acceleratorName := ""
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -194,7 +194,7 @@ func (e *Engine) optimize(ctx context.Context) error {
 	vaMap := make(map[string]*llmdVariantAutoscalingV1alpha1.VariantAutoscaling, len(activeVAs))
 	for i := range activeVAs {
 		va := activeVAs[i] // Copy to local variable to ensure stable pointer
-		vaMap[utils.GetVariantKey(va.Namespace, va.Name)] = &va
+		vaMap[utils.GetNamespacedKey(va.Namespace, va.Name)] = &va
 	}
 
 	// Create map to store current allocations populated during metrics collection
@@ -330,7 +330,7 @@ func (e *Engine) BuildVariantStates(
 
 		// Try to look up in provided map first (optimization)
 		if deployments != nil {
-			deploy, found = deployments[utils.GetDeploymentKey(va.Namespace, va.GetScaleTargetName())]
+			deploy, found = deployments[utils.GetNamespacedKey(va.Namespace, va.GetScaleTargetName())]
 		}
 
 		if !found {
@@ -529,11 +529,11 @@ func (e *Engine) RunSaturationAnalysis(
 		}
 
 		// Populate maps indexed by Deployment namespace/name
-		deploymentKey := utils.GetDeploymentKey(va.Namespace, va.GetScaleTargetName())
+		deploymentKey := utils.GetNamespacedKey(va.Namespace, va.GetScaleTargetName())
 		deployments[deploymentKey] = &deploy
 
 		// Populate maps indexed by VariantAutoscaling namespace/name
-		variantKey := utils.GetVariantKey(va.Namespace, va.Name)
+		variantKey := utils.GetNamespacedKey(va.Namespace, va.Name)
 		variantAutoscalings[variantKey] = va
 		variantCosts[variantKey] = cost
 	}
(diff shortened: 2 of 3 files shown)


clubanderson:
@WheelyMcBones you need to rebase - changes on https://github.com/llm-d-incubation/workload-variant-autoscaler/pull/585 affect you.

Jan 18, 23:08 UTC │ llm-d PR #548 merged, removing istioBench environment from helmfiles
Jan 19, 09:40 UTC │ PR #603 E2E test ran → failed (using old install.sh with BENCHMARK_MODE=true default) │
Jan 20            │ PR #585 merged, fixing the issue by changing BENCHMARK_MODE default to false          │
  
The fix in PR #585:

  # deploy/install.sh
  -BENCHMARK_MODE=${BENCHMARK_MODE:-"true"} # if true, updates to Istio config for benchmark
  +BENCHMARK_MODE=${BENCHMARK_MODE:-"false"} # if true, updates to Istio config for benchmark (istioBench env required in helmfile)

  # .github/workflows/ci-e2e-openshift.yaml
  +          # Disable benchmark mode - istioBench environment not available in llm-d helmfile
  +          BENCHMARK_MODE: "false"

  PR #603's test failed because it ran:
  - After llm-d removed istioBench
  - Before WVA fixed it with PR #585

  Solution: PR #603 should rebase on main (which now includes the PR #585 fix) and re-run the E2E test. The test should pass after rebasing.

clubanderson:
up to you - either way, you need that change

internal/controller/indexers.go
```diff
@@ -0,0 +1,173 @@
...
+	// PodOwnerKey is the index field name for looking up Pods by their controller owner.
+	// The index value is a composite key in the format "Kind/Name" (e.g., "ReplicaSet/my-rs")
+	// to uniquely identify owners and avoid collisions between different controller types.
+	PodOwnerKey = ".metadata.controller.kindName"
+
+	// RSOwnerKey is the index field name for looking up ReplicaSets by their owning Deployment.
+	RSOwnerKey = ".metadata.deploymentOwner"
+
+	// VAScaleTargetKey is the index field name for looking up VAs by their scale target.
+	// The index value is a composite key in the format "Kind/Name" (e.g., "Deployment/my-app")
```
asm582: Two replicasets in different namespaces can have the same key. How do we avoid a collision?

asm582 (changes requested):
Thanks for the PR, please address the reviews.

internal/controller/indexers.go
WheelyMcBones: Lookups are namespaced, so there should be no collisions, right?

internal/controller/indexers.go
```diff
@@ -0,0 +1,173 @@
...
+	// to uniquely identify scale targets and avoid collisions between different resource types.
+	VAScaleTargetKey = ".spec.scaleTargetRef.kindName"
+)
+
+// SetupIndexes registers custom indexes with the manager's cache.
+func SetupIndexes(ctx context.Context, mgr manager.Manager) error {
+	// Index Pods by their controller owner Kind/Name
+	if err := mgr.GetFieldIndexer().IndexField(ctx, &corev1.Pod{}, PodOwnerKey, func(obj client.Object) []string {
+		pod := obj.(*corev1.Pod)
```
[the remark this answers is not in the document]
asm582: We need to add a cache option to prevent seeing all pods in a production cluster across all namespaces.

internal/controller/indexers.go
WheelyMcBones: I could do so if you want, although as suggested by the docs, indexing over namespaces is already provided:

```go
// IndexerFunc knows how to take an object and turn it into a series
// of non-namespaced keys. Namespaced objects are automatically given
// namespaced and non-spaced variants, so keys do not need to include namespace.
type IndexerFunc func(Object) []string

// FieldIndexer knows how to index over a particular "field" such that it
// can later be used by a field selector.
type FieldIndexer interface {
	// IndexField adds an index with the given field name on the given object type
	// by using the given function to extract the value for that field.  If you want
	// compatibility with the Kubernetes API server, only return one key, and only use
	// fields that the API server supports.  Otherwise, you can return multiple keys,
	// and "equality" in the field selector means that at least one key matches the value.
	// The FieldIndexer will automatically take care of indexing over namespace
	// and supporting efficient all-namespace queries.
	IndexField(ctx context.Context, obj Object, field string, extractValue IndexerFunc) error
}
```
I also added a unit test to verify that we current have no conflicts. Let me know if we need to add the ns to the keys anyway.

internal/controller/indexers.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// scaleTargetIndexKey returns the composite index key for a scale target reference.
+func scaleTargetIndexKey(ref autoscalingv1.CrossVersionObjectReference) string {
+	return fmt.Sprintf("%s/%s", ref.Kind, ref.Name)
+}
+
+// ownerIndexKey returns the composite index key for an owner reference.
+func ownerIndexKey(owner *metav1.OwnerReference) string {
+	return fmt.Sprintf("%s/%s", owner.Kind, owner.Name)
```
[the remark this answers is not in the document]
WheelyMcBones: Done, thanks

asm582:
@WheelyMcBones OpenShift E2E's fail ?

PR #610, opened 2026-01-20
deploy: default BENCHMARK_MODE to false

ev-shindin:
## Summary
- Change the default value of `BENCHMARK_MODE` from `"true"` to `"false"` in the install script
- Add clarification that `istioBench` environment is required in helmfile when enabling benchmark mode
- Explicitly set `BENCHMARK_MODE=false` in CI e2e OpenShift workflow

## Background
The `BENCHMARK_MODE` setting switches the Istio gateway configuration to use the `istioBench` environment in the llm-d helmfile. However, this environment is not available in the standard llm-d helmfile, causing deployment failures when benchmark mode is enabled by default.

## Changes
1. **`deploy/install.sh`**: Changed default from `"true"` to `"false"` and updated comment to clarify the `istioBench` requirement
2. **`.github/workflows/ci-e2e-openshift.yaml`**: Added explicit `BENCHMARK_MODE: "false"` to both Model A1 and Model B deployment steps to ensure CI reliability

## Test plan
- [x] Verify install script deploys successfully with default settings
- [x] Verify CI e2e OpenShift workflow runs without benchmark mode errors
- [x] Verify benchmark mode can still be enabled via `BENCHMARK_MODE=true` when `istioBench` environment is available

### DIFF
--- a/test/e2e-openshift/sharegpt_scaleup_test.go
+++ b/test/e2e-openshift/sharegpt_scaleup_test.go
@@ -107,7 +107,7 @@ func getModelsToTest() []modelTestConfig {
 			deployment:        deploymentName,
 			gatewayService:    gatewayServiceName,
 			maxTokens:         400,  // Moderate tokens for ~3s requests, sustains queue during test
-			requestsPerWorker: 1000, // Balanced request count for moderate-length requests
+			requestsPerWorker: 1100, // 10% increase from original to sustain load slightly longer
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


WheelyMcBones:
Since `llm-d` does not support the benchmark config for Istio anymore, the `BENCHMARK_MODE` might as well be deleted completely, also removing the following check:
https://github.com/llm-d-incubation/workload-variant-autoscaler/blob/d0c4cd96244bbdbe863219320146336b18955813/deploy/install.sh#L497-L501

PR #648, opened 2026-01-27
🐛 Fix e2e tests by adding DECODE_REPLICAS override

clubanderson:
## Summary
- Add `DECODE_REPLICAS` environment variable support to `install.sh`
- Set `DECODE_REPLICAS=1` in OpenShift e2e workflow

## Problem
PR [llm-d/llm-d#619](https://github.com/llm-d/llm-d/pull/619) changed `decode.replicas` from 2 to 8, requiring 16 GPUs (8 pods × 2 GPUs each). This breaks e2e tests which don't have enough GPU resources available on the cluster.

## Solution
Add a `DECODE_REPLICAS` environment variable override in `deploy/install.sh`, similar to the existing `VLLM_MAX_NUM_SEQS` pattern. This allows e2e tests to start with 1 replica (2 GPUs) and let the HPA scale as needed based on WVA metrics.

## Test plan
- [ ] Run `/ok-to-test` on PR #644 after this is merged
- [ ] Verify deployment starts with 1 replica (2 GPUs)
- [ ] Verify HPA can scale based on WVA metrics

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -475,6 +475,8 @@ jobs:
           CONTROLLER_INSTANCE: ${{ env.WVA_NAMESPACE }}
           # vLLM max-num-seqs for e2e testing (lower = easier to saturate)
           VLLM_MAX_NUM_SEQS: ${{ env.MAX_NUM_SEQS }}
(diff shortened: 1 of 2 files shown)


asm582 (approved):
/lgtm

lionelvillard (approved):

PR #661, opened 2026-01-28
🐛 Increase HPA stabilization window to fix flaky e2e test

clubanderson:
## Summary
- Increase `HPA_STABILIZATION_SECONDS` from 30 to 240 in the CI workflow

## Problem
The "should maintain scaled state while load is active" test was flaky because:
- HPA `scaleDown.stabilizationWindowSeconds` = 30 seconds
- Test's `Consistently()` check = 30 seconds
- Same duration creates a race condition where scale-down can occur just as the stability check completes

## Solution
Increase the stabilization window to 240s, ensuring replicas stay stable throughout the entire 30s test period plus buffer time.

## Test plan
- [x] Review CI workflow changes
- [ ] Run `/ok-to-test` to trigger e2e tests

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -68,7 +68,7 @@ on:
       hpa_stabilization_seconds:
         description: 'HPA stabilization window in seconds'
         required: false
-        default: '30'
+        default: '240'
 
 jobs:
   # Gate: Check permissions and handle /ok-to-test for fork PRs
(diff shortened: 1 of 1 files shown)


asm582 (approved):
/lgtm

asm582:
> Thinking out loud here, but the use case for the test itself would be that WVA keeps on suggesting the same desired number of replicas under the same load scenario. If this is not the expected behavior anymore, probably the test itself should change.
> 
> Increasing the HPA stabilization window not only increases scale out/in times, but also hides potential unwanted behaviors of this kind in the tests.
> 
> cc: @asm582

@WheelyMcBones Thanks, can you create an issue about this? We need to make emulated tests similar to real GPU tests in future releases.

PR #677, opened 2026-02-04
Configuration Unification & Namespace-Local Overrides

mamy-CS:
This pr aims to simplify and consolidate config and also provide support for namespace-local overrides. Takes care of issues https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/664, https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/666 and https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/667 

Find below the major changes in a nutshell:
- single config struct with staticConfig and dynamicConfig replacing scattered config sources
- centralized loading with precedence (flags > env > ConfigMap > defaults)
- thread-safe access for concurrent reads/writes to dynamic config
- config passed to components instead of global singletons
- immutable parameter set and detection
- controller exits early if required config is missing or invalid
- namespace-local configmap overrides
- namespace-local > global (automatic fallback on deletion)
- track ns with VA resources and explicit opt-in (label namespaces with `wva.llmd.ai/config-enabled=true` to watch before VAs exist)
- controller watches well-known ConfigMap names across tracked namespaces
- namespace watch permission, added for label-based opt-in support
- predicate logic filters by well-known configmap names
- Updated config documentation 
- Removed unused accelerator-unit-costs configmap

Breaking change introduced***:
- configmap names are renamed with `wva-` prefix, existing deployments will need to rename their configmaps

### DIFF
--- a/test/e2e-saturation-based/e2e_saturation_test.go
+++ b/test/e2e-saturation-based/e2e_saturation_test.go
@@ -264,55 +264,58 @@ var _ = Describe("Test workload-variant-autoscaler - Saturation Mode - Single Va
 		})
 	})
 
-	Context("Before load - initial replica count", func() {
-		It("should have correct initial replica count before applying load", func() {
-			By("waiting for DesiredOptimizedAlloc to be populated")
-			Eventually(func(g Gomega) {
-				va := &v1alpha1.VariantAutoscaling{}
-				err := crClient.Get(ctx, client.ObjectKey{
-					Namespace: namespace,
-					Name:      name,
-				}, va)
-				g.Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Should be able to fetch VariantAutoscaling: %s", name))
+	// TODO: this test is flacky, re-enable once issue is fixed - needs more investigation.
+	/*
+		Context("Before load - initial replica count", func() {
+			It("should have correct initial replica count before applying load", func() {
+				By("waiting for DesiredOptimizedAlloc to be populated")
+				Eventually(func(g Gomega) {
+					va := &v1alpha1.VariantAutoscaling{}
+					err := crClient.Get(ctx, client.ObjectKey{
+						Namespace: namespace,
+						Name:      name,
+					}, va)
+					g.Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Should be able to fetch VariantAutoscaling: %s", name))
 
-				// Wait for DesiredOptimizedAlloc to be populated (ensures reconciliation loop is active)
-				g.Expect(va.Status.DesiredOptimizedAlloc.Accelerator).NotTo(BeEmpty(),
-					"DesiredOptimizedAlloc should be populated with accelerator info")
-				g.Expect(va.Status.DesiredOptimizedAlloc.NumReplicas).To(BeNumerically(">=", 0),
-					"DesiredOptimizedAlloc should have NumReplicas set")
-			}, 10*time.Minute, 10*time.Second).Should(Succeed())
+					// Wait for DesiredOptimizedAlloc to be populated (ensures reconciliation loop is active)
+					g.Expect(va.Status.DesiredOptimizedAlloc.Accelerator).NotTo(BeEmpty(),
+						"DesiredOptimizedAlloc should be populated with accelerator info")
+					g.Expect(va.Status.DesiredOptimizedAlloc.NumReplicas).To(BeNumerically(">=", 0),
+						"DesiredOptimizedAlloc should have NumReplicas set")
+				}, 10*time.Minute, 10*time.Second).Should(Succeed())
 
-			By("querying external metrics API")
-			Eventually(func(g Gomega) {
-				result, err := k8sClient.RESTClient().
-					Get().
-					AbsPath("/apis/external.metrics.k8s.io/v1beta1/namespaces/" + namespace + "/" + constants.WVADesiredReplicas).
-					DoRaw(ctx)
-				g.Expect(err).NotTo(HaveOccurred(), "Should be able to query external metrics API")
-				g.Expect(string(result)).To(ContainSubstring(constants.WVADesiredReplicas), "Metric should be available")
-				g.Expect(string(result)).To(ContainSubstring(name), "Metric should be for the correct variant")
-			}, 2*time.Minute, 5*time.Second).Should(Succeed())
+				By("querying external metrics API")
+				Eventually(func(g Gomega) {
+					result, err := k8sClient.RESTClient().
+						Get().
+						AbsPath("/apis/external.metrics.k8s.io/v1beta1/namespaces/" + namespace + "/" + constants.WVADesiredReplicas).
+						DoRaw(ctx)
+					g.Expect(err).NotTo(HaveOccurred(), "Should be able to query external metrics API")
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -0,0 +1,397 @@
...
+// For namespace-aware lookups, use GetSaturationConfigForNamespace instead.
+func (c *Config) GetSaturationConfig() map[string]interfaces.SaturationScalingConfig {
+	return c.GetSaturationConfigForNamespace("")
+}
+
+// GetSaturationConfigForNamespace returns the saturation scaling configuration for the given namespace.
+// Resolution order: namespace-local > global
+// Thread-safe. Returns a copy to prevent external modifications.
+// If namespace is empty, returns global config.
+func (c *Config) GetSaturationConfigForNamespace(namespace string) map[string]interfaces.SaturationScalingConfig {
```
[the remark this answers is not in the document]
mamy-CS: We can factor out common resolution logic and cache only the most frequently accessed, expensive configs to minimize race conditions and complexity.

internal/config/config.go
```diff
@@ -0,0 +1,533 @@
...
+	Static StaticConfig
+
+	// Dynamic configuration (can be updated via ConfigMap changes)
+	// Protected by mutex for thread-safe updates
+	mu      sync.RWMutex
+	Dynamic DynamicConfig
+
+	// Cache for expensive namespace-aware config lookups
+	// Always initialized (never nil) to avoid nil checks
+	cache configCache
```
lionelvillard: there is no need for a separate cache. Or to be precise, there is no need to keep the raw configuration here (raw = config before global vs namespace resolution), as it is already kept in the informer cache. 

Config should only expose resolved values.

internal/controller/variantautoscaling_controller.go
```diff
@@ -131,11 +119,16 @@ func (r *VariantAutoscalingReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+
 	// Skip if the VA is being deleted
 	if !va.DeletionTimestamp.IsZero() {
 		logger.Info("VariantAutoscaling is being deleted, skipping reconciliation",
 			"name", va.Name,
 			"namespace", va.Namespace)
+		// Untrack namespace when VA is deleted
+		r.untrackNamespace(va.Namespace)
 		return ctrl.Result{}, nil
 	}
```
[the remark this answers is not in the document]
mamy-CS: adding idempotent tracking to avoid potential race

internal/constants/metrics.go
```diff
@@ -93,4 +93,10 @@ const (
 	// ControllerInstanceLabelKey is the label key used to associate VAs with specific controller instances.
 	// Used for multi-controller isolation where each controller only manages VAs with matching labels.
 	ControllerInstanceLabelKey = "wva.llmd.ai/controller-instance"
+
+	// NamespaceConfigEnabledLabelKey is the label key used to opt-in namespaces for namespace-local ConfigMap overrides.
+	// When a namespace has this label set to "true", the controller will watch for namespace-local ConfigMaps
+	// even if no VariantAutoscaling resources exist in that namespace yet.
+	// This enables creating namespace-local ConfigMaps before VAs are created, avoiding race conditions.
+	NamespaceConfigEnabledLabelKey = "wva.llmd.ai/config-enabled"
```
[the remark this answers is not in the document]
mamy-CS: that metrics file is actually misleading. I was debating whether to rename it, keeping constants together or metrics constants separate. And there was already a constant for a label. Anywho, updated comment there for only metrics-related constants and moved labels to a separate file.

internal/controller/variantautoscaling_controller.go
```diff
@@ -274,6 +267,267 @@ func (r *VariantAutoscalingReconciler) handleDeploymentEvent(ctx context.Context
 	return requests
 }
 
+// trackNamespace increments the VA count for a namespace.
+// Thread-safe.
+func (r *VariantAutoscalingReconciler) trackNamespace(namespace string) {
```
lionelvillard: What the purpose of tracking namespaces? Is it to filter out excluded namespaces? If yes, have you considered configuring the controller runtime manager cache to only cache objects in namespaces that don't have the `exclude: true` (or whatever label you decide to have) label?

internal/config/config.go
```diff
@@ -0,0 +1,533 @@
...
+
+// EPPConfig holds EPP (Endpoint Pool) integration configuration.
+type EPPConfig struct {
+	// EPP metric reader bearer token for pod scraping
+	MetricReaderBearerToken string // EPP_METRIC_READER_BEARER_TOKEN
+}
+
+// NamespaceConfig holds configuration for a specific namespace.
+// This structure is used both for global defaults and namespace-local overrides.
+type NamespaceConfig struct {
```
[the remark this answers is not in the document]
lionelvillard: can't testing be done in this package only?

internal/controller/variantautoscaling_controller.go
lionelvillard: Got it. Here what I suggest to make it a bit cleaner:
- add a new namespace reconciler/handler (in namespace_reconciler/handler). 
- have the datastore keeping track of namespace to watch
- have the predicates use the datastore

That way all reconcilers (including the InferencePool reconciler) can be configured to watch only tracked ns.

I was looking for a solution where the informer cache does not include the non-tracked ns but it's not trivial. Premature optimization for sure.

How does that sound?

internal/config/helpers.go
```diff
@@ -43,9 +53,42 @@ func ParseIntFromConfig(data map[string]string, key string, defaultValue int, mi
...
-	if valStr := GetConfigValue(data, key, ""); valStr != "" {
+	if valStr := ConfigValue(data, key, ""); valStr != "" {
 		// Accept "true", "1", "yes" as true
 		return valStr == "true" || valStr == "1" || valStr == "yes"
 	}
 	return defaultValue
 }
+
+// Namespace returns the namespace from environment variable or default.
+func Namespace() string {
```
lionelvillard: Can you be a bit more specific? What about `SystemNamespace`?

internal/config/config.go
```diff
@@ -0,0 +1,765 @@
...
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	interfaces "github.com/llm-d-incubation/workload-variant-autoscaler/internal/interfaces"
+)
+
+// Config is the unified configuration structure for the WVA controller.
+// All fields are private and accessed via thread-safe getter methods.
+type Config struct {
```
lionelvillard: For a client PoV, it might be a bit overwhelming to handle a flat API, for instance `config.MetricAttr()` vs `config.Infra().MetricAttr()`. It's a minor (arguably) improvement to maybe consider in the future.

internal/controller/variantautoscaling_controller.go
```diff
@@ -149,6 +136,28 @@ func (r *VariantAutoscalingReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	// Fetch scale target Deployment
 	scaleTargetName := va.GetScaleTargetName()
+
+	// Handle empty scale target name (invalid configuration)
+	if scaleTargetName == "" {
```
lionelvillard: this is not possible 1. the scale target name field is mandatory 2. the k8s apiserver will reject va objects without a scaletargetref name.

PR #684, opened 2026-02-09
AI: first stab at claude.md file

lionelvillard:


asm582 (changes requested):
Why not use https://agents.md/ ?

lionelvillard:
> Why not use https://agents.md/ ?

because of this https://github.com/anthropics/claude-code/issues/6235

However we can create AGENTS.md and have CLAUDE.md including it, as described here: https://code.claude.com/docs/en/best-practices#configure-your-environment

asm582 (approved):
/lgtm

FYI: https://github.com/llm-d-incubation/workload-variant-autoscaler/issues/687 for creating agents.md in a later PR and integrating Claude.md

PR #713, opened 2026-02-12
config: read main config from file directly

lionelvillard:
### Summary

Replaces the Kubernetes ConfigMap-based main configuration loading with direct YAML file reading, simplifying startup and removing the need for a temporary Kubernetes client during config bootstrap. Dynamic ConfigMaps (saturation scaling, scale-to-zero) are still reconciled at runtime via a dedicated bootstrap step.

### Motivation

Loading the main configuration from a Kubernetes ConfigMap at startup required creating a temporary `client.Client` before the controller manager was ready, adding complexity and fragility. By reading from a mounted YAML file instead, the startup path becomes simpler and more predictable, while dynamic per-namespace ConfigMaps continue to be reconciled normally.

#### Config loading (`internal/config/`)
- `Load()` now accepts a file path instead of `context + client.Client`; reads config via Viper's file support instead of fetching a ConfigMap
- Removed `PrometheusConfig` struct from `interfaces/types.go`; Prometheus settings now accessed via individual getters on `Config`
- Moved `optimizationInterval` into `infrastructureConfig`; removed `UpdateOptimizationInterval()` (interval is now immutable after startup)
- Prometheus cache config parsed from Viper during load instead of from ConfigMap data at runtime
- Added `ConfigMapsBootstrapComplete` / `MarkConfigMapsBootstrapComplete` / `MarkConfigMapsBootstrapFailed` for readiness tracking

#### Controller (`internal/controller/`)
- Added `configmap_bootstrap.go`: new `BootstrapInitialConfigMaps()` method that pre-loads saturation and scale-to-zero ConfigMaps before manager start using an uncached API reader
- `ConfigMapReconciler` now uses `client.Reader` instead of `client.Client`
- Removed `handleMainConfigMap()` and `handleMutableParameters()` — the main ConfigMap is no longer reconciled at runtime
- Removed immutable-parameter-change detection logic (no longer needed)

#### Startup (`cmd/main.go`)
- Added `--config-file` CLI flag for specifying the YAML config path
- Flags are now registered as standalone (`flag.String(...)`) instead of bound to local variables (`flag.StringVar(...)`)
- Removed temporary `client.New()` call; config loads from file before manager creation
- Bootstrap of dynamic ConfigMaps happens via `configMapReconciler.BootstrapInitialConfigMaps()` using the manager's uncached API reader
- Readiness probe now gates on `ConfigMapsBootstrapComplete()`

#### TLS / Prometheus utils (`internal/utils/`)
- `CreateTLSConfig`, `ValidateTLSConfig`, `CreatePrometheusTransport`, `CreatePrometheusClientConfig` now accept `*config.Config` instead of `*interfaces.PrometheusConfig`

#### Helm / Kustomize / Deploy
- ConfigMap templates and deployment manifests updated to reflect file-based config mounting
- `deploy/install.sh` updated accordingly

### Removed

- `interfaces.PrometheusConfig` struct
- `Config.Prometheus()` aggregate getter
- `Config.UpdateOptimizationInterval()`
- `NewTestConfigWithPrometheus()`, `setOptimizationIntervalForTesting()`, `setPrometheusCacheConfigForTesting()`
- `ParsePrometheusConfigFromEnv()`, `ReadPrometheusCacheConfig()`, `ParsePrometheusCacheConfigFromData()`
- `handleMainConfigMap()`, `handleMutableParameters()`, `DetectImmutableParameterChanges()`
- `prometheus_test.go` (replaced by loader-level tests)
- `engine_test.go` for saturation engine (152 lines removed)

### Compatibility

- External config keys (e.g., `GLOBAL_OPT_INTERVAL`, `PROMETHEUS_BASE_URL`) are unchanged
- Config is now read from a YAML file instead of a ConfigMap — **deployment manifests must mount the config file**
- `GLOBAL_OPT_INTERVAL` is no longer mutable at runtime (set once at startup)

### Validation

- All existing and new unit tests pass (`make test`)

### Stats

- **27 files changed**, ~992 insertions, ~1895 deletions (net reduction of ~900 lines)

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -9,13 +9,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-const (
-
-	// ConfigFileKey is the key used in the ConfigMap to store the config file content
-	// when the main config is structured as a single YAML file.
-	ConfigFileKey = "config.yaml"
-)
-
 // flagBindings maps viper keys (= env var names = config file keys) to pflag names.
 var flagBindings = map[string]string{
 	"METRICS_BIND_ADDRESS":           "metrics-bind-address",


internal/config/loader.go
```diff
@@ -95,28 +84,18 @@ func loadConfig(ctx context.Context, cfg *Config, flagSet *flag.FlagSet, k8sClie
 	v.SetDefault("WVA_LIMITED_MODE", false)
 	v.SetDefault("SCALE_FROM_ZERO_ENGINE_MAX_CONCURRENCY", 10)
 	v.SetDefault("EPP_METRIC_READER_BEARER_TOKEN", "")
+	v.SetDefault("GLOBAL_OPT_INTERVAL", "60s")
```
mamy-CS: GLOBAL_OPT_INTERVAL change requires a pod start now. Should be in the future release notes, we should leave a note somewhere.

mamy-CS (approved):
good refactor, approving with some minor comments.

PR #717, opened 2026-02-13
🐛 Fix helmfile overwriting WVA with wrong monitoringNamespace

clubanderson:
## Summary
- Skip the WVA release in the llm-d helmfile when `DEPLOY_WVA=true` using `--selector kind!=autoscaling`
- Prevents the helmfile from overwriting the correctly-configured WVA installation

## Problem
The nightly E2E workflow fails at "Deploy infrastructure" with:

```
FAILED RELEASES:
workload-variant-autoscaler   llm-d-nightly-wva   oci://ghcr.io/llm-d-incubation/...

STDERR:
  Error: 1 error occurred:
    * namespaces "llm-d-monitoring" not found
```

**Root cause:** `deploy/install.sh` has two deployment paths for WVA:

1. `deploy_wva_controller()` (line ~1264) installs WVA from the **local chart** with correct env-specific values (`monitoringNamespace=openshift-user-workload-monitoring`)
2. `deploy_llm_d_infrastructure()` (line ~1271) clones the `llm-d/llm-d` repo and runs `helmfile apply`, which includes a WVA release using the **published OCI chart** with upstream defaults (`monitoringNamespace: llm-d-monitoring`)

Step 2 overwrites step 1, replacing `openshift-user-workload-monitoring` with `llm-d-monitoring`, which doesn't exist on the OpenShift cluster.

## Fix
When `DEPLOY_WVA=true`, add `--selector kind!=autoscaling` to the helmfile apply command to skip the WVA release (which is labeled `kind: autoscaling` in the helmfile). This lets the local chart installation from step 1 remain intact with the correct values.

## Test plan
- [ ] Re-trigger nightly E2E workflow — WVA should deploy without `llm-d-monitoring` namespace error
- [ ] Verify helmfile still deploys infra, gaie, and modelservice releases correctly
- [ ] Verify WVA controller is running with correct `monitoringNamespace` setting


Gregory-Pereira (approved):
/lgtm

PR #727, opened 2026-02-13
Enhance queueing model used by queue analyzer

atantawi:
Improve LLM queue analyzer model to handle simultaneous processing of prefill and decode in an iteration. The model uses three parameters which may be estimated either offline or online through a model tuner.
The model is used to

1. predict performance (TTFT and ITL) given a request load (requests/secs and average input and output tokens) and server configuration (max batch size); and/or
2. calculate the maximum throughput that attains a given performance (TTFT and/or ITL target values).


lionelvillard (approved):
LGTM.

Eventually we want this code to be integrated with the rest of the system.  The analyzer package may go under https://github.com/llm-d/llm-d-workload-variant-autoscaler/tree/main/internal/engines/analyzers. I don't know for the rest of the code. @atantawi do you have an idea?

PR #734, opened 2026-02-15
doc: update wva version in install doc

shuynh2017:
- In README.md, I suggest to point users to proper steps for helm method. If users just run the helm command as currently shown then it won't work since there are steps must be done before to be able to run helm command successfully.
- Update wva release to v0.5.1. This change will bring the install with kind method and helm method to the same version which is v0.5.0.


charts/workload-variant-autoscaler/README.md
```diff
@@ -59,7 +59,7 @@ helm ls -A
 ```
 export OWNER="llm-d"
 export WVA_PROJECT="llm-d-workload-variant-autoscaler"
-export WVA_RELEASE="v0.4.1"
+export WVA_RELEASE="v0.5.1"
```
lionelvillard: 0.5.1 hasn't been released yet. Please use v0.5.0

lionelvillard (commented):
@shuynh2017 please check your last commit it includes more changes you intended, right?

PR #737, opened 2026-02-15
e2e test: skip test case "Replica stability under constant load"

shuynh2017:
See comments within


shuynh2017:
Minor comment: if this testcase works consistently in CI pipeline then fine. I tried and reviewed, and don't see how this can work consistently.

Issue #741, opened 2026-02-16
Adding support for SLO-driven model-based scaling


atantawi:
Current autoscaler considers the utilization of the KV cache and the request queue length as indicators of saturation, and it accordingly scales up or down the number of replicas, without consideration of user-perceived performance metrics such as TTFT and ITL, nor potential specifications of targets (SLOs) for such metrics.

This issue addresses the need for an SLO-driven, model-based autoscaling. The reason for a performance model is to assess the number of replicas required to attain a given SLOs, based on the observer/predicted load on the servers.

Several tasks are envisioned to address this issue.

- [ ] Add some some basic functionality.

> - [ ] Add a queueing model to analyze and size a replica of an inference server. The model may be called to (1) analyze the steady state (server state over a period of time, such as a control cycle, as opposed to a given current state) and produce TTFT and ITL values for a given load (request rate and input/output token characteristics) and (2) size a server by providing the maximum request rate for a given load in order to attain TTFT and ITL SLOs. The queueing model may have a handful of internal parameters, which may be obtained off-line or on-line, preferably the latter. (PR #727 )
> - Add an on-line model tuner to continuously estimate the queueing model internal parameters using observations and standard filtering techniques. (PR #743 )

- [ ] Conduct experiments to validate queueing model and on-line model tuner.

- [ ] Add means to collect metrics from the Scheduler related to the offered request rate, as opposed to the successfully completed requests. The former should be used to size the server, whereas the latter should be used to tune/calibrate the queueing model.

- [ ] Add mechanism to specify SLOs. Two alternatives are possible: (1) explicit by modifying the APIs to allow user/admin to specify absolute/relative values for TTFT and ITL SLO targets, and (2) implicit by having WVA automatically setting target values based on observations and optimizing an internal objective.

- [ ] Optionally, add a predictor of load (request rate and token distributions) for the next control cycle, given the statistical history, potentially using time-series techniques.

- [ ] Add a new model-based scaling engine which employs above queueing model and on-line model tuner.

- [ ] Design and conduct e2e experiments to validate the behavior of the model-based scaling engine.

Notes:

- TTFT measure should be expressed as a relative measure to the number of prompt tokens in the requests, e.g. TTFT per 100 tokens.
- Queueing model need not be exact, as it is used solely to estimate maximum load to attain a given SLOs, as opposed to performance evaluation of performance metrics.

PR #743, opened 2026-02-16
Basic model tuner functionality

vishakha-ramani:
This PR relates to Issue #741
Specifically, it adds an on-line model tuner to continuously estimate the queueing model internal parameters using observations and standard filtering techniques.
A caller of the model tuner will be implemented in subsequent PRs.
cc @atantawi

### DIFF
--- a/internal/engines/analyzers/queueingmodel/tuner/defaults.go
+++ b/internal/engines/analyzers/queueingmodel/tuner/defaults.go
@@ -9,27 +9,6 @@ const DefaultEpsilon = 1e-6
 
 // Default tuner parameters
 const (
-	// scaling factor used to adjust the calculated measurement noise variance (> 0).
-	DefaultGammaFactor = 1.0
-
-	/*
-		DefaultErrorLevel,  DefaultTPercentile along with DefaultGammaFactor is used in initializing the measurement noise covariance matrix R.
-		Particularly, R is assumed to be diagonal, meaning measurement errors are independent across different measures ($z_i$).
-		The measurement of a performance metric of interest, say z_i, is expected to have a 95% Confidence Interval (CI) of +-5% of its mean value.
-		This means that the error (standard deviation) around z_i is expected to be 5% (DefaultErrorLevel).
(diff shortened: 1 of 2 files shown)


pkg/tuner/defaults.go
```diff
@@ -0,0 +1,51 @@
...
+		The measurement of a performance metric of interest, say z_i, is expected to have a 95% Confidence Interval (CI) of +-5% of its mean value.
+		This means that the error (standard deviation) around z_i is expected to be 5% (DefaultErrorLevel).
+		However, we assume that the measurement noise (error) follows a t-distribution.
+		Under the asymptotic limit, the t-distribution converges to a standard normal distribution, the critical value of which is 1.96 (DefaultTPercentile).
+	*/
+	DefaultErrorLevel  = 0.05
+	DefaultTPercentile = 1.96
+
+	// determines the limit on the amount of change in a state value per iteration. A small number results in the filter converging relatively slowly.
+	DefaultPercentChange = 0.05
```
atantawi: Some quantities that are defined in this file are not used anywhere: `DefaultPercentChange` `DefaultMinStateFactor` `DefaultMaxStateFactor` `TransientDelaySeconds` `BaseFactor`
I believe they belong to the caller of the Tuner, at a higher level. These quantities should be moved up in that layer.

lionelvillard:
> > @vishakha-ramani thanks for your PR!
> > @vishakha-ramani @atantawi as discussed, let's move all queuing model related code under `internal/engines/analyzers/queueing_model` (or `.../queue_model`). The tuner should go under `internal/engines/analyzers/queueing_model/tuner`, right?
> > @vishakha-ramani can you also fix The typo?
> 
> The queue model and model tuner, currently under `pkg` are very generic. Think of them as libraries. They have nothing to do with the notion of a variant, nor auto-scaling. If you'd ask me, I'd have them in an external github module that we could import. But, I could go either way.

There are many packages in this repo that are generic and that have nothing to do with the notion of variant, or autoscaling.

atantawi (changes requested):

PR #745, opened 2026-02-16
🐛 Fix Helm ownership collision for prometheus-ca ConfigMap in shared namespace

clubanderson:
## Summary

- Use `clusterResourceName` helper instead of `fullname` for the prometheus-ca ConfigMap in the shared monitoring namespace, making the name unique per release namespace

## Problem

When multiple WVA releases are installed on the same cluster from different namespaces (e.g., `llm-d-vezio-a` and `llm-d-vezio-b`), they both try to create a ConfigMap named `workload-variant-autoscaler-prometheus-ca` in `openshift-user-workload-monitoring`. Helm's ownership annotation (`meta.helm.sh/release-namespace`) rejects the second install:

```
Error: Unable to continue with install: ConfigMap "workload-variant-autoscaler-prometheus-ca"
in namespace "openshift-user-workload-monitoring" exists and cannot be imported into the current
release: invalid ownership metadata; annotation validation error: key
"meta.helm.sh/release-namespace" must equal "llm-d-vezio-b": current value is "llm-d-vezio-a"
```

## Fix

The chart already has a `clusterResourceName` helper that appends `.Release.Namespace` to the resource name — it's used by all ClusterRoles and ClusterRoleBindings. This PR applies the same helper to `prometheus-ca-configmap-prom.yaml` so the ConfigMap name becomes unique per namespace (e.g., `workload-variant-autoscaler-llm-d-vezio-a-prometheus-ca`).

The release-namespace copy (`prometheus-ca-configmap-wva.yaml`) is unchanged — the controller deployment mounts from that one.

## Test plan

- [ ] Install WVA from namespace A — ConfigMap created with namespace-qualified name
- [ ] Install WVA from namespace B on the same cluster — no Helm ownership collision
- [ ] Verify controller still mounts the CA cert correctly (from release-namespace copy)

### DIFF
--- a/test/e2e-openshift/sharegpt_scaleup_test.go
+++ b/test/e2e-openshift/sharegpt_scaleup_test.go
@@ -56,6 +56,14 @@ func sanitizeK8sName(name string) string {
 
 var lowLoad = numPrompts <= 2000 && requestRate <= 8
 
+// truncateString returns the first n characters of s, appending "..." if truncated.
+func truncateString(s string, n int) string {
+	if len(s) <= n {
+		return s
+	}
+	return s[:n] + "..."
+}
+
 // Load generation configuration constants
 // These values were tuned empirically to achieve ~2-3 replica scale-up without excessive scaling.
 // Original values (baseLoadWorkers=10, batchSize=50, batchSleepDuration=0.1) caused cascade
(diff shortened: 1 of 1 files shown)


clubanderson:
👋 @asm582 — all CI checks green. Could you approve when you get a chance? Thanks!

PR #749, opened 2026-02-17
doc: update wva version in install doc

shuynh2017:
- In README.md, I'd suggest to point users to proper steps for helm method. If users just run the helm command as currently shown then it won't work since there are steps must be done before to be able to run helm command successfully.
- Update wva release to v0.5.0.


mamy-CS (approved):
/lgtm thanks

PR #761, opened 2026-02-18
🐛 Fix KEDA APIService race condition in nightly E2E

clubanderson:
## Summary

- Add background APIService guard loop that re-patches `v1beta1.external.metrics.k8s.io` every 10s if KEDA reclaims it
- Guard starts when KEDA is detected during `deploy_prometheus_adapter()`, stops during cleanup
- Does **not** modify KEDA itself — just ensures the APIService always points to `prometheus-adapter`
- Increase external metrics API test timeout from 5min to 10min
- Increase scale-to-zero test timeout from 5min to 10min

## Problem

KEDA operator continuously reconciles the `v1beta1.external.metrics.k8s.io` APIService back to `keda-metrics-apiserver` within ~2 minutes. The existing one-shot patch in `install.sh` gets overwritten, causing the HPA to receive 404 errors when querying `wva_desired_replicas`. This makes the nightly E2E tests fail intermittently on OCP clusters with KEDA installed.

KEDA has no built-in option to disable external metrics registration ([issue #470](https://github.com/kedacore/keda/issues/470), open since 2019).

## Test plan

- [ ] Trigger WVA nightly on OCP — should no longer fail with KEDA APIService 404s
- [ ] Verify guard output shows re-patching in workflow logs when KEDA reclaims
- [ ] Verify guard stops cleanly during cleanup (no orphaned processes)
- [ ] Run multiple consecutive E2E runs to confirm stability

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -136,6 +136,59 @@ log_error() {
     exit 1
 }
 
+# APIService guard: background loop that continuously ensures the
+# v1beta1.external.metrics.k8s.io APIService points to prometheus-adapter.
+# On clusters with KEDA, the operator continuously reconciles the APIService
+# back to keda-metrics-apiserver, breaking HPA scaling for WVA.
+# This guard re-patches it every 10 seconds without modifying KEDA itself.
+APISERVICE_GUARD_PID=""
+
+start_apiservice_guard() {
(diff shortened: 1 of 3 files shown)


mamy-CS (approved):
/lgtm

PR #778, opened 2026-02-24
e2e: Add KEDA as optional scaler backend and refactor tests for dual-backend support

mamy-CS:
This PR is mainly about adding KEDA as a second scaler backend in emulated e2es. 
- Introduced `SCALER_BACKEND` (default prometheus-adapter; optional keda). With KEDA, tests create ScaledObjects (and use a ScaledObject builder) instead of HPAs; with Prometheus Adapter, behavior is unchanged (HPA + external metric).
- Fixtures and test layout update: Add scaled_object_builder.go and model_service_builder.go, and refactor e2e so the same scenarios (saturation, scale-to-zero, scale-from-zero, limiter, smoke, etc.) run for either backend via backend-aware scaler creation and assertions.

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -1624,7 +1624,12 @@ main() {
     fi
 
     # Deploy scaler backend: KEDA or Prometheus Adapter
+    # KEDA in this script is for kind-emulator e2e only; on OpenShift use the platform CMA / Prometheus Adapter.
     if [ "$SCALER_BACKEND" = "keda" ]; then
+        if [ "$ENVIRONMENT" != "kind-emulator" ]; then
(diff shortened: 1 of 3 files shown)


api/v1alpha1/variantautoscaling_types.go
```diff
@@ -121,6 +121,14 @@ const (
...
 )
 
+// Condition messages for MetricsAvailable
+const (
+	// MessageMetricsAvailable indicates metrics are available for scaling decisions
+	MessageMetricsAvailable = "Saturation metrics data is available for scaling decisions"
+	// MessageMetricsUnavailable indicates metrics are not available
+	MessageMetricsUnavailable = "No saturation metrics available - pods may not be ready or metrics not yet scraped"
+)
+
```
asm582: Why does KEDA integration need conditions? Can we avoid an API change?

PR #796, opened 2026-02-27
Add scaling-decision logging and E2E failure/debug dumps

mamy-CS:
Improves debuggability when scaling doesn’t behave as expected and when e2e tests fail or hang.

### DIFF
--- a/internal/engines/scalefromzero/engine.go
+++ b/internal/engines/scalefromzero/engine.go
@@ -267,8 +267,8 @@ func (e *Engine) processInactiveVariant(ctx context.Context, va wvav1alpha1.Vari
 	}
 
 	if !pendingRequestExist {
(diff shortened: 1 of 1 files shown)


internal/engines/scalefromzero/engine.go
```diff
@@ -263,7 +267,11 @@ func (e *Engine) processInactiveVariant(ctx context.Context, va wvav1alpha1.Vari
 	}
 
 	if !pendingRequestExist {
-		logger.V(logging.DEBUG).Info("No pending requests found in the flowcontrol queue - skipping scaling up from zero")
+		// Log at INFO so E2E and operators see why this VA stayed at zero (no deceptive silence).
```
lionelvillard: the scale from zero loop runs every 100ms, meaning this message will appear 10 times per second for each inactive VAs. This is flooding the log and that's why the verbosity level has been set to DEBUG.

Issue #748, opened 2026-02-17
ConfigMap bootstrapping does not load ConfigMap in all namespaces


lionelvillard:
### Problem
ConfigMap bootstrapping currently fails to load ConfigMap in all namespaces as intended.  
### Expected Behavior
When bootstrapping, the ConfigMap should be correctly loaded  in every controller-scoped namespace, ensuring consistent configuration before starting controllers

### Steps to Reproduce
1. Deploy the autoscaler in a cluster with multiple namespaces.
2. Trigger the ConfigMap bootstrapping process.
3. Observe that the ConfigMap is only loaded  in the system and watched namespace.

shuynh2017:
@lionelvillard, I can help out with this one?

PR #812, opened 2026-03-02
Bug fix: Load Configmap in all namespaces during bootstrap

shuynh2017:
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/748
- Ran lint, unit test, e2e-smoke

### DIFF
--- a/internal/controller/configmap_bootstrap.go
+++ b/internal/controller/configmap_bootstrap.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/config"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/constants"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


mamy-CS (commented):
Thanks for the fix and the thorough tests @shuynh2017. This is a nice improvement for all-namespaces mode.
One behavior worth calling out: during bootstrap we now scan all non-excluded namespaces for namespace-local cms, regardless of whether the namespace is tracked (has VAs) or has `wva.llmd.ai/config-enabled=true`. That’s a bit more permissive than the steady-state cm reconciler logic, which gates namespace-local watching on tracking/opt-in and exclusion.
This is probably fine and likely what we want, so that early namespace-local cms are honored, but it does create an interesting asymmetry:
At startup, we may load namespace-local config for a namespace that is neither tracked nor opted-in.
After startup, updates to that cm won’t be observed unless the namespace later becomes tracked or gets `wva.llmd.ai/config-enabled=true`.
Might be worth either:
Narrowing `namespacesToScan` to tracked OR config-enabled=true (still honoring `wva.llmd.ai/exclude`), or
Explicitly documenting that bootstrap does a one-time, best-effort sweep over all non-excluded namespaces, while ongoing updates still depend on tracking/opt-in.
Otherwise, the changes lgtm, unit tests for the new paths are clear and passed locally for me.

mamy-CS (approved):
@shuynh2017 thanks for adding the `wva.llmd.ai/config-enabled` gating.
Bootstrap in all-namespaces mode now only scans namespaces that have the opt-in label, which matches the reconciler’s steady-state behavior and keeps bootstrap from loading config for arbitrary namespaces. The new test that skips namespaces with `config-enabled=false` and the updated cases that set the label on included namespaces look good.
/lgtm

shuynh2017:
@mamy-CS thank you!

Issue #807, opened 2026-02-28
Add `minReplicas` and `maxReplicas` to VA Spec


ev-shindin:
**Scope:** CRD types, validation, enforcer integration

Add replica bounds to `VariantAutoscalingSpec`:

```go
type VariantAutoscalingSpec struct {
    ScaleTargetRef autoscalingv1.CrossVersionObjectReference `json:"scaleTargetRef"`
    ModelID        string                                     `json:"modelID"`
    VariantCost    string                                     `json:"variantCost,omitempty"`

    // MinReplicas is the lower bound for replicas. 0 enables scale-to-zero.
    // +kubebuilder:validation:Minimum=0
    // +kubebuilder:default=1
    MinReplicas *int32 `json:"minReplicas,omitempty"`

    // MaxReplicas is the upper bound for replicas.
    // +kubebuilder:validation:Minimum=1
    // +kubebuilder:default=10
    MaxReplicas int32 `json:"maxReplicas"`
}
```

**Changes required:**
- `api/v1alpha1/variantautoscaling_types.go` — Add fields with kubebuilder annotations
- `internal/engines/pipeline/enforcer.go` — Read min/max from VA spec instead of hardcoded logic. Enforcer currently ensures min=1 globally; should respect per-VA minReplicas
- `internal/engines/saturation/engine.go` — Pass min/max to optimizer. Currently the enforcer is the only min-replica guard; maxReplicas is only enforced by the GPU limiter
- `internal/interfaces/saturation_analyzer.go` — Add MinReplicas/MaxReplicas to `VariantReplicaState` or `VariantDecision` so pipeline components can access them
- CRD manifests regeneration (`make manifests`)
- Validation webhook (optional): ensure `minReplicas <= maxReplicas`
- Update printcolumns to show min/max

**Backward compatibility:**
- `minReplicas` defaults to 1 (current implicit behavior)
- `maxReplicas` defaults to 10 (matches current Helm chart default)
- Existing VAs without these fields work unchanged

PR #830, opened 2026-03-03
refactor(api): extract VariantAutoscalingConfigSpec for higher-level controller integration

### LINKED ISSUE #820: Extract inlineable VariantAutoscalingConfigSpec for higher-level controller integration
## Motivation

The current VA API works great standalone — users provide `scaleTargetRef`, `modelID`, and optional fields like `variantCost`. All three live in a single flat `VariantAutoscalingSpec`.

Higher-level controllers like [KServe](https://github.com/kserve/kserve) also create VA CRs, but in that context `scaleTargetRef` and `modelID` are **derived by the KServe controller** (from the Deployment and model it creates). Only the optional tuning fields like `variantCost` are user-facing.

Because required and optional fields are mixed in one struct, KServe **cannot import and embed `VariantAutoscalingSpec`** without exposing controller-internal fields to users. Instead, KServe must maintain its own mirror types, mapping logic, validation, and tests for every optional field — and update all of that each time WVA adds a new one.

## Proposal

Extract optional tuning fields into a separate, embeddable type:

```go
type VariantAutoscalingConfigSpec struct {
    // +optional
    // +kubebuilder:default="10.0"
    VariantCost string `json:"variantCost,omitempty"`
    // future optional fields go here
}

type VariantAutoscalingSpec struct {
    ScaleTargetRef autoscalingv1.CrossVersionObjectReference `json:"scaleTargetRef"`
    ModelID        string                                     `json:"modelID"`
    VariantAutoscalingConfigSpec `json:",inline"`
}
```

**This is fully backward compatible** — the CRD schema and YAML format are byte-for-byte identical. Standalone users see no change.

## How KServe would use it

**Before** (fragile — must mirror every field):

```go
type WVASpec struct {
    VariantCost string `json:"variantCost,omitempty"` // manually duplicated
    HPA  *HPAScalingSpec  `json:"hpa,omitempty"`
    KEDA *KEDAScalingSpec `json:"keda,omitempty"`
}
```

**After** (stays in sync automatically):

```go
type WVASpec struct {
    wvav1alpha1.VariantAutoscalingConfigSpec `json:",inline"` // embed upstream type
    HPA  *HPAScalingSpec  `json:"hpa,omitempty"`
    KEDA *KEDAScalingSpec `json:"keda,omitempty"`
}
```

Controller construction becomes a direct pass-through:

```go
va.Spec = wvav1alpha1.VariantAutoscalingSpec{
    ScaleTargetRef:                 derived,   // controller-derived
    ModelID:                        derived,   // controller-derived
    VariantAutoscalingConfigSpec:   userSpec,  // passed through as-is
}
```

New optional WVA fields automatically appear in KServe's CRD with **zero code changes**.

## Precedent

- **KEDA**: provides embeddable `AdvancedConfig` type
- **autoscaling/v2**: provides standalone `HorizontalPodAutoscalerBehavior` for embedding

## Summary

- Pure Go refactor — no CRD schema change, no breaking change
- Standalone WVA usage is completely unaffected
- Enables integrators to embed optional fields directly instead of maintaining fragile mirrors

Happy to contribute a PR if the team agrees with this direction.

vivekk16:
## Summary

Closes #820.

Extracts optional tuning fields from `VariantAutoscalingSpec` into a new standalone embeddable `VariantAutoscalingConfigSpec` type, enabling higher-level controllers (e.g. KServe) to embed it directly instead of maintaining fragile field mirrors that drift over time.

## Changes

- `api/v1alpha1/variantautoscaling_types.go`: add `VariantAutoscalingConfigSpec` with `VariantCost`; embed inline into `VariantAutoscalingSpec` via `json:",inline"`
- `api/v1alpha1/zz_generated.deepcopy.go`: add `DeepCopyInto`/`DeepCopy` for the new type
- `test/utils/e2eutils.go`, `test/e2e/fixtures/va_builder.go`: update struct literals to use the embedded type

## Backward Compatibility

- CRD schema is **byte-for-byte identical** — no re-install or migration required
- All existing `va.Spec.VariantCost` access continues to work via Go field promotion
- Standalone WVA usage is completely unaffected

## How integrators benefit

**Before** — must mirror and maintain every optional field manually:

```go
type WVASpec struct {
    VariantCost string `json:"variantCost,omitempty"` // manually duplicated
}
```

**After** — embeds upstream type, stays in sync automatically:

```go
type WVASpec struct {
    wvav1alpha1.VariantAutoscalingConfigSpec `json:",inline"`
}
```


vivekk16:
> > @ev-shindin @asm582
> > Can the `minReplicas` and `maxReplicas` fields being added in #807 live inside VariantAutoscalingConfigSpec rather than directly in VariantAutoscalingSpec? They are user-facing tuning knobs that integrators like KServe would want to expose, so the config struct seems like the right home for them.
> 
> Good question, before we decide that, have two questions:
> 
> * Can you describe briefly how Kserve is going to consume this and configure HPA or KEDA?
> * As a forward looking requirement how will we support LWS: https://github.com/kubernetes-sigs/lws/blob/2d92695ce7470e9ed536b5c3a226256de08b952e/docs/examples/vllm/GPU/lws.yaml#L6-L8

@asm582 For the first question
By "this" — do you mean `minReplicas`/`maxReplicas` specifically? or `VariantAutoscalingConfigSpec` as a whole?

Issue #831, opened 2026-03-03
CI: /trigger-e2e-full runs full e2e against latest main instead of PR head


mamy-CS:
When someone comments /trigger-e2e-full on a PR, the full e2e suite runs against the latest commit on main, not the PR’s head commit.

Cause
The workflow is triggered by issue_comment. In that case, the run is associated with the repository’s default branch (main), not the PR branch.
The e2e-tests job does a plain checkout with no ref:
  - name: Checkout source    uses: actions/checkout@v4
So it checks out whatever ref the run is using (main), not the PR head.
The check-code-changes job correctly checks out the PR head via steps.pr-info.outputs.pr_head_sha, but that ref is never passed to e2e-tests, so e2e still runs on main.
- Failures or passes may be misleading because the code under test is not the PR branch.

Expected behavior
Commenting /trigger-e2e-full on a PR should run the full e2e suite on that PR’s head commit.

PR #832, opened 2026-03-03
🐛 fix: checkout PR head SHA for issue_comment triggered workflows

clubanderson:
## Problem

Fixes https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/831

When `/trigger-e2e-full` is used on a PR, the workflow triggers via `issue_comment` event. For this event type, `github.sha` points to `main` (the default branch), **not** the PR head. This caused the `e2e-tests` job to build and test code from `main` instead of the PR changes.

## Root Cause

The `check-code-changes` job already correctly resolves the PR head SHA and checks it out, but `e2e-tests` did a bare `actions/checkout@v4` with no `ref:`, defaulting to `github.sha` (main for issue_comment events).

## Fix

- **check-code-changes**: Export `pr_head_sha` as a job output (already computed but not exposed)
- **e2e-tests**: Use `needs.check-code-changes.outputs.pr_head_sha` in checkout ref, with a validation step that fails fast if the SHA is empty (prevents silent fallback to main)
- **lint-and-test**: Skip entirely for `issue_comment` events — it already runs on `pull_request` triggers, so re-running on comments is unnecessary and would allow untrusted commenters to execute PR code
- **e2e-tests gate**: Tightened so `issue_comment` events only run when `check-full-tests` validates an approved trigger from a trusted collaborator AND `check-code-changes` succeeds
- **workflow_dispatch**: Preserved smoke-test path — runs when `run_full_tests=true` OR when code changes are detected (matching existing behavior)

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -256,14 +256,26 @@ jobs:
     needs: [lint-and-test, check-code-changes, check-full-tests]
     if: >-
       always() && (
-        (github.event_name == 'issue_comment' && needs.check-full-tests.outputs.run_full == 'true') ||
-        (github.event_name == 'workflow_dispatch' && needs.check-full-tests.outputs.run_full == 'true') ||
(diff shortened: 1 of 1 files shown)


mamy-CS (approved):
/lgtm

Issue #690, opened 2026-02-11
bug: Scale-to-zero Prometheus queries fail due to missing metric labels


mamy-CS:
Scale-to-zero enforcement fails because Prometheus queries for `vllm:request_success_total` expect `namespace` and `model_name` labels that are not present on the scraped metrics. This causes the query to return empty results, leading the enforcer to incorrectly scale deployments to zero. The `llm-d-sim` ServiceMonitor configuration `(test/utils/resources/llmdsim.go)` does not inject the required `namespace` and `model_name` labels into scraped metrics.

**Impact:**
E2E tests fail in scale-to-zero=true mode
Two tests affected:
- e2e_scale_to_zero_test.go: "should have correct initial replica count before applying load"
- e2e_saturation_test.go: "should have correct initial replica count before applying load"

**After the issue is fixed:**
Uncomment the disabled tests in:
- test/e2e-saturation-based/e2e_scale_to_zero_test.go
- test/e2e-saturation-based/e2e_saturation_test.go
- Run e2e tests with WVA_SCALE_TO_ZERO=true

shuynh2017:
I see it now after manually sending a request in:
```
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "__name__": "vllm:request_success_total",
          "container": "vllm",
          "endpoint": "metrics",
          "finish_reason": "length",
          "instance": "10.244.1.3:8200",
          "job": "llm-d-sim/ms-sim-llm-d-modelservice-decode-podmonitor",
          "model_name": "unsloth/Meta-Llama-3.1-8B",
          "namespace": "llm-d-sim",
          "pod": "ms-sim-llm-d-modelservice-decode-694c4766d8-7xt4h"
        },
        "value": [1772661228.485, "1"
        ]
      },
      {
        "metric": {
          "__name__": "vllm:request_success_total",
          "container": "vllm",
          "endpoint": "vllm",
          "finish_reason": "length",
          "instance": "10.244.1.3:8200",
          "job": "workload-variant-autoscaler-vllm",
          "model_name": "unsloth/Meta-Llama-3.1-8B",
          "namespace": "llm-d-sim",
          "pod": "ms-sim-llm-d-modelservice-decode-694c4766d8-7xt4h",
          "service": "workload-variant-autoscaler-vllm"
        },
        "value": [1772661228.485, "1"
        ]
      }
    ]
  }
}```

PR #846, opened 2026-03-05
Bug: restart prometheus-adapter for stable e2e test

shuynh2017:
The issue:
- The test case "should scale up under load" in scale_to_zeo_test.go is flaky, e.g. it scales ok, then stop next morning, etc. Observed:
  - WVA scaled up desired replicas correctly, but HPA failed to get the external metrics to scale up. Thus the test case would timeout and failed.
  - After restarting prometheus-adapter, this test case passes consistently.

A stable prometheus-adapter is essential for consistent e2e test so I propose to restart the adapter at beginning of the test suite.

Related: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/690

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -134,6 +134,52 @@ var _ = BeforeSuite(func() {
 		g.Expect(pods.Items).NotTo(BeEmpty(), "Prometheus pod not found")
 	}, 2*time.Minute, 5*time.Second).Should(Succeed(), "Prometheus should be running")
 
+	// prometheus-adapter exports the needed custom metrics for autoscaling, 
+	// but sometimes the metrics can't be obtained until prometheus-adaper is restarted.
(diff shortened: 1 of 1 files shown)


mamy-CS (commented):
Restarting the adapter in before suite is a good approach for stabilizing the test. Some comments before merge:
- Run the restart only when `cfg.ScalerBackend == "prometheus-adapter"`. With `SCALER_BACKEND=keda` there may be no prometheus-adapter pods, so the “wait for at least one ready” step can block for 3 minutes and then fail. Match the pattern used in `smoke_test.go` for the external metrics check.
- Please document that this is a workaround and that the underlying reason the adapter stops serving metrics (e.g. caching/discovery) is still worth investigating (Add a to do comment).
- Also make sure that doesn't run for openshift and only emulated, since the e2e suite is environment agnostic.

mamy-CS (approved):
/lgtm

Issue #777, opened 2026-02-24
Impact on max replicas on changing HPA stabilization window with llm-d-sim


asm582:
For a constant traffic of 10 RPS, WVA quickly scales to the max replicas when the HPA stabilization window is set to 0, but shows a staggered scale-up when the HPA stabilization window is set to 240. Careful tuning of this parameter is essential to achieve the desired scale-up performance, reducing request queuing time.

shuynh2017:
[experiment.md](https://github.com/user-attachments/files/25775976/experiment.md)
<img width="769" height="757" alt="Image" src="https://github.com/user-attachments/assets/ee13f500-6b8d-4235-b15b-0dc0d3f5e51a" />
<img width="761" height="562" alt="Image" src="https://github.com/user-attachments/assets/82efe2a5-9ffd-42a7-af97-706d0c1a9ec4" />
<img width="767" height="187" alt="Image" src="https://github.com/user-attachments/assets/36ea7b28-d626-408a-a09f-ed5a41ee3c55" />

PR #850, opened 2026-03-05
doc: stablizationWindow recommendations

shuynh2017:
Experiment results: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/777#issuecomment-4006706514


asm582 (dismissed):
/lgtm

it would have been better if we would have explicitly mentioned that the scaledown is further impacted by pod termination grace period set by the user: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

asm582 (approved):
/lgtm
/approve

PR #852, opened 2026-03-06
fix: cert name and make mount for optional to EPP pod + missing RBAC in "config" for OCP

zdtsw:
# Notes
- for OCP we need assume the monitoring stack is already up and running in the `openshift-monitoring` namespace
- esp. resource like thanos, clusterrole and configmap for cert (which is not about monitoring) are in place

# Changes
- keep only one replica for deployment via Kustomize
- add serect for SA as optional for mount 
  - if EPP set --metrics-endpoint-auth=false no need it
  - by default GIE has --metrics-endpoint-auth=true, we need secret to be mounted
- fix cert from CM openshift-service-ca.crt
- add missing RBAC on inferencepool and deployment/scale
- add missing rolebinding on controller for the monitoring view

ref https://issues.redhat.com/browse/RHOAIENG-52196

# Test
run `>cd llm-d-workload-variant-autoscaler/config/openshift; oc apply -f .`
check pod `workload-variant-autoscaler-controller-manager` is running in `workload-variant-autoscaler-system` with no error, except `"2026-03-06T10:42:06Z	INFO	saturation/engine.go:180	No active VariantAutoscalings found, skipping optimization" `is expected

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -21,7 +21,6 @@ spec:
     matchLabels:
       control-plane: controller-manager
       app.kubernetes.io/name: workload-variant-autoscaler
-  # k8s defaults to 1 replica, omit config here, but keep replica 2 in chart
   template:
     metadata:
       annotations:


config/openshift/epp-metrics-volume-patch.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: system
+spec:
+  template:
+    spec:
+      volumes:
+      - name: epp-metrics-token
+        secret:
+          secretName: epp-metrics-token
+          defaultMode: 420
+          optional: true  # wont fail deployment if secret doest not exist
```
pierDipi: if this is not actually optional I'd not do this as it will recover once it gets created

zdtsw:
@mamy-CS can you take a look at this PR?

lionelvillard:
/lgtm

@pierDipi ok to merge?

PR #863, opened 2026-03-09
build: sync Helm chart CRD from config/crd/bases on make manifests

vivekk16:
## Why

The CRD lives in two places:
- `config/crd/bases/` — generated by controller-gen, used by kustomize
- `charts/workload-variant-autoscaler/crds/` — used by Helm

Previously, `make manifests` only updated `config/crd/bases/`. The Helm chart CRD had to be manually synced, which is easy to forget and leads to silent drift — the Helm-deployed cluster gets a stale schema while the kustomize path gets the correct one.

## What

Adds a `cp` step to the `manifests` target so both files are always updated together in a single `make manifests` run.

## Impact

No functional change to the CRD schema itself. Pure tooling fix.


asm582 (commented):
/lgtm
/approve

PR #931, opened 2026-03-24
Update go to 1.25.0

shuynh2017:
https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/929


.github/workflows/ci-e2e-openshift.yaml
mamy-CS: Adding `GOTOOLCHAIN: auto `here should solve the openshift issue I believe. Currently set to local

mamy-CS (changes requested):

mamy-CS (approved):
/lgtm

PR #942, opened 2026-03-27
test/e2e: migrate from Ensure* to Create* pattern for test isolation

### LINKED ISSUE #939: e2e: limit/remove Ensure* helpers to avoid cross-test resource mutation illusion
We should reduce the use of EnsureService / EnsureXX helpers in test/e2e/fixtures and specs. Audit current usage of Ensure* in `test/e2e/fixtures/` and in `specs`.
- Prefer Create* everywhere the spec intends single-owner control.
- Guarantee isolation via unique names per spec (and/or labels) instead of delete/recreate.
- If we still need leftover from previous run handling, keep it centralized (BeforeSuite cleanup) rather than per-spec Ensure*.

**Why this matters**
- The current Ensure* pattern (delete/recreate to make setup idempotent) can create the wrong mental model that multiple independent ginkgo specs safely manipulate the same shared resources.
- Even if tests are not parallel today, this can become flaky or misleading later
- Deterministic correctness e2e should maintain a strong isolation story (one spec owns one resource set)

**Acceptance criteria**
- Every spec’s setup should clearly reflect ownership (no delete/recreate ensure semantics unless explicitly justified and documented).
- If two specs could ever collide (names/labels), they must use unique identities rather than relying on Ensure*.
This will improve determinism and clarity of the e2e suite by making the isolation model explicit.

sats-23:
Fixes #939 

Replace delete-and-recreate Ensure* helpers with fail-if-exists Create* functions across all e2e test files. This eliminates the illusion that multiple tests can safely share resources and establishes clear single- owner semantics for test resources.

Key changes:
- Introduce utils.NewTestResourceNames() for unique resource naming
- Migrate 59 Ensure* calls to Create* across 7 test files
- Consolidate cleanup from scattered DeferCleanup to AfterAll blocks
- Enhance BeforeSuite to clean leftover resources from interrupted runs
- Add comprehensive documentation and migration guides

Each test now owns uniquely-named resources, preventing cross-test mutations and enabling future parallel test execution. The only remaining Ensure* calls are for transient load generation utilities, which are documented as justified exceptions.

### DIFF
--- a/docs/developer-guide/e2e-justified-ensure-usage.md
+++ b/docs/developer-guide/e2e-justified-ensure-usage.md
@@ -1,0 +1,209 @@
+# Justified Ensure* Usage in E2E Tests
+
+This document catalogs the remaining `Ensure*` function usage in the e2e test suite and provides justification for each case.
+
+## Overview
+
+As part of the test isolation refactoring effort, we've established that `Create*` functions with unique resource names should be preferred over `Ensure*` functions. However, certain scenarios justify the use of `Ensure*` semantics.
+
+## Justification Criteria
+
+`Ensure*` usage is justified when:
+
+1. **Testing Recovery Behavior**: The test explicitly validates controller recovery when resources are deleted/recreated
+2. **Shared Test Infrastructure**: Resources are truly shared across all tests (e.g., namespace-level configuration)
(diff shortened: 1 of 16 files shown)


sats-23:
Hey @mamy-CS 
PTAL at the PR

Issue #809, opened 2026-02-28
[Phase 2 - API bump candidate] Remove `Accelerator` from Status


ev-shindin:
**Scope:** CRD types, controller, actuator, metrics

Remove `Accelerator` field from `OptimizedAlloc` — it duplicates the VA label `inference.optimization/acceleratorName`.

```go
// Before:
type OptimizedAlloc struct {
    LastRunTime metav1.Time `json:"lastRunTime,omitempty"`
    Accelerator string      `json:"accelerator"`       // REMOVE
    NumReplicas int         `json:"numReplicas"`
}

// After:
type OptimizedAlloc struct {
    LastRunTime metav1.Time `json:"lastRunTime,omitempty"`
    NumReplicas *int32      `json:"numReplicas,omitempty"`  // Pointer, see Issue 4
}
```

**Changes required:**
- `api/v1alpha1/variantautoscaling_types.go` — Remove `Accelerator` field from `OptimizedAlloc`
- `internal/actuator/actuator.go` — Read accelerator from VA label instead of `Status.DesiredOptimizedAlloc.Accelerator` for metric emission
- `internal/controller/variantautoscaling_controller.go` — Remove accelerator assignment in status update: `va.Status.DesiredOptimizedAlloc.Accelerator = ...`
- `internal/engines/saturation/engine.go` — Remove code that writes accelerator to decision/status
- `internal/metrics/metrics.go` — Read accelerator type from VA label `inference.optimization/acceleratorName`
- CRD manifests regeneration
- E2E test fixtures — Update VA builder and assertion helpers

**Backward compatibility:**
- API version bump to `v1alpha2` recommended (breaking change to status schema)
- Alternatively, keep field with deprecation annotation in v1alpha1 and stop populating it

vivekk16:
Yes please @shuynh2017

PR #955, opened 2026-03-30
Chore: Deprecate desiredOptimizedAlloc.accelerator

shuynh2017:
Requirements: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/809
Proposal: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/809#issuecomment-4155097077
- We will first mark as deprecated (soft deprecation) with message what to use instead of this deprecated field.

> Deprecating a required field in a Kubernetes CRD with Kubebuilder involves a multi-step process for backward compatibility, primarily by relaxing the requirement in the current version and marking it as deprecated, or by introducing a new API version

- Once this PR is approved, we open an issue to remove it (breaking change) for the next API version bump.

Notes:
- Kubebuilder has this annotation `// +kubebuilder:deprecatedversion:warning=`, but this is for the whole api, not for just a field.
- The annotation `Deprecated` used in this PR is according to recommended practice.
- The only visible change to users is:
```
$ kubectl explain VariantAutoscaling.status.desiredOptimizedAlloc
GROUP:      llmd.ai
KIND:       VariantAutoscaling
VERSION:    v1alpha1

FIELD: desiredOptimizedAlloc <Object>


DESCRIPTION:
    DesiredOptimizedAlloc indicates the target optimized allocation based on
    autoscaling logic.

FIELDS:
  accelerator   <string>
    Accelerator is the type of accelerator for the optimized allocation.
    Deprecated: This field is deprecated and will be removed in a future                 <=====
    version. Use node selector or node affinity from scale target instead.

  lastRunTime   <string>
    LastRunTime is the timestamp of the last optimization run.

  numReplicas   <integer>
    NumReplicas is the number of replicas for the optimized allocation.
    nil means no optimization decision has been made yet.
```


ev-shindin (changes requested):
Look at my comment. Also ensure that both CRD YAML files are generated by `make manifests` from the Go types + kubebuilder markers.

ev-shindin (dismissed):

ev-shindin (approved):
Great work! Thanks!

PR #967, opened 2026-04-02
opt in for keda, env specific

mamy-CS:
treat KEDA as already provided by the cluster on generic Kubernetes and openshift, and only verify it’s there instead of installing it by default.
Opt-in: use an env flag when you want the script to install or upgrade KEDA with Helm

### DIFF
--- a/test/e2e/config.go
+++ b/test/e2e/config.go
@@ -10,7 +10,9 @@ type E2EConfig struct {
 	testconfig.SharedConfig
 
 	// Feature gates
-	ScaleToZeroEnabled bool // HPAScaleToZero feature gate
+	// ScaleToZeroEnabled: env SCALE_TO_ZERO_ENABLED — assume native HPA may use minReplicas=0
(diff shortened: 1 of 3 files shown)


test/e2e/config.go
```diff
@@ -56,7 +56,7 @@ func LoadConfigFromEnv() E2EConfig {
 	// OpenShift clusters typically don't have the HPAScaleToZero feature gate
 	// enabled, so attempting to create HPAs with minReplicas=0 will fail with:
 	//   "spec.minReplicas: Invalid value: 0: must be greater than or equal to 1"
-	// Override the env var to prevent test failures on OpenShift.
+	// Clear SCALE_TO_ZERO_ENABLED here so e2e uses SCALER_BACKEND=keda (ScaledObject) for scale-from-zero.
```
lionelvillard: ~this comment is wrong. Please remove.~

Let me rephrase: scale from zero is possible even in a cluster without HPAScaleToZero and KEDA. Scale to zero is not possible.

PR #971, opened 2026-04-02
upgrade infra versions

mamy-CS:
- Aligns Kind/e2e installs with published llm-d-inference-scheduler v0.7.x (and matching routing sidecar) when using llm-d guide charts from LLM_D_RELEASE, without building custom EPP images.
- v0.3.0 -> v0.5.1 for LLM_D_RELEASE.

Note: helm upgrade is the right long term mechanism once we’re on a chart+values that emit v0.7-safe args for example and flowControl in config. upgrading without changing chart/version mostly repeats the same broken yaml. These patches are the bridge until that chart lands.
```

Ran 60 of 63 Specs in 1018.360 seconds
SUCCESS! -- 60 Passed | 0 Failed | 0 Pending | 3 Skipped
--- PASS: TestE2E (1018.36s)
PASS
ok      github.com/llm-d/llm-d-workload-variant-autoscaler/test/e2e     1019.030s

```

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -414,6 +414,8 @@ jobs:
       MODEL_B_RELEASE: model-b-${{ github.run_id }}
       # Use the image built in the previous job
       WVA_IMAGE_TAG: ${{ needs.build-image.outputs.image_tag }}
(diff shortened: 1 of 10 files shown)


lionelvillard:
not needed anymore

PR #979, opened 2026-04-03
fix: pass SCALER_BACKEND=keda to OpenShift E2E test step

### LINKED ISSUE #978: e2e flaky query external metric tests
Environment: OpenShift, full E2E (make test-e2e-full).
- Three specs fail with the same pattern: `“Should be able to query external metrics API” → server error Code: 500, Reason: InternalError, message like “an error on the server ("unknown") has prevented the request from succeeding”.`
- Affected areas (from run): Basic VA lifecycle, Basic VA lifecycle with LeaderWorkerSet, Basic VA lifecycle with LeaderWorkerSet (single-node).
- Model decode deployments (e.g. `ms-inference-scheduling-llm-d-modelservice-decode`) can time out or stay non-ready.
- Diagnostics showed ImagePullBackOff / ErrImagePull for `ghcr.io/llm-d/llm-d-cuda:v0.6.0` (manifest unknown).

here most recent run: https://github.com/llm-d/llm-d-workload-variant-autoscaler/actions/runs/23950871105/job/69858041963

cc: @shuynh2017

clubanderson:
## Summary
- The CI deploys infrastructure with `SCALER_BACKEND: keda` (line 697), but the "Run OpenShift E2E tests" step did not pass this env var to the tests
- Without it, the test defaults to `prometheus-adapter` and queries the external metrics API directly — KEDA serves that API and returns 500 for unknown metrics instead of 404
- The test only tolerates `NotFound` (404), so it fails with `InternalError` (500)
- With `SCALER_BACKEND=keda` set, the test takes the KEDA-aware branch and checks for `ScaledObject` resources instead

Fixes #978

## Test plan
- [ ] OpenShift E2E CI passes (the three LWS specs should no longer 500)
- [ ] Verify `SCALER_BACKEND: keda` appears in the CI log output

### DIFF
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -1046,6 +1046,8 @@ jobs:
           LLMD_NAMESPACE_B: ${{ env.LLMD_NAMESPACE_B }}
           GATEWAY_NAME: infra-inference-scheduling-inference-gateway-istio
           DEPLOYMENT: ms-inference-scheduling-llm-d-modelservice-decode
+          # Must match the scaler backend used during infrastructure deployment
+          SCALER_BACKEND: keda
           # Pass WVA_RELEASE_NAME so test can filter for current run's resources
           WVA_RELEASE_NAME: ${{ env.WVA_RELEASE_NAME }}
           # Controller instance label must match what the controller was deployed with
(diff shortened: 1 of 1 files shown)


mamy-CS (approved):
/lgtm, maybe worth checking Copilot review in setting SCALER_BACKEND: keda once in the workflow

PR #981, opened 2026-04-05
Add optimization loop performance metrics

### LINKED ISSUE #914: Optimization Loop Performance Metrics
Add timing and throughput metrics for the optimization loop.

**Metrics:**
```
wva_optimization_duration_seconds   histogram {status}
wva_models_processed_total          counter   {}
```

**Implementation:**
- Wrap `optimize()` with `time.Since(start)` observation
- Status labels: `success`, `error`, `partial`
- Histogram buckets: `{0.01, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}` seconds
- Increment `wva_models_processed_total` for each model processed in a cycle

**Acceptance Criteria:**
- [ ] Duration histogram emitted per optimization cycle
- [ ] Models-processed counter incremented correctly
- [ ] Unit tests verify histogram observation and counter increment

jia-gao:
## Summary

Add timing and throughput metrics for the optimization loop. Closes #914.

### New Metrics

| Metric | Type | Labels | Description |
|--------|------|--------|-------------|
| `wva_optimization_duration_seconds` | Histogram | `status` | Duration of each optimization cycle |
| `wva_models_processed_total` | Counter | — | Total models processed across cycles |

Histogram buckets: `{0.01, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}` seconds

Status labels: `success`, `error`

### Implementation

- `internal/constants/metrics.go` — New metric name and label constants
- `internal/metrics/metrics.go` — Register histogram + counter in `InitMetrics()`, add `ObserveOptimizationDuration()` and `IncrModelsProcessed()` methods on `MetricsEmitter`
- `internal/engines/saturation/engine.go` — Instrument `optimize()` with a deferred timing observation that captures duration and status (success/error), and tracks models processed per cycle
- `internal/metrics/optimization_metrics_test.go` — Unit tests

### Tests

```
=== RUN   TestObserveOptimizationDuration
--- PASS: TestObserveOptimizationDuration (0.00s)
=== RUN   TestIncrModelsProcessed
--- PASS: TestIncrModelsProcessed (0.00s)
=== RUN   TestObserveOptimizationDuration_NilSafety
--- PASS: TestObserveOptimizationDuration_NilSafety (0.00s)
PASS
```

### Acceptance Criteria (from issue)

- [x] Duration histogram emitted per optimization cycle
- [x] Models-processed counter incremented correctly
- [x] Unit tests verify histogram observation and counter increment

### DIFF
--- a/internal/constants/metrics.go
+++ b/internal/constants/metrics.go
@@ -113,8 +113,8 @@ const (
 	// Labels: status (success, error)
 	WVAOptimizationDurationSeconds = "wva_optimization_duration_seconds"
 
-	// WVAModelsProcessedTotal is a gauge that tracks the number of models processed in the last optimization cycle.
-	WVAModelsProcessedTotal = "wva_models_processed_total"
+	// WVAModelsProcessed is a gauge that tracks the number of models processed in the last optimization cycle.
+	WVAModelsProcessed = "wva_models_processed"
 )
 
 // Metric Label Names
--- a/internal/constants/metrics.go
+++ b/internal/constants/metrics.go
@@ -108,6 +108,13 @@ const (
 	// WVADesiredRatio is a gauge that tracks the ratio of desired to current replicas.
 	// Labels: variant_name, namespace, accelerator_type
 	WVADesiredRatio = "wva_desired_ratio"
+
+	// WVAOptimizationDurationSeconds is a histogram that tracks the duration of each optimization cycle.
+	// Labels: status (success, error, partial)
+	WVAOptimizationDurationSeconds = "wva_optimization_duration_seconds"
+
+	// WVAModelsProcessedTotal is a counter that tracks the total number of models processed across optimization cycles.
+	WVAModelsProcessedTotal = "wva_models_processed_total"
 )
 
 // Metric Label Names
@@ -120,4 +127,5 @@ const (
 	LabelReason             = "reason"
 	LabelAcceleratorType    = "accelerator_type"
 	LabelControllerInstance = "controller_instance"
+	LabelStatus             = "status"
 )
--- a/internal/constants/metrics.go
+++ b/internal/constants/metrics.go
@@ -110,10 +110,10 @@ const (
 	WVADesiredRatio = "wva_desired_ratio"
 
 	// WVAOptimizationDurationSeconds is a histogram that tracks the duration of each optimization cycle.
-	// Labels: status (success, error, partial)
+	// Labels: status (success, error)
 	WVAOptimizationDurationSeconds = "wva_optimization_duration_seconds"
 
-	// WVAModelsProcessedTotal is a counter that tracks the total number of models processed across optimization cycles.
+	// WVAModelsProcessedTotal is a gauge that tracks the number of models processed in the last optimization cycle.
 	WVAModelsProcessedTotal = "wva_models_processed_total"
 )
 
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -190,9 +190,7 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 			status = "error"
 		}
 		metrics.ObserveOptimizationDuration(time.Since(start).Seconds(), status)
-		if modelsProcessed > 0 {
-			metrics.SetModelsProcessed(modelsProcessed)
-		}
+		metrics.SetModelsProcessed(modelsProcessed)
 	}()
 
 	logger := ctrl.LoggerFrom(ctx)
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -315,10 +315,7 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 	} else {
 		logger.Info("No scaling decisions to apply, updating VA status with metrics")
 	}
-	if err := e.applySaturationDecisions(ctx, allDecisions, vaMap, currentAllocations); err != nil {
-		logger.Error(err, "Failed to apply saturation decisions")
-		return err
-	}
+	e.applySaturationDecisions(ctx, allDecisions, vaMap, currentAllocations)
 
 	logger.Info("Optimization completed successfully",
 		"mode", "saturation-only",
@@ -889,7 +886,7 @@ func (e *Engine) applySaturationDecisions(
 	decisions []interfaces.VariantDecision,
 	vaMap map[string]*llmdVariantAutoscalingV1alpha1.VariantAutoscaling,
 	currentAllocations map[string]*interfaces.Allocation,
-) error {
+) {
 	logger := ctrl.LoggerFrom(ctx)
 	// Create a map of decisions for O(1) lookup
 	// Use namespace/variantName as key to match vaMap and avoid collisions
@@ -1120,8 +1117,6 @@ func (e *Engine) applySaturationDecisions(
 				"reason", reason)
 		}
 	}
-
-	return nil
 }
 
 // emitSafetyNetMetrics emits fallback metrics when saturation analysis fails.
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -103,9 +103,6 @@ type Engine struct {
 	// AnalyzerResults. Selected per-cycle based on enableLimiter config:
 	// CostAwareOptimizer (unlimited) or GreedyByScoreOptimizer (limited).
 	optimizer pipeline.ScalingOptimizer
-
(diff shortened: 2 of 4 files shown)


jia-gao:
Note on the `status` label: The issue spec mentions `partial` as a possible status value. The current implementation only emits `success` or `error` because `optimize()` either completes fully or returns an error — there's no partial completion path today.

If a `partial` status is needed in the future (e.g., when some model groups succeed but others fail within a single cycle), it can be added without breaking the histogram schema since it's just a new label value.

internal/metrics/metrics.go
```diff
@@ -76,6 +79,21 @@ func InitMetrics(registry prometheus.Registerer) error {
 		baseLabels,
 	)
 
+	optimizationDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    constants.WVAOptimizationDurationSeconds,
+			Help:    "Duration of optimization loop cycles in seconds",
+			Buckets: []float64{0.01, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
+		},
+		[]string{constants.LabelStatus},
```
ev-shindin: All existing WVA metrics conditionally add a `controller_instance` label when the `CONTROLLER_INSTANCE` env var is set (see `InitMetrics` lines 45-48 where `baseLabels` and `scalingLabels` get the extra label appended). The new `optimizationDuration` histogram only has `LabelStatus` and `modelsProcessedTotal` has no labels at all. In HA deployments with multiple controller instances, these metrics will collide.

internal/engines/saturation/engine.go
```diff
@@ -102,6 +103,9 @@ type Engine struct {
 	// AnalyzerResults. Selected per-cycle based on enableLimiter config:
 	// CostAwareOptimizer (unlimited) or GreedyByScoreOptimizer (limited).
 	optimizer pipeline.ScalingOptimizer
+
+	// metricsEmitter emits optimization loop performance metrics (duration, models processed).
+	metricsEmitter *metrics.MetricsEmitter
```
ev-shindin: The `Engine` struct didn't previously have a `metricsEmitter` — the existing `MetricsEmitter` is used downstream in `applySaturationDecisions`. Adding a second instance creates two separate entry points for metric emission. Since `MetricsEmitter` is a stateless empty struct backed by package-level vars, this works but is a bit confusing. Consider either:
- Package-level helper functions (no emitter instance needed), or
- Receiving the emitter as a `NewEngine` parameter so there's a single instance

internal/metrics/metrics.go
```diff
@@ -133,6 +157,23 @@ func (m *MetricsEmitter) EmitReplicaScalingMetrics(ctx context.Context, va *llmd
 	return nil
 }
 
+// ObserveOptimizationDuration records the duration of an optimization cycle with the given status.
+// Status should be one of: "success", "error", "partial".
```
ev-shindin: The comment documents `partial` as a valid status, but the engine code only ever emits `success` or `error`. Either remove `partial` from the comment until it's implemented, or add a note that it's reserved for future use — otherwise someone might grep for `"partial"` and be confused when it's never emitted.

ev-shindin (changes requested):
Please address review comments

internal/metrics/metrics.go
jia-gao: Good catch, missed this pattern. Added controller_instance conditionally to the histogram labels, same as the existing metrics. The gauge (modelsProcessedGauge) is a singleton (no labels) so it won't collide in HA — it
 reflects the local controller's last cycle.

internal/metrics/metrics.go
```diff
@@ -76,6 +79,25 @@ func InitMetrics(registry prometheus.Registerer) error {
...
+	}
+	optimizationDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    constants.WVAOptimizationDurationSeconds,
+			Help:    "Duration of optimization loop cycles in seconds",
+			Buckets: []float64{0.01, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10},
+		},
+		optimizationDurationLabels,
+	)
+	modelsProcessedGauge = prometheus.NewGauge(
```
ev-shindin: The `optimizationDuration` histogram correctly gets the `controller_instance` label when in HA mode, but `modelsProcessedGauge` is a plain `prometheus.Gauge` with no labels at all. In an HA deployment with multiple controller instances, this gauge will collide across instances — you won't be able to distinguish which controller reported the count.

It should be a `*prometheus.GaugeVec` with a conditional `controller_instance` label, matching the pattern used for `optimizationDuration` and all other metrics in the file

internal/constants/metrics.go
```diff
@@ -108,6 +108,13 @@ const (
 	// WVADesiredRatio is a gauge that tracks the ratio of desired to current replicas.
 	// Labels: variant_name, namespace, accelerator_type
 	WVADesiredRatio = "wva_desired_ratio"
+
+	// WVAOptimizationDurationSeconds is a histogram that tracks the duration of each optimization cycle.
+	// Labels: status (success, error, partial)
+	WVAOptimizationDurationSeconds = "wva_optimization_duration_seconds"
+
+	// WVAModelsProcessedTotal is a counter that tracks the total number of models processed across optimization cycles.
+	WVAModelsProcessedTotal = "wva_models_processed_total"
```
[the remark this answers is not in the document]
ev-shindin: Still be "_total". Please fix.

internal/engines/saturation/engine.go
```diff
@@ -180,7 +181,20 @@ func (e *Engine) StartOptimizeLoop(ctx context.Context) {
...
+func (e *Engine) optimize(ctx context.Context) (retErr error) {
+	start := time.Now()
+	var modelsProcessed int
+	defer func() {
+		status := "success"
+		if retErr != nil {
+			status = "error"
+		}
+		metrics.ObserveOptimizationDuration(time.Since(start).Seconds(), status)
+		if modelsProcessed > 0 {
```
ev-shindin: When `len(activeVAs) == 0`, the function returns early (line 210 in current main) with `return nil` — before `modelGroups` is computed. The defer fires with `modelsProcessed = 0` and `status = "success"`, but `SetModelsProcessed` is skipped because of the `if modelsProcessed > 0` guard in the defer block. This means the gauge retains its stale previous value. If the system had 5 models last cycle and now has 0, the gauge still shows 5.

Consider calling `SetModelsProcessed(0)` unconditionally (remove the `> 0` guard), or always set it in the defer regardless.

internal/metrics/optimization_metrics_test.go
```diff
@@ -0,0 +1,144 @@
+/*
+Copyright 2025 The llm-d Authors
```
ev-shindin: We are in 2026 :)

ev-shindin (approved):
Good progress. Please address review comments.

ev-shindin (changes requested):
Please address review comments

internal/metrics/metrics.go
jia-gao: Good catch. Converted `modelsProcessedGauge` to `*prometheus.GaugeVec` with a conditional `controller_instance` label, matching the `optimizationDuration` pattern. In non-HA mode the label slice is empty; in HA mode the label is appended, so instances will not collide.

internal/constants/metrics.go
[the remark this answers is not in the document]
jia-gao: Fixed. Renamed the constant to `WVAModelsProcessed` and the metric name to `wva_models_processed` — since it is a gauge, the `_total` suffix was wrong per Prometheus naming conventions. Updated all references.

internal/engines/saturation/engine.go
jia-gao: Removed it, so the defer now always calls `SetModelsProcessed(modelsProcessed)`. On the `len(activeVAs) == 0` early return, `modelsProcessed` is still its zero value, so the gauge correctly drops to 0 instead of retaining the stale previous cycle count.

internal/metrics/optimization_metrics_test.go
jia-gao: Ha, fixed 🙂

jia-gao:
> Please address review comments

thanks for the comments! fixed them

ev-shindin (approved):
Great Work!

ev-shindin:
@jia-gao Please rebase this PR.

jia-gao:
Rebased onto latest `main` (new tip `c3f36af`) — no conflicts. Ready for re-review @ev-shindin, thanks!

PR #990, opened 2026-04-12
docs: Document acceleratorName resolution on VariantAutoscaling

### LINKED ISSUE #855: acceleratorName label requirement on VariantAutoscaling is not documented
## Summary

The `inference.optimization/acceleratorName` label on `VariantAutoscaling` resources is a functional requirement for WVA — without it, WVA cannot build allocations, skips status updates, and never emits `wva_desired_replicas` metrics. However, this requirement is not documented anywhere. Users creating VAs without Helm have no way to know this label is needed without reading WVA source code.

## Impact

Without the label, WVA silently fails across multiple code paths:

1. **`BuildAllocationFromMetrics`** (`internal/utils/allocation.go:31-35`) — Returns a hard error, so WVA cannot build an allocation for the VA:
   ```go
   if val, ok := va.Labels["inference.optimization/acceleratorName"]; ok && val != "" {
       acc = val
   } else {
       return interfaces.Allocation{},
           fmt.Errorf("missing or empty acceleratorName label on VariantAutoscaling object: %s", va.Name)
   }
   ```

2. **Saturation engine** (`internal/engines/saturation/engine.go:883-887`) — Skips status update with `MetricsAvailable=False`, so no `DesiredOptimizedAlloc` is set.

3. **Metric emission / safety net** (`internal/engines/saturation/engine.go:1071-1074`) — Skips metric emission entirely, so HPA/KEDA never receives a scaling signal:
   ```
   "Safety net: skipping metric emission - no accelerator name available"
   ```

The net result is that a VA without this label is accepted but effectively does nothing.

## Documentation gaps

- The label appears in some YAML examples (`deploy/README.md`, `docs/tutorials/demo.md`, `config/samples/variantautoscaling-integration.yaml`) but is never explicitly called out as required.
- The CRD reference (`docs/user-guide/crd-reference.md`) only documents `scaleTargetRef`, `modelID`, and `variantCost`. No mention of required labels.
- The Helm chart silently handles this by injecting the label from `.Values.va.accelerator`, so Helm users don't encounter the issue.

## Existing TODO in code

There is already a TODO in the codebase acknowledging this should become a CRD field:

```go
// internal/engines/saturation/engine.go:1057-1058
// TODO: remove this checks when we will move to a new version of the CRD
// with required accelerator field
```

Until that CRD change happens, the label remains the only mechanism and should be documented accordingly.

## How this was discovered

While modifying a controller (KServe LLMISVC) that creates `VariantAutoscaling` resources programmatically, the label requirement was not apparent from the CRD types, API documentation, or sample YAMLs. It was only discovered by tracing the runtime error in WVA source code.

## Suggestion

Document the `inference.optimization/acceleratorName` label as a **required** label in the documentation, configuration guide, and `deploy/README.md`.

jia-gao:
## Summary

Document how WVA resolves the accelerator name for a `VariantAutoscaling`. Closes #855.

## Background

After #882 merged, WVA resolves the accelerator name in two steps:

1. **Auto-discovery** from the target Deployment/LWS pod template by reading `nodeSelector` / `nodeAffinity` for GPU product keys:
   - `nvidia.com/gpu.product`
   - `amd.com/gpu.product-name`
   - `cloud.google.com/gke-accelerator`
2. **Fallback** to the `inference.optimization/acceleratorName` label on the VA

If **both** fail, WVA silently skips status updates and metric emission for the variant — HPA/KEDA never receives a scaling signal. The failure is silent at the API level (the VA is accepted), but the controller log contains `accelerator name not found in scale target nodeSelector/nodeAffinity or VA label`.

This resolution logic was added in #882 but was never explicitly documented in user-facing docs, so users who hit the failure mode had no way to diagnose or fix it without reading source. Thanks to @ev-shindin for the #882 context on the issue thread.

## Changes

- **`docs/user-guide/configuration.md`** — New "Accelerator Name Resolution" subsection under "Enabling Autoscaling for a Model Deployment". Explains the two-step lookup, the failure log signature, the Helm note, and includes a complete manual YAML example.

- **`docs/user-guide/troubleshooting.md`** — New section "Accelerator Name Cannot Be Resolved" under "Why is my VariantAutoscaling not being reconciled?" with:
  - `kubectl` commands to inspect the scale target `nodeSelector`
  - `kubectl` command to inspect the VA fallback label
  - Two log signatures users will see
  - Resolution via either pinning GPU type on the Deployment **or** setting the label

- **`deploy/README.md`** — Info callout on the existing manual VA creation example explaining the resolution order. Links to the new configuration.md section.

- **`config/samples/variantautoscaling-integration.yaml`** — Inline comment annotating the label as a fallback, not a hard requirement, with the auto-discovery keys spelled out.

## Out of scope

- `docs/user-guide/crd-reference.md` is generated from CRD types (`CRD_OUTPUT` in the Makefile) and cannot be edited directly.
- The existing `TODO: remove this checks when we will move to a new version of the CRD with required accelerator field` at `internal/engines/saturation/engine.go` still applies as a future direction but is out of scope for this doc-only fix.

## Note on earlier framing

An earlier revision of this PR framed the label as **required**. That was accurate pre-#882 and matched the original issue description, but after #882 the label is a fallback — auto-discovery takes priority. All text has been rewritten to reflect the current behavior.

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -58,7 +58,7 @@ Every `VariantAutoscaling` must resolve to an **accelerator name** (e.g., `A100`
 
 2. **`inference.optimization/acceleratorName` label on the VariantAutoscaling** (fallback). Used if auto-discovery fails or the scale target has no GPU node selector.
 
-If **both** fail, WVA cannot build an allocation and will skip status updates and metric emission for the variant — HPA/KEDA will never receive a scaling signal. The failure is silent at the API level (the VA is accepted), but WVA controller logs contain:
+If **both** fail, WVA cannot build a full allocation. It will still set the `MetricsAvailable=False` condition on the VA (visible via `kubectl get va <name> -o yaml`), but it skips the full target-replicas calculation and does not emit scaling metrics — HPA/KEDA will never receive a scaling signal. The WVA controller logs contain:
 
 ```
 accelerator name not found in scale target nodeSelector/nodeAffinity or VA label
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -45,6 +45,47 @@ Choose between the following approaches:
 - [With HPA](hpa-integration.md) - Use Kubernetes HPA for autoscaling based on WVA's custom metrics
 - [With KEDA](keda-integration.md) - Use KEDA for autoscaling based on WVA's custom metrics
 
+### Accelerator Name Resolution
+
+Every `VariantAutoscaling` must resolve to an **accelerator name** (e.g., `A100`, `H100`, `L40S`) so WVA can build allocations, match against cluster inventory, and label emitted Prometheus metrics. WVA resolves the accelerator name in the following order:
+
+1. **Auto-discovery from the scale target** (preferred). WVA inspects the target `Deployment`/`LeaderWorkerSet` pod template for these `nodeSelector` / `nodeAffinity` keys:
+   - `nvidia.com/gpu.product`
+   - `amd.com/gpu.product-name`
+   - `cloud.google.com/gke-accelerator`
+
+   The first matching value is used.
+
+2. **`inference.optimization/acceleratorName` label on the VariantAutoscaling** (fallback). Used if auto-discovery fails or the scale target has no GPU node selector.
+
+If **both** fail, WVA cannot build an allocation and will skip status updates and metric emission for the variant — HPA/KEDA will never receive a scaling signal. The failure is silent at the API level (the VA is accepted), but WVA controller logs contain:
+
+```
+accelerator name not found in scale target nodeSelector/nodeAffinity or VA label
+```
+
+> **Helm users:** The chart injects the label automatically from `.Values.va.accelerator` (used as a fallback) but the comment in `values.yaml` notes it is optional — auto-discovery takes priority.
+
+**Recommendation:** For deployments that pin GPU type via `nodeSelector` or `nodeAffinity`, auto-discovery works out of the box. For deployments without GPU node selectors (e.g., simulators, tests, or shared clusters), set the label explicitly:
+
+```yaml
+apiVersion: llmd.ai/v1alpha1
+kind: VariantAutoscaling
+metadata:
+  name: my-variant
+  namespace: llm-d
+  labels:
+    # Fallback used if the target Deployment has no nvidia.com/gpu.product
+    # (or equivalent AMD/GKE) nodeSelector/nodeAffinity entry.
+    inference.optimization/acceleratorName: A100
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: my-vllm-deployment
+  modelID: "meta-llama/Llama-3.1-8B"
+```
+
 ## Operating Mode
 
 WVA operates in **saturation mode**.
--- a/docs/user-guide/troubleshooting.md
+++ b/docs/user-guide/troubleshooting.md
@@ -154,7 +154,67 @@ Fix the validation errors and reapply the resource.
 
 ---
 
-### 6. Controller Not Running
+### 6. Accelerator Name Cannot Be Resolved
+
+WVA needs to determine the accelerator (GPU) type for each `VariantAutoscaling`. It resolves this in two steps:
+
+1. **Auto-discovery** from the target `Deployment`/`LeaderWorkerSet` pod template by reading `nodeSelector` / `nodeAffinity` keys: `nvidia.com/gpu.product`, `amd.com/gpu.product-name`, or `cloud.google.com/gke-accelerator`.
+2. **Fallback** to the `inference.optimization/acceleratorName` label on the VA.
+
+If **both** fail, WVA silently skips status updates and metric emission for this variant — HPA/KEDA never receives a scaling signal. The VA appears healthy (no error events), but the resource effectively does nothing.
+
+**Check the scale target's node selector:**
+
+```bash
+# Get the scale target name from the VA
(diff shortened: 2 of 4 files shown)


docs/user-guide/troubleshooting.md
```diff
@@ -154,7 +154,67 @@ Fix the validation errors and reapply the resource.
...
+**Check the scale target's node selector:**
+
+```bash
+# Get the scale target name from the VA
+TARGET=$(kubectl get va <va-name> -n <namespace> -o jsonpath='{.spec.scaleTargetRef.name}')
+
+# Inspect the deployment's nodeSelector for any GPU product keys
+kubectl get deployment "$TARGET" -n <namespace> \
+  -o jsonpath='{.spec.template.spec.nodeSelector}'
+```
```
ev-shindin, line 186: The diagnostic command above checks `nodeSelector` only.
But the code at `internal/utils/utils.go:387-391` also checks `nodeAffinity` (both `requiredDuringScheduling` and `preferredDuringScheduling`). A deployment might use affinity rules instead of nodeSelector. Consider adding:

```bash
# Also check nodeAffinity for GPU product keys
kubectl get deployment "$TARGET" -n <namespace> \
  -o jsonpath='{.spec.template.spec.affinity.nodeAffinity}'

File: docs/user-guide/troubleshooting.md [hunk 2]
```diff
@@ -154,7 +154,67 @@ Fix the validation errors and reapply the resource.
...
 ---
 
-### 6. Controller Not Running
+### 6. Accelerator Name Cannot Be Resolved
+
+WVA needs to determine the accelerator (GPU) type for each `VariantAutoscaling`. It resolves this in two steps:
+
+1. **Auto-discovery** from the target `Deployment`/`LeaderWorkerSet` pod template by reading `nodeSelector` / `nodeAffinity` keys: `nvidia.com/gpu.product`, `amd.com/gpu.product-name`, or `cloud.google.com/gke-accelerator`.
+2. **Fallback** to the `inference.optimization/acceleratorName` label on the VA.
+
```
ev-shindin, line 163: The code at `internal/engines/saturation/engine.go:977-989` shows that when `acceleratorName == ""`, WVA does **not** completely skip status updates — it still sets the `MetricsAvailable=False` condition via `DecisionCache.Set()` with `MetricsAvailable: false, MetricsReason: ReasonMetricsMissing`. It only skips the full allocation/target-replicas update.

You say *"The VA appears healthy (no error events)"* — while this technically true (no k8s events), the `MetricsAvailable` condition will be `False`, which is observable via `kubectl get va <name> -o yaml`. This is actually useful diagnostic info that should be mentioned.

docs/user-guide/troubleshooting.md
```diff
@@ -154,7 +154,67 @@ Fix the validation errors and reapply the resource.
...
+Either pin GPU type via the target deployment's `nodeSelector`:
+
+```yaml
+# Deployment spec.template.spec.nodeSelector
+nodeSelector:
+  nvidia.com/gpu.product: A100
+```
+
+or set the label on the VA as an explicit override:
```
ev-shindin: The label does NOT override auto-discovery — it's only used when auto-discovery fails. Should read "as a fallback" for consistency.

docs/user-guide/configuration.md
```diff
@@ -45,6 +45,47 @@ Choose between the following approaches:
...
+1. **Auto-discovery from the scale target** (preferred). WVA inspects the target `Deployment`/`LeaderWorkerSet` pod template for these `nodeSelector` / `nodeAffinity` keys:
+   - `nvidia.com/gpu.product`
+   - `amd.com/gpu.product-name`
+   - `cloud.google.com/gke-accelerator`
+
+   The first matching value is used.
+
+2. **`inference.optimization/acceleratorName` label on the VariantAutoscaling** (fallback). Used if auto-discovery fails or the scale target has no GPU node selector.
+
+If **both** fail, WVA cannot build an allocation and will skip status updates and metric emission for the variant — HPA/KEDA will never receive a scaling signal. The failure is silent at the API level (the VA is accepted), but WVA controller logs contain:
```
ev-shindin: The code at `internal/engines/saturation/engine.go:977-989` shows that when `acceleratorName == ""`, WVA does **not** completely skip status updates — it still sets the `MetricsAvailable=False` condition via `DecisionCache.Set()` with `MetricsAvailable: false, MetricsReason: ReasonMetricsMissing`. It only skips the full allocation/target-replicas update.

ev-shindin (changes requested):
Good documentation PR that fills a real gap — accelerator resolution was undocumented despite being a common source of silent failures. Please address review comments.

docs/user-guide/troubleshooting.md
jia-gao: Good catch — added a separate `kubectl get deployment` command for `affinity.nodeAffinity` alongside the existing `nodeSelector` check.

docs/user-guide/troubleshooting.md
jia-gao: You're right — corrected. WVA doesn't fully skip status updates; it still sets `MetricsAvailable=False` via `DecisionCache.Set()`. Updated the text and added a `kubectl get va -o jsonpath` command to check the condition as the first diagnostic step.

docs/user-guide/troubleshooting.md
jia-gao: Fixed — changed to "fallback" for consistency.

docs/user-guide/configuration.md
jia-gao: Same fix as the troubleshooting comment — updated to clarify that `MetricsAvailable=False` is still set, only the full allocation/target-replicas update is skipped.

ev-shindin (approved):
Great work! Thanks!

Issue #987, opened 2026-04-10
Saturation Analyzer v1 docs do not match implementation


lionelvillard:
# Saturation Analyzer Doc vs Code Comparison

## Context
Comparing `docs/user-guide/saturation-analyzer.md` against the V1 saturation analyzer implementation in `internal/saturation/analyzer.go` and `internal/config/saturation_scaling.go`.

### 1. Scale-Down Safety Simulation — Algorithm Differs
**Doc (lines 77-93):** Describes summing total load across non-saturated replicas, then dividing by `remaining_replicas`:
```
total_kv_load = Σ kv_cache_usage_i
avg_kv_after_removal = total_kv_load / remaining_replicas
```

**Code ([analyzer.go:250-269](internal/saturation/analyzer.go#L250-L269)):** Uses average spare capacity (already computed) and a scale factor `N/(N-1)`:
```go
avgKvLoad := config.KvCacheThreshold - avgSpareKv
scaleFactor := float64(nonSaturatedCount) / float64(remainingCount)
avgKvAfterRemoval := avgKvLoad * scaleFactor
remainingSpareKv := config.KvCacheThreshold - avgKvAfterRemoval
```

**Impact:** Mathematically equivalent, but the doc implies iterating over replicas again while the code reuses precomputed averages. The doc's pseudo-code is misleading about the implementation approach. The doc should be updated to match the code's formula.

### 2. `CalculateCapacityTargets` — Transition Blocking Logic Missing from Docs
**Doc (lines 110-116):** States the rule as `desired ≠ 0 AND desired ≠ current → target = desired`. No mention of model-level transition blocking.

**Code ([analyzer.go:316-372](internal/saturation/analyzer.go#L316-L372)):** The actual logic is:
1. Check if ANY variant has `desired ≠ 0 && desired ≠ current` OR `metricsCount ≠ current` → entire model is "in transition"
2. If model in transition: ALL variants are blocked (no scale-up or scale-down for any variant), with preserved desired or current as target
3. Only if model is stable do scale-up/scale-down decisions proceed

**Impact:** This is a significant behavioral difference. The doc makes it sound like preservation is per-variant, but the code blocks the *entire model* when *any* variant is transitioning. The doc also doesn't mention the `metricsCount ≠ current` check at all.

### 3. Cascade Scaling Prevention — Doc Describes Per-Variant Skip, Code Has Model-Level Block
[...]

PR #994, opened 2026-04-13
feat: add V1 default thresholds and partial override support for satu…

### LINKED ISSUE #993: V1 saturation thresholds have no hardcoded defaults and per-model overrides don't support partial fields
### Problem

The V1 (percentage-based) saturation analyzer has two issues with its configuration:

#### 1. No hardcoded defaults for V1 thresholds

When the `wva-saturation-scaling-config` ConfigMap is missing or has no `default` entry, the V1 threshold fields (`kvCacheThreshold`, `queueLengthThreshold`, `kvSpareTrigger`, `queueSpareTrigger`) default to zero. This causes:

- Every replica to appear saturated (`kv_cache_usage >= 0` is always true)
- Continuous scale-up triggering (`avg_spare < 0` is always true)
- The documented defaults (0.80, 5, 0.10, 3) are never actually applied by the code

The `ApplyDefaults()` method only fills V2 fields (`ScaleUpThreshold`, `ScaleDownBoundary`) when `IsV2()` is true. V1 fields are never defaulted.

**Code:** `internal/config/saturation_scaling.go` — `ApplyDefaults()` (line 117)

#### 2. Per-model overrides replace the entire default config

`resolveSaturationConfig()` in `internal/engines/saturation/engine.go` returns the first matching config without merging:

```go
if cfg, ok := configMap[modelID+"#"+namespace]; ok {
    cfg.ApplyDefaults()
    return cfg  // override replaces default entirely
}
```

If a per-model override specifies only `kvCacheThreshold: 0.85`, all other fields (`queueLengthThreshold`, `kvSpareTrigger`, `queueSpareTrigger`) will be zero — not inherited from the `default` entry. This contradicts the documented behavior and forces operators to duplicate all fields in every override.

### Expected behavior

1. When the ConfigMap is missing, V1 thresholds should default to the documented values:
   - `kvCacheThreshold`: 0.80
   - `queueLengthThreshold`: 5
   - `kvSpareTrigger`: 0.10
   - `queueSpareTrigger`: 3

2. Per-model overrides should support partial fields — unspecified fields should inherit from the `default` ConfigMap entry.

### Proposed fix

1. Add V1 default constants and apply them in `ApplyDefaults()` when fields are zero
2. Add a `Merge(override)` method to `SaturationScalingConfig` that overlays non-zero fields
3. Update `resolveSaturationConfig()` to start from `default`, merge the override on top, then call `ApplyDefaults()`

Related issue #987

ev-shindin:
## Summary

Add hardcoded V1 default thresholds and partial override support for saturation scaling config, addressing the code changes requested in #987.

Previously, V1 saturation thresholds (`kvCacheThreshold`, `queueLengthThreshold`, `kvSpareTrigger`, `queueSpareTrigger`) defaulted to zero when the ConfigMap was missing or had no `default` entry. This caused all replicas to appear saturated and triggered continuous scale-up. Per-model overrides also required specifying all fields — omitted fields were zero, not inherited from the `default` config entry.

### V1 Default Thresholds (`internal/config/saturation_scaling.go`)

- Add exported constants: `DefaultKvCacheThreshold=0.80`, `DefaultQueueLengthThreshold=5`, `DefaultKvSpareTrigger=0.10`, `DefaultQueueSpareTrigger=3`
- Update `ApplyDefaults()` to fill zero-valued V1 fields unconditionally (V2 defaults remain gated by `IsV2()`)

### Partial Override via Merge (`internal/config/saturation_scaling.go`)

- Add `SaturationScalingConfig.Merge(override)` method that overlays non-zero fields from an override onto the receiver
- Covers all fields: V1 thresholds, V2 thresholds, `AnalyzerName`, `Priority`, `EnableLimiter`, `Analyzers`, `ModelID`, `Namespace`

### Config Resolution (`internal/engines/saturation/engine.go`)

- Rewrite `resolveSaturationConfig()` to:
  1. Start from the `default` ConfigMap entry (or zero-value if absent)
  2. Merge the model-specific `{modelID}#{namespace}` override on top
  3. Call `ApplyDefaults()` last to fill any remaining zero fields
- Per-model entries now only need to specify the fields they want to change

### Tests

- `internal/config/saturation_scaling_test.go`: 9 new test cases
  - `ApplyDefaults`: V1 defaults applied when not V2; explicit V1 values preserved
  - `Merge`: partial overlay, full overlay, empty overlay, V2 fields, analyzers list, ModelID/Namespace
- `internal/engines/saturation/engine_v2_test.go`: updated + 1 new test case
  - Verify merge-onto-default behavior, V1 defaults when map is empty, partial single-field override

## Known limitation

`Merge` uses zero-value detection (`!= 0` / `!= ""`) to distinguish "set" from "unset" fields. This means boolean fields like `EnableLimiter` cannot be overridden from `true` to `false` via a per-model override. However this is expected. Limiter either enabled or disabled for all models.

## Test plan

- [x] `go test ./internal/config/...` — 50 passed
- [x] `go test ./internal/engines/saturation/...` (envtest via WSL) — 18 passed
- [x] `go build ./...` — clean
- [x] Verify existing e2e tests still pass in CI

### DIFF
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -49,22 +49,25 @@ import (
 )
 
 // resolveSaturationConfig resolves config for a model.
-// Lookup: "{modelID}#{namespace}" → "default" → zero-value with defaults.
+// Starts from the "default" entry (or zero-value), then merges the model-specific
+// override "{modelID}#{namespace}" on top (if present). This allows per-model
+// overrides to specify only the fields they want to change.
+// ApplyDefaults is called last to fill any remaining zero-valued fields.
 func resolveSaturationConfig(
 	configMap map[string]config.SaturationScalingConfig,
 	modelID, namespace string,
 ) config.SaturationScalingConfig {
-	if cfg, ok := configMap[modelID+"#"+namespace]; ok {
-		cfg.ApplyDefaults()
-		return cfg
+	// Start with default config as base
+	base := config.SaturationScalingConfig{}
+	if defaultCfg, ok := configMap["default"]; ok {
+		base = defaultCfg
 	}
-	if cfg, ok := configMap["default"]; ok {
-		cfg.ApplyDefaults()
-		return cfg
+	// Overlay model-specific override if present (non-zero fields win)
+	if override, ok := configMap[modelID+"#"+namespace]; ok {
+		base.Merge(override)
 	}
-	cfg := config.SaturationScalingConfig{}
-	cfg.ApplyDefaults()
-	return cfg
+	base.ApplyDefaults()
+	return base
 }
 
 type Engine struct {
--- a/internal/config/saturation_scaling.go
+++ b/internal/config/saturation_scaling.go
@@ -106,17 +106,40 @@ func (c *SaturationScalingConfig) IsV2() bool {
 	return len(c.Analyzers) > 0 || c.AnalyzerName == "saturation"
 }
 
+// V1 analyzer default thresholds, applied when fields are omitted from YAML config.
+const (
+	DefaultKvCacheThreshold     = 0.80
+	DefaultQueueLengthThreshold = 5.0
+	DefaultKvSpareTrigger       = 0.10
+	DefaultQueueSpareTrigger    = 3.0
+)
+
 // V2 analyzer default thresholds, applied when fields are omitted from YAML config.
 const (
 	DefaultScaleUpThreshold  = 0.85
 	DefaultScaleDownBoundary = 0.70
 )
 
-// ApplyDefaults fills in zero-valued V2 fields with their defaults.
+// ApplyDefaults fills in zero-valued fields with their defaults.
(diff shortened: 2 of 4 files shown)


internal/engines/saturation/engine.go
```diff
@@ -49,22 +49,25 @@ import (
 )
 
 // resolveSaturationConfig resolves config for a model.
-// Lookup: "{modelID}#{namespace}" → "default" → zero-value with defaults.
+// Starts from the "default" entry (or zero-value), then merges the model-specific
+// override "{modelID}#{namespace}" on top (if present). This allows per-model
+// overrides to specify only the fields they want to change.
+// ApplyDefaults is called last to fill any remaining zero-valued fields.
 func resolveSaturationConfig(
```
lionelvillard, line 56: @mamy-CS can you PTAL? IIRC the config package already provides such features. If not it should.

lionelvillard (approved):

PR #997, opened 2026-04-13
fix: add EPP RBAC for InferenceModelRewrite (required by v0.7.0)

kahilam:
## Summary

- EPP v0.7.0 (built on `gateway-api-inference-extension v1.3.0`) always watches `InferenceModelRewrite` resources on startup via `InferenceModelRewriteReconciler`. There is no feature gate to disable this.
- The `inferencepool` Helm chart v1.0.1 (shipped with `llm-d v0.3.0`) does not include `inferencemodelrewrites` in the EPP Role, causing the EPP to crash-loop when its image is patched to v0.7.0:
  ```
  Failed to watch *v1alpha2.InferenceModelRewrite:
  inferencemodelrewrites.inference.networking.x-k8s.io is forbidden:
  User "system:serviceaccount:...:gaie-inference-scheduling-epp"
  cannot list resource "inferencemodelrewrites"
  ```
- Adds a supplemental Role + RoleBinding granting the EPP service account read-only (`get`, `list`, `watch`) access to `inferencemodelrewrites`, applied alongside the existing image and ConfigMap patches in `deploy/lib/infra_llmd.sh`.
- The upstream chart [v1.3.0 already includes this permission](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/v1.3.0/config/charts/inferencepool/templates/rbac.yaml); this backfills it for the older chart version.

## Test plan

- [ ] Trigger `/benchmark openshift` on a PR to verify EPP no longer crash-loops
- [ ] Verify EPP pod reaches `Ready` state and Gateway returns HTTP 200
- [ ] Confirm benchmark runs end-to-end through Gateway connectivity check


kahilam:
Closing in favor of a fresh PR. The RBAC fix was merged via #998. Will create a new PR with remaining fixes.

PR #1007, opened 2026-04-14
feat: infrastructure support for multi-model deployment and isolated gateway routing

asm582:
This PR enables deploying multiple models in a single namespace. PR was implemented using the Gemini code model.


asm582:
This PR and #1008, #1009 tried to add multi-model infra setup and test with script consolidation. The consolidation touched install scripts that are consumed by other make `* test` commands. This caused the testing surface to increase due to the tight coupling between the benchmark setup and existing `*-test` commands, leading to tests failing in CI that do not fail locally; this could still leave non-determinism in the test suite. Hence, I changed the design in #1014  and am closing these PRs for now.

PR #1008, opened 2026-04-14
feat: secure prometheus auth transport for OpenShift telemetry

asm582:
This PR adds a mechanism to pass a token so that multi-model `initial` perf metrics can be printed during benchmark execution, related to #1007.  PR was implemented using the Gemini code model.


asm582:
Refer: https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1007#issuecomment-4252618880

PR #1010, opened 2026-04-14
chore: benchmark test run

kahilam:
Dummy PR to trigger benchmark test on latest main.


asm582:
Closing this PR, wondering if we can run it with draft PRs.

kahilam:
## Investigation Summary: `/benchmark openshift` Gateway 500 Failure

**Problem:** The Gateway connectivity check always fails with HTTP 500 from `istio-envoy` (empty body).

**Root cause:** The `llm-d-infra` chart (v1.4.0) creates the Gateway with `istio.io/enable-inference-extproc: "true"`, which requires Istio to natively support InferencePool-based ext_proc routing. The Istio/OSSM version on the CI OpenShift cluster doesn't appear to support this feature.

**What was tried:**
1. Changed `LLM_D_RELEASE` default from `v0.3.0` to `main` (aligned CRDs with GA API)
2. Patched EPP `--pool-group` to match the InferencePool API group (`inference.networking.k8s.io`)

Neither fixed the 500 — the issue is at the Istio layer, not our deployment scripts.

**Note:** The `benchmark-openshift` job has never succeeded in CI history (18+ consecutive failures), confirming this is a pre-existing infrastructure issue.

**Next step:** Need to confirm the Istio/OSSM version on the CI cluster. InferencePool ext_proc support requires Istio 1.24+.

---
_Converting future test PRs to draft per @asm582's suggestion._

PR #1026, opened 2026-04-19
fix: use sentinel accelerator name instead of silently stopping scaling

### LINKED ISSUE #995: Controller silently stops scaling when accelerator name cannot be resolved
### Problem

When a `VariantAutoscaling` resource has no accelerator name available — neither from the target Deployment's `nodeSelector`/`nodeAffinity` nor from the `inference.optimization/acceleratorName` VA label — the controller silently stops functioning for that variant:

1. **Status updates are skipped** — `DesiredOptimizedAlloc` is never set (no target replicas, no last run time)
2. **Prometheus metrics are never emitted** — `wva_desired_replicas`, `wva_current_replicas`, `wva_desired_ratio` are absent
3. **HPA/KEDA never receives a scaling signal** — autoscaling is completely broken

The VA is accepted by the API server with no error events. The only signal is a `MetricsAvailable=False` condition and a log line:

```
accelerator name not found in scale target nodeSelector/nodeAffinity or VA label
```

This is particularly problematic in **homogeneous GPU clusters** (e.g., all H100s) where deployments often don't set `nodeSelector` because there's only one GPU type. Users must discover the issue by reading controller logs.

### Root cause

The accelerator name is used as a required Prometheus label (`accelerator_type`) on all WVA output metrics. The code treats a missing accelerator as a hard stop and skips the entire metric emission and status update flow.

Relevant code paths:

- `internal/engines/saturation/engine.go:977-996` — skips status update with `continue` when `acceleratorName == ""`
- `internal/engines/saturation/engine.go:1177-1181` — safety-net metric emission also skips: `"Safety net: skipping metric emission - no accelerator name available"`
- `internal/utils/allocation.go:37-39` — `BuildAllocationFromMetrics` returns error when accelerator is empty

The accelerator name is **not used in the core scaling decision logic**. The V1 saturation analyzer (`internal/saturation/analyzer.go`) makes scale-up/down decisions based purely on KV cache and queue metrics. The V2 cost-aware optimizer uses `AcceleratorName` only to propagate it downstream — cost is a separate field.

However, the accelerator name **is used by the GPU limiter** when limited mode is enabled (`enableLimiter: true`). The `typeAllocator` in `internal/engines/pipeline/type_inventory.go:348-351` looks up `decision.AcceleratorName` to match against per-GPU-type inventory discovered from cluster nodes. It returns an error if the field is empty and returns 0 GPUs if the type has no matching pool. This means the limiter cannot constrain scaling decisions without knowing the GPU type.

### Expected behavior

The controller should continue to function correctly (emit metrics, update status, enable HPA/KEDA scaling) even when the accelerator name cannot be resolved from the deployment or VA label.

### Proposed fix

Use a default sentinel value `"unknown"` when accelerator name cannot be resolved, instead of skipping the entire flow.

**Changes needed:**

1. **Add a constant** in `internal/constants/constants.go` or `internal/utils/variant.go`:
   ```go
   const DefaultAcceleratorName = "unknown"
   ```

2. **Update `GetAcceleratorNameFromScaleTarget`** in `internal/utils/utils.go` to return the sentinel instead of `""` when all lookups fail. Alternatively, handle this at the call sites.

3. **Remove the `continue` guards** in `internal/engines/saturation/engine.go:977` and `engine.go:1177` that skip processing when accelerator is empty — they become unreachable once the sentinel is in place.

4. **Update `BuildAllocationFromMetrics`** in `internal/utils/allocation.go:37-39` to accept the sentinel value instead of returning an error.

5. **Log a warning** (not skip) when the sentinel is used, so operators know auto-discovery didn't find a GPU type:
   ```
   WARN  Using default accelerator name "unknown" - set nodeSelector on Deployment or inference.optimization/acceleratorName label on VA
   ```

6. **Update the GPU limiter** in `internal/engines/pipeline/type_inventory.go:343-369` — when `AcceleratorName` is `"unknown"`, resolve it against the inventory at allocation time:

   - **Homogeneous cluster** (inventory has exactly one GPU type): map `"unknown"` to that type. The limiter works normally with correct per-type limits, and `decision.AcceleratorName` gets corrected in place so downstream status updates and Prometheus metrics also get the real GPU type label.
   - **Heterogeneous cluster** (inventory has multiple GPU types): can't determine which pool to debit, so treat as unlimited (return all requested GPUs) and log a warning.

   **Behavior by cluster type:**

   | Cluster | Inventory | `"unknown"` resolves to | Limiter behavior |
   |---------|-----------|------------------------|------------------|
   | Homogeneous (all H100) | `{"H100": 16}` | `"H100"` | Full per-type limiting works; `AcceleratorName` corrected to `"H100"`; metrics emit `accelerator_type="H100"` |
   | Heterogeneous (H100 + A100) | `{"H100": 8, "A100": 8}` | `""` (can't resolve) | Treat as unlimited, log warning |
   | No inventory / limiter off | empty | N/A | Limiter not called |

   The key benefit for homogeneous clusters: `decision.AcceleratorName` gets **corrected in place** to the real GPU type, so the limiter effectively becomes a second auto-discovery mechanism. Downstream status updates and Prometheus metrics will get the correct `accelerator_type` label instead of `"unknown"`.

**Impact:**

- Metrics will be emitted as `accelerator_type="unknown"` initially — HPA/KEDA will receive scaling signals immediately
- **With limiter enabled in a homogeneous cluster:** the limiter resolves `"unknown"` to the real GPU type, so metrics will emit the correct `accelerator_type` label (e.g., `"H100"`) and per-type GPU limits are enforced normally
- **With limiter enabled in a heterogeneous cluster:** `"unknown"` variants are treated as unlimited with a warning — operators must set `nodeSelector` or the VA label for proper per-type limiting
- Status updates will proceed normally
- No breaking changes — existing deployments with auto-discovery or the VA label continue to work identically

jia-gao:
## Summary

Fixes #995 — Controller silently stops scaling when accelerator name cannot be resolved.

When a `VariantAutoscaling` resource has no accelerator name available (neither from the target Deployment's `nodeSelector`/`nodeAffinity` nor from the VA label), the controller was silently skipping all status updates and metric emission. This is common in **homogeneous GPU clusters** where deployments don't set `nodeSelector` because there's only one GPU type.

### Changes

| File | Change |
|------|--------|
| `internal/constants/constants.go` | Add `DefaultAcceleratorName = "unknown"` sentinel (used only inside the limiter pipeline) |
| `internal/utils/utils.go` | `GetAcceleratorNameFromScaleTarget` returns sentinel instead of `""` when all lookups fail |
| `internal/utils/utils_test.go` | Updated tests to expect sentinel instead of empty string |
| `internal/utils/allocation.go` | `BuildAllocationFromMetrics` no longer errors on empty accelerator |
| `internal/engines/pipeline/default_limiter.go` | Add `resolveUnknownAccelerators` step that maps empty/unknown → real type in homogeneous clusters before usage counting |
| `internal/engines/pipeline/type_inventory.go` | `TryAllocate` resolves in homogeneous clusters (safety net), returns 0 in heterogeneous |
| `internal/engines/pipeline/type_inventory_test.go` | Updated + added test cases for both cluster types |
| `internal/engines/saturation/engine.go` | Continue status updates without accelerator, emit K8s Warning event, skip metric emission when unresolved, never persist sentinel to status |

### How it works

**Homogeneous cluster (e.g. all H100):**
The limiter's `resolveUnknownAccelerators` sees one GPU type in inventory and resolves the decision's `AcceleratorName` from `"unknown"` → `"H100"` *before* usage counting and allocation. The resolved type flows through to status updates and Prometheus metrics. Full per-type limiting works correctly.

**Heterogeneous cluster (e.g. H100 + A100):**
Can't determine which pool — `TryAllocate` returns 0 so the decision keeps current replicas without over-allocating. Status updates continue but metric emission is skipped. Operator must set `nodeSelector` or the VA label.

**No limiter / limiter disabled:**
Status updates continue with empty accelerator. K8s Warning event emitted. Metric emission skipped to avoid incorrect `accelerator_type` label.

### Design constraints

The `"unknown"` sentinel is **never persisted** to VA status or used as a Prometheus label. It exists only transiently inside the limiter pipeline. This avoids:
- Duplicate Prometheus series (switching from a real label to "unknown" and back)
- Sticky sentinel in status that prevents self-healing
- Incorrect GPU capacity accounting in `calculateUsedGPUs`

### Deviation from proposed fix

The issue proposes "treat as unlimited (return all requested GPUs)" for heterogeneous clusters when the accelerator is unresolved. This implementation returns 0 instead, because returning all requested GPUs without debiting any per-type pool allows double-allocation: the same physical GPUs can be granted once via the unlimited path for the unresolved decision, and again from a typed pool for a subsequent known-type decision. Returning 0 is safer — the variant keeps its current replicas and the operator is notified via the K8s Warning event to set `nodeSelector` or the VA label. Open to discussion if the maintainers prefer a different trade-off here.

The issue also states that metrics will be emitted as `accelerator_type="unknown"`. This implementation skips metric emission instead when the accelerator is unresolved, because emitting under `"unknown"` creates a duplicate Prometheus series alongside any existing real-type series for the same VA. Since HPA/KEDA queries filter only by `variant_name`/`namespace` (not `accelerator_type`), duplicate series cause ambiguous scaling signals or query errors. Skipping emission is safer — status updates still flow so the controller can reconcile, and the operator is notified via the K8s Warning event.

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `gofmt` passes (imports sorted correctly)
- [x] `go test ./internal/utils/...` passes (updated test expectations)
- [x] `go test ./internal/engines/pipeline/...` — all specs pass (including new test cases)
- [ ] CI e2e tests (require kubebuilder/Kind)


internal/engines/saturation/engine.go
```diff
@@ -969,35 +971,34 @@ func (e *Engine) applySaturationDecisions(
...
-				MetricsMessage:   llmdVariantAutoscalingV1alpha1.MessageMetricsUnavailable,
-			})
-			// Trigger reconciler to apply the condition
-			common.DecisionTrigger <- event.GenericEvent{
-				Object: &updateVa,
+		// Emit a K8s event when accelerator cannot be resolved so operators
+		// can see the problem without digging through controller logs.
+		if acceleratorName == constants.DefaultAcceleratorName {
+			if e.Recorder != nil {
+				e.Recorder.Eventf(&updateVa, corev1.EventTypeWarning, "AcceleratorNotResolved",
```
ev-shindin: This runs inside `applySaturationDecisions` which is called on every optimization cycle (default: every few seconds). An unresolved VA will generate Warning events **continuously** — potentially hundreds per hour. Kubernetes event deduplication by the API server aggregates identical events, but the `lastTimestamp` updates each time, which may still create noise in event sinks.

Consider rate-limiting the event — e.g., only emit on the **first** occurrence per VA, or only when the condition transitions from resolved → unresolved:

Or use `e.Recorder.Event` (non-formatted) which deduplicates better than `Eventf`.

internal/engines/pipeline/default_limiter.go
```diff
@@ -67,6 +74,26 @@ func (l *DefaultLimiter) Limit(ctx context.Context, decisions []*interfaces.Vari
 	return nil
 }
 
+// resolveUnknownAccelerators maps empty or "unknown" accelerator names to the
+// real GPU type when the inventory has exactly one type (homogeneous cluster).
+// This must run before calculateUsedGPUs so existing replicas are counted
+// against the correct pool, and before TryAllocate so new allocations use it.
+func (l *DefaultLimiter) resolveUnknownAccelerators(decisions []*interfaces.VariantDecision) {
```
ev-shindin, line 84: The new `resolveUnknownAccelerators` method has no direct test. The `typeAllocator` tests (`type_inventory_test.go:362-374`) cover the safety-net in `TryAllocate`, but the primary resolution path in the limiter isn't tested.

internal/engines/saturation/engine.go
ev-shindin, line 1: The sentinel check `== constants.DefaultAcceleratorName` appears in many places. Consider extract a function.

ev-shindin (changes requested):
Thanks. Please address review comments.

jia-gao:
Addressed all three review comments in 5d2dd56:

1. **Event rate-limiting**: Switched from `Eventf` to `Event` with a fixed message string so the K8s API server deduplicates identical events (aggregates with updated count instead of creating new entries each cycle).

2. **Test for `resolveUnknownAccelerators`**: Added direct unit tests covering homogeneous (resolves to real type), heterogeneous (unchanged), and empty inventory (unchanged).

3. **Extract helper**: Added `constants.IsAcceleratorResolved(name)` that returns `true` when name is not empty and not the sentinel. Replaced all scattered `== "" || == constants.DefaultAcceleratorName` checks across `engine.go`, `default_limiter.go`, and `type_inventory.go`.

internal/engines/saturation/engine.go
jia-gao: Switched from `Eventf` to `Event` with a fixed message string — the K8s API server deduplicates identical events by aggregating with an updated count rather than creating new entries each cycle. This should eliminate the noise in event sinks.

internal/engines/pipeline/default_limiter.go
jia-gao: Added direct unit tests for `resolveUnknownAccelerators` covering three cases: homogeneous cluster (empty/"unknown" → real type), heterogeneous cluster (names unchanged), and empty inventory (names unchanged).

internal/engines/saturation/engine.go
jia-gao: Extracted `constants.IsAcceleratorResolved(name)` helper — returns `true` when name is not empty and not the sentinel. Replaced all 8 inline checks across `engine.go`, `default_limiter.go`, and `type_inventory.go`.

ev-shindin: **The sentinel is leaking into CRD status, contradicting the engine's documented invariant.**

The function correctly sanitizes its in-memory copy above. But:

1. `applySaturationDecisions` does **not** write to the API server here. It only writes to the cache + triggers the controller.
2. The cache write uses the **raw** `acceleratorName` (the sentinel `"unknown"`), not `statusAccelerator`.
3. The controller reads the cache via `common.DecisionToOptimizedAlloc(decision)` at `internal/controller/variantautoscaling_controller.go:169`, which at `internal/engines/common/cache.go:56` returns `d.AcceleratorName` verbatim.
4. The controller writes status at `variantautoscaling_controller.go`:

```go
if accelerator != "" {  // "unknown" != "" → TRUE
    va.Status.DesiredOptimizedAlloc = OptimizedAlloc{
        Accelerator: accelerator, // "unknown" goes to API server
    }
}
```

**Result:** `kubectl get va -o yaml` will show `accelerator: unknown` after a reconcile, in violation of the engine's stated invariant.
**Fix (preferred — sanitize once, at the source):**

```go
common.DecisionCache.Set(va.Name, va.Namespace, interfaces.VariantDecision{
    ...
    AcceleratorName: statusAccelerator,  // already sanitized above
    ...
})
```

**Optional defense-in-depth at the controller** (`variantautoscaling_controller.go`):

```go
if constants.IsAcceleratorResolved(accelerator) {
    va.Status.DesiredOptimizedAlloc = ...
}
```

ev-shindin: For a VA where `hasAllocation == true` (metrics were collected) but the accelerator is unresolved, this reports `metricsAvailable: true` — but the actual metric emission was **skipped** at line 1054 because the accelerator is unresolved:
The CRD status will then show `MetricsAvailable=True` while no `wva_*` Prometheus series actually exist for that variant, leaving HPA/KEDA-side reasoning inconsistent with controller-side reporting.

internal/engines/saturation/engine.go
```diff
@@ -969,35 +971,37 @@ func (e *Engine) applySaturationDecisions(
...
+		if !constants.IsAcceleratorResolved(acceleratorName) {
+			if e.Recorder != nil {
+				e.Recorder.Event(&updateVa, corev1.EventTypeWarning, "AcceleratorNotResolved",
+					"Cannot resolve accelerator type from Deployment nodeSelector/nodeAffinity or VA label "+
+						utils.AcceleratorNameLabel+". "+
+						"Set nodeSelector on Deployment or add the label to the VariantAutoscaling resource. "+
+						"HPA/KEDA metrics will not be emitted until the accelerator is resolved.")
 			}
-			continue
+			logger.Info("Accelerator name not resolved - status will be updated but metrics will not be emitted",
```
ev-shindin: The K8s event at line 981 above is deduplicated by the API server, but this Info-level log line emits on every optimization cycle for every unresolved VA. With multiple unresolved VAs and a ~30s optimization interval, that's noisy in production logs.
Please downgrade log to `Debug`

internal/engines/saturation/engine.go
```diff
@@ -969,35 +971,37 @@ func (e *Engine) applySaturationDecisions(
...
-			common.DecisionTrigger <- event.GenericEvent{
-				Object: &updateVa,
+		// Emit a K8s event when accelerator cannot be resolved so operators
+		// can see the problem without digging through controller logs.
+		// Uses Event (not Eventf) with a fixed message for better event deduplication
+		// by the API server — identical events are aggregated with an updated count
+		// rather than creating new entries each optimization cycle.
+		if !constants.IsAcceleratorResolved(acceleratorName) {
+			if e.Recorder != nil {
+				e.Recorder.Event(&updateVa, corev1.EventTypeWarning, "AcceleratorNotResolved",
```
ev-shindin: The new K8s event emission has no unit test. It's trivial to add with `record.NewFakeRecorder`.

internal/engines/pipeline/type_inventory.go
```diff
@@ -346,9 +347,24 @@ func (a *typeAllocator) TryAllocate(decision *interfaces.VariantDecision, gpusRe
...
-			decision.Namespace, decision.VariantName)
+	if !constants.IsAcceleratorResolved(accType) {
+		// Normally resolved by DefaultLimiter.resolveUnknownAccelerators before
+		// reaching here. This is a safety net for direct callers.
+		if len(a.remainingByType) == 1 {
+			for resolvedType := range a.remainingByType {
+				decision.AcceleratorName = resolvedType
+				accType = resolvedType
+			}
+		} else {
```
ev-shindin, line 307: The K8s Warning event surfaces this at the engine level, but a debug log at this exact site would help anyone tracing a "why isn't this VA scaling?" investigation through the limiter.

internal/engines/saturation/engine.go
```diff
@@ -969,35 +971,37 @@ func (e *Engine) applySaturationDecisions(
...
-				MetricsAvailable: false,
-				MetricsReason:    llmdVariantAutoscalingV1alpha1.ReasonMetricsMissing,
-				MetricsMessage:   llmdVariantAutoscalingV1alpha1.MessageMetricsUnavailable,
-			})
-			// Trigger reconciler to apply the condition
-			common.DecisionTrigger <- event.GenericEvent{
-				Object: &updateVa,
+		// Emit a K8s event when accelerator cannot be resolved so operators
+		// can see the problem without digging through controller logs.
+		// Uses Event (not Eventf) with a fixed message for better event deduplication
```
ev-shindin: This conflates two unrelated things. The Kubernetes API server's event dedup keys on the `(involvedObject, source, type, reason, message, related)` tuple within a time window — **it does not care whether you called `Event` or `Eventf`.** Both methods produce structurally identical Event objects when the resulting message string is identical.

The actual fix here is correct (using a constant message instead of `Eventf` with per-cycle formatted args), but the comment misattributes *why* it works. A future reader could conclude that switching back to `Eventf` is what would re-introduce the bug, or that `Event` has some inherent dedup advantage.

ev-shindin (commented):
Thanks. Please address review comments.

asm582:
@jia-gao I am trying to understand the issue better. llmd modelservice always has `nodeSelector/nodeAffinity` , while we do not control how users deploy llmd model servers, but can we solve the issue by having nodeselector/affinity?

internal/engines/saturation/engine.go
jia-gao: Good catch — confirmed the leak path. Fixed at the source: the cache write now uses the already-sanitized accelerator value, so the controller never sees the sentinel. Also added a defense-in-depth gate at the controller's status-write site (and the patch-base helper at line 276), so even if a future code path bypasses sanitization, the sentinel can't be persisted.

jia-gao: Fixed. The cache now reports `metricsAvailable` only when accelerator resolution actually succeeded — same predicate the metric-emission branch uses to decide whether to skip. Controller-side reporting and HPA/KEDA-side reasoning are now consistent, and the surrounding comment explains the gating.

internal/engines/saturation/engine.go
jia-gao: Done — downgraded to debug level. The Warning event remains the operator-facing signal; the log is now only useful for trace-level investigation.

internal/engines/saturation/engine.go
jia-gao: Added. Extracted the inline emission into a small helper on `Engine`, then added three focused unit tests using `FakeRecorder`:

- asserts the event is Warning-typed with the expected reason and remediation message
- asserts a nil recorder is a safe no-op
- asserts repeated emissions produce identical event strings (the contract that lets the API server aggregate them)

internal/engines/pipeline/type_inventory.go
jia-gao: Added at the heterogeneous-cluster fallback. Now anyone tracing "why isn't this VA scaling?" through the limiter sees the skip locally, without having to chase it back up to the engine-level Warning event.

internal/engines/saturation/engine.go
jia-gao: You're right, the previous comment misattributed the cause. Rewrote it: the dedup works because the message is a constant string, not because we picked `Event` over `Eventf`. The new test pinning identical-message dedup will fail if a future contributor switches back to per-cycle interpolation, so the contract is enforced rather than just documented.

jia-gao:
@asm582 — fair question. You're right that `llmd-modelservice` always sets nodeSelector, so for that path the unresolved branch is unreachable and this PR doesn't change behavior.

The case this addresses is everything *outside* modelservice: hand-rolled VAs, test/eval clusters, in-progress migrations, third-party deployers. Today those hit a silent scaling stop — controller logs nothing user-visible and scaling just halts (#995 was exactly this). This PR converts that into a Warning event + visible status condition + best-effort scaling in the homogeneous case. Purely additive on top of modelservice.

If the project's stance is that non-modelservice deployments aren't supported, the cleaner long-term answer is admission-time validation requiring nodeSelector or the VA label. Happy to file that as a follow-up. WDYT?

internal/engines/saturation/engine.go
```diff
@@ -969,35 +971,34 @@ func (e *Engine) applySaturationDecisions(
...
+		// is empty, and the limiter resolves it to the real type for decisions.
 		numReplicas := int32(targetReplicas)
+		statusAccelerator := acceleratorName
+		if !constants.IsAcceleratorResolved(statusAccelerator) {
+			statusAccelerator = ""
+		}
 		updateVa.Status.DesiredOptimizedAlloc = llmdVariantAutoscalingV1alpha1.OptimizedAlloc{
 			NumReplicas: &numReplicas,
-			Accelerator: acceleratorName,
+			Accelerator: statusAccelerator,
```
ev-shindin, line 1306: The engine never writes `va.Status` to the API server — there are zero `Status().Update`/`Status().Patch` calls in this file. The reconciliation path is **cache → DecisionTrigger channel → controller** (the comment at `engine.go:1073-1074` says exactly this: "This avoids any API server interaction from the Engine"). The controller fetches a fresh VA from the cluster, then reads the cache for the decision and patches via `r.Status().Patch(...)` at `variantautoscaling_controller.go:177` and `:260`.

That has two consequences worth fixing:

1. **The comment at lines 989-991 is wrong about the unresolved-accelerator case.** The comment promises that "Status is updated even without accelerator so that `DesiredOptimizedAlloc.NumReplicas` and `LastRunTime` remain current". But the controller's matching code at `variantautoscaling_controller.go:231-235` does the opposite — when the cached accelerator is empty (sentinel sanitized to ""), it discards the decision's `numReplicas` and `lastRunTime` and preserves `originalVA.Status.DesiredOptimizedAlloc`:

   ```go
   } else {
       // When we have a partial decision (no accelerator yet), explicitly preserve
       // the existing DesiredOptimizedAlloc from the fetched object to avoid
       // sending zero-valued struct in the patch which would fail CRD validation.
       va.Status.DesiredOptimizedAlloc = originalVA.Status.DesiredOptimizedAlloc
   }
   ```

   So for a VA with an unresolved accelerator, **`NumReplicas` and `LastRunTime` in status do not refresh** until the accelerator becomes resolvable — directly contradicting the engine's comment.

2. **The local mutation at lines 999-1003 is dead code.** Nothing reads `updateVa.Status.DesiredOptimizedAlloc` after this assignment. The cache write at `engine.go:1099-1108` uses `statusAccelerator`, `targetReplicas`, and a fresh `metav1.Now()` directly from local scope. The metric-emission branch at `engine.go:1056-1058` reads `updateVa.Name`, not its status fields. The conditions are set from raw fields, not from `updateVa.Status.DesiredOptimizedAlloc`.

I suggest delete lines 999-1003 entirely. The cache already carries `targetReplicas` (→ `NumReplicas`), `metav1.Now()` (→ `LastRunTime`), and `statusAccelerator`. Update the comment block at lines 988-993 to describe the cache-write invariant instead.

internal/engines/saturation/engine_event_test.go
ev-shindin, line 1: The new test file uses standard `testing`, while the rest of the package is Ginkgo. Can you please align the test file with project style?
The tests themselves are good — clear naming, doc comments per test, asserting the right contracts. This is purely about house style.

internal/engines/pipeline/type_inventory.go
```diff
@@ -346,9 +350,30 @@ func (a *typeAllocator) TryAllocate(decision *interfaces.VariantDecision, gpusRe
...
+			// Heterogeneous cluster — can't determine which pool to debit.
+			// Note: #995 proposed "treat as unlimited (return all requested GPUs)"
+			// but that allows the same physical GPUs to be double-allocated: once
+			// here for the unresolved decision, and again from a typed pool for a
+			// subsequent known-type decision. Returning 0 is safer — the variant
+			// keeps its current replicas without over-allocating. Operator must
+			// set nodeSelector or the VA label for scaling in mixed-GPU clusters.
+			ctrl.Log.WithName("typeAllocator").V(logging.DEBUG).Info(
```
ev-shindin: `TryAllocate(decision *interfaces.VariantDecision, gpusRequested int)` (signature at line 347) doesn't take a `context.Context`, so the new debug log uses the package-level `ctrl.Log`. This works, but it loses request-scoped trace context (logger fields injected upstream via `ctrl.LoggerFrom(ctx)`) — which means an operator tracing "why did variant X not scale this cycle?" through the controller's structured logs won't see this line correlated with the rest of the cycle's output.
Can you please thread `ctx` into `TryAllocate` (and `Limit` / `ResourceAllocator.TryAllocate` if needed). This is small signature change, but it brings this hot path into line with the rest of the engine.

ev-shindin (commented):
Almost done! Please adrees my comments.

asm582:
> @asm582 — fair question. You're right that `llmd-modelservice` always sets nodeSelector, so for that path the unresolved branch is unreachable and this PR doesn't change behavior.
> 
> The case this addresses is everything _outside_ modelservice: hand-rolled VAs, test/eval clusters, in-progress migrations, third-party deployers. Today those hit a silent scaling stop — controller logs nothing user-visible and scaling just halts (#995 was exactly this). This PR converts that into a Warning event + visible status condition + best-effort scaling in the homogeneous case. Purely additive on top of modelservice.
> 
> If the project's stance is that non-modelservice deployments aren't supported, the cleaner long-term answer is admission-time validation requiring nodeSelector or the VA label. Happy to file that as a follow-up. WDYT?

Thanks, trying to understand better:

```
The case this addresses is everything outside modelservice: hand-rolled VAs, test/eval clusters, in-progress migrations, third-party deployers.
``` 

VAs are deployed outside of modelservice today; the expectation is that accelerator type should be provided. This could be either upfront and forcing the deployment pods to land on the upfront hardware type via labels/affinity or waiting for the scheduler to pick a node and later fire a VA by referring to what GPU type the scheduler actually assigned. What am I missing?

```
This PR converts that into a Warning event + visible status condition + best-effort scaling in the homogeneous case.
```
What does best-effort scaling mean?

ev-shindin:
@asm582 I see at least 3 use cases where this PR could help.
1. Testing clusters. We already have problems in the past. When you not define nodeAffinity or forgot provide VA label WVA doesn't work
2. Cluster with homogeneous GPU. Again nodeAffinity and VA labels are not straghtforward.
3. Managed cluster that provide access to the specific GPU types for the specific tenant. Again user should not care on nodeAffinity or GPU label on VA.

IMO the situation when WVA will not work because user don't put some label on VA is not good, so I prefer to improve UX.

internal/engines/pipeline/type_inventory.go
jia-gao: Done in a81fa11 — threaded ctx through `TryAllocate` (interface + `typeAllocator` + greedy caller) and swapped the skip log to `ctrl.LoggerFrom(ctx).WithName("typeAllocator")`. Test mocks (`mockTypeAllocator`, `simpleAllocator`) and the direct test calls got the new signature too.

internal/engines/saturation/engine_event_test.go
jia-gao: Migrated to Ginkgo in a81fa11 — three `It` blocks under one `Describe`, shared recorder/engine/va via `BeforeEach`. Same contracts, just BDD style now.

internal/engines/saturation/engine.go
jia-gao: Comment fixed in a81fa11, but I think the mutation itself shouldn't be deleted — `EmitMetrics` actually reads from it.

`act.EmitMetrics(ctx, &updateVa)` at line 1059 pulls fields straight off `updateVa.Status.DesiredOptimizedAlloc`:
- `actuator.go:59` — early-return if `NumReplicas == nil`
- `actuator.go:65` — `desiredReplicas := *…NumReplicas`
- `actuator.go:80` — `…Accelerator` as the metric label

So if we drop lines 999-1003, the first cycle skips emission entirely (NumReplicas still nil), and later cycles emit whatever the controller last persisted instead of the target we just computed.

You're right that the old comment was misleading though — it framed this as Status going to the API server, which it isn't. New comment spells out what's actually happening: local staging for `EmitMetrics`, with CRD persistence handled by the cache → controller path below.

If you'd rather we make the data flow explicit, I can refactor `EmitMetrics` to take `desiredReplicas` + `accelerator` directly — then the lines really are dead and we can delete them. Heads up that touches the `Actuator` interface and ~7 tests in `actuator_test.go`. Lmk which way you want it.

ev-shindin (commented):
I am going to approve this PR. @asm582 could you answer my comment above?

asm582:
> @asm582 — fair question. You're right that `llmd-modelservice` always sets nodeSelector, so for that path the unresolved branch is unreachable and this PR doesn't change behavior.
> 
> The case this addresses is everything _outside_ modelservice: hand-rolled VAs, test/eval clusters, in-progress migrations, third-party deployers. Today those hit a silent scaling stop — controller logs nothing user-visible and scaling just halts (#995 was exactly this). This PR converts that into a Warning event + visible status condition + best-effort scaling in the homogeneous case. Purely additive on top of modelservice.
> 
> If the project's stance is that non-modelservice deployments aren't supported, the cleaner long-term answer is admission-time validation requiring nodeSelector or the VA label. Happy to file that as a follow-up. WDYT?

To me, it is worth adding admission time validation.

asm582:
> @asm582 I see at least 3 use cases where this PR could help.
> 
> 1. Testing clusters. We already have problems in the past. When you not define nodeAffinity or forgot provide VA label WVA doesn't work
> 2. Cluster with homogeneous GPU. Again nodeAffinity and VA labels are not straghtforward.
> 3. Managed cluster that provide access to the specific GPU types for the specific tenant. Again user should not care on nodeAffinity or GPU label on VA.
> 
> IMO the situation when WVA will not work because user don't put some label on VA is not good, so I prefer to improve UX.

re 1: All our deployments and benchmarks use helm charts and guides. In fact, llmd as a project promotes usage of helm and well-lit path to stream line deployment, so I dont know if this is a real use case. 
re 2: Creating streamlined installation of modelservers which is part of the issue should be out of scope of WVA
re 3: Adding node and affinity is exactly Kubernetes to mechanism to target a specific GPU type. we are yet to see a concrete use case where the whole llmd system with WVA that works with heterogeneous GPUs

ev-shindin:
@jia-gao please rebase

jia-gao:
> @jia-gao please rebase

rebased, thanks!

asm582:
Thanks @jia-gao , can you please sign-off commits to pass DCO, below is a way to do it:

```
To add your Signed-off-by line to every commit in this branch:

Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
In your local branch, run: git rebase HEAD~7 --signoff
Force push your changes to overwrite the branch: git push --force-with-lease origin fix/995-default-accelerator-name
```

jia-gao:
Thanks @asm582 — rebased with `--signoff` and force-pushed, DCO is green now.

asm582:
> Thanks @asm582 — rebased with `--signoff` and force-pushed, DCO is green now.

Thanks, sorry, but I'm seeing conflicts now. 

```
This branch has conflicts that must be resolved
Use the [web editor](https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1026/conflicts) or the command line to resolve conflicts before continuing.

internal/constants/constants.go
internal/engines/pipeline/type_inventory.go
```

ev-shindin:
@jia-gao can you  please rebase it again?

jia-gao:
> @jia-gao can you please rebase it again?

rebased!

ev-shindin (approved):
Thanks @jia-gao!

PR #1028, opened 2026-04-20
api: add scalingBackend field to VariantAutoscaling CRD

vivekk16:
## Summary

Adds a new optional `scalingBackend` field to `VariantAutoscalingSpec` that allows users to opt into controller-managed HPA or KEDA ScaledObject backends. When the field is nil, the controller continues in metrics-only mode (fully backward compatible).

## Changes

- **`ScalingBackendSpec`**: top-level discriminator with `type` enum (`hpa` / `keda`), `HPAConfig`, and `KEDAConfig`
- **`HPAConfig`**: wraps `autoscalingv2.HorizontalPodAutoscalerBehavior` for scale-up/down
  policy tuning on the generated HPA
- **`KEDAConfig`**: exposes `pollingInterval`, `cooldownPeriod`, `initialCooldownPeriod`, `idleReplicaCount`, plus `fallback` and `advanced` using native KEDA Go types (`kedav1alpha1.Fallback`, `kedav1alpha1.AdvancedConfig`)
- **`EffectiveScalingBackendType()`**: helper that returns `""` for nil (metrics-only), defaults to `"hpa"` when type is empty
- **Condition constants**: `TypeScalingBackendReady` with reasons `ScalingBackendReconciled`, `KEDANotInstalled`, `ScalingBackendError`, `UnsupportedScaleTarget`
- **KEDA dependency**: added `github.com/kedacore/keda/v2 v2.19.0` with `replace` directives to pin `k8s.io/*`, `controller-runtime`, and `prometheus/*` to WVA's existing versions
- Regenerated `zz_generated.deepcopy.go` and CRD manifests

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./api/...` passes
- [x] `make generate` and `make manifests` produce clean output


vivekk16:
@lionelvillard @asm582 @ev-shindin PTAL

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -16,6 +17,99 @@ type VariantAutoscalingConfigSpec struct {
...
+// Scaling backend type constants.
+const (
+	ScalingBackendTypeHPA  = "hpa"
+	ScalingBackendTypeKEDA = "keda"
+)
+
+// ScalingBackendSpec selects how the controller materializes a scaler for this VA.
+// When nil on the parent VariantAutoscalingSpec, the controller only emits
+// Prometheus metrics (metrics-only / legacy mode).
+type ScalingBackendSpec struct {
```
ev-shindin, line 29: No CEL cross-validation between `Type` and sub-configs — `hpa:` ignored when `type: keda`...
An operator can set `type: keda` but also populate `hpa: { behavior: ... }`. The API server accepts this silently — there's no `XValidation` CEL rule rejecting the mismatch. The HPA config block is then ignored at runtime (the controller will read only the KEDA config), which can confuse operators who think they're tuning HPA behavior.

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -16,6 +17,99 @@ type VariantAutoscalingConfigSpec struct {
...
+	Advanced *kedav1alpha1.AdvancedConfig `json:"advanced,omitempty"`
+}
+
+// EffectiveScalingBackendType returns the resolved backend type.
+// Returns empty string when sb is nil (metrics-only mode).
+func EffectiveScalingBackendType(sb *ScalingBackendSpec) string {
+	if sb == nil {
+		return ""
+	}
+	if sb.Type == "" {
```
ev-shindin, line 107: The `Type` field has `+kubebuilder:validation:Required` (line 36), so the API server rejects empty strings at admission time. The `sb.Type == ""` branch in `EffectiveScalingBackendType` is defensive — it can only fire if the function is called from code that bypasses API validation (e.g., unit tests constructing structs directly, or old objects in etcd before the validation was added).

Two options:
- **(a)** Keep the defensive default as-is — robust against data-corruption scenarios. Add a brief comment explaining why.
- **(b)** Remove the empty-check and let it return empty string — callers must handle unexpected values anyway.

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -16,6 +17,99 @@ type VariantAutoscalingConfigSpec struct {
...
+
+	// Advanced configures KEDA advanced options such as HPA behavior overrides
+	// and replica restoration. Maps to ScaledObject.spec.advanced.
+	// +kubebuilder:validation:Optional
+	Advanced *kedav1alpha1.AdvancedConfig `json:"advanced,omitempty"`
+}
+
+// EffectiveScalingBackendType returns the resolved backend type.
+// Returns empty string when sb is nil (metrics-only mode).
+func EffectiveScalingBackendType(sb *ScalingBackendSpec) string {
```
ev-shindin, line 103: No unit test for this function.

api/v1alpha1/zz_generated.deepcopy.go
ev-shindin, line 1: `TestDeepCopyIndependence` tests that mutating a copy doesn't affect the original, but it doesn't exercise the new `ScalingBackend *ScalingBackendSpec` field. Since `ScalingBackend` contains nested pointers (`*HPAConfig` → `*Behavior`, `*KEDAConfig` → `*int32` fields, `*Fallback`, `*AdvancedConfig`), a deep-copy bug there could produce shared mutation. The generated `zz_generated.deepcopy.go` looks correct by inspection, but a test gives confidence:

ev-shindin (changes requested):
Hi. Thanks for PR. Please address review comments.

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -16,6 +17,99 @@ type VariantAutoscalingConfigSpec struct {
...
+// Prometheus metrics (metrics-only / legacy mode).
+type ScalingBackendSpec struct {
+	// Type is the scaling backend: "hpa" creates a controller-managed
+	// autoscaling/v2 HorizontalPodAutoscaler backed by the wva_desired_replicas
+	// external metric; "keda" creates a controller-managed keda.sh/v1alpha1
+	// ScaledObject with a prometheus trigger on the same metric.
+	// +kubebuilder:validation:Enum=hpa;keda
+	// +kubebuilder:default=hpa
+	// +kubebuilder:validation:Required
+	Type string `json:"type"`
```
lionelvillard, line 37: As discussed, there is no documented use case for allowing this type of choice at the VA level. Instead, I suggest to add a namespace-scoped configuration parameter.

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -16,6 +17,99 @@ type VariantAutoscalingConfigSpec struct {
...
+
+	// HPA contains HPA-specific configuration. Used when Type is "hpa".
+	// +kubebuilder:validation:Optional
+	HPA *HPAConfig `json:"hpa,omitempty"`
```
lionelvillard, line 41: I would remove the level of indirection and inline the common HPA/KEDA fields

api/v1alpha1/variantautoscaling_types.go
```diff
@@ -16,6 +17,99 @@ type VariantAutoscalingConfigSpec struct {
...
+
+	// Fallback configures KEDA fallback behavior when a scaler fails.
+	// Maps to ScaledObject.spec.fallback.
+	// +kubebuilder:validation:Optional
+	Fallback *kedav1alpha1.Fallback `json:"fallback,omitempty"`
+
+	// Advanced configures KEDA advanced options such as HPA behavior overrides
+	// and replica restoration. Maps to ScaledObject.spec.advanced.
+	// +kubebuilder:validation:Optional
+	Advanced *kedav1alpha1.AdvancedConfig `json:"advanced,omitempty"`
```
lionelvillard, line 98: I would not include any of these fields are they are not used by WVA. 

Kubernetes 1.22+ does a much better job at tracking multiple object owners allowing shared responsibility.  That may be the way to go.

go.mod
```diff
@@ -1,24 +1,52 @@
 module github.com/llm-d/llm-d-workload-variant-autoscaler
 
-go 1.25.0
+go 1.25.6
```
lionelvillard, line 3: why updating go version?

go.mod
```diff
@@ -1,24 +1,52 @@
...
-	github.com/onsi/gomega v1.39.0
+	github.com/onsi/gomega v1.39.1
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.89.0
 	github.com/spf13/viper v1.21.0
 	github.com/stretchr/testify v1.11.1
 	gonum.org/v1/gonum v0.17.0
-	k8s.io/apimachinery v0.34.5
-	k8s.io/client-go v0.34.5
-	sigs.k8s.io/controller-runtime v0.22.5
+	k8s.io/apimachinery v0.35.0
```
lionelvillard, line 13: this change should be done in a separate PR. One concern per PR.

asm582:
@vivekk16, can we vendor types that we absolutely need inside WVA instead of upgrading go, wdyt?

lionelvillard:
@vivekk16 any progress?

vivekk16:
@lionelvillard Will define what "stack" means first, then proceed from there.

@asm582 Using upstream KEDA API types as a dependency reduces the maintenance burden of replicating them, and gives users the flexibility to customize ScaledObjects directly, which was a common request from KServe users.

vivekk16:
Closing this in favor of VA deprecation.

PR #1032, opened 2026-04-20
tighten e2e fixture ensure correctness

mamy-CS:
Tightens e2e fixture correctness so `Ensure*` helpers only no-op when resources are both ready and match the requested config. Also removes hardcoded InfObj identity paths (adds named variants with default compatibility) and makes InfPool group detection fail fast instead of silently defaulting.

Local full e2e pass:
```
[AfterSuite] PASSED [0.050 seconds]

Ran 60 of 63 Specs in 1018.360 seconds
SUCCESS! -- 60 Passed | 0 Failed | 0 Pending | 3 Skipped
--- PASS: TestE2E (1018.36s)
PASS
ok      github.com/llm-d/llm-d-workload-variant-autoscaler/test/e2e     1019.030s
```

### DIFF
--- a/test/e2e/fixtures/lws_builder.go
+++ b/test/e2e/fixtures/lws_builder.go
@@ -3,6 +3,7 @@ package fixtures
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -17,13 +18,14 @@ import (
 // EnsureModelServiceLWS creates or replaces a LeaderWorkerSet for model service (idempotent for test setup).
 func EnsureModelServiceLWS(ctx context.Context, crClient client.Client, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int, groupSize int32) error {
 	lwsName := name + "-decode"
+	desiredLWS := buildModelServiceLWS(namespace, name, poolName, modelID, useSimulator, maxNumSeqs, groupSize)
 
 	// Check if LWS already exists
 	existingLWS := &lwsv1.LeaderWorkerSet{}
 	err := crClient.Get(ctx, client.ObjectKey{Name: lwsName, Namespace: namespace}, existingLWS)
 	if err == nil {
 		// LWS exists, check if it's ready
-		if existingLWS.Status.ReadyReplicas > 0 {
+		if existingLWS.Status.ReadyReplicas > 0 && modelServiceLWSMatchesDesired(existingLWS, desiredLWS) {
 			return nil
 		}
 		// Not ready, delete and recreate
@@ -31,10 +33,8 @@ func EnsureModelServiceLWS(ctx context.Context, crClient client.Client, k8sClien
 		deleteErr := crClient.Delete(ctx, existingLWS, &client.DeleteOptions{
 			PropagationPolicy: &propagationPolicy,
 		})
-		if deleteErr != nil && !errors.IsNotFound(deleteErr) {
-			if !errors.IsConflict(deleteErr) {
-				return fmt.Errorf("delete existing LWS %s: %w", lwsName, deleteErr)
-			}
+		if deleteErr != nil && !errors.IsNotFound(deleteErr) && !errors.IsConflict(deleteErr) {
+			return fmt.Errorf("delete existing LWS %s: %w", lwsName, deleteErr)
 		}
 		// Wait for deletion
 		waitCtx, cancel := context.WithTimeout(ctx, 2*time.Minute)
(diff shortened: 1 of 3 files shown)


test/e2e/fixtures/lws_builder.go
```diff
@@ -53,19 +53,27 @@ func EnsureModelServiceLWS(ctx context.Context, crClient client.Client, k8sClien
...
-	lws := buildModelServiceLWS(namespace, name, poolName, modelID, useSimulator, maxNumSeqs, groupSize)
-	err = crClient.Create(ctx, lws)
+	err = crClient.Create(ctx, desiredLWS)
 	if err != nil && errors.IsAlreadyExists(err) {
 		propagationPolicy := metav1.DeletePropagationForeground
-		_ = crClient.Delete(ctx, lws, &client.DeleteOptions{
+		_ = crClient.Delete(ctx, desiredLWS, &client.DeleteOptions{
 			PropagationPolicy: &propagationPolicy,
 		})
 		time.Sleep(2 * time.Second)
```
asm582, line 62: not related to this PR, but why do we have hardcoded sleep, can we use Eventually or waitforcond blocks?

test/e2e/fixtures/model_service_builder.go
```diff
@@ -39,20 +40,19 @@ func DeleteModelService(ctx context.Context, k8sClient *kubernetes.Clientset, na
 func EnsureModelService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int) error {
 	appLabel := name + "-decode"
 	deploymentName := appLabel
+	desiredDeployment := buildModelServiceDeployment(namespace, name, poolName, modelID, useSimulator, maxNumSeqs)
 
 	existingDeployment, err := k8sClient.AppsV1().Deployments(namespace).Get(ctx, deploymentName, metav1.GetOptions{})
 	if err == nil {
```
asm582, line 46: Why do we check err as nil?

test/e2e/fixtures/model_service_builder.go
```diff
@@ -71,11 +70,20 @@ func EnsureModelService(ctx context.Context, k8sClient *kubernetes.Clientset, na
...
 			return fmt.Errorf("timeout waiting for deployment %s to be deleted before recreate: %w", deploymentName, waitErr)
 		}
-		_, err = k8sClient.AppsV1().Deployments(namespace).Create(ctx, deployment, metav1.CreateOptions{})
+		_, err = k8sClient.AppsV1().Deployments(namespace).Create(ctx, desiredDeployment, metav1.CreateOptions{})
 	}
 	return err
 }
 
+func modelServiceDeploymentMatchesDesired(existing, desired *appsv1.Deployment) bool {
+	if existing == nil || desired == nil {
```
asm582: Can avoid passing nils and simplify logic.

test/e2e/fixtures/model_service_builder.go
```diff
@@ -71,11 +70,20 @@ func EnsureModelService(ctx context.Context, k8sClient *kubernetes.Clientset, na
 		if waitErr := WaitUntilDeploymentDeleted(ctx, k8sClient, namespace, deploymentName, 2*time.Minute); waitErr != nil {
 			return fmt.Errorf("timeout waiting for deployment %s to be deleted before recreate: %w", deploymentName, waitErr)
 		}
-		_, err = k8sClient.AppsV1().Deployments(namespace).Create(ctx, deployment, metav1.CreateOptions{})
+		_, err = k8sClient.AppsV1().Deployments(namespace).Create(ctx, desiredDeployment, metav1.CreateOptions{})
 	}
 	return err
 }
 
+func modelServiceDeploymentMatchesDesired(existing, desired *appsv1.Deployment) bool {
```
asm582: This function is repeated.

asm582 (changes requested):
Address review.

test/e2e/fixtures/lws_builder.go
mamy-CS: yes needs to be addressed, future pr

test/e2e/fixtures/model_service_builder.go
mamy-CS: will be addressed in future pr sweep

test/e2e/fixtures/model_service_builder.go
mamy-CS: Where is it repeated? Are you talking about `modelServiceLWSMatchesDesired`? If so, they’re repeated in structure, but they compare different resource types with different relevant fields. This is more of a nit, can be extracted with a helper but would keep as is.

mamy-CS:
The openshift e2e failure is expected, it's a cluster issue unrelated to pr changes @clubanderson fyi here for example as well. Full e2es on kind pass.

asm582 (approved):
/lgtm
/approve

It is agreed that subsequent PRs will include additional modifications to e2es to make them more robust and simpler.

PR #1036, opened 2026-04-21
 feat: add P/D role-aware grouping to V1 saturation analyzer path

### LINKED ISSUE #1031: V1 saturation analyzer does not support P/D disaggregation
### Problem

The V1 (percentage-based) saturation analyzer groups variants **by model only** (`GroupVariantAutoscalingByModel`). In P/D-disaggregated deployments — where a model is served by separate prefill and decode variants — this produces incorrect scaling behavior:

1. **Spare capacity averaging mixes roles.** Prefill variants typically have different KV-cache usage patterns than decode variants. Averaging them together produces a saturation signal that doesn't accurately represent either stage. The controller may over-scale one role and under-scale the other.

2. **Transition blocking is too broad.** If any variant in the model is transitioning (desired ≠ current or metrics ≠ current), `CalculateSaturationTargets` blocks ALL variants of the model from scaling. In a P/D deployment, a prefill variant scaling up should not prevent decode variants from responding to their own saturation signal.

3. **Scale-up/down targets the wrong variant.** The cheapest-variant selection for scale-up operates across all variants of the model. If the decode pipeline is saturated but the cheapest variant is a prefill variant, the controller scales up prefill — which doesn't relieve the decode bottleneck.

### Root cause

`internal/engines/saturation/engine.go` — `optimizeV1` iterates `modelGroups` (keyed by `modelID`) and passes all of a model's VAs to `RunSaturationAnalysis` as a single group. The V1 analyzer at `internal/saturation/analyzer.go` processes this group generically — it doesn't know about roles.

The V2 and queueing-model paths already handle P/D disaggregation via `RoleCapacities` in `AnalyzerResult`, but V1 was never updated.

### Expected behavior

Within each model, variants should be sub-grouped by role (prefill, decode, both). Each role group should get its own:

- Saturation analysis (spare KV-cache and queue capacity averaged only within the role)
- Transition blocking (only blocks variants of the same model + role)
- Scale-up decision (cheapest variant within the same role)
- Scale-down decision (most expensive variant within the same role)
- Scale-to-zero enforcement (evaluated per role group)

For non-disaggregated models (all variants have role `"both"` or `""`), behavior should be unchanged — a single role group is produced.

### Proposed fix

1. After `prepareModelData` collects metrics once per model (preserving the single Prometheus query pass), sub-group `variantStates` by role using a new `groupByRole` helper.

2. Filter replica metrics per role group via `filterReplicaMetricsByVariants` (no re-query — just filter the already-collected metrics by variant name membership).

3. Run `AnalyzeModelSaturation` + `CalculateSaturationTargets` + `convertSaturationTargetsToDecisions` per role group instead of per model.

4. The V1 saturation analyzer itself (`internal/saturation/analyzer.go`) does not need to change — it already operates generically on whatever group of variants it receives.

### Acceptance criteria

- [ ] `groupByRole` sub-groups states by normalized role (`""` and `"both"` treated identically)
- [ ] `filterReplicaMetricsByVariants` correctly filters metrics to the role's variant set
- [ ] Per-role saturation analysis produces independent spare-capacity averages
- [ ] Per-role transition blocking: prefill transitioning does not block decode scaling
- [ ] Per-role scale-up targets the cheapest variant within the role
- [ ] Per-role scale-down targets the most expensive variant within the role
- [ ] Non-disaggregated models behave identically to before (single group, no logic change)
- [ ] `Role` field set on V1 `VariantDecision` for downstream observability
- [ ] Unit tests for all helper functions
- [ ] Existing envtest suite passes unchanged (backward compatibility)
- [ ] `go build ./...` clean, `golangci-lint` 0 issues

### Related

- V2 path already handles P/D via `AnalyzerResult.RoleCapacities` — this brings V1 to parity
- V1 is on the deprecation path — but P/D support is needed until V1 is fully removed

ev-shindin:
### Summary

The V1 saturation analyzer grouped variants by model only. In P/D-disaggregated deployments (separate prefill and decode variants serving the same model), this produced incorrect scaling: spare capacity averaging mixed roles, transition blocking was too broad, and scale-up/down targeted the wrong pipeline stage.

This PR sub-groups variants by `(model, role)` for saturation analysis and scaling decisions, while keeping scale-to-zero enforcement at the model level (both P/D stages must scale together).

### Problem

In a P/D deployment with `prefill-h100` and `decode-l4` variants serving `llama-70b`:

1. **Mixed averaging** — prefill KV-cache usage averaged with decode KV-cache usage produces a signal that doesn't accurately represent either stage
2. **Over-broad blocking** — a prefill variant transitioning (desired ≠ current) blocked ALL variants of the model from scaling, including decode variants that could have responded to their own saturation
3. **Wrong target** — scale-up picked the cheapest variant across the entire model, not within the saturated pipeline stage (e.g., cheapest prefill variant when decode was the bottleneck)

### Design: three-level enforcement

```
Per role:   analyze → targets → decisions    (independent per role)
Per model:  scale-to-zero enforcement        (all roles together)
Per cycle:  GPU limiter                      (all models together)
```

- **Saturation analysis, transition blocking, scale-up/down** are per `(model, role)` — prefill and decode scale independently based on their own metrics.
- **Scale-to-zero** is per model — scaling prefill to zero while keeping decode (or vice versa) makes the model non-functional, so both stages must scale together. The enforcer sees all decisions for the model across all roles.
- **GPU limiter** is per cycle — unchanged, applied globally after all models.

### Changes

**`internal/engines/saturation/engine.go`**

- `optimizeV1` refactored: after `prepareModelData` collects metrics once per model (single Prometheus query pass preserved), variants are sub-grouped by role via `groupByRole`. Each role group runs `AnalyzeModelSaturation` + `CalculateSaturationTargets` + `convertSaturationTargetsToDecisions` independently.
- Transition blocking now operates per `(model, role)` — prefill transitioning does not block decode.
- Scale-up targets the cheapest variant within the role; scale-down targets the most expensive within the role.
- Scale-to-zero enforcement applied per model after role loop — `hasMinReplicasAboveZero` checks all variants regardless of role, enforcer sees all decisions across roles.
- Safety-net metrics emitted per role group on analysis failure (via `filterVAsByVariantStates`).
- Role group iteration is deterministic (sorted keys via `sortedRoleKeys`).
- `scaleToZeroConfig` read once per model (hoisted out of role loop).
- `Role` field set on V1 `VariantDecision` (was previously unset).
- Removed dead `RunSaturationAnalysis` method (no callers after refactor).

New helper functions:
- `groupByRole(states)` — sub-groups by normalized role (`""` and `"both"` treated identically)
- `filterReplicaMetricsByVariants(metrics, states)` — filters metrics to a role's variant set without re-querying Prometheus
- `filterVAsByVariantStates(vas, states)` — maps role group back to source VAs for safety-net emission
- `sortedRoleKeys(groups)` — deterministic iteration order
- `normalizeRole(role)` — `""` → `"both"`
- `variantNames(states)` — for logging

**`internal/engines/saturation/role_grouping_test.go`** (new)

16 unit tests covering all helper functions:
- `TestNormalizeRole` — 4 cases (empty, both, prefill, decode)
- `TestGroupByRole` — 4 cases (non-P/D, P/D, mixed, empty)
- `TestFilterReplicaMetricsByVariants` — 4 cases (single role, all, none)
- `TestFilterVAsByVariantStates` — 3 cases (single role, all, empty)
- `TestVariantNames` — 1 case

### Backward compatibility

For non-disaggregated models (all variants have role `"both"` or `""`):
- `groupByRole` produces a single entry `{"both": [all states]}`
- `filterReplicaMetricsByVariants` returns all metrics (identity)
- Single call to `AnalyzeModelSaturation` with the same inputs → same analysis
- Single call to `CalculateSaturationTargets` with the same states → same targets, same transition blocking
- Scale-to-zero applied to the single group's decisions → same as before
- **Behavior is identical to pre-refactor**

Verified by: existing envtest suite passes unchanged.

### Test plan

- [x] `go build ./...` — clean
- [x] `go test ./internal/engines/saturation/...` (envtest via WSL) — all pass (existing suite + 16 new unit tests)
- [x] `go test ./internal/saturation/...` — pass
- [x] `golangci-lint run ./internal/engines/saturation/...` — 0 issues
- [x] CI pipelines green on PR

### DIFF
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -19,6 +19,7 @@ package saturation
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strconv"
 	"time"
 
@@ -317,6 +318,12 @@ func (e *Engine) optimize(ctx context.Context) error {
 
 // optimizeV1 runs the V1 percentage-based saturation analysis path (saturation-percentage-based).
 // Processes each model independently: analyze → enforce → convert → limiter.
+//
+// For P/D disaggregation: within each model, variants are sub-grouped by role
+// (prefill, decode, both). Each role group gets its own saturation analysis,
+// transition blocking, and scale-up/down decisions. This ensures that a prefill
+// variant transitioning doesn't block decode scaling, and that spare-capacity
+// averaging doesn't mix semantically different workload stages.
 func (e *Engine) optimizeV1(
 	ctx context.Context,
 	modelGroups map[string][]llmdVariantAutoscalingV1alpha1.VariantAutoscaling,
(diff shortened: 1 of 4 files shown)


lionelvillard:
### Code Review

Found 6 issues:

**Go Code Quality**

1. `RunSaturationAnalysis` removed without deprecation notice — `internal/engines/saturation/engine.go`
   Being exported (`*Engine.RunSaturationAnalysis`), it is part of the package's public surface. Any caller outside the file (test binary, integration harness) that imports the `saturation` package and calls it will break at compile time. If the removal is intentional, at minimum document it in the PR description (the PR description does not mention the removal — the "Changes" section only says "Removed dead `RunSaturationAnalysis` method"). If the method has external callers in any release branch, a `// Deprecated:` notice should precede the removal.
   > AGENTS.md: "Every exported name should have a doc comment"

2. Double `ApplyDefaults()` call on the resolved config — [`engine.go#L354-L355`](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/b18d547ce3a9a33f709ccbc287b449e83ebb73bf/internal/engines/saturation/engine.go#L354-L355)
   `resolveSaturationConfig` already calls `ApplyDefaults()` on every return path. The extra call added in the new V1 path is dead code that misleads readers into thinking defaults are not applied inside the helper.
   ```go
   saturationConfig := resolveSaturationConfig(saturationConfigMap, modelID, namespace)
   saturationConfig.ApplyDefaults()  // redundant
   ```

**Test Coverage**

3. The `optimizeV1` role-grouping loop is entirely untested at the unit level — `internal/engines/saturation/engine.go`
   `role_grouping_test.go` tests only the four helper functions. The new per-role iteration in `optimizeV1` — running `AnalyzeModelSaturation` per role group, merging per-role decisions into `modelDecisions`, and passing `modelDecisions` (all roles combined) into `ScaleToZeroEnforcer` — is exercised by zero unit tests. The envtest integration suite bypasses this path because it returns no replica metrics, so `prepareModelData` returns nil and skips the loop.
   Suggested additions for a `TestOptimizeV1RoleGrouping` table:
   - **P/D model**: two VAs for the same model (`role=prefill`, `role=decode`) — assert two independent `AnalyzeModelSaturation` calls and both decisions in output.
   - **homogeneous-role model**: all VAs have no role — assert single-group behavior is unchanged.
   - **scale-to-zero with mixed roles**: `ScaleToZeroEnforcer` receives merged decisions from all groups.

4. Safety-net metrics scoped to a per-role failure are untested — `internal/engines/saturation/engine.go`
   The PR scopes `emitSafetyNetMetrics` to only the failing role group's VAs. There is no test where one role group analysis fails while another succeeds, so the scoping cannot be verified. Suggested: add a case to `TestOptimizeV1RoleGrouping` where the prefill-group stub returns an error while decode succeeds; assert that safety-net emission is called only for prefill VAs and decode decisions still propagate.

5. `Role` field on `VariantDecision` is never asserted — [`engine_test.go`](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/b18d547ce3a9a33f709ccbc287b449e83ebb73bf/internal/engines/saturation/engine_test.go)
   The PR adds `Role: state.Role` to `convertSaturationTargetsToDecisions`. No existing test checks `decision.Role`. Extend the existing `convertSaturationTargetsToDecisions` test to populate `Role` in `variantStates` and assert the propagated value.

**Security**

No issues found. Checked RBAC scope, role string sanitization (already allowlisted in `getRoleFromScaleTarget`), secret handling, and the `RunSaturationAnalysis` removal.

<sub>If this review was useful, react with 👍. Otherwise, react with 👎.</sub>

mamy-CS:
rebase @ev-shindin please

ev-shindin:
@lionelvillard its waiting for your review.

lionelvillard (approved):

PR #1042, opened 2026-04-21
Fix: forward WVA_NS/LLMD_NS in undeploy, always run helm uninstall and export namespace for wva metrics

asm582:
- Makefile: pass WVA_NS and LLMD_NS to deploy/install.sh --undeploy so custom namespaces are respected during undeployment
- cleanup.sh: move helm uninstall commands outside the EXAMPLE_DIR check so Helm releases are always cleaned up, even when the cloned llm-d repo directory does not exist locally
- Added namespace filter to expose WVA metrics when installed namespace-scoped.

Co-authored by Gemini.


mamy-CS (dismissed):
/lgtm

config/samples/prometheus-adapter-values-ocp.yaml
```diff
@@ -12,7 +12,7 @@ rules:
     name:
       matches: "^wva_desired_replicas"
       as: "wva_desired_replicas"
-    metricsQuery: 'wva_desired_replicas{<<.LabelMatchers>>}'
+    metricsQuery: 'max by (exported_namespace, variant_name) (wva_desired_replicas{<<.LabelMatchers>>})'
```
mamy-CS, line 15: not a correct fix here, this returns the highest value. Correct fix is appropriately using exported_namespace, and controller_instance

asm582:
Closing in favor of #1049

Issue #1044, opened 2026-04-22
Incorrect scaling behavior with HPA when multiple WVA installs exists on cluster


asm582:
HPA scaling up without load:

```
[15s] replicas: spec=3 ready=1 | va=1 hpa=1→3 | kv=0.0000 qd=0.0 epp_qd=0.0
    [30s] replicas: spec=3 ready=1 | va=1 hpa=3→3 | kv=0.0000 qd=0.0 epp_qd=0.0
    [45s] replicas: spec=4 ready=1 | va=1 hpa=3→4 | kv=0.0079 qd=0.0 epp_qd=0.0
    [60s] replicas: spec=4 ready=1 | va=1 hpa=4→4 | kv=0.0079 qd=0.0 epp_qd=0.0
    [75s] replicas: spec=4 ready=1 | va=1 hpa=4→4 | kv=0.8523 qd=0.0 epp_qd=0.0
    [90s] replicas: spec=4 ready=1 | va=1 hpa=4→4 | kv=0.8523 qd=0.0 epp_qd=0.0
    [105s] replicas: spec=4 ready=1 | va=1 hpa=4→4 | kv=0.9988 qd=137.0 epp_qd=0.0
    [120s] replicas: spec=4 ready=1 | va=1 hpa=4→4 | kv=0.4994 qd=68.5 epp_qd=0.0
    [135s] replicas: spec=4 ready=3 | va=1 hpa=4→4 | kv=0.2499 qd=60.2 epp_qd=0.0
    [150s] replicas: spec=4 ready=3 | va=1 hpa=4→4 | kv=0.2499 qd=60.2 epp_qd=28.0
    [165s] replicas: spec=4 ready=4 | va=1 hpa=4→4 | kv=0.2496 qd=50.0 epp_qd=98.0
    [180s] replicas: spec=4 ready=4 | va=1 hpa=4→4 | kv=0.2496 qd=50.0 epp_qd=21.0
    [195s] replicas: spec=4 ready=4 | va=1 hpa=4→4 | kv=0.2331 qd=0.0 epp_qd=0.0
    [210s] replicas: spec=4 ready=4 | va=1 hpa=4→4 | kv=0.3188 qd=0.0 epp_qd=0.0
    [225s] replicas: spec=4 ready=4 | va=1 hpa=4→4 | kv=0.4204 qd=0.0 epp_qd=0.0
[...]

PR #1049, opened 2026-04-22
multi tenant patches and guards

mamy-CS:
Addresses https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1044
- add hpa selector safety (safer defaults for shared clusters) 
- Added namespace-isolation and dual-controller isolation e2e coverage 

Generated with the help of 🤖 [Claude Code](https://claude.com/claude-code)


cmd/main.go
```diff
@@ -374,6 +376,28 @@ func main() {
 		}
 	}
 
+	controllerInstance := os.Getenv(metrics.ControllerInstanceEnvVar)
```
asm582: I don't think we need this. Can't we enforce the controller to watch a desired namespace?

asm582 (changes requested):
Address review

deploy/lib/cleanup.sh
```diff
@@ -29,6 +29,12 @@ undeploy_prometheus_adapter() {
     # Stop the APIService guard if running
     stop_apiservice_guard
 
+    if [ "${ALLOW_SHARED_METRICS_MUTATION:-false}" != "true" ]; then
```
asm582: why is this needed, I can do `DEPLOY_PROMETHEUS_ADAPTER-= false` , correct?

deploy/lib/scaler_runtime.sh
```diff
@@ -262,24 +262,33 @@ deploy_prometheus_adapter() {
...
         else
             log_warning "external.metrics.k8s.io APIService points to '$current_svc' in '$current_ns'"
-            log_info "Patching APIService to point to Prometheus Adapter in $MONITORING_NAMESPACE"
-            kubectl patch apiservice "$EXTERNAL_METRICS_APISERVICE_NAME" --type=merge -p "{
-                \"spec\": {
-                    \"insecureSkipTLSVerify\": true,
-                    \"service\": {
-                        \"name\": \"$PROMETHEUS_ADAPTER_SERVICE_NAME\",
-                        \"namespace\": \"$MONITORING_NAMESPACE\"
+            if [ "${ALLOW_SHARED_METRICS_MUTATION:-false}" != "true" ]; then
```
asm582: can DEPLOY_PROMETHEUS_ADAPTER-= false avoid this code?

asm582 (changes requested):
address simplification changes.

test/e2e/smoke_test.go
```diff
@@ -203,6 +239,277 @@ var _ = Describe("Smoke Tests - Infrastructure Readiness", Label("smoke", "full"
...
+			if primaryNamespace == "" {
+				primaryNamespace = cfg.LLMDNamespace
+			}
+			if primaryNamespace == "" {
+				// SharedConfig normally sets LLMD namespace, but keep a defensive default to avoid invalid names.
+				primaryNamespace = "llm-d-sim"
+			}
+			if secondaryNamespace == "-mt" {
+				secondaryNamespace = primaryNamespace + "-mt"
+			}
```
asm582: The test should explicitly mention the namespace. Simplify this code.

asm582 (changes requested):

test/e2e/smoke_test.go
```diff
@@ -20,6 +24,38 @@ import (
...
+func splitImage(image string) (string, string) {
+	lastColon := strings.LastIndex(image, ":")
+	lastSlash := strings.LastIndex(image, "/")
+	if lastColon == -1 || lastColon < lastSlash {
+		return image, "latest"
+	}
+	return image[:lastColon], image[lastColon+1:]
+}
+
+func resolveWVAChartPath() (string, error) {
```
asm582: why is this function required?

test/e2e/smoke_test.go
```diff
@@ -203,6 +239,252 @@ var _ = Describe("Smoke Tests - Infrastructure Readiness", Label("smoke", "full"
 		})
 	})
 
+	Context("External metrics namespace isolation", Serial, Ordered, func() {
```
asm582, line 226: Both new tests should also check HPA status.

asm582 (changes requested):
changes needed

test/e2e/smoke_test.go
mamy-CS: Initially kept this test metric-only because the bug was specifically about selector isolation in the external metrics API (variant_name + exported_namespace), so validating the returned series directly was the most focused check. Based on your feedback I added hpa status checks so the test now covers the end-to-end path.

test/e2e/smoke_test.go
mamy-CS: This was for a deterministic e2e run, regardless of where the e2e dir was run from but pinned the path for e2e to simplify further.

asm582 (dismissed):
/lgtm
/approve

Issue #1005, opened 2026-04-14
[Autoscaling] feat: Throughput Analyzer — proactive rate-based autoscaling


deanlorenz:
## Summary

Add a **Throughput Analyzer** that provides a rate-based (velocity) scaling signal complementing the existing Saturation Analyzer. Unlike the Saturation Analyzer, which detects overload only after KV cache or queue depth crosses a threshold, the Throughput Analyzer detects when the demand trajectory will exhaust capacity *before* saturation occurs — enabling anticipatory scaling **30–60 seconds earlier**.

Related: #834 (parent epic), #901 (experiments), #902 (initial v0), #903 (benchmarks).

---

## Background

The Saturation Analyzer (V1/V2) makes scaling decisions from **instantaneous state** — current KV cache utilization and queue depth. By the time saturation is visible, users are already experiencing degraded latency.

The Throughput Analyzer tracks **rate signals** instead: token arrival rate vs. observed throughput capacity per replica. When arrival rate exceeds service rate, it produces a positive `RequiredCapacity` signal — even before saturation is detectable.

### Relationship to Saturation Analyzer V2

| Aspect | Saturation Analyzer V2 | Throughput Analyzer |
|---|---|---|
| Signal type | Instantaneous state (level) | Rate of change (velocity) |
| Units | Tokens (absolute capacity) | Tokens/sec (throughput) |
| Detects | Current overload | Approaching overload |
| Strength | Accurate when saturated | Early detection during traffic ramps |
| Weakness | Reactive — detects after the fact | May overshoot on transient spikes |
| Scale-up trigger | Demand vs. capacity headroom | Demand rate vs. throughput capacity |
| Scale-down trigger | Spare capacity | Sustainable processing rate |

The two analyzers run **in parallel** using the multi-analyzer framework (Proposal 4):
- **Scale-up**: OR logic — either analyzer can trigger
- **Scale-down**: AND logic — both analyzers must agree

---

## Design: Triple-Channel Model

Input tokens (prefill) and output tokens (decode) have fundamentally different computational costs in vLLM — treating them as equivalent produces meaningless throughput numbers. The Throughput Analyzer uses a **triple-channel model**, tracking demand and supply separately for each:

| Channel | What it measures | Unit |
|---|---|---|
| T1 — Request rate | Requests arriving at the model server | Requests/sec (RPS) |
| T2 — Prefill rate | Input tokens being processed (prompt ingestion) | Prefill tokens/sec |
[...]

PR #1051, opened 2026-04-23
collector/registration: register throughput analyzer queries (PR-1, PR-2)

deanlorenz:
## Summary

Registers the Prometheus queries and collector fields for the Throughput Analyzer's
metrics collection layer (PR-1 + PR-2 from the TA plan, combined into one PR since
the query set is cohesive and there is no useful intermediate state).

**Queries registered** in `RegisterThroughputAnalyzerQueries`:

| Query | Metric | Purpose |
|-------|--------|---------|
| `generation_token_rate` | `vllm:request_generation_tokens_sum` | μ_dec^obs per pod (1m rate) |
| `kv_usage_instant` | `vllm:kv_cache_usage_perc` | k* — current KV utilization (instantaneous, no max_over_time) |
| `vllm_request_rate` | `vllm:request_generation_tokens_count` | λ_dec fallback when EPP is not deployed (req/s) |

**Design principle:** only queries that are *not* already covered by
`RegisterSaturationQueries` or `RegisterQueueingModelQueries` are registered here.
All other TA inputs (KV capacity, ITL, output/input token averages, prefix cache hit
rate, scheduler dispatch rate) are read from `interfaces.ReplicaMetrics` fields
that those existing registrations populate.

**Collector wiring:** `GenerationTokenRate`, `KvUsageInstant`, and
`VLLMRequestRate` fields added to `podMetricData` and wired in
`internal/collector/replica_metrics.go`.

**Not included:** the next PRs (PR-3 through PR-5) will add the `ThroughputAnalyzer`
struct, ITL model calibration, and μ_dec vs λ_dec scaling signal under
`internal/engines/analyzers/throughput/`.

## Related

- Part of #1005

## Test plan

- [x] `make test` passes (unit tests verify all 3 TA-exclusive queries register,
  have `QueryTypePromQL` type, build templates with correct metric names and rate
  windows, and that calling `RegisterThroughputAnalyzerQueries` twice on the same
  registry panics)
- [x] No changes to existing analyzers (saturation, queueing model, scale-to-zero)


docs/developer-guide/throughput-analyzer.md
```diff
@@ -0,0 +1,220 @@
...
+predicts how much decode throughput the current replica fleet can sustain at a given KV cache
+operating point, and scales before demand exceeds that supply.
+
+**Key concepts:**
+- **μ_dec** — decode token supply: how many tokens/sec the fleet can generate, estimated from
+  KV cache occupancy and a calibrated inter-token latency (ITL) model
+- **λ_dec** — decode token demand: how many tokens/sec the scheduler is dispatching to this model
+
+> **Status:** This document covers the metrics collection layer (PR-1, PR-2) and the internal
+> state management package (PR-3). The ITL model fit, μ_dec supply estimation, and scaling
```
ev-shindin: It looks like PR-3 is not covered here....

docs/developer-guide/throughput-analyzer.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Metrics Collector (`internal/collector/replica_metrics.go`)**  
+Populates all `interfaces.ReplicaMetrics` fields in a single `Refresh()` call covering all
+12 registered queries. The three TA-exclusive fields are:
+`GenerationTokenRate`, `KvUtilization`, `VLLMRequestRate`.
+The remaining TA fields (`TotalKvCapacityTokens`, `AvgITL`, `AvgOutputTokens`, `AvgInputTokens`,
+`PrefixCacheHitRate`, `ArrivalRate`) are populated by saturation and queueing model queries.
+
+**Throughput Analyzer (`internal/engines/analyzers/throughput/`)**  
+*(PR-3 complete; ITL model and scaling signal in PR-4)*
```
ev-shindin: This section describes `internal/engines/analyzers/throughput/`, `ShapeTracker`, `ObservationWindow`, and `ThroughputAnalyzer` as PR-3 deliverables. None of those exist in this diff — the PR title and description scope this to PR-1 + PR-2.

Either:
- Trim the doc to describe only what's in this PR (queries + collector fields + the TA-notation reference table), and ship the architecture section with the PR that introduces those types; **or**
- Move the architecture section under a clearly-marked "Pending (PR-3 / PR-4)" callout so a maintainer doesn't grep for a `ShapeTracker` that doesn't yet exist.

docs/developer-guide/throughput-analyzer.md
ev-shindin, line 1: This file reads mostly as developer documentation — μ_dec / λ_dec model derivation, internal package paths, OLS calibration plan, query design rationale. Documentation in this project organized into:

> 1. **Developer Documentation** — Architecture decisions … usually in the `docs/developer-guide/` subdirectory
> 2. **Administrator Documentation** — Installation and configuration …
> 3. **End-User Documentation** — Usage guides and examples …

There's almost nothing here an end-user or operator does (no install, no configure, no operate). Suggest moving to `docs/developer-guide/throughput-analyzer.md`. Optional split: keep a short user-guide page about *enabling* the analyzer once PR-4 lands, and put the model derivation in developer-guide.

internal/collector/registration/throughput_analyzer_test.go
```diff
@@ -0,0 +1,103 @@
...
+	Context("when prometheus source is registered", func() {
+		BeforeEach(func() {
+			metricsSource := prometheus.NewPrometheusSource(ctx, mockAPI, prometheus.DefaultPrometheusSourceConfig())
+			err := registry.Register("prometheus", metricsSource)
+			Expect(err).NotTo(HaveOccurred())
+			RegisterThroughputAnalyzerQueries(registry)
+			queryList = registry.Get("prometheus").QueryList()
+		})
+
+		It("should not panic during registration", func() {
```
ev-shindin: This case creates a *second* fresh registry and asserts that the first registration on it doesn't panic. But the surrounding `BeforeEach` already calls `RegisterThroughputAnalyzerQueries(registry)`, and the next test (`should register exactly the three TA-exclusive queries`) implicitly proves that succeeded — so this case is redundant.

What would actually be valuable here is testing that calling `RegisterThroughputAnalyzerQueries` **twice on the same registry** panics (since `MustRegister` panics on duplicate names) — or doesn't, depending on the intended idempotency. Either rewrite to that semantic, or drop the test.

internal/collector/replica_metrics.go
```diff
@@ -391,6 +399,69 @@ func (c *ReplicaMetricsCollector) CollectReplicaMetrics(
 		}
 	}
 
+	// Process generation token rate results (tokens/sec) — throughput analyzer μ_dec^obs
+	if result := results[registration.QueryGenerationTokenRate]; result != nil {
```
ev-shindin, line 489: As we discussed:  three new blocks (`GenerationTokenRate`, `KvTokensUsed`, `VLLMRequestRate`) are syntactically near-clones — same `result == nil` guard, same `pod` → `pod_name` label fallback, same `if podData[podName] == nil { ... }` init, same NaN/Inf/range validation, differing only in the field assigned and the validator predicate. This matches the pre-existing pattern in this file (every other query handler uses the same shape), so it's *consistent* — flagging as observation only, not blocking. A future cleanup could extract a `forEachPodValue(result, validate, assign)` helper that all 12 query handlers in this file would benefit from.

All three new blocks (and existing ones throughout this file) try `value.Labels["pod"]` then `value.Labels["pod_name"]`. If both labels are present but disagree, `pod` silently wins. A one-time top-of-file comment documenting which underlying PromQL query returns which label (and why both fallbacks exist) would help future readers — but again, pre-existing pattern, so flagging as observation only.

ev-shindin (commented):
Great work! There are some minor things to address.

docs/developer-guide/throughput-analyzer.md
deanlorenz: Fixed in 04f6571. The status callout now reads:

> **Status (PR-1/PR-2 — query registration + collector wiring):** Three PromQL queries registered; `GenerationTokenRate`, `KvUsageInstant`, and `VLLMRequestRate` fields wired in the collector. Full analyzer implementation (ShapeTracker, ObservationWindow, ITL model, `ThroughputAnalyzer`) and engine wiring are pending PR-3 through PR-5.

Also renamed `QueryKvTokensUsed` → `QueryKvUsageInstant` and `KvUtilization` → `KvUsageInstant` throughout the doc, and added pending markers to ITL Model Calibration, Supply Estimation, Demand Estimation, and Scaling Signal sections.

docs/developer-guide/throughput-analyzer.md
deanlorenz: Fixed in 04f6571. Added a note block at the top of the Architecture section:

> **Note:** The Query Registration and Metrics Collector components below are implemented in PR-1/PR-2. All other components — Package Structure, ShapeTracker, ObservationWindow, ITLModel, ThroughputAnalyzer, Analysis Pipeline, and Data Flow — are **pending PR-3/PR-4**.

Pending markers also added to the ITL Model Calibration, Supply Estimation, Demand Estimation, and Scaling Signal sections.

docs/developer-guide/throughput-analyzer.md
deanlorenz: Fixed in fb8f9c4. Moved to `docs/developer-guide/throughput-analyzer.md`.

internal/collector/registration/throughput_analyzer_test.go
deanlorenz: Will fix. The redundant case will be replaced with a duplicate-registration panic test (`MustRegister` called twice on the same registry with the same name). Addressing in the next commit.

internal/collector/replica_metrics.go
deanlorenz: Will fix. A top-of-file comment explaining the `pod`/`pod_name` label fallback pattern will be added to `replica_metrics.go`. The three new TA blocks follow the same pattern as the existing ones. Addressing in the next commit.

internal/collector/registration/throughput_analyzer_test.go
deanlorenz: Fixed in e84207d. Replaced the redundant case with a duplicate-registration panic test:

```go
It("should panic when RegisterThroughputAnalyzerQueries is called twice on the same registry", func() {
    // MustRegister panics on duplicate names; calling the function a second
    // time on the same registry (queries already registered by BeforeEach)
    // must trigger that panic.
    Expect(func() {
        RegisterThroughputAnalyzerQueries(registry)
    }).To(Panic())
})
```

internal/collector/replica_metrics.go
deanlorenz: Fixed in e84207d. Added a top-of-file doc comment to `replica_metrics.go` explaining the fallback pattern:

```go
// # Pod label fallback
//
// Every processing block in Refresh() extracts a pod identity from Prometheus
// labels using a two-step fallback:
//
//     podName := value.Labels["pod"]
//     if podName == "" {
//         podName = value.Labels["pod_name"]
//     }
//
// vLLM metrics are typically scraped via a PodMonitor or ServiceMonitor that
// applies the Prometheus operator's default target-relabeling, which produces
// a "pod" label. Some scrape configurations (e.g., raw Prometheus scrape jobs,
// kube-state-metrics-style configs) instead expose the pod identity as
// "pod_name". The fallback handles both conventions so the collector works
// regardless of how the Prometheus scrape is configured.
```

internal/collector/registration/throughput_analyzer.go
```diff
@@ -0,0 +1,136 @@
...
+//
+// TA notation → ReplicaMetrics field (query / registration):
+//
+//	KV_max  (total KV token capacity) → TotalKvCapacityTokens  (QueryCacheConfigInfo       / RegisterSaturationQueries)
+//	ITL_obs (observed ITL, seconds)   → AvgITL                 (QueryAvgITL                / RegisterQueueingModelQueries)
+//	OL      (avg output tokens)       → AvgOutputTokens        (QueryAvgOutputTokens       / RegisterSaturationQueries)
+//	IL      (avg input tokens)        → AvgInputTokens         (QueryAvgInputTokens        / RegisterSaturationQueries)
+//	H%      (prefix cache hit rate)   → PrefixCacheHitRate     (QueryPrefixCacheHitRate    / RegisterSaturationQueries)
+//	λ_req   (per-pod arrival rate)    → ArrivalRate            (QuerySchedulerDispatchRate / RegisterQueueingModelQueries)
+//	         λ_dec = ArrivalRate × AvgOutputTokens, computed in analyzer
```
ev-shindin: The developer guide spells it as `Σ_r (rate_r × tokens_r)` — per-replica product, then sum. These are mathematically different when OL varies across replicas:
- `sum(rate) × avg(tokens)` — assumes uniform OL, weights all rates by the average
- `Σ_r rate_r × tokens_r` — each replica's rate weighted by its own OL

For heterogeneous replicas (which is the whole point of the variant-aware analyzer) the two diverge.

internal/collector/registration/throughput_analyzer.go
```diff
@@ -0,0 +1,136 @@
...
+	// same value, max = avg — the choice has no effect on correctness.
+	// Source: vllm:kv_cache_usage_perc (gauge)
+	QueryKvUsageInstant = "kv_usage_instant"
+
+	// QueryVLLMRequestRate is the query name for the vLLM-side request completion
+	// rate per pod (req/s), derived from the generation tokens histogram count.
+	//
+	// Used as a fallback for λ_dec estimation when EPP/scheduler metrics are
+	// unavailable (ArrivalRate == 0 for all pods). The analyzer computes:
+	//   λ_dec_vllm = sum(VLLMRequestRate) × avg(AvgOutputTokens)
```
ev-shindin: Same as above: 
The developer guide spells it as `Σ_r (rate_r × tokens_r)` — per-replica product, then sum. These are mathematically different when OL varies across replicas:
- `sum(rate) × avg(tokens)` — assumes uniform OL, weights all rates by the average
- `Σ_r rate_r × tokens_r` — each replica's rate weighted by its own OL

For heterogeneous replicas (which is the whole point of the variant-aware analyzer) the two diverge.

internal/collector/registration/throughput_analyzer.go
```diff
@@ -0,0 +1,136 @@
...
+//
+// μ_dec is computed using a linear ITL model:
+//
+//	ITL(k)   = A·k + B            (calibrated from AvgITL × k* pairs over time)
+//	IL_eff   = IL × (1 - H%)
+//	KV_req   = IL_eff + OL/2
+//	N_dec(k) = k × KV_max / KV_req
+//	μ_dec    = N_dec(k_sat) / ITL(k_sat)
+//
+// λ_dec primary:  sum(ArrivalRate) × avg(AvgOutputTokens)         [EPP deployed]
```
ev-shindin: Same as above: 
The developer guide spells it as `Σ_r (rate_r × tokens_r)` — per-replica product, then sum. These are mathematically different when OL varies across replicas:
- `sum(rate) × avg(tokens)` — assumes uniform OL, weights all rates by the average
- `Σ_r rate_r × tokens_r` — each replica's rate weighted by its own OL

For heterogeneous replicas (which is the whole point of the variant-aware analyzer) the two diverge.

internal/collector/registration/throughput_analyzer.go
```diff
@@ -0,0 +1,136 @@
...
+		Template:    `max by (pod) (vllm:kv_cache_usage_perc{namespace="{{.namespace}}",model_name="{{.modelID}}"})`,
+		Params:      []string{source.ParamNamespace, source.ParamModelID},
+		Description: "Instantaneous KV cache utilization per pod (0.0–1.0), used as k* in the ITL model",
+	})
+
+	// Per-pod vLLM request completion rate (req/s).
+	// Derived from the generation tokens histogram _count (increments once per
+	// completed request). Used as a fallback for λ_dec when EPP/scheduler metrics
+	// are unavailable; the analyzer falls back to:
+	//   λ_dec_vllm = sum(VLLMRequestRate) × avg(AvgOutputTokens)
```
ev-shindin: Same as above: 
The developer guide spells it as `Σ_r (rate_r × tokens_r)` — per-replica product, then sum. These are mathematically different when OL varies across replicas:
- `sum(rate) × avg(tokens)` — assumes uniform OL, weights all rates by the average
- `Σ_r rate_r × tokens_r` — each replica's rate weighted by its own OL

For heterogeneous replicas (which is the whole point of the variant-aware analyzer) the two diverge.

docs/developer-guide/throughput-analyzer.md
```diff
@@ -0,0 +1,464 @@
...
+       │ vllm:request_generation_tokens_sum      (QueryGenerationTokenRate   → GenerationTokenRate)
+       │ vllm:kv_cache_usage_perc                (QueryKvUsageInstant          → KvUsageInstant)
+       │ vllm:request_generation_tokens_count    (QueryVLLMRequestRate       → VLLMRequestRate)
+       │ vllm:cache_config_info                  (QueryCacheConfigInfo       → TotalKvCapacityTokens)
+       │ vllm:inter_token_latency_seconds_*      (QueryAvgITL               → AvgITL)
+       │ vllm:request_generation_tokens_*        (QueryAvgOutputTokens       → AvgOutputTokens)
+       │ vllm:request_prompt_tokens_*            (QueryAvgInputTokens        → AvgInputTokens)
+       │ vllm:prefix_cache_hits/queries          (QueryPrefixCacheHitRate    → PrefixCacheHitRate)
+       │ inference_extension_scheduler_*         (QuerySchedulerDispatchRate → ArrivalRate)
+       │ inference_extension_flow_control_*      (QueryFlowControlQueueSize  → QueueSize)
```
ev-shindin: `QueryFlowControlQueueSize` is not registered anywhere in the codebase. The actual constant is `QuerySchedulerQueueSize` (referenced at `internal/collector/replica_metrics.go:696` and `:697`).

internal/collector/registration/throughput_analyzer.go
```diff
@@ -0,0 +1,136 @@
...
+	// ArrivalRate (via QuerySchedulerDispatchRate) as the primary demand source.
+	// Source: vllm:request_generation_tokens_count (histogram _count counter)
+	QueryVLLMRequestRate = "vllm_request_rate"
+)
+
+// RegisterThroughputAnalyzerQueries registers the three TA-exclusive queries.
+// It must be called once at engine startup alongside other analyzer registrations.
+//
+// Registered queries:
+//   - QueryGenerationTokenRate — μ_dec^obs: observed decode token rate per pod
```
ev-shindin, line 73: Can you please align list?

internal/collector/replica_metrics.go
ev-shindin: This `Refresh()` call now fans out 12 queries spanning saturation, queueing, and throughput analyzers. If the refresh fails, the error message blames "saturation metrics" specifically, which is misleading once a non-saturation query is in the mix. Suggested rewording: "failed to refresh replica metrics".

Pre-existing wording, but this PR widened the scope of the call, so it's a fair time to update.

ev-shindin (commented):
Please address review comments.

ev-shindin:
Hi @deanlorenz . Please check my comments. Also please rebase your PR.

internal/collector/registration/throughput_analyzer.go
deanlorenz: Fixed in 900c94c. All four formula comments now use `Σ_{r∈V}(rate_r × tokens_r)`, making the per-variant scope explicit — `r` ranges over replicas of variant V only, which matches the analyzer implementation (`computeDemand` receives `variantMetrics`, a slice scoped to one variant). Also updated the equivalent formulas in the developer guide (lines 111, 140–141, 371, 379, 387) to use the same notation.

internal/collector/registration/throughput_analyzer.go
deanlorenz: Same as above — fixed in 900c94c.

internal/collector/registration/throughput_analyzer.go
deanlorenz: Same as above — fixed in 900c94c.

internal/collector/registration/throughput_analyzer.go
deanlorenz: Same as above — fixed in 900c94c.

docs/developer-guide/throughput-analyzer.md
deanlorenz: Fixed in 900c94c. `QueryFlowControlQueueSize` replaced with `QuerySchedulerQueueSize` (the correct constant, used at `replica_metrics.go:643`).

internal/collector/registration/throughput_analyzer.go
deanlorenz: Fixed in 900c94c. All three entries now align their `—` at the same column.

internal/collector/replica_metrics.go
deanlorenz: Fixed in 900c94c. Updated to "failed to refresh replica metrics" to reflect the wider query scope.

ev-shindin (dismissed):
Thanks @deanlorenz . Please rebase again, then merge.

ev-shindin (dismissed):

ev-shindin:
@deanlorenz . it looks like you need to change something very small in the PR and commit (may be in doc), otherwise I can not re-trigger `signed-commits` action.

ev-shindin (approved):

PR #1052, opened 2026-04-23
engines/analyzers/throughput: add state management package (PR-3)

deanlorenz:
## Summary

- Implements the per-variant calibration state layer for the Throughput Analyzer (`internal/engines/analyzers/throughput/`)
- `ShapeTracker` — tracks current workload shape (IL, OL, IL_eff, KV_req); detects >20% shift and triggers observation window reset
- `ObservationWindow` — rolling window of (k, ITL_obs) pairs per replica; reports `Ready()` when ≥10 samples with ≥0.30 k-spread are accumulated
- `ThroughputAnalyzer` — implements `interfaces.Analyzer`; groups replicas by variant, runs sanity checks, updates shape and window each cycle; zero scaling signal until PR-4 adds the OLS fit
- 78 Ginkgo tests covering all components and multi-cycle integration
- Updates `docs/user-guide/throughput-analyzer.md` with PR-3 component descriptions and diagram fixes

**Scope:** state management only — no OLS regression, no μ_dec/λ_dec computation, no scaling signal (PR-4).

## Dependencies

> ⚠️ This PR is stacked on #1051 (PR-1/PR-2: query registration). Please review/merge that first. The diff against `main` includes both until #1051 lands.

## Related

- Part of #1005

## Test plan

- [ ] `make test` passes (78 new tests in `internal/engines/analyzers/throughput/`)
- [ ] `go build ./...` clean

### DIFF
--- a/docs/developer-guide/throughput-analyzer.md
+++ b/docs/developer-guide/throughput-analyzer.md
@@ -239,8 +239,8 @@ Per-variant state is minimal:
 
 | Field | What it holds |
 |---|---|
-| `ShapeTracker` | Three EMA floats: IL, OL, prefix hit rate |
-| `ObservationWindow` | Rolling slice of ≤ 10 `(k*, ITL_obs)` pairs + timestamps |
+| `ShapeTracker` | Current workload shape snapshot (IL, OL, hit rate); overwritten each cycle |
+| `ObservationWindow` | Rolling slice of ≤ 20 `(k*, ITL_obs)` pairs + timestamps |
 | `lastSanityReport` | Most recent sanity check result |
 | `lastObservedAt` | Timestamp of last observation |
 
@@ -479,7 +479,7 @@ use the same decode-rate framework.
 | `DefaultKSat` | 0.85 | KV utilization at which μ_dec_sat is evaluated |
 | `DefaultBaselineITLSec` | 0.006 | B in tier-2 ITL model (H100 near-zero-load baseline) |
 | `DefaultQueueDrainFactor` | 2.0 | Bounds queueing time to ≤ factor × ITL(k_sat) × OL |
-| `DefaultWindowMaxSize` | 100 | Max (k*, ITL) pairs in ObservationWindow |
+| `DefaultWindowMaxSize` | 20 | Max (k*, ITL) pairs in ObservationWindow |
 | `DefaultObservationMaxAge` | 30m | Observations older than this are pruned |
 | `DefaultMinSamples` | 10 | Minimum samples for OLS Ready flag |
 | `DefaultMinKSpread` | 0.30 | Minimum k-spread for OLS Ready flag |
--- a/docs/developer-guide/throughput-analyzer.md
+++ b/docs/developer-guide/throughput-analyzer.md
@@ -197,11 +197,26 @@ The remaining TA fields (`TotalKvCapacityTokens`, `AvgITL`, `AvgOutputTokens`, `
 `PrefixCacheHitRate`, `ArrivalRate`) are populated by saturation and queueing model queries.
 
 **ShapeTracker (`shape_tracker.go`)**  
-Maintains the current workload shape bucket `(IL, OL, IL_eff, KVreq)`. Detects shape changes
+Maintains the current workload shape bucket `(IL, OL, H%, IL_eff, KVreq)`. Detects shape changes
 (>20% shift in IL or OL) and triggers observation window reset.
 
-- `IL_eff = IL × (1 − PrefixCacheHitRate)` — effective input length after prefix cache
+- `IL_eff = IL × (1 − H%)` — effective input length after prefix cache
 - `KVreq = IL_eff + OL/2` — time-averaged KV footprint per decode request
+
+**Shape dimensions — design note:**  
+The tracker stores IL, OL, and H% but change detection uses only IL and OL (see `Within()`).
+H% is stored because it feeds IL_eff and KVreq; a H%-only change does not reset the window.
+
+The minimal sufficient representation for KVreq is `(OL, IL_eff)` rather than `(IL, OL, H%)`
+separately — we track all three because it is not yet clear whether IL and H% should be treated
+as independent shape dimensions or collapsed into IL_eff alone.
+
+Likewise, the slope A in `ITL(k) = A·k + B` may be independent of H%: A captures how quickly
+decode latency grows with KV load, which is driven by hardware and concurrency, not by the
+fraction of input tokens served from cache. We do not have enough data to confirm this yet.
+If it holds, a H%-only shift would warrant updating only B (via a new Tier-2 constrained fit)
+while carrying over the previous A — avoiding a full window reset.  
+*(This is a potential future optimization; the current implementation resets nothing on H%-only changes.)*
 
 **ObservationWindow (`observation_window.go`)**  
 Rolling window of `(k*, ITL_obs)` pairs collected per replica per cycle. Filters observations
--- a/docs/developer-guide/throughput-analyzer.md
+++ b/docs/developer-guide/throughput-analyzer.md
@@ -15,10 +15,10 @@ operating point, and scales before demand exceeds that supply.
 - **λ_dec** — decode token demand: how many tokens/sec the scheduler is dispatching to this model
 - **ITL(k)** — inter-token latency as a function of KV utilization k: fitted as `A·k + B` via OLS
 
-> **Status (PR-1/PR-2 — query registration + collector wiring):** Three PromQL queries
-> registered; `GenerationTokenRate`, `KvUsageInstant`, and `VLLMRequestRate` fields wired in the
-> collector. Full analyzer implementation (ShapeTracker, ObservationWindow, ITL model,
-> `ThroughputAnalyzer`) and engine wiring are pending PR-3 through PR-5.
+> **Implementation status:** Query registration, collector wiring (three PromQL queries +
+> `GenerationTokenRate`, `KvUsageInstant`, `VLLMRequestRate` fields), ShapeTracker,
+> ObservationWindow, and SanityReport are implemented. ITL model fitting, supply/demand estimation,
+> scaling signal (full `ThroughputAnalyzer`), and engine wiring are not yet implemented.
 
 ## Table of Contents
 
@@ -162,9 +162,9 @@ namespace filtering limitation of the scheduler metric.
 
 ## Architecture
 
-> **Note:** The Query Registration and Metrics Collector components below are implemented in
-> PR-1/PR-2. All other components — Package Structure, ShapeTracker, ObservationWindow, ITLModel,
-> ThroughputAnalyzer, Analysis Pipeline, and Data Flow — are **pending PR-3/PR-4**.
+> **Note:** Query Registration, Metrics Collector, Package Structure, ShapeTracker,
+> ObservationWindow, and SanityReport are implemented. ITLModel, full ThroughputAnalyzer,
+> Analysis Pipeline, and Data Flow are **not yet implemented**.
 
 ### Package Structure
 
@@ -342,7 +342,7 @@ On leader failover the incoming leader starts with an empty analyzer. During war
 
 ## ITL Model Calibration
 
-> **Pending: PR-3/PR-4.**
+> **Not yet implemented.**
 
 The ITL model `ITL(k) = A·k + B` captures how inter-token latency grows with KV cache
 utilization k. It is calibrated independently per variant (different hardware → different A, B).
@@ -374,7 +374,7 @@ zero-replica fallback using the last successful tier-1 fit. It is not wired into
 
 ## Supply Estimation
 
-> **Pending: PR-3/PR-4.**
+> **Not yet implemented.**
 
 Per replica `r`:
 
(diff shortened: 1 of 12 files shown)


asm582:
I would like to understand how state management is done. FYI, WVA currently runs in HA mode.

deanlorenz:
> I would like to understand how state management is done. FYI, WVA currently runs in HA mode.

Good question. The state is **in-memory only** — `ThroughputAnalyzer` holds a `map[string]*variantState` (keyed by `namespace|modelID|variantName`) protected by a `sync.Mutex`. Nothing is persisted to etcd or Kubernetes.

**In HA mode:**

The engine loops are already gated on leader election (`cmd/main.go` — "Only start when leader"). The `ThroughputAnalyzer` will follow the same pattern when it is wired into `main.go` (PR-5). On a leader failover, the new leader starts a fresh `ThroughputAnalyzer` with empty `variantStates`. During re-warm-up (until the observation window accumulates ≥10 `(k*, ITL)` samples with ≥0.30 k-spread), the TA emits no scaling signal — `RC = 0, SC = 0`. The saturation analyzer continues to provide signals throughout.

**Why this is acceptable:**

The observation window already clears itself whenever the workload shape shifts >20% (ShapeTracker). A leader failover is handled the same way — it causes a temporary warm-up gap, not an incorrect signal. The expected warm-up duration is short (order of minutes at normal traffic levels, bounded by the 50ms metric scrape interval × sample requirements).

Persisting state to a ConfigMap or lease annotation would add significant complexity for a benefit that only matters during failover. Happy to revisit this tradeoff if there's a specific concern about the warm-up gap in your deployment.

deanlorenz:
There is almost no state. Per variant, `ThroughputAnalyzer` holds three EMA floats (ShapeTracker) and a rolling slice of ≤ 10 `(k*, ITL)` pairs (ObservationWindow) — all in-memory, nothing persisted.

In HA mode the engine loop runs only on the leader, so the state is local to that process. On failover the new leader warms up from scratch; during that time (a few minutes at normal traffic) the TA emits no signal and the saturation analyzer covers. No external store needed.

Documented in the new [State and High Availability](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/TA2/docs/developer-guide/throughput-analyzer.md#state-and-high-availability) section.

internal/engines/analyzers/throughput/analyzer.go
```diff
@@ -0,0 +1,217 @@
...
+// across reconcile cycles. It implements interfaces.Analyzer; in the current phase
+// it produces no scaling signal (RequiredCapacity=0, SpareCapacity=0). The OLS fit,
+// μ_dec supply estimation, and λ_dec vs μ_dec scaling signal are added in PR-4.
+//
+// State is tracked per variant (keyed by "namespace|modelID|variantName") because
+// different variants may run on different hardware with different ITL coefficients,
+// and all replicas of the same variant are expected to share OL, IL, and KV_max.
+type ThroughputAnalyzer struct {
+	mu            sync.Mutex
+	variantStates map[string]*variantState
```
ev-shindin, line 24: `variantStates` is populated by `getOrCreateVariantState` but never pruned. As `VariantAutoscaling` resources come and go (autoscale-down/up, VA delete/create, namespace teardown), the map accumulates stale entries indefinitely.

**Subtle correctness implication**: when a `(namespace, modelID, variantName)` triple is deleted and later recreated, `ShapeTracker.Observe` will see a "shape change" against the *stale* shape from the previous lifetime — not a fresh warm-up. That's a latent bug, not just a memory issue.

The `lastObservedAt` field on `variantState` is already tracked but unused for cleanup.

docs/developer-guide/throughput-analyzer.md
```diff
@@ -0,0 +1,498 @@
...
+
+### State and High Availability
+
+`ThroughputAnalyzer` is stateful across reconcile cycles: it accumulates `(k*, ITL)` observations until the window is ready to fit the ITL model. The state is **in-memory only** — a `map[string]*variantState` held inside the analyzer instance, with no persistence to etcd or Kubernetes.
+
+Per-variant state is minimal:
+
+| Field | What it holds |
+|---|---|
+| `ShapeTracker` | Three EMA floats: IL, OL, prefix hit rate |
```
ev-shindin: `ShapeTracker.Observe` in `shape_tracker.go:36` does `t.current = next` — direct overwrite, not an exponentially-weighted moving average. The stored state is the *current* (IL, OL, prefix_hit_rate) tuple plus derived `ILeff` / `KVreq`, reset when the new tuple falls outside the ±20% tolerance band.

docs/developer-guide/throughput-analyzer.md
```diff
@@ -0,0 +1,498 @@
...
+| `ObservationWindow` | Rolling slice of ≤ 10 `(k*, ITL_obs)` pairs + timestamps |
```
ev-shindin: The window cap is `DefaultWindowMaxSize = 20` (`constants.go:13-17`). The number `10` is `DefaultMinSamples` — the `Ready()` threshold, a different parameter.

internal/engines/analyzers/throughput/analyzer.go
```diff
@@ -0,0 +1,217 @@
...
+//  5. Prunes observations older than DefaultObservationMaxAge.
+//
+// Returns a map of variantName → SanityReport for logging. An empty SanityReport
+// (report.OK() == true) means that variant's metrics were healthy this cycle.
+func (a *ThroughputAnalyzer) Observe(
+	ctx context.Context,
+	modelID, namespace string,
+	metrics []interfaces.ReplicaMetrics,
+) map[string]SanityReport {
+	now := time.Now()
```
ev-shindin: `Observe` reads wall clock directly. The window-level tests in `observation_window_test.go` pass explicit timestamps, but the analyzer-level tests can't exercise time-based pruning with synthetic time.

Two common patterns:
- Inject `k8s.io/utils/clock.Clock` (fake clock available for tests).
- Accept `now time.Time` as a parameter to `Observe`.

internal/engines/analyzers/throughput/observation_window.go
```diff
@@ -0,0 +1,103 @@
...
+		minKSpread:   minKSpread,
+		minK:         minK,
+		maxK:         maxK,
+	}
+}
+
+// Add appends a (k, itl) observation if k ∈ [minK, maxK] and itl > 0.
+// When the window is at capacity, the oldest observation is evicted first.
+func (w *ObservationWindow) Add(k, itl float64, ts time.Time) {
+	if k < w.minK || k > w.maxK {
```
ev-shindin, line 44: Silently dropping observations outside `[0.15, 0.85]` is the right *policy* (those are unreliable for the linear ITL fit), but it produces a confusing failure mode: if a workload spends all its time at `k > 0.85` (saturated) or `k < 0.15` (very lightly loaded), the window never fills and `Ready()` stays false forever — with no log or counter showing why.
I think you should log this and/or add a metric.

internal/engines/analyzers/throughput/analyzer.go
```diff
@@ -0,0 +1,217 @@
...
+func averageShapeMetrics(metrics []interfaces.ReplicaMetrics) (il, ol, hitRate float64) {
+	var sumIL, sumOL, sumHitRate float64
+	var count float64
+	for _, m := range metrics {
+		if m.AvgInputTokens <= 0 || m.AvgOutputTokens <= 0 {
+			continue
+		}
+		sumIL += m.AvgInputTokens
+		sumOL += m.AvgOutputTokens
+		sumHitRate += m.PrefixCacheHitRate
```
ev-shindin: Simple arithmetic mean of `PrefixCacheHitRate` across replicas. With cache-aware routing (EPP prefix routing), fleet hit rates converge in steady state — so this is fine *eventually*. But during cache warm-up — exactly the period where this analyzer is most useful — one hot replica at 0.8 and three cold replicas at 0.0 averages to 0.2, which understates the *effective* fleet hit rate.

A request-rate-weighted average (`sumHitRate += m.PrefixCacheHitRate * m.RequestRate`, normalized) would reflect what the routing layer actually sees.

internal/engines/analyzers/throughput/analyzer.go
```diff
@@ -0,0 +1,217 @@
...
+		LastSanityReport: state.lastSanityReport,
+	}, true
+}
+
+// --- helpers ---
+
+// variantKey builds the map key for a variant. The pipe delimiter is safe because
+// Kubernetes resource names follow DNS rules and cannot contain "|".
+func variantKey(namespace, modelID, variantName string) string {
+	return fmt.Sprintf("%s|%s|%s", namespace, modelID, variantName)
```
ev-shindin: `namespace` and `variantName` are k8s resource names, so DNS-safe → no `|`. But `modelID` is operator-provided (e.g. `ibm/granite-13b`, `meta-llama/Llama-3.1-8B-Instruct`), *not* constrained by k8s naming. A model ID containing `|` would silently collide.

internal/collector/registration/throughput_analyzer_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/source"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/source/prometheus"
+)
+
+var _ = Describe("RegisterThroughputAnalyzerQueries", func() {
+	var (
```
ev-shindin, line 14: `ThroughputAnalyzer` holds a `sync.Mutex` (`analyzer.go:24`) and serializes access. The 15 `It` blocks in this file cover correctness paths but none exercises concurrent `Observe` calls. A single `It("is safe for concurrent observation across variants"...)` with `sync.WaitGroup`-driven goroutines would lock the contract in, and `go test -race` would catch any future regressions.

ev-shindin (changes requested):
please address review comments

internal/engines/analyzers/throughput/analyzer.go
deanlorenz: Fixed. Added a pruning pass at the end of `Observe()`: entries where `now.Sub(state.lastObservedAt) > 2×DefaultObservationMaxAge` (60 min) are deleted from `variantStates`. A VA that disappears stops updating `lastObservedAt`; after 60 min the stale entry is evicted so a recreated VA gets a clean warm-up rather than a false shape-change signal against the old shape.

docs/developer-guide/throughput-analyzer.md
deanlorenz: Fixed. The doc now reads: *"Current workload shape snapshot (IL, OL, hit rate); overwritten each cycle"* — matching the `t.current = next` direct overwrite in `shape_tracker.go`.

docs/developer-guide/throughput-analyzer.md
deanlorenz: Fixed both. State table "≤ 10" → "≤ 20"; config table `DefaultWindowMaxSize` value 100 → 20. Both now match `constants.go`. The 10 was `DefaultMinSamples` (the `Ready()` threshold) — a different constant that I conflated with the window cap.

internal/engines/analyzers/throughput/analyzer.go
deanlorenz: Fixed using the `now time.Time` parameter approach. `Observe` now takes `now time.Time` as its second argument; `Analyze` passes `time.Now()` at the call site. Existing tests updated; time-based pruning scenarios can now be written with synthetic timestamps.

internal/engines/analyzers/throughput/observation_window.go
deanlorenz: Fixed. `ObservationWindow.Add()` now emits a `DEBUG` log with `k`, `minK`, and `maxK` when an observation is dropped outside the range. A workload stuck at k > 0.85 (saturated) or k < 0.15 (idle) will surface this in operator logs immediately rather than leaving `Ready()` perpetually false with no indication.

internal/engines/analyzers/throughput/analyzer.go
deanlorenz: Fixed. `averageShapeMetrics` now computes a `VLLMRequestRate`-weighted mean for IL, OL, and hit rate. A replica serving 80% of requests contributes proportionally rather than being diluted by idle replicas at 0.0 hit rate. Falls back to an unweighted mean when all eligible replicas report zero `VLLMRequestRate`.

internal/engines/analyzers/throughput/analyzer.go
deanlorenz: Fixed. Separator changed from `"|"` to `"\x00"` (null byte). Neither k8s resource names nor operator-provided model IDs can contain a null byte, so collisions are impossible regardless of what the operator puts in `modelID`. Removed the misleading comment about DNS-safety.

internal/collector/registration/throughput_analyzer_test.go
deanlorenz: Added `Describe("Observe — concurrent safety")` to `analyzer_test.go` (the file with the 15 `It` blocks): 10 goroutines call `Observe` concurrently across 3 variant names while a parallel goroutine calls `VariantState`, all joined on a `WaitGroup`. Passes `go test -race` cleanly.

ev-shindin (approved):

PR #1058, opened 2026-04-23
fix(collector): wrong ITL metric name and dispatch rate pod label

### LINKED ISSUE #1056: fix(collector): wrong ITL metric name and dispatch rate pod label in queueing model
## Problem

Two bugs in the queueing model metrics collector prevent the EKF tuner from receiving valid observations:

### 1. Wrong ITL metric name

`internal/collector/registration/queueing_model.go` queries `vllm:time_per_output_token_seconds` for inter-token latency, but vLLM exports `vllm:inter_token_latency_seconds`. As a result, the ITL query returns no data, and `env.Valid()` fails (ITL=0), so the EKF tuner never updates its parameters.

### 2. Wrong pod label for dispatch rate

`internal/collector/replica_metrics.go` keys dispatch rate results by the `pod` label (the EPP scrape target) instead of `pod_name` (the vLLM endpoint pod). Since vLLM TTFT/ITL metrics are keyed by the vLLM pod name, the arrival rate and latency metrics never join, producing a `ReplicaMetrics` entry where `ArrivalRate > 0` but `AvgTTFT == 0` and `AvgITL == 0`.

## Fix

- Change ITL query to use `vllm:inter_token_latency_seconds_{sum,count}`
- Change dispatch rate pod label lookup to prefer `pod_name` over `pod`

Both fixes are in the collector layer only — no changes to analyzer, tuner, or interfaces.

vishakha-ramani:
## Summary

Fixes two bugs in the queueing model metrics collector that prevented the EKF tuner from receiving valid observations. Both fixes are confined to the collector layer.

- **ITL metric name**: Changed from `vllm:time_per_output_token_seconds` to `vllm:inter_token_latency_seconds` (the name actually exported by vLLM). Without this, the ITL query returned no data → `env.Valid()` failed → tuner never updated.

- **Dispatch rate pod label**: `inference_extension_scheduler_attempts_total` carries both `pod` (EPP scrape target) and `pod_name` (vLLM endpoint). The collector was keying by `pod` (EPP name), so arrival rate and latency metrics never joined. Fixed to prefer `pod_name`, falling back to `pod`.

## Test plan

- [ ] Deploy with a vLLM server and send traffic; verify WVA logs show `Queueing model observed vs. predicted` with non-zero `arrivalRate_rps`, `observedTTFT_ms`, `observedITL_ms`
- [ ] Verify EKF `updateCount` increments each cycle (no longer stuck at 0)
- [ ] Verify `inference_extension_scheduler_attempts_total` dispatch rate joins correctly with vLLM pod metrics

Closes #1056

### DIFF
--- a/internal/collector/registration/queueing_model.go
+++ b/internal/collector/registration/queueing_model.go
@@ -18,7 +18,7 @@ const (
 	QueryAvgTTFT = "avg_ttft"
 
 	// QueryAvgITL is the query name for average inter-token latency per pod (in seconds).
-	// Source: vllm:time_per_output_token_seconds histogram
+	// Source: vllm:inter_token_latency_seconds histogram
 	QueryAvgITL = "avg_itl"
 )
 
@@ -46,7 +46,7 @@ func RegisterQueueingModelQueries(sourceRegistry *source.SourceRegistry) {
 	})
 
 	// Average time-to-first-token per pod (seconds).
-	// Uses histogram _sum/_count from vLLM over a 1m rate window.
+	// Uses histogram rate(sum[1m]) / rate(count[1m]) over a 1m sliding window.
 	// Used by queueing model tuner as the observed TTFT for Kalman filter updates.
 	registry.MustRegister(source.QueryTemplate{
 		Name:     QueryAvgTTFT,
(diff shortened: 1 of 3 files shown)


asm582 (dismissed):
Thanks for this PR, @vishakha-ramani. We would like to have tests. For this PR, we can start by adding unit tests; a simple assertion on the PromQL template string would help prevent regression

mamy-CS (commented):
lgtm, I don’t see any regressions in the changed paths. But iterating and agree as a follow-up, I’d suggest adding a small unit test around these query/label assumptions so we catch upstream metric-name or label drift earlier.

vishakha-ramani:
Added unit tests in the second commit: assertions on the PromQL template strings (ITL metric name, TTFT metric name, dispatch rate pod label) and a negative test that the old `time_per_output_token_seconds` name is absent. All 10 tests pass.

PR #1065, opened 2026-04-27
add configurable defaults for fixtures services and model workloads

mamy-CS:
- Add minimal fixture cleanup while preserving existing e2e defaults.
- Our e2e fixtures encode environment conventions that work for the main CI path. This change keeps those defaults intact
- Also aligns fixture behavior checks and wait logic with k8s machinery helpers for consistency.

### DIFF
--- a/test/e2e/fixtures/infra_builder.go
+++ b/test/e2e/fixtures/infra_builder.go
@@ -13,16 +13,87 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type infraFixtureConfig struct {
+	serviceSuffix             string
+	serviceMonitorSuffix      string
+	servicePortName           string
+	serviceMonitorMetricsPath string
+	releaseLabelValue         string
+	testLabelValue            string
+}
+
+// InfraOption overrides fixture conventions used by Service and ServiceMonitor helpers.
+type InfraOption func(*infraFixtureConfig)
+
+// WithServiceSuffix overrides the generated Service name suffix.
+func WithServiceSuffix(suffix string) InfraOption {
+	return func(cfg *infraFixtureConfig) {
+		if suffix != "" {
+			cfg.serviceSuffix = suffix
+		}
+	}
+}
+
+// WithServiceMonitorSuffix overrides the generated ServiceMonitor name suffix.
+func WithServiceMonitorSuffix(suffix string) InfraOption {
+	return func(cfg *infraFixtureConfig) {
+		if suffix != "" {
+			cfg.serviceMonitorSuffix = suffix
+		}
+	}
+}
+
+// WithServicePortName overrides the Service port name used by ServiceMonitor endpoints.
+func WithServicePortName(portName string) InfraOption {
+	return func(cfg *infraFixtureConfig) {
+		if portName != "" {
+			cfg.servicePortName = portName
+		}
+	}
+}
+
+// WithServiceMonitorDefaults overrides monitor-specific convention values.
+func WithServiceMonitorDefaults(metricsPath, releaseLabelValue string) InfraOption {
+	return func(cfg *infraFixtureConfig) {
+		if metricsPath != "" {
+			cfg.serviceMonitorMetricsPath = metricsPath
+		}
+		if releaseLabelValue != "" {
+			cfg.releaseLabelValue = releaseLabelValue
+		}
+	}
+}
+
+func defaultInfraFixtureConfig() infraFixtureConfig {
+	return infraFixtureConfig{
+		serviceSuffix:             serviceNameSuffix,
+		serviceMonitorSuffix:      serviceMonitorNameSuffix,
+		servicePortName:           defaultServicePortName,
+		serviceMonitorMetricsPath: defaultServiceMonitorMetricsPath,
+		releaseLabelValue:         kubePrometheusStackReleaseLabelValue,
+		testLabelValue:            defaultTestResourceLabelValue,
+	}
+}
+
+func resolveInfraFixtureConfig(opts ...InfraOption) infraFixtureConfig {
+	cfg := defaultInfraFixtureConfig()
+	for _, opt := range opts {
+		opt(&cfg)
+	}
+	return cfg
+}
+
 // CreateService creates a Kubernetes Service for the model server. Fails if the service already exists.
-func CreateService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, appLabel string, port int) error {
-	service := buildService(namespace, name, appLabel, port)
+func CreateService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, appLabel string, port int, opts ...InfraOption) error {
+	service := buildService(namespace, name, appLabel, port, opts...)
 	_, err := k8sClient.CoreV1().Services(namespace).Create(ctx, service, metav1.CreateOptions{})
 	return err
 }
 
 // DeleteService deletes the Kubernetes Service. Idempotent; ignores NotFound.
-func DeleteService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name string) error {
-	serviceName := name + "-service"
+func DeleteService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name string, opts ...InfraOption) error {
+	cfg := resolveInfraFixtureConfig(opts...)
+	serviceName := name + cfg.serviceSuffix
 	err := k8sClient.CoreV1().Services(namespace).Delete(ctx, serviceName, metav1.DeleteOptions{})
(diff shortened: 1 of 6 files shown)


test/e2e/fixtures/infra_builder.go
```diff
@@ -57,40 +87,41 @@ func EnsureService(ctx context.Context, k8sClient *kubernetes.Clientset, namespa
...
 			},
 			Type: corev1.ServiceTypeClusterIP,
 		},
 	}
 }
 
 // CreateServiceMonitor creates a ServiceMonitor for Prometheus. Fails if it already exists.
-func CreateServiceMonitor(ctx context.Context, crClient client.Client, monitoringNamespace, targetNamespace, name, appLabel string) error {
-	serviceMonitor := buildServiceMonitor(monitoringNamespace, targetNamespace, name, appLabel)
+func CreateServiceMonitor(ctx context.Context, crClient client.Client, monitoringNamespace, targetNamespace, name, appLabel string, opts ...InfraOption) error {
```
asm582: As I understand, if these functions are only to be used by e2e, why are they exported? Is there a contract breach?

asm582 (changes requested):
Address review

test/e2e/fixtures/infra_builder.go
mamy-CS: Good question. These helpers are exported because test/e2e imports test/e2e/fixtures as a separate package. There is no intended contract breach.

test/e2e/fixtures/infra_builder.go
```diff
@@ -13,16 +13,46 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type infraFixtureConfig struct {
+	servicePortName string
+}
+
```
asm582: This is a good idea, but too much for one field. Unless we have methods that are hard to read with like 5-6 config options, we can safely delay this approach.

test/e2e/fixtures/lws_builder.go
```diff
@@ -3,22 +3,22 @@ package fixtures
...
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	lwsv1 "sigs.k8s.io/lws/api/leaderworkerset/v1"
 )
 
 // EnsureModelServiceLWS creates or replaces a LeaderWorkerSet for model service (idempotent for test setup).
-func EnsureModelServiceLWS(ctx context.Context, crClient client.Client, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int, groupSize int32) error {
-	lwsName := name + "-decode"
-	desiredLWS := buildModelServiceLWS(namespace, name, poolName, modelID, useSimulator, maxNumSeqs, groupSize)
+func EnsureModelServiceLWS(ctx context.Context, crClient client.Client, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int, groupSize int32, opts ...ModelServiceOption) error {
```
asm582: Who is passing options here?

test/e2e/fixtures/infra_builder.go
```diff
@@ -31,8 +61,8 @@ func DeleteService(ctx context.Context, k8sClient *kubernetes.Clientset, namespa
 }
 
 // EnsureService creates or replaces the Service (idempotent for test setup).
-func EnsureService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, appLabel string, port int) error {
-	serviceName := name + "-service"
+func EnsureService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, appLabel string, port int, opts ...InfraOption) error {
```
asm582: Who is passing options here?

test/e2e/fixtures/model_service_conventions.go
```diff
@@ -0,0 +1,63 @@
...
+	defaultModelServiceContainerPort  = int32(8000)
+	defaultHFTokenSecretName          = "llm-d-hf-token"
+	defaultHFTokenSecretKey           = "HF_TOKEN"
+	defaultModelServiceLabelValue     = "ms-sim-llm-d-modelservice"
+	defaultInferenceServingLabelValue = "true"
+	defaultGuideLabelValue            = "workload-autoscaling"
+	defaultTestResourceLabelValue     = "true"
+)
+
+type modelServiceFixtureConfig struct {
```
asm582: infraFixtureConfig and modelServiceFixtureConfig look identical to me.

asm582 (dismissed):
I'm not sure the options pattern is worth using at this point. I vote for simplification. Only e2e is a consumer of these methods, and all callers are under our control.

mamy-CS:
@asm582 I initially added small optional config hooks to make fixture defaults overrideable for different e2e environments/ use of fixtures by others, such as benchmark. After your review, I simplified it down since we don’t have active callers needing that flexibility yet.

test/e2e/fixtures/lws_builder.go
mamy-CS: opts removed

test/e2e/fixtures/infra_builder.go
mamy-CS: removed

test/e2e/fixtures/model_service_conventions.go
mamy-CS: simplified further

lionelvillard:
### Code Review

Found 2 issues:

**Go Code Quality**

1. Two distinct label keys (`"llm-d.ai/inferenceServing"` and `"llm-d.ai/inference-serving"`) share a single constant `defaultInferenceServingLabelValue`. The name implies it belongs to the first key only, making the abstraction misleading — a future value change to one label won't be independent of the other. Consider a neutral name like `labelValueTrue = "true"` or introduce separate constants per key. — [`test/e2e/fixtures/model_service_conventions.go`](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/cb7a172f325c15db54adf243d3ccb0a1836419bf/test/e2e/fixtures/model_service_conventions.go#L26)

2. `defaultModelServiceContainerPort = int32(8000)` declares a typed constant, which is atypical in Go. Untyped numeric constants widen to the required type at each use site; pinning it to `int32` prevents use in `int` contexts without an explicit conversion. Prefer `defaultModelServiceContainerPort = 8000` and let the compiler coerce it at the struct-literal site. — [`test/e2e/fixtures/model_service_conventions.go#L21`](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/cb7a172f325c15db54adf243d3ccb0a1836419bf/test/e2e/fixtures/model_service_conventions.go#L21)

test/e2e/fixtures/infra_builder.go
```diff
@@ -124,21 +124,21 @@ func EnsureServiceMonitor(ctx context.Context, crClient client.Client, monitorin
 }
 
 func buildServiceMonitor(monitoringNamespace, targetNamespace, name, appLabel string) *promoperator.ServiceMonitor {
-	serviceMonitorName := name + "-monitor"
+	serviceMonitorName := name + serviceMonitorNameSuffix
```
asm582: I see a suffix added inside the function. Why can't the caller send a baseName+suffix to the function? This should also reduce the defaults defined, I think

mamy-CS: intentionally keeping suffix composition inside the fixture so that we centralize naming conventions and avoid call site duplication/errors. The helper contract is base name in, canonical resource name out.

lionelvillard: I think it's fine as soon as the function contract does not "lie". The parameter `name` should probably be renamed to `baseName` to avoid confusion.

test/e2e/fixtures/infra_builder.go
```diff
@@ -58,24 +58,24 @@ func EnsureService(ctx context.Context, k8sClient *kubernetes.Clientset, namespa
 }
 
 func buildService(namespace, name, appLabel string, port int) *corev1.Service {
-	serviceName := name + "-service"
+	serviceName := name + serviceNameSuffix
```
lionelvillard: `buildService` function signature is misleading as it does not create a service named `name`. Either call it `baseName` or change implementation.

mamy-CS: ok then, I will rename to baseName in service/serviceMonitor helpers

test/e2e/fixtures/infra_builder.go
mamy-CS: renamed to baseName

lionelvillard (approved):
more work to do IMO but good enough for now

mamy-CS:
> more work to do IMO but good enough for now

yes, small focused prs

PR #1067, opened 2026-04-28
test(e2e): add saturation V2 smoke tier (path, scale-up, scale-down)

### LINKED ISSUE #965: Add saturation v2 (token based) e2es

ev-shindin:
## Summary

Adds three end-to-end smoke tests for the saturation **V2** engine path. 

The tests assert that:

1. **Path selection works** — setting `analyzerName: "saturation"` routes the engine through V2 and produces a positive `DesiredOptimizedAlloc.NumReplicas` with `MetricsAvailable=True`.
2. **Scale-up triggers** — when the simulator emits `kv-cache-usage = 0.4` (above `scaleUpThreshold = 0.30`), V2 recommends more than 1 replica.
3. **Scale-down triggers** — with the deployment scaled to 2 replicas and a raised `scaleDownBoundary = 0.50`, V2 recommends fewer than 2 replicas.

All three are labeled `Label("smoke", "full")`, so they run in both `make test-e2e-smoke` and `make test-e2e-full`.

## What's in this PR

| File | Change | Purpose |
|---|---|---|
| `test/e2e/saturation_v2_test.go` | **new (+269 LOC)** | The three V2 smoke tests, with constants, BeforeAll, AfterAll. |
| `test/e2e/fixtures/model_service_builder.go` | +19 / -2 | Adds `CreateModelServiceWithExtraArgs`; refactors `buildModelServerArgs` into a thin wrapper that appends caller-supplied args to `buildModelServerBaseArgs`. |
| `test/e2e/fixtures/lws_builder.go` | +1 / -1 | Internal-signature follow-through (passes `nil` extra args). |

**Zero production code changes.** `CreateModelService` and `EnsureModelService` keep their existing signatures — the new variant is additive.

## How the simulator is driven

V2's primary demand signal is `vllm:kv_cache_usage_perc × effectiveCapacity`. 
This revision sidesteps the runtime-fidelity gap by configuring the simulator's `--fake-metrics` flag, [documented at the simulator repo](https://github.com/llm-d/llm-d-inference-sim/blob/main/docs/configuration.md#fake-metrics-configuration). Per the docs:

> When specified, only these fake metrics will be reported — real metrics and fake metrics will never be reported together.

The smoke deployment's simulator container is launched with:

```json
{"kv-cache-usage":0.4,"running-requests":1,"waiting-requests":0}
```

V2 reads `kv-cache-usage = 0.4` deterministically on every reconcile cycle. No load trigger is needed for the scale-up assertion.

### Calibration math (documented inline at `saturation_v2_test.go:19-42`)

With `effectiveCapacity ≈ 6553` (deployment-derived; the simulator does not emit `vllm:cache_config_info`, so V2 falls back to its capacity-store estimate):

### Helpers reused

From `saturation_analyzer_path_test.go`:
- `buildSaturationConfigYAMLWithThresholds(...)` — already accepts V2 thresholds.
- `expectAnalyzerPathLog("V2", modelID)` — controller-log polling.
- `waitForPositiveDesiredAllocation(...)` — `Eventually` poll on a positive `NumReplicas`.
- `upsertSaturationConfigEntry(...)` and `saturationConfigMapName()` / `saturationConfigMapForRecreate(...)` — ConfigMap snapshot/restore plumbing.

From `test/e2e/fixtures/`:
- `EnsureService`, `EnsureServiceMonitor`, `EnsureVariantAutoscalingWithDefaults`, `DeleteModelService`.

`runThresholdTriggerJob` is **not** used by this test — the fake-metrics approach makes load generation unnecessary.

### New fixture surface

```go
// CreateModelServiceWithExtraArgs is like CreateModelService but appends
// additional CLI args to the model-server (simulator or vLLM) container.
// Used by tests that need to inject engine-specific configuration such as
// --fake-metrics.
func CreateModelServiceWithExtraArgs(...) error
```

The internal `buildModelServerArgs(modelID, useSimulator, maxNumSeqs, extraArgs)` is a one-line wrapper around the renamed `buildModelServerBaseArgs(...)` that returns `append(base, extraArgs...)`. Existing callers (`CreateModelService`, `EnsureModelService`, `buildModelServiceLWS`) pass `nil` and behave identically to before.

### Lifecycle

- **`BeforeAll`** snapshots the existing saturation ConfigMap, deploys the model service with `--fake-metrics` injected, creates Service + ServiceMonitor + VA, waits for one ready replica, and installs the V2 saturation config.
- **`AfterAll`** restores the original ConfigMap (delete-then-recreate from the deep-copied snapshot, mirroring V1's pattern) and best-effort deletes all created resources.

The V2 saturation config is installed in `BeforeAll` rather than inside the first `It` so the suite's behavior does not depend on `Ordered` execution to set up V2's preconditions.

## Test plan

- [x] `go vet ./test/e2e/...` clean
- [x] `go build ./...` clean
- [x] `go test -c -o /dev/null ./test/e2e` clean (compile-only check)
- [x] `gofmt -l test/e2e/saturation_v2_test.go test/e2e/fixtures/model_service_builder.go` returns empty
- [x] CI: `make test-e2e-smoke` passes against kind + simulator
- [x] CI: `make test-e2e-full` includes the new tests and passes

### DIFF
--- a/test/e2e/fixtures/model_service_builder.go
+++ b/test/e2e/fixtures/model_service_builder.go
@@ -20,7 +20,14 @@ import (
 // It does not create a Kubernetes Service; callers must use CreateService or EnsureService
 // (typically naming the Service name + "-service") to expose the deployment.
 func CreateModelService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int) error {
-	deployment := buildModelServiceDeployment(namespace, name, poolName, modelID, useSimulator, maxNumSeqs)
+	return CreateModelServiceWithExtraArgs(ctx, k8sClient, namespace, name, poolName, modelID, useSimulator, maxNumSeqs, nil)
+}
+
+// CreateModelServiceWithExtraArgs is like CreateModelService but appends additional
+// CLI args to the model-server (simulator or vLLM) container. Used by tests that
+// need to inject engine-specific configuration such as --fake-metrics.
+func CreateModelServiceWithExtraArgs(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int, extraArgs []string) error {
+	deployment := buildModelServiceDeployment(namespace, name, poolName, modelID, useSimulator, maxNumSeqs, extraArgs)
 	_, err := k8sClient.AppsV1().Deployments(namespace).Create(ctx, deployment, metav1.CreateOptions{})
 	return err
 }
@@ -40,7 +47,7 @@ func DeleteModelService(ctx context.Context, k8sClient *kubernetes.Clientset, na
 func EnsureModelService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int) error {
 	appLabel := name + decodeNameSuffix
 	deploymentName := appLabel
-	desiredDeployment := buildModelServiceDeployment(namespace, name, poolName, modelID, useSimulator, maxNumSeqs)
+	desiredDeployment := buildModelServiceDeployment(namespace, name, poolName, modelID, useSimulator, maxNumSeqs, nil)
 
 	existingDeployment, err := k8sClient.AppsV1().Deployments(namespace).Get(ctx, deploymentName, metav1.GetOptions{})
 	if err != nil {
@@ -83,13 +90,13 @@ func modelServiceDeploymentMatchesDesired(existing, desired appsv1.Deployment) b
 		apiequality.Semantic.DeepEqual(existing.Spec.Template.Spec, desired.Spec.Template.Spec)
 }
 
-func buildModelServiceDeployment(namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int) *appsv1.Deployment {
+func buildModelServiceDeployment(namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int, extraArgs []string) *appsv1.Deployment {
 	appLabel := name + decodeNameSuffix
 	image := defaultModelServiceSimulatorImage
 	if !useSimulator {
 		image = defaultModelServiceRuntimeImage
 	}
-	args := buildModelServerArgs(modelID, useSimulator, maxNumSeqs)
+	args := buildModelServerArgs(modelID, useSimulator, maxNumSeqs, extraArgs)
 	labels := map[string]string{
 		"app":                        appLabel,
 		"llm-d.ai/inferenceServing":  defaultLabelValueTrue,
@@ -202,7 +209,11 @@ func buildModelServiceResources(useSimulator bool) corev1.ResourceRequirements {
 	}
 }
 
-func buildModelServerArgs(modelID string, useSimulator bool, maxNumSeqs int) []string {
+func buildModelServerArgs(modelID string, useSimulator bool, maxNumSeqs int, extraArgs []string) []string {
+	return append(buildModelServerBaseArgs(modelID, useSimulator, maxNumSeqs), extraArgs...)
+}
+
+func buildModelServerBaseArgs(modelID string, useSimulator bool, maxNumSeqs int) []string {
 	if useSimulator {
 		// Simulator is configured to be deliberately slow so that Prometheus
 		// can observe non-zero KV-cache and queue metrics between scrapes (every 15s).
--- a/test/e2e/fixtures/model_service_builder.go
+++ b/test/e2e/fixtures/model_service_builder.go
@@ -26,6 +26,13 @@ func CreateModelService(ctx context.Context, k8sClient *kubernetes.Clientset, na
 // CreateModelServiceWithExtraArgs is like CreateModelService but appends additional
 // CLI args to the model-server (simulator or vLLM) container. Used by tests that
 // need to inject engine-specific configuration such as --fake-metrics.
+//
+// The caller is responsible for ensuring extraArgs are valid for the chosen
+// runtime — this function performs no flag validation. Engine-specific flags
+// (e.g., --fake-metrics, which exists only on llm-d-inference-sim) will cause
+// the container to crash-loop if passed to the wrong runtime. Tests that use
+// simulator-only flags should gate their suite on `cfg.UseSimulator` and Skip
+// otherwise.
 func CreateModelServiceWithExtraArgs(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int, extraArgs []string) error {
 	deployment := buildModelServiceDeployment(namespace, name, poolName, modelID, useSimulator, maxNumSeqs, extraArgs)
 	_, err := k8sClient.AppsV1().Deployments(namespace).Create(ctx, deployment, metav1.CreateOptions{})
--- a/test/e2e/saturation_v2_test.go
+++ b/test/e2e/saturation_v2_test.go
@@ -18,32 +18,24 @@ import (
 
 // V2 smoke calibration via the simulator's --fake-metrics flag.
 //
-// The simulator (llm-d-inference-sim) emits a fixed set of vLLM Prometheus
-// metric values rather than tracking them from real traffic. Setting
-// kv-cache-usage to 0.4 gives V2 a deterministic input that:
+// kv-cache-usage = 0.4 is the operating point that deterministically exercises
+// both arcs of the V2 cost-aware optimizer with the suite's chosen thresholds:
 //
-//   - Crosses scaleUpThreshold=0.30 at 1 replica:
-//     replicaDemand = 0.4 * effectiveCapacity (~6553) ≈ 2621 tokens
-//     requiredCapacity = 2621/0.30 - 6553 = 2184 > 0 → scale-up
+//   - At 1 replica with scaleUpThreshold = 0.30, replicaDemand crosses the
+//     threshold and the optimizer's required-capacity signal becomes positive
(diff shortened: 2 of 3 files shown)


mamy-CS:
> ## Summary
> Adds three end-to-end smoke tests for the saturation **V2** engine path. This is the first PR in a five-PR series that establishes V2 e2e coverage — see `plan-e2e-saturation-v2.md` for the full roadmap.
> 
> The tests assert that:
> 
> 1. **Path selection works** — setting `analyzerName: "saturation"` routes the engine through V2 and produces a positive `DesiredOptimizedAlloc.NumReplicas` with `MetricsAvailable=True`.
> 2. **Scale-up triggers** — bounded load above `scaleUpThreshold` drives V2 to recommend more replicas than the current baseline.
> 3. **Scale-down triggers** — when the deployment is at 2 replicas and load is below `scaleDownBoundary`, V2 recommends fewer than 2.
> 
> All three are labeled `Label("smoke", "full")`, so they run in both `make test-e2e-smoke` and `make test-e2e-full`.
> 
> ## What's in this PR
> * **One new file:** `test/e2e/saturation_v2_test.go` (~266 LOC).
> * **Zero production code changes.**
> * **Zero fixture changes** — every helper this test needs already exists in `saturation_analyzer_path_test.go` and `test/e2e/fixtures/`.
> 
> ## Implementation notes
> ### Helpers reused (no new helpers introduced)
> From `saturation_analyzer_path_test.go`:
> 
> * `buildSaturationConfigYAMLWithThresholds(...)` — already accepts `scaleUpThreshold` + `scaleDownBoundary` parameters used by V2.
> * `expectAnalyzerPathLog("V2", modelID)` — polls controller logs for `"Processing model (V2)"`.
> * `runThresholdTriggerJob(...)` — bounded-volume curl Job that crosses thresholds deterministically.
> * `waitForPositiveDesiredAllocation(...)` — `Eventually` poll on `DesiredOptimizedAlloc.NumReplicas > 0`.
> * `upsertSaturationConfigEntry(...)` — patches a YAML entry in the global saturation ConfigMap.
> * `saturationConfigMapName()` / `saturationConfigMapForRecreate(...)` — name resolution and snapshot/restore plumbing.
> 
> From `test/e2e/fixtures/`:
> 
> * `CreateModelService`, `EnsureService`, `EnsureServiceMonitor`, `EnsureVariantAutoscalingWithDefaults`, `DeleteModelService`.
> 
> ### Lifecycle
> * **`BeforeAll`** snapshots the existing saturation ConfigMap, creates the model service / service / ServiceMonitor / VA, waits for one ready replica, and installs the V2 saturation config. Doing the V2 config install in `BeforeAll` (rather than inside the first `It`) means the suite's behavior does not depend on `Ordered` execution to set up V2's preconditions for later tests.
> * **`AfterAll`** restores the original ConfigMap (delete-then-recreate from the deep-copied snapshot, mirroring V1's pattern) and best-effort deletes all created resources.
> 
> ### Calibration values
> The smoke thresholds are deliberate:
> 
> Constant	Value	Rationale
> `v2SmokeScaleUpThreshold`	`0.30`	Aggressive enough that the existing `runThresholdTriggerJob` (6 × 400-token requests) reliably crosses it against the simulator.
> `v2SmokeScaleDownBoundary`	`0.20`	Low enough that any load produces a deeply negative `SpareCapacity = supply − demand/0.20 = supply − 5×demand`, suppressing scale-down during the scale-up test.
> Scale-down test thresholds	`scaleUp=0.85`, `scaleDown=0.50`	Canonical ordering (`scaleUp > scaleDown`) so the scale-down test isn't asking V2 to reason about an inverted configuration.
> The V1-specific knobs (`kvCacheThreshold=0.80`, `queueLengthThreshold=50`, `kvSpareTrigger=0.10`, `queueSpareTrigger=5`) are filled with safe defaults; V2's token-based path ignores them.
> 
> **Calibration may need refinement after the first CI run** — if the simulator's `vllm:request_generation_tokens_*` rates don't accumulate enough for the 0.30 threshold to fire, lowering it further is a simple follow-up commit.
> 
> ## Test plan
> * [x]  `go vet ./test/e2e/...` clean
> * [x]  `go build ./...` clean
> * [x]  `go test -c -o /dev/null ./test/e2e` clean (compile-only check)
> * [x]  `gofmt -l test/e2e/saturation_v2_test.go` returns empty
> * [ ]  CI: `make test-e2e-smoke` passes against kind + simulator
> * [ ]  CI: `make test-e2e-full` includes the new tests and passes
> * [ ]  If TC for scale-up misses its trigger in CI, lower `v2SmokeScaleUpThreshold` and re-run
> 
> Static checks above pass locally; CI provides the runtime calibration validation.

@ev-shindin where is plan-e2e-saturation-v2.md?

ev-shindin:
> @ev-shindin where is plan-e2e-saturation-v2.md?

@mamy-CS  updated PR description (also changed PR itself)

test/e2e/saturation_v2_test.go
```diff
@@ -0,0 +1,278 @@
...
+	var (
+		modelID         string
+		cmName          string
+		cmNamespace     string
+		cmKey           string
+		cmOriginal      *corev1.ConfigMap
+		cmExistedBefore bool
+	)
+
+	BeforeAll(func() {
```
mamy-CS, line 80: This suite appends `--fake-metrics`, which only applies to `llm-d-inference-sim`. Please Skip(...) at the start of BeforeAll when not in sim mode, so OpenShift-style runs with real vLLM don’t get a broken Deployment.

test/e2e/fixtures/model_service_builder.go
```diff
@@ -20,7 +20,14 @@ import (
...
 // (typically naming the Service name + "-service") to expose the deployment.
 func CreateModelService(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int) error {
-	deployment := buildModelServiceDeployment(namespace, name, poolName, modelID, useSimulator, maxNumSeqs)
+	return CreateModelServiceWithExtraArgs(ctx, k8sClient, namespace, name, poolName, modelID, useSimulator, maxNumSeqs, nil)
+}
+
+// CreateModelServiceWithExtraArgs is like CreateModelService but appends additional
+// CLI args to the model-server (simulator or vLLM) container. Used by tests that
+// need to inject engine-specific configuration such as --fake-metrics.
+func CreateModelServiceWithExtraArgs(ctx context.Context, k8sClient *kubernetes.Clientset, namespace, name, poolName, modelID string, useSimulator bool, maxNumSeqs int, extraArgs []string) error {
```
mamy-CS, line 36: must ensure extraArgs are valid for the chosen runtime correct? since this API does not validate flags.

test/e2e/saturation_v2_test.go
```diff
@@ -0,0 +1,278 @@
...
+	// chosen so the cost-aware optimizer's scale-down rule fires, V2
+	// recommends a smaller target. Uses canonical-ordering thresholds
+	// (scaleUpThreshold > scaleDownBoundary):
+	//
+	//   scaleUpThreshold  = 0.95 (high, so kv=0.4 demand does not trigger scale-up)
+	//   scaleDownBoundary = 0.85 (chosen so spareCapacity at 2 replicas exceeds
+	//                             one full per-replica capacity — see calibration
+	//                             comment on v2SmokeFakeMetricsJSON for the math)
+	//
+	// Assertion is "<2" rather than "==1" because V2 routes through the
```
mamy-CS: Is 0 replicas actually reachable here given VA minReplicas=1 from EnsureVariantAutoscalingWithDefaults, or does status sometimes show 0 while spec stays at 1? If 1 is the only realistic outcome, would Equal(1) be clearer and less alarming to readers?

test/e2e/saturation_v2_test.go
```diff
@@ -0,0 +1,278 @@
...
+// (cost_aware_optimizer.go: replicasToRemove := floor(spare / perReplicaCapacity)).
+// At kv=0.4 with 2 replicas, that lower bound is scaleDownBoundary ≥ 0.8.
+//
+// The simulator docs note that --fake-metrics replaces real metric emission
+// entirely, so traffic against the model service has no effect on these
+// values — that's the point: the test exercises V2's decision logic against
+// known inputs rather than depending on the simulator's runtime fidelity.
+//
+// Format documented at:
+//
```
mamy-CS: Do we want this math in-repo long-term, or is a short summary + link to a docs good for maintainability?

test/e2e/saturation_v2_test.go
ev-shindin: Skip message names the env var (`USE_SIMULATOR`) to match the existing `scale_from_zero_test.go:160` convention so OpenShift-style runs against real vLLM can read the skip and know exactly what to flip.

test/e2e/fixtures/model_service_builder.go
ev-shindin: Right — and validating per-runtime would mean baking either an llm-d-inference-sim CLI grammar or a vLLM CLI grammar into the test fixtures, which I don't think we want. Done in commit 23055e12 by tightening the doc-comment contract instead.

test/e2e/saturation_v2_test.go
ev-shindin: Done in commit 23055e12

test/e2e/saturation_v2_test.go
ev-shindin: Done in commit 23055e12. The 30-line math block on `v2SmokeFakeMetricsJSON` is trimmed to a qualitative summary stating what kv-cache-usage=0.4 exercises at 1 and 2 replicas, citing `cost_aware_optimizer.go`'s scale-down formula by symbol (`math.Floor(remaining / vc.PerReplicaCapacity)`), and linking upstream to the simulator docs for the `--fake-metrics` JSON format.

I think creating doc will require a separate PR

mamy-CS (approved):
/lgtm /approve

Issue #913, opened 2026-03-23
Error Tracking Metrics


ev-shindin:
Add a counter to track errors by component, replacing the need to parse logs for error rates.

**Metrics:**
```
wva_errors_total                    counter {component, error_type}
```

**Implementation:**
- Components: `collector`, `analyzer`, `optimizer`, `limiter`, `enforcer`, `controller`
- Increment on every `logger.Error()` call in engine and controller
- Consider a helper `metrics.RecordError(component, errorType string)` to keep call sites clean

**Acceptance Criteria:**
- [ ] Error counter emitted for each component
- [ ] At least one error path per component has metric increment
- [ ] Unit tests verify counter increments on error paths

shuynh2017:
@ev-shindin, qq: just to make sure, what would be the `error_type`, is it just `msg string` in `logger.Error` ? e.g.;
```
logger.Error(err, "Failed to find VA for deployment event using index",
			"deployment", deploy.Name,
			"namespace", deploy.Namespace)
logger.Error(err, "Unable to fetch VariantAutoscaling",
			"name", req.Name,
			"namespace", req.Namespace)
logger.Error(err, "Failed to get scale target "+va.Spec.ScaleTargetRef.Kind,
			"name", scaleTargetName,
			"namespace", va.Namespace)
```

PR #1070, opened 2026-04-29
feat: Observability - Error Tracking Metrics

shuynh2017:
Ref:  https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/913. 

## Notes:
- For calls such as logger.V(logger.....)Error()... I choose to call metrics.RecordError() without any if condition since we might not want to add too much to our log but we should always record errors by metrics.
- There are some components such as analyzer, optimize, limiter, there are no logger.Error
- e2e can be done for some cases, but is difficult to do (or to do reliably) for other cases. So e2e test is limited.

## Sample Error metrics

1. Metadata
```
 "wva_errors_total": [
      {
        "type": "counter",
        "unit": "",
        "help": "Total number of errors by component"
      }
    ]
```
2. Sample query from promethteus after e2e test for `controller`:
```

  "metric": "wva_errors_total",
  "labels": {
    "component": "controller",
    "container": "manager",
    "endpoint": "https",
    "error_type": "Failed to parse saturation scaling config entry",
    "instance": "10.244.2.15:8443",
    "job": "workload-variant-autoscaler-metrics",
    "namespace": "workload-variant-autoscaler-system",
    "pod": "workload-variant-autoscaler-controller-manager-75968fcb5c-gq5mj",
    "service": "workload-variant-autoscaler-metrics"
  },
  "value": "3"
}
```

## Sample Dashboard Using Metric
image


lionelvillard:
### Code Review

Found 7 issues:

**Go Code Quality**

1. `error_type` local variable uses snake_case throughout — violates AGENTS.md naming convention — [pod_scraping_source.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/collector/source/pod/pod_scraping_source.go#L236), [pod_va_mapper.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/collector/source/pod_va_mapper.go#L89), [configmap_bootstrap.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/configmap_bootstrap.go#L23), [configmap_helpers.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/configmap_helpers.go#L33), [configmap_reconciler.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/configmap_reconciler.go#L63), [variantautoscaling_controller.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/variantautoscaling_controller.go#L133), [enforcer.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/engines/pipeline/enforcer.go#L89)
   > AGENTS.md: "Use **MixedCaps** or **mixedCaps** rather than underscores for multi-word names"

   Rename all `error_type :=` declarations to `errorType :=` across the 7 production files. Note that `pod_va_mapper.go` already correctly uses `const errorType` in the switch block — the `error_type :=` declarations in the same and other files should match that style.

2. Redundant import alias `metrics "…/internal/metrics"` — the package name is already `metrics`, so the alias is noise — [configmap_bootstrap.go#L10](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/configmap_bootstrap.go#L10), [configmap_helpers.go#L30](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/configmap_helpers.go#L30), [variantautoscaling_controller.go#L44](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/variantautoscaling_controller.go#L44)

3. New `Component*` constants block in [constants.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/constants/constants.go#L63-L69) has no group doc comment — only a trailing inline comment on the first constant
   > AGENTS.md: "Every exported name should have a doc comment" / "Start comments with the name being described"

**Security**

4. User-controlled CRD field embedded in Prometheus label — unbounded cardinality attack vector — [variantautoscaling_controller.go#L190](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/variantautoscaling_controller.go#L190)

   ```go
   error_type := "Failed to get scale target " + va.Spec.ScaleTargetRef.Kind
   metrics.RecordError(ctx, constants.ComponentController, error_type)
   ```

   `va.Spec.ScaleTargetRef.Kind` is an unconstrained `string` in the CRD schema (no `enum`/`maxLength`). Any user with `create`/`update` on `variantautoscalings` can create objects with arbitrary unique `.spec.scaleTargetRef.kind` values, generating a new `wva_errors_total` time series per value. A sustained stream of novel values can exhaust Prometheus memory (classic high-cardinality label DoS).

   Fix: use a fixed label `"Failed to get scale target"` (the kind is already captured in the structured log field next to it) — or restrict `scaleTargetRef.kind` to a `+kubebuilder:validation:Enum=Deployment;LeaderWorkerSet` marker in the API types.

**Test Coverage**

5. `configmap_reconciler_test.go` "Metrics Recording" context only calls `metrics.RecordError` directly — it never drives the reconciler through a failing `Get()`, so `configmap_reconciler.go:63` (`"Failed to get ConfigMap"`) is not integration-tested — [configmap_reconciler_test.go#L487](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/configmap_reconciler_test.go#L487)

   The `failingK8sClient` helper defined in `configmap_bootstrap_test.go` already has `failGet` support — wire it up here.

6. `variantautoscaling_controller_test.go` "should record error when failing to update status" creates no metric assertion, and with envtest `Status().Patch()` succeeds (does not error), so `RecordError("Failed to update VariantAutoscaling status")` is never reached in the test — [variantautoscaling_controller_test.go#L560](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/variantautoscaling_controller_test.go#L560)

   Either add a mock that fails `Status().Patch()` and assert the metric, or rename the test to reflect what it actually verifies.

**Library Reuse**

7. Four near-identical `getErrorMetricCount`/`getErrorMetricValue` helpers are copy-pasted across test files — [pod_scraping_source_test.go#L1022](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/collector/source/pod/pod_scraping_source_test.go#L1022), [enforcer_test.go#L228](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/engines/pipeline/enforcer_test.go#L228), [configmap_bootstrap_test.go#L450](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/14cdd37a2cc70f57857d42fb3c982e2c0a8bc6b6/internal/controller/configmap_bootstrap_test.go#L450)

   `metrics_test.go` already uses `testutil.ToFloat64(errorsTotal.With(labels))` from `github.com/prometheus/client_golang/prometheus/testutil` (already in go.mod). For tests inside `package metrics` that can access `errorsTotal`, that one-liner replaces every helper. For external-package tests, `testutil.GatherAndCount` or `testutil.GatherAndCompare` covers the same ground. The inconsistent label-matching logic in the helpers (e.g., `configmap_bootstrap_test.go`'s version matches if any expected label matches, rather than all) also risks false-positive test results.

docs/integrations/prometheus.md
```diff
@@ -1,6 +1,6 @@
 # Prometheus Integration
 
-WVA integrates with Prometheus to collect metrics from vLLM inference servers and expose custom autoscaling metrics. This guide covers Prometheus configuration, metric collection, and security best practices.
+WVA integrates with Prometheus to collect metrics from vLLM inference servers and expose custom autoscaling metrics. In addition, WVA also emits internal metrics for observability (See [Internal Metrics](#internal-metrics))). This guide covers Prometheus configuration, metric collection, and security best practices.
```
ev-shindin: Three closing parens at the end: `))` is one too many. Should be `(See [Internal Metrics](#internal-metrics))`.

docs/integrations/prometheus.md
```diff
@@ -241,4 +241,35 @@ abs(wva_desired_replicas - wva_current_replicas)
 
 # Scaling frequency by reason
 rate(wva_replica_scaling_total[5m]) by (reason)
+```
+
+## Internal Metrics
+### Error Metric
+Errors in WVA controller is available via `wva_errors_total` metric which has the following metadata:
```
ev-shindin: Should be: "Errors ... are available"

go.mod
```diff
@@ -118,5 +119,5 @@ require (
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.2 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
-	sigs.k8s.io/yaml v1.6.0 // indirect
+	sigs.k8s.io/yaml v1.6.0
```
ev-shindin: The diff moves `sigs.k8s.io/yaml v1.6.0 // indirect` → `sigs.k8s.io/yaml v1.6.0` (direct), but looks like no `.go` file in the PR imports `sigs.k8s.io/yaml` directly.

internal/metrics/metrics.go
```diff
@@ -416,6 +429,27 @@ func (m *MetricsEmitter) EmitReplicaMetrics(ctx context.Context, va *llmdOptv1al
 	return nil
 }
 
+// RecordError records an error metric for the specified component and error type
+// Callers MUST invoke InitMetrics before this method (the package-level
+// metric vars are nil otherwise, and the Set calls below would panic).
+// InitMetricsAndEmitter is the recommended construction path because it
+// performs the registration before returning the emitter.
+func RecordError(ctx context.Context, component, errorType string) {
```
ev-shindin: `ctx` is accepted but never read. Either drop it or use it by adding a `log.FromContext(ctx).V(logging.DEBUG).Info("RecordError called with uninitialized counter, dropping", "component", component, "errorType", errorType)` on the nil-guard path

ev-shindin (changes requested):
Issue #913's acceptance criterion says: *"At least one error path per component has metric increment."* The PR ships six `Component*` constants but only three are wired.  Either reword `Close #913` → `Refs #913` and reopen a tracker for the missing analyzer/optimizer/limiter wiring, or add a doc note in `prometheus.md` explicitly listing which components are wired today vs reserved for future use. Either makes the partial-completion honest; "Close #913" as written sets up an "I thought this was done" misunderstanding.

shuynh2017:
> Issue #913's acceptance criterion says: _"At least one error path per component has metric increment."_ The PR ships six `Component*` constants but only three are wired. Either reword `Close #913` → `Refs #913` and reopen a tracker for the missing analyzer/optimizer/limiter wiring, or add a doc note in `prometheus.md` explicitly listing which components are wired today vs reserved for future use. Either makes the partial-completion honest; "Close #913" as written sets up an "I thought this was done" misunderstanding.
- Updated Close -> Ref
- Added to `prometheus.md`. I didn't find any logger.Error() in other components so I think we should update `prometheus.md`:

## Operational Metrics
### `wva_errors_total`
- **Type**: Counter
- **Description**: Total number of errors by WVA components
- **Labels**:
  - `component`: Name of the component. The components are `collector`, `analyzer`, `optimizer`, `limiter`, `enforcer`, and `controller`. Currently, this metric is available for `collector`, `enforcer`, `controller`. It will be available for other components as these components handle errors
  - `error_type`: Short description of the error
- **Use Case**: Track errors in WVA by components

docs/integrations/prometheus.md
shuynh2017: fixed and updated to follow existing style as in `Replica Management Metrics
`

internal/metrics/metrics.go
shuynh2017: removed `ctx` from the function params to be consistent with other similar functions in this file.

go.mod
shuynh2017: added back `// indirect`

ev-shindin (approved):

PR #1075, opened 2026-04-30
config: allow opt-in HTTP Prometheus endpoints

### LINKED ISSUE #1074: ValidateTLSConfig unconditionally rejects http:// Prometheus URLs — no way to opt out
`internal/utils/tls.go` (`ValidateTLSConfig`) rejects any Prometheus base URL that doesn't start with `https://`. There is no configuration flag to allow plain HTTP.

What's the problem? E.g. `kube-prometheus-stack` exposes Prometheus on plain HTTP (port 9090) by default, and enabling TLS requires non-trivial manual configuration that is unnecessary for most deployments. Adding TLS to Prometheus purely to satisfy this check is undesirable. WVA does not seem to send any really sensitive data from my understanding. Additionally, some deployments might attempt to use Istio/mTLS instead of HTTPS connection.

**Current workaround** (from our dev deployment): deploy a dedicated nginx pod with a self-signed cert-manager certificate that terminates TLS and proxies to plain-HTTP Prometheus, then set `PROMETHEUS_TLS_INSECURE_SKIP_VERIFY=true`. This satisfies the URL check but provides no value, it's pure overhead to work around the validation.

**Request:** Add a flag (e.g. `--prometheus-allow-http` / `PROMETHEUS_ALLOW_HTTP=true`, or simply allow `http://` URLs when `insecureSkipVerify` is already set) to permit plain HTTP Prometheus connections for clusters that don't expose Prometheus over TLS.

sats-23:
Fixes #1074

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -419,15 +419,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Always validate TLS configuration since HTTPS is required
+	// Validate Prometheus transport configuration before creating the client.
 	if err := utils.ValidateTLSConfig(cfg); err != nil {
-		setupLog.Error(err, "TLS configuration validation failed - HTTPS is required")
+		setupLog.Error(err, "Prometheus transport configuration validation failed")
 		os.Exit(1)
 	}
 
 	setupLog.Info("Initializing Prometheus client",
 		"address", cfg.PrometheusBaseURL(),
-		"tlsEnabled", true,
+		"tlsEnabled", cfg.PrometheusBaseURL()[:5] == "https",
+		"allowHTTP", cfg.PrometheusAllowHTTP(),
 	)
 
 	// Create Prometheus client with TLS support
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -427,7 +428,7 @@ func main() {
 
 	setupLog.Info("Initializing Prometheus client",
 		"address", cfg.PrometheusBaseURL(),
-		"tlsEnabled", cfg.PrometheusBaseURL()[:5] == "https",
+		"tlsEnabled", strings.HasPrefix(cfg.PrometheusBaseURL(), "https"),
 		"allowHTTP", cfg.PrometheusAllowHTTP(),
 	)
 
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -162,6 +162,7 @@ func loadConfig(cfg *Config, flagSet *flag.FlagSet, configFilePath string) error
 	cfg.prometheus.bearerToken = v.GetString("PROMETHEUS_BEARER_TOKEN")
 	cfg.prometheus.tokenPath = v.GetString("PROMETHEUS_TOKEN_PATH")
 	cfg.prometheus.insecureSkipVerify = v.GetBool("PROMETHEUS_TLS_INSECURE_SKIP_VERIFY")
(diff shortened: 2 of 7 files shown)


lionelvillard:
### Code Review

**Go Code Quality**
- [ ] `cfg.PrometheusBaseURL()[:5]` panics on strings shorter than 5 chars — `cmd/main.go`

  Replace with `strings.HasPrefix(cfg.PrometheusBaseURL(), "https")`. This also eliminates a reimplementation of `strings.HasPrefix` that is already available in the standard library.

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/b0fcd006392bc08168092458d4a646c50110e053/cmd/main.go#L429-L432

- [ ] `CreateTLSConfig` doc comment is now factually wrong — `internal/utils/tls.go`

  The comment still says "TLS is always enabled for HTTPS-only support" after this PR adds HTTP support. Update it to reflect that the function is only called for HTTPS endpoints.

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/b0fcd006392bc08168092458d4a646c50110e053/internal/utils/tls.go#L16-L20

**Test Coverage**
- [ ] `PROMETHEUS_ALLOW_HTTP` missing from `testConfigFromEnv` cleanup keys — `internal/utils/tls_test.go`

  Any test that sets `PROMETHEUS_ALLOW_HTTP` will leak the env var to subsequent tests. Add it to the `keys` slice in `testConfigFromEnv`. Also missing: a test case for the new `default:` branch (unknown scheme such as `ftp://`).

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/b0fcd006392bc08168092458d4a646c50110e053/internal/utils/tls_test.go#L21-L30

- [ ] `TestLoad_PrometheusAllowHTTPFromEnv` does not check `err` from `Load()` — `internal/config/loader_test.go`

  If `Load()` returns an error, `cfg` is nil and the next line panics. Every other test in the file calls `t.Fatalf` on a non-nil error; this one is missing that guard.

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/b0fcd006392bc08168092458d4a646c50110e053/internal/config/loader_test.go#L417-L440

- [ ] `TestLoad_Defaults` missing assertion that `PrometheusAllowHTTP()` defaults to `false` — `internal/config/loader_test.go`

  This is a security-sensitive default; a future regression changing it to `true` would go undetected without an explicit assertion.

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/b0fcd006392bc08168092458d4a646c50110e053/internal/config/loader_test.go#L1-L50

**Security**
- [ ] Bearer token sent in plaintext when `PROMETHEUS_ALLOW_HTTP=true` — `internal/utils/prometheus_transport.go`

  `bearerTokenRoundTripper` adds `Authorization: Bearer <token>` on every request regardless of scheme. When HTTP is opted into and a token is configured, credentials are transmitted in cleartext. Fix: refuse to send the `Authorization` header when `req.URL.Scheme == "http"`, or reject the combination of `allowHTTP=true` + any bearer token at startup.

- [ ] TLS cert fields silently ignored for HTTP endpoints — `internal/utils/tls.go`

  `CreateTLSConfig` is still called for HTTP URLs (via `CreatePrometheusTransport`). A `tls.Config` is built and attached to the transport, but Go's `http.Transport` silently ignores it for plain HTTP. An operator combining `PROMETHEUS_ALLOW_HTTP=true` with `PROMETHEUS_CA_CERT_PATH` will believe mTLS is active when it is not. Emit a warning (or return an error) when TLS cert fields are set alongside an HTTP URL.

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/b0fcd006392bc08168092458d4a646c50110e053/internal/utils/tls.go#L88-L97

sats-23:
Hey @lionelvillard 
The comments have been addressed, PTAL

PR #1078, opened 2026-05-01
improve pr-review skill: idempotent comments with checkbox diff

lionelvillard:
## Summary

- `/pr-review` now checks for an existing Claude Code review comment on the PR (identified by the `🤖 Generated with [Claude Code]` signature) before posting
- **First run**: creates a new comment as before, but using markdown checkboxes (`- [ ]`) instead of a numbered list
- **Re-run**: updates the existing comment in-place via `gh api PATCH`, diffing old issues against new findings:
  - `- [x] ~~issue~~` for items that are no longer found (resolved)
  - `- [ ] issue` for items still present
  - `- [ ] issue *(new)*` for newly discovered items
- Added `Bash(gh api:*)` to `allowed-tools` so the PATCH call is permitted

## Test plan

- [ ] Run `/pr-review` on a PR with no existing Claude comment → new comment created with checkboxes
- [ ] Fix one of the reported issues, then re-run `/pr-review` on the same PR → existing comment updated, fixed item marked `[x]`
- [ ] Run `/pr-review` on a PR where all issues have been fixed → all items marked resolved

---

Assisted-by: Claude Code (claude-sonnet-4-6)

### DIFF
--- a/.claude/skills/pr-review/SKILL.md
+++ b/.claude/skills/pr-review/SKILL.md
@@ -2,7 +2,7 @@
 name: pr-review
 description: Review an open pull request for this project. Checks Go code quality (AGENTS.md), test coverage, and Kubernetes security. Posts findings as a GitHub PR comment (idempotent: creates on first run, updates with resolved/new diff on re-runs). Use when the user asks to review a PR. Invoke with /pr-review [PR-number] or /pr-review to auto-detect the current branch's PR.
 disable-model-invocation: true
-allowed-tools: Bash(gh pr view:*), Bash(gh pr diff:*), Bash(gh pr comment:*), Bash(gh pr list:*), Bash(gh api repos/*/issues/*/comments:*), Bash(gh api repos/*/issues/comments/*:*), Bash(git rev-parse:*), Bash(git log:*), Agent, TodoWrite
+allowed-tools: Bash(gh pr view:*), Bash(gh pr diff:*), Bash(gh pr comment:*), Bash(gh pr list:*), Bash(gh api repos/*/issues/*/comments:*), Bash(gh api repos/*/issues/comments/*:*), Bash(git rev-parse:*), Bash(git log:*), Bash(mktemp:*), Agent, TodoWrite
 ---
 
 # PR Review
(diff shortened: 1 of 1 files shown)


lionelvillard:
### Code Review

**Test Coverage / Logic**
- [x] ~~"All resolved" template missing `<!-- claude-pr-review -->` sentinel~~ *(resolved)*
- [x] ~~Heredoc placeholder `<updated body>` is never explicitly replaced~~ *(resolved)*
- [x] ~~Matching key has no fallback for issues without a `#Lnn` anchor~~ *(resolved)*
- [x] ~~No-output case from detection query is never defined as first-run~~ *(resolved)*

- [x] Heredoc instruction says "replace `<actual comment body>` **(and the `EOF` line)**" — the `EOF` terminator must be kept as-is; removing or replacing it produces a malformed heredoc. Fix: "replace only the `<actual comment body>` placeholder line; leave the `EOF` terminator intact." *(new)*

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/9fb0b19e39c51ef9f58ceeab1cadc751e99bb48c/.claude/skills/pr-review/SKILL.md#L109

- [x] Sentinel change silently breaks idempotency for pre-existing review comments *(new)* — PRs that already have a review comment posted by the old skill (which starts with `### Code Review`, not `<!-- claude-pr-review -->`) will never match the new `startswith` filter, causing every re-run to post a duplicate instead of updating. Fix: add a fallback in Step 1 — if the sentinel query returns no output, also check `contains("Generated with [Claude Code]")` and adopt that comment as the existing review; or document this as a known one-time duplicate on the first re-run after the upgrade.

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/9fb0b19e39c51ef9f58ceeab1cadc751e99bb48c/.claude/skills/pr-review/SKILL.md#L29-L34

- [x] Word-overlap computation is ambiguous: anchor tokens contaminate the word set *(new)* — matching rules 2–4 compute overlap on the full bullet line, which includes the trailing `` — `file.go#L42` `` suffix. Those tokens pollute the word set and can push overlap below 80% for semantically identical issues that differ only in wording. Fix: strip the `` — `...` `` suffix (the anchor reference) from each bullet before tokenizing.

  https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/9fb0b19e39c51ef9f58ceeab1cadc751e99bb48c/.claude/skills/pr-review/SKILL.md#L116-L122

PR #1092, opened 2026-05-05
va removal proposal

lionelvillard:
see proposal

### DIFF
--- a/docs/proposals/deprecate-va-crd.md
+++ b/docs/proposals/deprecate-va-crd.md
@@ -114,11 +114,6 @@ This migration is an opportunity to re-evaluate whether the existing analyzers a
 
 **Current approach:** Online Kalman filter learning service parameters (alpha, beta) from TTFT/ITL. M/G/1 model computes max throughput per replica under SLO constraints.
 
-**Questions:**
-1. **Is the Kalman filter justified?** If EPP provides stable TTFT/ITL metrics, a simpler exponential moving average or Prometheus `rate()` may suffice.
-2. **Is M/G/1 the right model?** vLLM uses continuous batching, not queue-then-serve. A throughput-ceiling model (observed max RPS at acceptable latency) may be more practical.
-3. **Should SLO targets drive scaling?** Alternatively, scale on utilization and let the scheduler handle SLO routing entirely in EPP.
-
 ### Recommendation
 
 Start with the **V1 saturation analyzer logic** (KV cache % + queue depth thresholds with spare-capacity triggers) for Phase 0-1. This is the simplest, most proven approach and maps directly to Prometheus recording rules. Defer token-capacity (V2) and queueing-model refinements to Phase 2-3 where WVA computes what PromQL cannot.
--- a/docs/proposals/deprecate-va-crd.md
+++ b/docs/proposals/deprecate-va-crd.md
@@ -72,7 +72,7 @@ The WVA system consists of:
 
 | # | Feature | Complexity | Strategy |
 |---|---------|-----------|----------|
-| 1 | Basic saturation scaling (KV cache + queue threshold) | Simple | Prometheus recording rules + HPA/KEDA |
+| 1 | V1 saturation scaling (KV cache % + queue depth thresholds) | Simple | Prometheus recording rules + HPA/KEDA (primary path) |
 | 2 | Token-based capacity (V2 analyzer) | Medium | Needs reassessment (see below) |
 | 3 | Cost-aware multi-variant optimization | Hard | WVA coordinator adjusting HPA bounds |
 | 4 | GPU-limited fair-share | Hard | Same coordinator with GPU inventory awareness |
@@ -89,9 +89,15 @@ The WVA system consists of:
 
 This migration is an opportunity to re-evaluate whether the existing analyzers are the right approach for an HPA/KEDA-native model.
 
-### V1 Saturation Analyzer
+### V1 Saturation Analyzer (Percentage-Based)
 
-Already deprecated. Will be removed — not ported.
+**Current approach:** Uses two metrics — KV cache usage (0.0-1.0) and queue length (integer count) — compared against configurable thresholds. Scale-up triggers when average spare capacity (threshold - usage) drops below a spare trigger. Scale-down includes a redistribution safety check: simulates removing one replica by applying a `N/(N-1)` load scale factor and verifying spares remain positive.
+
+**Configuration:** `KvCacheThreshold`, `QueueLengthThreshold`, `KvSpareTrigger`, `QueueSpareTrigger`, `MinNonSaturatedReplicasForScaleDown = 2`
+
+**Assessment:** This is the most natural fit for recording rules + HPA/KEDA. The logic is simple threshold arithmetic on percentage metrics that Prometheus already exposes. The redistribution safety check (scale-down guard) is slightly more involved but can be approximated with a recording rule that accounts for per-replica load redistribution.
+
+**Port strategy:** Express directly as Prometheus recording rules. This becomes the primary scaling signal for Phase 0-1.
 
 ### V2 Token-Based Analyzer
 
@@ -115,7 +121,7 @@ Already deprecated. Will be removed — not ported.
 
 ### Recommendation
 
-Start with **utilization-based scaling** (KV cache usage + queue depth) for Phase 0-1. This is expressible in recording rules today. Defer token-capacity and queueing-model refinements to Phase 2-3 where WVA computes what PromQL cannot.
+Start with the **V1 saturation analyzer logic** (KV cache % + queue depth thresholds with spare-capacity triggers) for Phase 0-1. This is the simplest, most proven approach and maps directly to Prometheus recording rules. Defer token-capacity (V2) and queueing-model refinements to Phase 2-3 where WVA computes what PromQL cannot.
 
 ---
 
@@ -157,39 +163,74 @@ metadata:
 
 ### Prometheus Recording Rules
 
-**Tier 1 — Basic Saturation Signals:**
+**Tier 1 — V1 Saturation Signals (per-deployment):**
 ```yaml
 groups:
 - name: llmd_autoscaling
   interval: 30s
   rules:
+  # Peak KV cache usage per deployment (1m window catches bursts between scrapes)
   - record: llmd:variant_kv_saturation
     expr: |
       max by (namespace, model_name, deployment) (
         max_over_time(vllm:kv_cache_usage_perc[1m])
       )
 
+  # Peak queue depth per deployment
   - record: llmd:variant_queue_depth
     expr: |
       max by (namespace, model_name, deployment) (
         max_over_time(vllm:num_requests_waiting[1m])
       )
+
+  # Spare KV capacity (positive = headroom, negative = saturated)
+  # KvCacheThreshold default: 0.80
+  - record: llmd:variant_kv_spare
+    expr: |
+      0.80 - llmd:variant_kv_saturation
+
+  # Spare queue capacity
+  # QueueLengthThreshold default: 5
+  - record: llmd:variant_queue_spare
+    expr: |
+      5 - llmd:variant_queue_depth
 ```
 
-**Tier 2 — Utilization-Based Desired Replicas:**
+**Tier 2 — V1 Desired Replicas (scale-up/down logic):**
 ```yaml
+  # Current replica count per deployment
+  - record: llmd:variant_current_replicas
+    expr: |
+      count by (namespace, model_name, deployment) (
+        vllm:kv_cache_usage_perc
+      )
+
+  # Desired replicas based on V1 saturation logic:
+  # Scale-up when spare < trigger (KvSpareTrigger default: 0.10)
+  # Uses ratio: current_replicas * (usage / target) to derive needed count
   - record: llmd:variant_desired_replicas
     expr: |
       ceil(
-        count by (namespace, model_name, deployment) (vllm:kv_cache_usage_perc)
-        * max by (namespace, model_name, deployment) (
-            max_over_time(vllm:kv_cache_usage_perc[1m])
-          )
+        llmd:variant_current_replicas
+        * llmd:variant_kv_saturation
         / 0.80
       )
+
+  # Scale-down guard: simulated load after removing one replica (N/(N-1) factor)
+  # Only safe when redistributed spare remains positive
+  - record: llmd:variant_scaledown_safe
+    expr: |
+      (
+        0.80 - (
+          llmd:variant_kv_saturation
+          * llmd:variant_current_replicas
+          / (llmd:variant_current_replicas - 1)
+        )
+      ) > 0
+      and llmd:variant_current_replicas > 2
 ```
 
-Note: This is the simplified utilization-based approach. Full token-capacity formulas, if needed, are computed by WVA and emitted as metrics.
+Note: Threshold values (0.80, 5, etc.) are defaults. These can be parameterized per-deployment via separate recording rules or overridden by the WVA coordinator in Phase 2-3. Full token-capacity formulas (V2), if needed, are computed by WVA and emitted as metrics.
 
 **Tier 3 — Scale-from-Zero Signal:**
 ```yaml
--- a/docs/proposals/deprecate-va-crd.md
+++ b/docs/proposals/deprecate-va-crd.md
@@ -1,0 +1,341 @@
+# Proposal: Deprecate VariantAutoscaling CRD in Favor of HPA/KEDA
+
+**Authors:** [TBD]
+**Status:** Draft
+**Created:** 2026-05-05
+**Last Updated:** 2026-05-05
+
+---
+
+## Problem Statement
+
+The Workload Variant Autoscaler (WVA) introduces a custom VariantAutoscaling CRD and a dedicated controller that acts as an intermediary between metrics sources (vLLM, EPP) and the actual scaling mechanism (HPA/KEDA). Today the flow is:
+
+```
+vLLM/EPP metrics → Prometheus → WVA controller → wva_desired_replicas metric → HPA/KEDA → scale
+```
+
+This indirection adds operational complexity:
+- A custom CRD that operators must learn and manage
+- A dedicated controller with its own failure modes, status management, and actuator logic
+- Tight coupling between optimization logic and the CRD lifecycle
+- Duplication of what HPA/KEDA already does (metrics → scaling decision → actuation)
+
+Much of the WVA logic can be expressed as Prometheus recording rules consumed directly by HPA/KEDA, reducing the architecture to:
+
+```
+vLLM/EPP metrics → Prometheus recording rules → HPA/KEDA → scale
+```
+
+---
+
+## Goals
+
+1. Eliminate the VariantAutoscaling CRD as the primary user-facing API
+2. Make HPA or KEDA ScaledObject the entry point for autoscaling configuration
+3. Represent variant cost as an annotation on HPA/KEDA rather than a CRD field
+4. Enable simple deployments (single variant per model) to work without any custom controller
+5. Retain advanced features (multi-variant cost optimization, queueing model) in a simplified WVA that operates on standard HPA/KEDA objects
+6. Reassess existing analyzers for fitness in the new architecture
+
+## Non-Goals
+
+- Rewriting HPA or KEDA — we use them as-is
+- Building a general-purpose autoscaling framework
+- Changing how vLLM exposes metrics (though we may request upstream improvements)
+- Replacing the EPP — we consume its existing metrics
+
+---
+
+## Background
+
+### Current WVA Architecture
+
+The WVA system consists of:
+
+1. **VariantAutoscaling CRD** (`llmd.ai/v1alpha1`): Fields include `scaleTargetRef`, `modelID`, `minReplicas`, `maxReplicas`, `variantCost`
+2. **Saturation Engine** (30s optimization loop): Collects vLLM metrics, runs analyzers (V1/V2 saturation, queueing model), runs optimizers (cost-aware, GPU-limited fair-share), applies enforcers (scale-to-zero)
+3. **Scale-from-Zero Engine**: Monitors EPP `inference_extension_flow_control_queue_size` to wake scaled-to-zero deployments
+4. **Metrics Emission**: Emits `wva_desired_replicas`, `wva_current_replicas`, `wva_desired_ratio`
+5. **External Actuation**: HPA (via prometheus-adapter) or KEDA ScaledObject reads `wva_desired_replicas` and performs actual scaling
+
+### Metrics Consumed
+
+| Source | Metrics |
+|--------|---------|
+| vLLM pods | `vllm:kv_cache_usage_perc`, `vllm:num_requests_waiting`, `vllm:cache_config_info`, `vllm:request_prompt_tokens_*`, `vllm:request_generation_tokens_*`, `vllm:request_success_total` |
+| EPP (scheduler) | `inference_extension_flow_control_queue_size`, `inference_extension_scheduler_attempts_total`, TTFT/ITL metrics |
+
+---
+
+## Feature Inventory & Porting Assessment
+
+| # | Feature | Complexity | Strategy |
+|---|---------|-----------|----------|
+| 1 | Basic saturation scaling (KV cache + queue threshold) | Simple | Prometheus recording rules + HPA/KEDA |
+| 2 | Token-based capacity (V2 analyzer) | Medium | Needs reassessment (see below) |
+| 3 | Cost-aware multi-variant optimization | Hard | WVA coordinator adjusting HPA bounds |
+| 4 | GPU-limited fair-share | Hard | Same coordinator with GPU inventory awareness |
+| 5 | Scale-to-zero | Simple | KEDA `idleReplicaCount: 0` / HPA `minReplicas: 0` |
+| 6 | Scale-from-zero | Medium | KEDA trigger on EPP flow control queue > 0 |
+| 7 | P/D disaggregation | Medium | Separate HPA/ScaledObject per role |
+| 8 | Queueing model (Kalman filter + M/G/1) | Hard | Needs reassessment (see below) |
+| 9 | Safety net metrics | Eliminated | No intermediary = no intermediary failure |
+| 10 | ConfigMap-based configuration | Simple | Moves to recording rule params + annotations |
+
+---
+
+## Analyzer Reassessment
+
+This migration is an opportunity to re-evaluate whether the existing analyzers are the right approach for an HPA/KEDA-native model.
+
+### V1 Saturation Analyzer
+
+Already deprecated. Will be removed — not ported.
+
+### V2 Token-Based Analyzer
+
+**Current approach:** Computes per-replica capacity in tokens using `cache_config_info` labels (`num_gpu_blocks`, `block_size`), average input/output tokens, and prefix cache hit rate. Derives k1 (memory-bound) and k2 (compute-bound) capacity.
+
+**Problem:** `cache_config_info` exposes capacity as labels on an info metric, not numeric gauges. The k1/k2 formula cannot be expressed in PromQL (you cannot do arithmetic on label values).
+
+**Options:**
+1. **Simplify to utilization-based:** Use `kv_cache_usage_perc` directly as a utilization signal. Recording rule: `desired = current_replicas * (kv_usage / target_utilization)`. No absolute token capacity needed.
+2. **Request upstream vLLM change:** Expose `num_gpu_blocks` and `block_size` as numeric gauge metrics. Then the formula is expressible in PromQL.
+3. **Keep in WVA:** WVA computes token capacity and emits `llmd:variant_per_replica_capacity` as a metric for recording rules to reference.
+
+### Queueing Model Analyzer
+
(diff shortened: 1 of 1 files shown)


docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+
+```
+vLLM/EPP metrics → Prometheus recording rules → HPA/KEDA → scale
+```
+
+---
+
+## Goals
+
+1. Eliminate the VariantAutoscaling CRD as the primary user-facing API
```
shuynh2017: do we need an acceptable upgrade story - this may be a go or no go?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+This indirection adds operational complexity:
+- A custom CRD that operators must learn and manage
+- A dedicated controller with its own failure modes, status management, and actuator logic
+- Tight coupling between optimization logic and the CRD lifecycle
+- Duplication of what HPA/KEDA already does (metrics → scaling decision → actuation)
+
+Much of the WVA logic can be expressed as Prometheus recording rules consumed directly by HPA/KEDA, reducing the architecture to:
+
+```
+vLLM/EPP metrics → Prometheus recording rules → HPA/KEDA → scale
```
shuynh2017: WVA controller is gone from arch?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+3. Represent variant cost as an annotation on HPA/KEDA rather than a CRD field
+4. Enable simple deployments (single variant per model) to work without any custom controller
+5. Retain advanced features (multi-variant cost optimization, queueing model) in a simplified WVA that operates on standard HPA/KEDA objects
+6. Reassess existing analyzers for fitness in the new architecture
+
+## Non-Goals
+
+- Rewriting HPA or KEDA — we use them as-is
+- Building a general-purpose autoscaling framework
+- Changing how vLLM exposes metrics (though we may request upstream improvements)
```
asm582: Add capacity management as a non-goal for now

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+vLLM/EPP metrics → Prometheus recording rules → HPA/KEDA → scale
+```
+
+---
+
+## Goals
+
+1. Eliminate the VariantAutoscaling CRD as the primary user-facing API
+2. Make HPA or KEDA ScaledObject the entry point for autoscaling configuration
+3. Represent variant cost as an annotation on HPA/KEDA rather than a CRD field
```
shuynh2017: the word `variant` has been implied `variantautoscaling` object/resource, it's confusing here?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+## Problem Statement
+
+The Workload Variant Autoscaler (WVA) introduces a custom VariantAutoscaling CRD and a dedicated controller that acts as an intermediary between metrics sources (vLLM, EPP) and the actual scaling mechanism (HPA/KEDA). Today the flow is:
+
+```
+vLLM/EPP metrics → Prometheus → WVA controller → wva_desired_replicas metric → HPA/KEDA → scale
+```
+
+This indirection adds operational complexity:
+- A custom CRD that operators must learn and manage
```
asm582: Developers also have to maintian backward compatiblity

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+### Metrics Consumed
+
+| Source | Metrics |
+|--------|---------|
+| vLLM pods | `vllm:kv_cache_usage_perc`, `vllm:num_requests_waiting`, `vllm:cache_config_info`, `vllm:request_prompt_tokens_*`, `vllm:request_generation_tokens_*`, `vllm:request_success_total` |
+| EPP (scheduler) | `inference_extension_flow_control_queue_size`, `inference_extension_scheduler_attempts_total`, TTFT/ITL metrics |
+
+---
+
+## Feature Inventory & Porting Assessment
```
asm582: Should we consider mapping out FMA integration in this section?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+
+| # | Feature | Complexity | Strategy |
```
asm582: Porting complexity has three levels: simple, hard, and medium. How are these levels achieved? Is it based on LoC changes analysis? What about NFR requirements?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+
+### V2 Token-Based Analyzer
+
+**Current approach:** Computes per-replica capacity in tokens using `cache_config_info` labels (`num_gpu_blocks`, `block_size`), average input/output tokens, and prefix cache hit rate. Derives k1 (memory-bound) and k2 (compute-bound) capacity.
+
+**Problem:** `cache_config_info` exposes capacity as labels on an info metric, not numeric gauges. The k1/k2 formula cannot be expressed in PromQL (you cannot do arithmetic on label values).
+
+**Options:**
+1. **Simplify to utilization-based:** Use `kv_cache_usage_perc` directly as a utilization signal. Recording rule: `desired = current_replicas * (kv_usage / target_utilization)`. No absolute token capacity needed.
+2. **Request upstream vLLM change:** Expose `num_gpu_blocks` and `block_size` as numeric gauge metrics. Then the formula is expressible in PromQL.
```
asm582: Is option 2 a viable option? What about back-level vLLM servers, which do not expose new metrics?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+
+### Recommendation
+
+Start with the **V1 saturation analyzer logic** (KV cache % + queue depth thresholds with spare-capacity triggers) for Phase 0-1. This is the simplest, most proven approach and maps directly to Prometheus recording rules. Defer token-capacity (V2) and queueing-model refinements to Phase 2-3 where WVA computes what PromQL cannot.
+
+---
+
+## Proposed Solution
+
+### Architecture Overview
```
asm582: It is not clear whether the two arch modes operate in isolation or in tandem.

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+    llmd.ai/variant-cost: "40.0"
+    llmd.ai/gpus-per-replica: "4"
+    llmd.ai/role: "decode"                  # "prefill" | "decode" | "both"
+    llmd.ai/priority: "1.0"                 # Fair-share priority
+    llmd.ai/scale-to-zero-retention: "10m"  # Idle period before scale-to-zero
+```
+
+### Prometheus Recording Rules
+
+**Tier 1 — V1 Saturation Signals (per-deployment):**
```
asm582: need an example with LWS

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+
+**Deliverables:**
+
+WVA evolves into a lightweight coordinator (no CRD) that:
+- Discovers work by watching HPA/ScaledObjects annotated with `llmd.ai/model-id` and `llmd.ai/variant-cost`
+- Groups HPAs by model-id
+- Reads `llmd:variant_desired_replicas` and per-replica capacity from Prometheus
+- Runs cost-aware allocation (cheap variants get higher maxReplicas, expensive get lower)
+- **Does NOT do scaling** — only adjusts HPA `spec.minReplicas` / `spec.maxReplicas` bounds
+- GPU-limited mode: reads node accelerator inventory, enforces global GPU ceiling
```
asm582: Why does cost optimization need GPU-limited mode?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,377 @@
...
+### What Still Requires WVA (Without the CRD)
+
+| Logic | WVA needed? | Why |
+|-------|------------|-----|
+| Basic saturation, scale-to/from-zero, P/D | **No** | Pure recording rules + HPA/KEDA |
+| Cost-aware multi-variant | **Yes** | Cross-deployment comparison |
+| GPU-limited fair-share | **Yes** | Global constraint enforcement |
+| Queueing model | **Yes** | Stateful parameter learning |
+
+WVA packages both the cost coordinator and queueing model as components of a single binary. It is much simpler than today's WVA: no CRD, no full optimization engine pipeline, no actuator/status management. It only adjusts HPA min/max bounds and emits computed metrics.
```
asm582: How can the queuing model consume cost optimization?

kaushikmitr:
This is great, operational complexity has been one of the major concerns we've noticed with WVA, and dropping the CRD in favor of annotations on the ScaledObject is a real reduction in user-facing surface. ALso +1 on the phasing and on keeping the analyzers/optimizer untouched.

One thing I'd like to push on, since this proposal opens the door: based on our customer use cases, a lot of the scaling we see is intra-cluster and often single-variant. For those users, the WVA controller deployment itself is still an extra moving part, even after the CRD goes away. Can we reduce the complexity further by splitting responsibilities between EPP and WVA?

Concretely:

Per-pool desired-replicas computation moves into EPP (as a plugin ). EPP exposes epp_desired_replicas directly. This kills the epp-saturation analyzer in WVA and lets EPP do it?
WVA controller stays, but only for cluster-wide concerns: cost-aware fair-share, GPU budget allocation across variants, and the V1/V2/queueing-model analyzers for non-EPP users. For single-variant or non-cost-optimizing users, WVA becomes optional, KEDA scales directly on the EPP metric.
This complements rather than competes with the CRD removal: the annotation-based discovery is exactly what makes WVA cleanly optional for the simple case. Multi-variant cost optimization is genuinely cluster-wide and doesn't fit in EPP, so WVA still earns its keep where it's needed, but a lot of users wouldn't need to deploy it at all (AFAIK).

Curious whether this lines up with where you were already heading, or whether there's a reason to keep per-pool sizing in WVA that I'm missing.

asm582:
FYI, I resolved all the comments, and the proposal has gone through changes, and now I think it is simplified.

lionelvillard:
thanks @kaushikmitr for you feedback. Some comments.

> Can we reduce the complexity further by splitting responsibilities between EPP and WVA?

Yes absolutely.

> Per-pool desired-replicas computation moves into EPP (as a plugin ). EPP exposes epp_desired_replicas directly. This > kills the epp-saturation analyzer in WVA and lets EPP do it?

That's an interesting idea worth exploring IMO.

> WVA controller stays, but only for cluster-wide concerns: ...
> //
> Curious whether this lines up with where you were already heading, or whether there's a reason to keep per-pool sizing in WVA that I'm missing.

Per-pool vs cluster-wide resonates a lot with me. However, I didn't go as far as thinking about packaging scaling-related metrics in EPP. Again, interesting idea worth exploring.

mamy-CS:
This def improves adoption and reduces api surface area. A couple of nits/ questions, more of transitional rigor like than architecture. 
- The proposal doesn’t define the canonical identity key (target ref? model-id? namespace+variant name?) 
- Phase 1 says VA + annotated ScaledObject/HPA run simultaneously, but doesn’t define what happens if both represent the same target/model
- The idea that wva manages via annotations is clear to me at a high level, but not operationally. What exactly is wva allowed to mutate (if anything), and what remains solely user/KEDA/HPA-owned?
- we might also need to add/ plan for validation errors
Other than tightening the plan proposal for migration, this direction makes sense to me.

lionelvillard:
@thanks @mamy-CS for looking at the proposal and comments.

> * The proposal doesn’t define the canonical identity key (target ref? model-id? namespace+variant name?)

Can you expand and explain what's the concern here?  

> * Phase 1 says VA + annotated ScaledObject/HPA run simultaneously, but doesn’t define what happens if both represent the same target/model

Good point. This is no different from having 2 HPA objects targeting the same scaleTargetRef, right?


> * The idea that wva manages via annotations is clear to me at a high level, but not operationally. What exactly is wva allowed to mutate (if anything), and what remains solely user/KEDA/HPA-owned?

To be documented. Thanks to SSA, ownership is at the field level with safeguarding. 

For KEDA, triggers array is not keyed based on the `name` fields, instead it is keyed based on the `type` field. Looks like a bug to me TBH. It limits what we can do but not a deal breaker.

> * we might also need to add/ plan for validation errors

yes. I'm adding this to the proposal, thanks.

>   Other than tightening the plan proposal for migration, this direction makes sense to me.
 great!

shuynh2017:
Just a thought: VA currently seems to only to hold data/configuration which may be can be moved to configmaps - and this may be cleaner than annotation to hpa/scaledobject - keeping a level of indirection may be cleaner here. In addition, containing replicas information in VA causes mismatching count and is confusing.

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+There are two levels of pluggability, each targeting a different entry point:
+
+### Level 1 — The ScaledObject/HPA as Integration Point (Low Barrier)
+
+The ScaledObject or HPA is the stable interface. Any metric producer can drive scaling by writing a compatible Prometheus metric. WVA is one such producer, not the only one. EPP is another producer. Teams that don't need multi-variant coordination, cost-aware optimization, or SLO guarantees never need to deploy WVA at all:
+
+| Scenario | What to deploy | Entry barrier |
+|----------|----------------|---------------|
+| Single model, load-based scaling | ScaledObject with Prometheus trigger | Low — standard KEDA |
+| Single model, saturation-aware | Recording rules + ScaledObject trigger | Low — PromQL only |
```
shuynh2017, line 61: Single model, saturation-aware, WVA also has some additional value-adds right?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+The trade-off is explicit: **Level 2 has a higher barrier** — understanding WVA's internal engine interface, contributing Go code, and running WVA as a dependency. Level 1 requires only YAML. Both are valid; the choice depends on how much of WVA's infrastructure the team wants to reuse versus build independently.
+
+### Granular Feature Opt-In
+
+Features within WVA remain modular and annotation-driven. Omitting an annotation means the feature is skipped or defaults are used:
+
+| Feature | Annotation required |
+|---------|-------------------|
+| WVA management | `llm-d.ai/managed: "true"` |
+| Multi-variant grouping | `llm-d.ai/model-id` |
```
shuynh2017, line 83: why don't we call this `llm-d.ai/variant-group`, `model-id` is overloaded.

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+| Cost-aware optimization | `llm-d.ai/variant-cost` |
```
shuynh2017, line 84: may be we can change the term `cost` to something that's less direct, e.g. `rank`?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+| GPU-limited fair-share | `llm-d.ai/gpus-per-replica` |
+| Scale-to-zero | `llm-d.ai/scale-to-zero-retention` |
+| P/D disaggregation role | `llm-d.ai/role` |
+
+This granularity means a single-model deployment that only needs saturation-based scaling pays no configuration cost for features it doesn't use.
```
shuynh2017, line 89: do we mean `configuration cost` or `optimization cost`?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+- Changing how KEDA/HPA consume `wva_desired_replicas`
+- Replacing vLLM or EPP metrics
+
+---
+
+## Proposed Solution
+
+### Discovery via Annotations
+
+WVA watches ScaledObjects (or HPAs) with the `llm-d.ai/managed: "true"` annotation. A ScaledObject without this annotation is invisible to WVA and can use any trigger it wants. Adding the annotation opts into WVA management; additional annotations enable specific features. Configuration currently in the VA CRD moves to annotations on the ScaledObject:
```
shuynh2017, line 116: WVA will only read from these scaledObject, and will not **manage**, i.e., update, delete right?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+
+WVA's internal pipeline remains the same:
+1. **Discovery**: Watch annotated ScaledObjects/HPAs instead of VA CRDs
+2. **Collection**: Query Prometheus for vLLM/EPP metrics (same queries, same collector)
+3. **Analysis**: Run saturation engine (V1, V2, queueing model — unchanged)
+4. **Optimization**: Cost-aware, GPU-limited fair-share (unchanged)
+5. **Expose**: Expose `wva_desired_replicas{variant_name="granite-13b", exported_namespace="production"}` (unchanged)
+6. **Actuation**: None — KEDA/HPA reads the metric and scales (unchanged)
+
+The ScaledObject now serves dual duty: it carries WVA annotations (discovery + configuration) and defines the KEDA trigger (unchanged). WVA reads the annotations; KEDA reads `spec.triggers`. No additional objects needed.
```
shuynh2017, line 162: WVA also reads the `scaleTargetRef` ?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+
+```yaml
+apiVersion: keda.sh/v1alpha1
+kind: ScaledObject
+metadata:
+  name: granite-13b-scaler
+  namespace: production
+  annotations:
+    llm-d.ai/managed: "true"
+    llm-d.ai/model-id: "ibm/granite-13b"
```
shuynh2017, line 126: Asked above: is  `model-id` the same as `mode-id` sent to the chat/completions request?

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+      metricType: "AverageValue"
+```
+
+### Annotation Schema
+
+Annotations are placed on the ScaledObject (or HPA if not using KEDA):
+
+| Annotation | Required | Default | Description |
+|-----------|----------|---------|-------------|
+| `llm-d.ai/managed` | Yes | — | Opt-in for WVA management |
```
shuynh2017, line 148: may want to consider adding a prefix such as `llm-d.ai/wva-managed`

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+
+**Goal:** WVA supports both VA CRDs and annotated ScaledObjects/HPAs simultaneously.
+
+**Deliverables:**
+- Add annotation-based discovery (watching ScaledObjects/HPAs) alongside existing CRD reconciler
+- WVA exposes `wva_desired_replicas` for both VA-discovered and annotation-discovered variants
+- Validation: annotated ScaledObject produces same metric output as equivalent VA CRD
+- Documentation for annotation-based setup
+
+**Success Criteria:** Users can run either mode; existing VA CRD deployments continue working unchanged.
```
shuynh2017, line 207: If there's both CRD and annotation for a variant, log warning and CRD is disregard?

lionelvillard:
> Just a thought: VA currently seems to only to hold data/configuration which may be can be moved to configmaps - and this may be cleaner than annotation to hpa/scaledobject - keeping a level of indirection may be cleaner here. 

On the other hand, this level of indirection causes friction. I suggest to start with annotations and reevaluate. 

A good use case for configmap is namespace-scoped and clusters-scope configuration. 

> In addition, containing replicas information in VA causes mismatching count and is confusing.

what do you mean?

docs/proposals/deprecate-va-crd.md
lionelvillard: yes that's correct, like limited-mode.

docs/proposals/deprecate-va-crd.md
lionelvillard: It's `model-id` because that's what VA defines today. We discussed to rename it to `inference-pool` to be aligned with llm-d terminology.

docs/proposals/deprecate-va-crd.md
lionelvillard: +1.  `priority` is also possible.

shuynh2017:
> > Just a thought: VA currently seems to only to hold data/configuration which may be can be moved to configmaps - and this may be cleaner than annotation to hpa/scaledobject - keeping a level of indirection may be cleaner here.
> 
> On the other hand, this level of indirection causes friction. I suggest to start with annotations and reevaluate.
> 
> A good use case for configmap is namespace-scoped and clusters-scope configuration.
> 
> > In addition, containing replicas information in VA causes mismatching count and is confusing.
> 
> what do you mean?

The configmap comment has been answered in the proposal. As for ` In addition, containing replicas information in VA causes mismatching count and is confusing.`, basically I am in agreement of removing VA to streamline.

docs/proposals/deprecate-va-crd.md
lionelvillard: configuration cost is correct. Here the concern is UX. Less configuration equals better UX.

docs/proposals/deprecate-va-crd.md
asm582: Yes, that's an implementation detail we can agree on.

docs/proposals/deprecate-va-crd.md
lionelvillard: That's partially correct. WVA does not create/delete HPA/KEDA objects but it can potentially manage some fields. See https://kubernetes.io/docs/reference/using-api/server-side-apply/ for more info. For instance, complex triggers may be populated by WVA.

docs/proposals/deprecate-va-crd.md
asm582: Yes, model-id is the same field also used by HuggingFace to download the model. this is from actually remnant from previous modelservice dev/design activity. it should now represent inferencepool which is in sync with EPP.

docs/proposals/deprecate-va-crd.md
lionelvillard: Yes and no, all depends what features are enabled.

asm582: Yes, `scaleTargetRef` is part of phase 1 where dual mode is supported, if thats what you are asking.

docs/proposals/deprecate-va-crd.md
lionelvillard: see answer above

docs/proposals/deprecate-va-crd.md
asm582: sure.

lionelvillard: `llm-d.ai` is good enough IMO. Even `wva.llm-d.ai` is too much.

docs/proposals/deprecate-va-crd.md
lionelvillard: yes! great idea! (@mamy-CS also called out this issue).

docs/proposals/deprecate-va-crd.md
shuynh2017: If there's plan to modify the resources, I think we should call this out very early, i.e. we don't want to push this deep in the doc and surprise users. As an example, populating triggers can be tricky especially when they don't work right away.

docs/proposals/deprecate-va-crd.md
ev-shindin: `priority` is bad I reserved it for model priorities in multi-model support. Putting it here will be confusing.

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+
+3. **Use labels instead of annotations** — Labels have character restrictions and are indexed (performance cost). Annotations are the Kubernetes convention for non-identifying metadata.
+
+4. **Annotate Deployments instead of ScaledObjects** — Deployments don't know about scaling policy; the ScaledObject/HPA is the natural place for scaling-related metadata since WVA's output feeds into it directly.
+
+---
+
+## Open Questions
+
+1. Should WVA watch only KEDA ScaledObjects, or also plain HPAs? Proposal: support both — watch any object with the `llm-d.ai/managed` annotation that has a `scaleTargetRef`.
```
ev-shindin, line 261: This will add some complexity because we can have both HPA and KEDA for same Deployment/LWS. But this is implementation details.

docs/proposals/deprecate-va-crd.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Alternatives Considered
+
+1. **Keep the CRD but make it optional** — Adds complexity (two discovery paths permanently) without ever removing the CRD burden. The dual-mode in Phase 1 is temporary.
+
+2. **Move everything to a ConfigMap** — ConfigMaps don't have per-object identity or namespace locality. Annotations on the ScaledObject/HPA keep configuration co-located with the scaling object.
+
+3. **Use labels instead of annotations** — Labels have character restrictions and are indexed (performance cost). Annotations are the Kubernetes convention for non-identifying metadata.
+
+4. **Annotate Deployments instead of ScaledObjects** — Deployments don't know about scaling policy; the ScaledObject/HPA is the natural place for scaling-related metadata since WVA's output feeds into it directly.
```
ev-shindin, line 255: What about another direction we followed - user just create VA and WVA creates KEDA/HPA object for this VA. This also will simplify UX?

mamy-CS:
Thanks @lionelvillard 
> @thanks @mamy-CS for looking at the proposal and comments.
> 
> > * The proposal doesn’t define the canonical identity key (target ref? model-id? namespace+variant name?)
> 
> Can you expand and explain what's the concern here?
> 
Since we are keeping both initially during transition I was wondering best way to identify so that the initial dual-mode until full migration can’t duplicate metrics or optimizer rows. 
> > * Phase 1 says VA + annotated ScaledObject/HPA run simultaneously, but doesn’t define what happens if both represent the same target/model
> 
> Good point. This is no different from having 2 HPA objects targeting the same scaleTargetRef, right?
> 
Partly, but I dont think its the same problem. multiple HPAs on the same object is a known footgun (competing writes to .spec.replicas). Bt now WVA adds another layer, it does not only scale the deployment. it also emits the desire replica target. Doesnt dual discovery without rules mean complexity (releated issue as above)
> > * The idea that wva manages via annotations is clear to me at a high level, but not operationally. What exactly is wva allowed to mutate (if anything), and what remains solely user/KEDA/HPA-owned?
> 
> To be documented. Thanks to SSA, ownership is at the field level with safeguarding.
> 
Yes, SSA is a good implementation detail after the product rule but mutating vs non-mutating WVA must be stated first I think.
> For KEDA, triggers array is not keyed based on the `name` fields, instead it is keyed based on the `type` field. Looks like a bug to me TBH. It limits what we can do but not a deal breaker.
> 
> > * we might also need to add/ plan for validation errors
> 
> yes. I'm adding this to the proposal, thanks.
> 
> > Other than tightening the plan proposal for migration, this direction makes sense to me.
> > great!

docs/proposals/deprecate-va-crd.md
lionelvillard: noted. Let's make sure we leave the door open for both options (ie. let's not document it)

docs/proposals/deprecate-va-crd.md
lionelvillard: this is a common k8s problem.

docs/proposals/deprecate-va-crd.md
lionelvillard: well this proposal is the opposite: users just create HPA/KEDA objects. Only one objects, not 2 (VA, HPA/KEDA). And sometimes they use WVA.

deanlorenz: WVA uses KEDA/HPA in a very specific way with a specific setup (eg the trigger). Having the WVA configuration on top of the HPA/KEDA config may lead potential conflicts with custom HPA/KEDA setups. That is, the propagation of constraints should be from WVA to KEDA rather than the other way around.

If there is a strong push towards a single object it should be VA, it becomes the only visible API. WVA can generate any HPA/KEDA artifacts it needs (the "natural" direction of config constraints).

deanlorenz:
I am late to the party, but would like to capture a few thoughts:
- IMHO, the pluggability discussion is orthogonal to the VA CRD issue and should be captured as a separate issue
  - Level 1 already works: if you don't want WVA don't create VA CRDs. The CRD doesn't gate any other scaling approach. 
  - Level 2 is an engine interface concern — it works identically whether variants are discovered via CRD or annotation.  
  
- I did not understand how the operational burden is improved by this. 
  - Users provide the same data either way (model ID, cost, min/max replicas). 
  - User may perceive an extra burden due to loss of CRD semantics (schema validation, status, `kubectl get va -A` discoverability, LFM)
  - An HPA admin would still need to understand WVA to understand the annotations. WVA owned config essentially bleeds into KEDA/HPA objects as annotations that are visible to all tooling and persist if WVA is uninstalled.

As said, the pluggability discussion is very important and deserves its own doc.

lionelvillard:
thanks @deanlorenz for taking the time to read this proposal. 

I generally agree with what you wrote. This proposal is really about providing a coherent story when bringing HPA/KEDA without WVA-optimizer into the picture, and how to seamlessly transition from without to with WVA by "just adding a new metric". There is a lot more to it but that the gist of it.

mamy-CS (approved):
/lgtm
/approve

Issue #1082, opened 2026-05-04
Observability: Add per-VA freshness `up` gauge for saturation metrics


ev-shindin:
### Problem

PR #933 introduced five saturation/capacity gauges that are recorded only when the optimization cycle produces a fresh decision for a variant:

```
wva_saturation_utilization
wva_spare_capacity
wva_required_capacity
wva_kv_cache_tokens_used
wva_kv_cache_tokens_capacity
```

The recording site at `internal/engines/saturation/engine.go:1181-1183` is gated:

```go
if hasDecision {
    act.RecordSaturationMetrics(ctx, decision)
}
```

When `hasDecision` is false (e.g. the analyzer ran but the optimizer produced no decision for this variant — sparse traffic, transient metric gap, role filter excluded the variant, etc.), the gauges keep their last-recorded values. Prometheus eventually marks the series stale (default ~5 minutes after last sample), but in the gap between the missed cycle and the staleness marker firing, **dashboards and alerts cannot tell "the system says utilization is 0.42" from "the system has not updated utilization in 4 minutes and 0.42 is just the stalest sample"**.

We deliberately did not add a `Delete*` call in this path (see PR #933 review thread): deleting the series would lose the time-series history, and the gauges are still semantically meaningful as "last known value" — the missing piece is a *freshness signal* alongside them, not deletion.

### Proposal

Add a per-VA freshness gauge published on the same emission cycle as the saturation gauges:

```
wva_saturation_metrics_up{variant_name, namespace}    gauge   1.0 = fresh this cycle, 0.0 = stale
```

Recording:

- Set to `1.0` whenever `RecordSaturationMetrics` is called (i.e. when the cycle produced a fresh decision for the variant).
- Set to `0.0` for variants the analyzer is aware of but did **not** produce a fresh decision for in this cycle (so dashboards see an explicit "stale" signal rather than series absence).
- The metric is removed when the VA is deleted (covered separately by `issue-wire-delete-saturation-metrics.md`).

This is cheaper, less invasive, and more honest than either:
- (a) deleting the gauges (loses last-known-value semantics; dashboards see series gaps that look identical to scrape failures), or
- (b) relying on Prometheus' implicit staleness marker (5-minute lag; not visible in PromQL boolean form without `absent()` tricks per series).

### Open design questions

[...]

PR #933, opened 2026-03-25
Feat: Add saturation and capacity Prometheus metrics for WVA observability

ev-shindin:
### Summary
- Add 5 new Prometheus gauges (`wva_saturation_utilization`, `wva_spare_capacity`, `wva_required_capacity`, `wva_kv_cache_tokens_used`, `wva_kv_cache_tokens_total`) emitted per variant after each scaling decision
- Enrich `VariantDecision` with `Utilization`, `SpareCapacity`, `RequiredCapacity`, `KvCacheTokensUsed`, and `KvCacheTokensTotal` fields for both V1 and V2 engine paths
- Emit metrics via `EmitSaturationMetrics` in `applySaturationDecisions` for every variant with a scaling decision

### Details

**New metrics** (emitted per reconciliation cycle per variant):

| Metric | Labels | Description |
|--------|--------|-------------|
| `wva_saturation_utilization` | variant_name, namespace, accelerator_type | Utilization ratio (0.0-1.0) |
| `wva_spare_capacity` | variant_name, namespace, accelerator_type | Spare capacity (0.0-1.0) |
| `wva_required_capacity` | variant_name, namespace | Model-level required capacity (>0 = scale-up needed) |
| `wva_kv_cache_tokens_used` | variant_name, namespace | Sum of KV cache tokens in use |
| `wva_kv_cache_tokens_total` | variant_name, namespace | Sum of KV cache token capacity |

**V1 path:** `enrichDecisionsFromReplicaMetrics` aggregates per-pod `ReplicaMetrics` by variant to compute utilization (avg KV cache usage) and KV token sums. `RequiredCapacity` is 1.0 if `shouldScaleUp`, else 0.0.

**V2 path:** `Utilization`, `SpareCapacity`, and `RequiredCapacity` are populated from `AnalyzerResult` in `buildDecisionsWithOptimizer`. `enrichDecisionsWithKvTokenData` adds KV token sums from `ReplicaMetrics`.

**Files changed:**
- `internal/constants/metrics.go` — 5 new metric constant definitions
- `internal/interfaces/saturation_analyzer.go` — 4 new fields on `VariantDecision`
- `internal/metrics/metrics.go` — metric registration and `EmitSaturationMetrics()`
- `internal/metrics/metrics_test.go` — 236-line test suite (4 test cases)
- `internal/engines/saturation/engine.go` — `enrichDecisionsFromReplicaMetrics`, `enrichDecisionsWithKvTokenData`, emission call in `applySaturationDecisions`
[...]

PR #1094, opened 2026-05-06
plan: KEDA-only v1 saturation scaling plan 

lionelvillard:


docs/plans/keda-direct-vllm.md
```diff
@@ -0,0 +1,581 @@
...
+    # threshold = queueLengthThreshold - queueSpareTrigger = 5 - 3 = 2
+    threshold: '2'
+    metricType: "Value"
+```
+
+### Edge Case: All Pods Saturated
+
+When all pods are saturated, the filtering produces an empty set and the recording rule returns no data. KEDA sees no metric and does not fire. WVA V1 would scale up in this case (avgSpare=0 < trigger).
+
+Mitigation: keep the Phase 1 rules deployed alongside Phase 2 as a fallback trigger, or accept this as a known divergence (if all pods are saturated, the system is already in trouble and other mechanisms like KEDA's `fallback` config will maintain minimum replicas).
```
ev-shindin, line 255: **This is the worst possible failure mode.** When all pods are saturated, the cluster is in trouble — every other autoscaler (HPA, KEDA, WVA) fires harder. Your design fails *closed* (no scale-up) at exactly the moment scale-up is most needed.

The "mitigation" of running Phase 1 alongside Phase 2 doesn't compose cleanly:
- Phase 1's trigger fires on `avg(all_pods) > T₁ − S₁`, which is over-aggressive in mixed-saturation scenarios (the very thing Phase 2 was added to fix).
- KEDA OR-s the triggers, so deploying both gives you Phase 1's behavior whenever Phase 1's threshold is crossed — Phase 2 becomes ornamental.

Concretely, this proposal's all-saturated handling becomes "use Phase 1 always" or "miss the surge entirely". Neither is acceptable parity.

docs/plans/keda-direct-vllm.md
```diff
@@ -0,0 +1,581 @@
...
+
+| Pod | kv_usage | queue | Saturated? |
+|-----|----------|-------|-----------|
+| A   | 0.40     | 1     | No |
+| B   | 0.35     | 1     | No |
+| C   | 0.45     | 1     | No |
+
+**WVA V1**: avgKvLoad = 0.40, scaleFactor = 3/2 = 1.5, afterRemoval = 0.60, remainingSpare = 0.20 ≥ 0.10 → **safe**
+**KEDA**: avg = 0.40. Ratio = 0.40/0.70 = 0.571 < 1. HPA wants scale-down, policy limits to -1 → **scale-down to 2**
+
```
ev-shindin, line 497: The scope table downgrades scale-down to "HPA policy (-1/min)", but this is not a small difference — WVA V1's scale-down algorithm is *conservative by design*:

> WVA V1: avgKvLoad = 0.40, scaleFactor = 3/2 = 1.5, afterRemoval = 0.60, remainingSpare = 0.20 ≥ 0.10 → safe

WVA V1 simulates `n → n−1` and checks the *projected* spare capacity after removal — refusing to scale down if doing so would saturate the survivors. The HPA policy (`-1/min`) in your plan does no such check; it simply ramps replicas down whenever the trigger metric falls below threshold.

This produces real, observable failures:
1. **Pre-saturation oscillation.** With 3 healthy pods at `kv=0.40` (average 0.40, threshold 0.70), HPA scales to 2. After scale-down, the remaining 2 pods inherit the 3-pod load: `kv ≈ 0.40 · (3/2) = 0.60`. If demand grows even slightly, kv crosses 0.70 and the system thrashes back up (1-minute cooldown after ramp-down). WVA V1's `remainingSpare ≥ trigger` check rejects this scale-down up front.
2. **Burst-trailing scale-down.** When a burst subsides, KEDA sees the average fall and ramps down 1 replica/min. WVA's projection-based approach holds replicas longer because it accounts for `(n−1)`-replica saturation under residual demand.

docs/plans/keda-direct-vllm.md
```diff
@@ -0,0 +1,581 @@
...
+Let N = set of non-saturated replicas, T₁ = kvCacheThreshold, S₁ = kvSpareTrigger.
+
+```
+avgSpareKv < S₁
+⟺ (1/|N|) * Σᵣ∈N (T₁ - kv(r)) < S₁
+⟺ T₁ - (1/|N|) * Σᵣ∈N kv(r) < S₁
+⟺ avg(kv, N) > T₁ - S₁
+```
+
+### KEDA Trigger Equivalence
```
ev-shindin, line 438: Strictly speaking, KEDA computes a **desired replica count per trigger** and uses **max** across them (this is the underlying HPA semantics — each trigger produces a target, the controller picks the largest). For binary scale-up/no-scale-up under a single threshold, max-of-targets is equivalent to OR-of-conditions, so the claim is correct in spirit.

But the plan never defines what "scale up" means in replica-count terms. WVA V1 produces a single integer scale-up step (one or more replicas, depending on internal logic). Your KEDA design lets the HPA scale-up policy determine the step size:

```yaml
scaleUp:
  policies:
  - type: Pods
    value: 1
    periodSeconds: 15
```

So KEDA fires "+1 replica every 15s" while the trigger remains crossed.

docs/plans/keda-direct-vllm.md
```diff
@@ -0,0 +1,581 @@
...
+---
+
+## Scope Limitations (vs. full WVA V1)
+
+| Aspect | WVA V1 | KEDA (Phase 1) | KEDA (Phase 2) |
+|--------|--------|----------------|----------------|
+| Scale-up trigger | Exact spare analysis | Conservative (more aggressive) | Exact parity |
+| Scale-down | N/(N-1) simulation | HPA policy (-1/min) | HPA policy (-1/min) |
+| All-saturated | Scales up | Scales up (natural) | No data (needs fallback) |
+| Multi-variant cost | Scales cheapest | One ScaledObject per deployment | Same |
```
ev-shindin, line 507: The table lists this as "One ScaledObject per deployment" without further commentary. This understates the loss.

WVA's value proposition includes **cost-aware optimization across heterogeneous accelerator pools** — given an SLO violation in `granite-13b`, WVA can route additional traffic to the cheaper accelerator type *first*, scale H100 only when A100 saturates, etc. The MILP-driven roadmap (PR #933 + PR #1051 + ongoing throughput-analyzer work) deepens this.

KEDA-per-deployment scales each `Deployment` independently. There is no notion of:
- Heterogeneous-pool routing
- Inter-variant cost objectives
- P/D disaggregation role-awareness
- Coordinated scaling under shared GPU pool capacity

For users whose primary goal is "replace WVA with KEDA," this plan delivers — but for users whose problem is "I have 4 model variants on a mixed H100/A100 pool with cost constraints," KEDA-direct is a step backward.

The plan should explicitly say: **this proposal is for single-variant or homogeneous-pool deployments; multi-variant cost optimization continues to require WVA.** Otherwise readers will assume "KEDA is enough" when it isn't for the use case the platform is being designed around.

docs/plans/keda-direct-vllm.md
ev-shindin, line 1: Before diving into the technical issues, a step back: **the plan's headline goal is "V1 equivalence", but V1 is the analyzer the project is in the middle of replacing**. V2 (the token-budget saturation analyzer) is already in production for users who opted in, and the throughput analyzer  currently in flight (PRs #933, #1051, and the Halfin-Whitt staffing roadmap) goes further still. Anchoring a new integration path on V1 pulls operational weight backward at exactly the moment the project's analytical surface is moving forward.

**What V1 doesn't capture, that V2 already does:**
- *Shape sensitivity.* V1's `kv_usage` threshold is shape-blind: the same `k = 0.7` means very different ITL on IL=256 vs IL=2048 traffic. V2's token-budget math (`KV_req = IL·(1−H%) + OL/2`, capacity = `k_sat·KV_max / KV_req`) corrects this. Encoding V1 thresholds in KEDA recording rules re-introduces the very gap V2 was designed to close.
- *Heterogeneous pods.* V1 averages percentages across pods with possibly different `KV_max` (e.g., paged-attention configs differing across replicas). V2 normalizes to tokens. The KEDA recording rule `avg by (model_name, namespace) (max by (pod) (max_over_time(...)))` quietly inherits V1's blind spot.
- *Prefix-cache awareness.* V2 reads `H%` and discounts effective IL. V1 (and this KEDA plan) does not.

**Why this matters more, not less, given project direction:**
Productionizing KEDA-direct against V1 cements V1 as a long-lived migration target — making it harder, not easier, to deprecate later, and forking integration energy away from the V2 roadmap. The plan's own "Scope Limitations" table is symptomatic: scale-down degraded, all-saturated mishandled, multi-variant cost lost. **Each of those gaps is harder to accept in V2's richer model than in V1's coarser one — which is exactly why this plan picked V1 to begin with.** That is a smell, not a justification.

**Predictive scaling is foreclosed entirely.**
Saturation-based autoscaling — V1 *or* V2 — is purely reactive: by the time the threshold is crossed, the surge that triggered it is already on the floor and the new replica is several minutes from being ready. Reactive systems are bounded below by `T_decision + T_cold_start`, which on commodity Kubernetes is routinely 5–8 minutes for vLLM (image pull, GPU node provisioning, weight load, warmup). A KEDA-on-saturation pipeline cannot close this gap structurally, no matter how clean its math is.

The project's parallel roadmap on workload prediction (issue #1082-tracked freshness work, the cold-start-aware predictor proposal, the per-`modelId` rate + shape forecasting) is precisely what *can* close this gap — by lifting demand into the supply/demand comparison `T_cold_start` ahead of when the surge actually arrives, via an element-wise `max(current, predicted)`. The throughput model is the natural carrier for that signal because its inputs (rate, IL, OL, H%) are exactly what a predictor would forecast.

Anchoring the public KEDA integration on saturation-only signals removes the hook predictive scaling would attach to. If users adopt KEDA-direct and tune their dashboards/alerts/cost-models around it, retrofitting predictive signals later means re-plumbing a downstream tooling layer that's already settled. **The architectural choice in this PR forecloses an entire category of capability the project is otherwise positioning itself to deliver.** That is worth pricing into the merge decision, not deferring to "future work."

ev-shindin (changes requested):
I don't think this suggestion as strategic. We are making large step backward, limiting oursef with Prometheus and KEDA capbilities and actually not gain something here. I suggest go over WVA installation and UI simplification, keeping existent logic in.

docs/plans/keda-direct-vllm.md
lionelvillard: Thanks, this is very informative. This PR is about providing an example of what can de done with KEDA without requiring an extra component. And this is only scratching the surface.

lionelvillard:
thanks @ev-shindin. I'm closing this PR as it gives the impression of duplicating existing WVA features, which is not the intent.

PR #1095, opened 2026-05-06
fix(GH): rename file after llm-d-infra changed

zdtsw:
ref https://github.com/llm-d/llm-d-infra/pull/153

cc @vivekk16


ev-shindin (commented):
Sorry I overlooked this. The issue is alrteady closed bby #1115

PR #1098, opened 2026-05-06
docs: remove How to Run section from benchmark.md

kahilam:
Removes the How to Run section from docs/benchmark.md as it is not needed in this file.


kahilam:
@asm582 removed how to run section.

asm582 (approved):

Issue #872, opened 2026-03-10
EPIC: refactor/cleanup install.sh and usage


shuynh2017:
Tasks:
- Doc: review and reduce visibility of install.sh as it's should only be used by developers/pipelines. The official installation path is via `helm`.
- Refactor: Split install.sh into infrastructure (llmd, prometheus, WVA, etc ...) and application (VAs)  - may be 2 scripts.
- Reduce: scripts to minimum (variables, Openshift/K8S, etc ...)
- Any other?
@asm582 @lionelvillard

lionelvillard:
@mamy-CS here is the issue related to deploy scripts simplification.

lionelvillard:
@mamy-CS @shuynh2017 here a plan for simplifying install.sh. Each step can be a PR (I think).


# install.sh Simplification Plan

## Background

`deploy/install.sh` currently orchestrates six distinct concerns in a single script:

1. WVA controller deployment (Helm)
2. Prometheus / monitoring stack
3. Scaler backend (KEDA or Prometheus Adapter)
4. Scaler / analyzer tuning (KV/queue thresholds, HPA stabilization)
5. HPA / VariantAutoscaling CR deployment
6. llm-d infrastructure — EPP, Gateway control plane, ModelService, vLLM configuration, HuggingFace token, repo cloning

The stated scope of `install.sh` is to **bootstrap infrastructure for running e2e tests**. It should not install inference servers, user-facing autoscaling resources, or fine-tuning knobs that belong to the application layer. The current bundling inflates the script with ~55 environment variables and ~150-line orchestration logic that mixes infrastructure concerns with inference-server deployment details.

A secondary `INFRA_ONLY=true` flag was introduced to skip VA/HPA chart CRs during e2e runs, but this was a patch: even in infra-only mode the script still calls `deploy_llm_d_infrastructure` (skipping only the ModelService helmfile release via a selector). The EPP and Gateway remain entangled inside `install.sh`.

---

## Proposed Changes

### Step 1 — Extract llm-d deployment into its own script

Create `deploy/install-llmd-infra.sh` by extracting all content currently in `deploy/lib/infra_llmd.sh` and the related orchestration in `install_core.sh`. This script owns:

- llm-d repo cloning
- HuggingFace token secret creation
- Gateway CRD installation and control plane setup (Istio / kgateway)
- EPP + ModelService deployment via helmfile
[...]

PR #1099, opened 2026-05-06
Phase 1 - refactor(deploy): install script simplification

mamy-CS:
Addresses - https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/872
This pr is focused on simplifying the install script and extracting the llm-d infra installation only (moving into deprecation soon, to use upstream installation). Specifically follows the proposed plan [here](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/872#issuecomment-4299924320)
Consecutive PRs will also deprecate the Helm installation, moving to Kustomize (matching upstream)

Main changes in this pr
- WVA install is slimmer: chart VA/HPA and capacity knobs aren’t crammed through install.sh
- llm-d installation is extracted out, not part of the install script, which now focuses on wva + monitoring
- No more interactive prompts in deploy scripts: missing tools fail unless bypassed

🤖 Generated with the help of [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci-benchmark.yaml
+++ b/.github/workflows/ci-benchmark.yaml
@@ -287,17 +287,18 @@ jobs:
       - name: Deploy WVA and llm-d infrastructure
         env:
           ENVIRONMENT: openshift
+          # Cluster provides gateway — skip install-llmd gateway helmfile.
           INSTALL_GATEWAY_CTRLPLANE: "false"
-          E2E_TESTS_ENABLED: "true"
           NAMESPACE_SCOPED: "false"
           LLMD_NS: ${{ env.LLMD_NAMESPACE }}
           WVA_NS: ${{ env.WVA_NAMESPACE }}
           CONTROLLER_INSTANCE: ${{ env.WVA_NAMESPACE }}
-          DEPLOY_VA: "false"
-          DEPLOY_HPA: "false"
+          # Benchmark does not use chart VA/HPA from install — workloads and autoscalers come from the benchmark job.
           DECODE_REPLICAS: "1"
           MONITORING_NAMESPACE: openshift-user-workload-monitoring
+          # See OpenShift e2e: user-workload Prometheus cannot bearer-auth to WVA /metrics.
           WVA_METRICS_SECURE: "false"
+          # Full capacity-scaling chart values (install.sh no longer passes these).
           KV_CACHE_THRESHOLD: "0.90"
           QUEUE_LENGTH_THRESHOLD: "10"
           KV_SPARE_TRIGGER: "0.05"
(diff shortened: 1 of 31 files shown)


mamy-CS:
openshift ci should pass once merged to main

lionelvillard:
### Code Review

**Test Coverage**
- [ ] `USE_SIMULATOR` flipped to `false` in the scale-from-zero full-e2e job; the `false` branch of `buildModelServiceDeployment` (GPU volume mounts, `HF_TOKEN` env var, non-simulator image) has no unit test — a regression in that path would only surface during a live OpenShift run — `test/e2e/fixtures/model_service_builder.go`
- [ ] `LLMD_PATCH_EPP_FLOW_CONTROL` conditional in `deploy/lib/infra_llmd.sh` is exercised only by the full e2e suite; `smoke-deploy-scripts` exits at `--help` before any conditional logic runs — `deploy/lib/infra_llmd.sh`

**Security**
- [x] `$CURRENT_MODEL_LABEL` is interpolated directly into a `kubectl patch --type=merge -p` JSON string; a model label value containing `"` or `\` would break the JSON payload. Fix: build the patch with `jq -n --arg label "$CURRENT_MODEL_LABEL" '{…}'` — `deploy/lib/infra_llmd.sh`
- [x] `$MODEL_ID` and `$ACTUAL_DEFAULT_MODEL` are interpolated into a `yq eval` expression string; values containing `"` or `|` (valid in HF model names) silently mis-configure the values file. Fix: pass as yq env vars (`MODEL_ID="$MODEL_ID" yq eval '(.. | select(. == env(MODEL_ID))) = env(MODEL_ID)'`) — `deploy/lib/infra_llmd.sh`
- [x] `$LLM_D_INFERENCE_SCHEDULER_IMG` is shell-concatenated into a `kubectl patch --type=json -p` JSON literal; a value containing `"` or `\` produces malformed JSON. Fix: escape via `jq` or use `kubectl set image` instead — `deploy/lib/infra_llmd.sh`
- [ ] `$DECODE_DEPLOY` is populated by `grep … | head -1` from any deployment in the namespace ending in `-decode`, then passed directly to `kubectl scale "deployment/$DECODE_DEPLOY"` and `kubectl rollout status`; a parallel CI job injecting a deployment with a crafted name into the shared namespace could redirect the scale. Fix: validate against `[[ "$DECODE_DEPLOY" =~ ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$ ]]` before use — `.github/workflows/ci-e2e-openshift.yaml`

mamy-CS:
@lionelvillard tightened what makes sense for this pr scope

Issue #918, opened 2026-03-23
Kubernetes Events for Scaling Decisions


ev-shindin:
Emit Kubernetes events on VariantAutoscaling resources for key scaling decisions and errors. Currently only `ServiceMonitorDeleted` warning events are emitted.

**Events:**

| Event | Type | Reason | When |
|-------|------|--------|------|
| Scale-up decision | Normal | `ScaledUp` | Target replicas increased |
| Scale-down decision | Normal | `ScaledDown` | Target replicas decreased |
| Limiter constrained | Warning | `ResourceConstrained` | Decision was limited by GPU availability |
| Metrics unavailable | Warning | `MetricsUnavailable` | Prometheus query failed or returned no data |
| Scale-to-zero applied | Normal | `ScaledToZero` | Enforcer set replicas to 0 (no active requests) |
| Optimization error | Warning | `OptimizationFailed` | Analysis/optimizer error for this VA |

**Implementation:**
- Use `r.Recorder.Eventf()` in `applySaturationDecisions()` and the reconciler
- Rate-limit: at most 1 event per VA per optimization cycle to avoid flooding the API server. The exception is when there's a scaling event and the decision is limited in which case there will be both scaling event and `Limiter constrained` events.
- Include relevant context in event message (e.g., "Scaled up from 2 to 4 replicas, saturation=0.85")

**Acceptance Criteria:**
- [ ] 6 event types emitted on appropriate conditions
- [ ] Events visible via `kubectl describe va <name>`
- [ ] Events are rate-limited (no flooding on rapid cycles)
- [ ] Unit/integration tests verify event emission

lionelvillard:
@shuynh2017 FYI (as discussed).

shuynh2017:
@lionelvillard @ev-shindin, can we consider also having these events to be sent to OTEL events if OTEL endpoint is specified?

shuynh2017:
start working on this issue ...

PR #1107, opened 2026-05-07
feat: Observability - K8S Events for Scaling Decisions

shuynh2017:
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/918
# Kubernetes Events for Scaling Decisions
Emit Kubernetes events on VariantAutoscaling resources for key scaling decisions and errors. Currently only `ServiceMonitorDeleted` warning events are emitted.

**Events:**

| Event | Type | Reason | When |
|-------|------|--------|------|
| Scale-up decision | Normal | `ScaledUp` | Target replicas increased |
| Scale-down decision | Normal | `ScaledDown` | Target replicas decreased |
| Limiter constrained | Warning | `ResourceConstrained` | Decision was limited by GPU availability |
| Metrics unavailable | Warning | `MetricsUnavailable` | Prometheus query failed or returned no data |
| Scale-to-zero applied | Normal | `ScaledToZero` | Enforcer set replicas to 0 (no active requests) |
| Optimization error | Warning | `OptimizationFailed` | Analysis/optimizer error for this VA |

**Implementation:**
- [x] Use `r.Recorder.Eventf()` in `applySaturationDecisions()` and the reconciler
- [x] Rate-limit: at most 1 event per VA per optimization cycle to avoid flooding the API server. The exception is when there's a scaling event and the decision is limited in which case there will be both scaling event and Limiter constrained events.
- [x] Include relevant context in event message (e.g., "Scaled up from 2 to 4 replicas, saturation=0.85")
  -  For `ScaledUp`, `ScaledDown`,  `ResourceConstrained`, `decision.Reason` is used. 
  -  For `MetricsUnavailable`, example reasons are `Failed to collect metrics for model`, `No saturation metrics available for model`.
  - For `ScaledToZero`, reason is `scalefromzero mode: pending request - scale-up`.

**Acceptance Criteria:**
- [x] 6 event types emitted on appropriate conditions
- [x] Events visible via `kubectl describe va <name>`
- [x] Events are rate-limited (no flooding on rapid cycles): **at most 1 event per VA per optimization cycle**
- [x] Unit test 
- [x] integration tests verify event emission

**Metrics**
- [x] **ScaledUp/Down:**
- Sample:
  ```
  Events:
  Type     Reason              Age                  From                                           Message
  ----     ------              ----                 ----                                           -------
  Normal   ScaledUp            13m (x4 over 15m)    workload-variant-autoscaler-saturation-engine
  Normal   ScaledUp            12s (x14 over 7m4s)  workload-variant-autoscaler-saturation-engine
  ```

- [ ] **ResourceConstrained**
- Sample:

- [x] **MetricsUnavailable**

- Sample output when disable Prometheus service:
  ```
  Events:
    Type     Reason              Age                  From                                           Message
    ----     ------              ----                 ----                                           -------
    Warning  MetricsUnavailable  8m20s (x2 over 12m)  workload-variant-autoscaler-saturation-engine  Failed to collect metrics for model
    Warning  MetricsUnavailable  6m34s                workload-variant-autoscaler-saturation-engine  No saturation metrics available for model
  ```
  
- [ ] **ScaledToZero**
- Sample:
  
- [x] **OptimizationFailed**
- Sample: from code inserted to issue the event. (It's not easy to trigger an error)
```
Events:
  Type     Reason              Age    From                                           Message
  ----     ------              ----   ----                                           -------
 Warning  OptimizationFailed  23s (x8 over 3m49s)  workload-variant-autoscaler-saturation-engine  Processing model (V1)
```


lionelvillard:
### Code Review

**Go Code Quality**
- [ ] `CollectReplicaMetrics` exported doc comment does not start with the identifier name — `internal/collector/replica_metrics.go#L55`
  > AGENTS.md: "Every exported name should have a doc comment" / "Start comments with the name being described." Replace `// This function wraps collectReplicaMetrics to make sure to always record K8S events if needed.` with `// CollectReplicaMetrics wraps collectReplicaMetrics and records a K8s warning event...`

- [ ] Nil panic risk in `recordMetricsUnavailableEvent` and `recordOptimizationFailedEvent`: both call `recorder.Eventf` without a nil guard; tests that pass `nil` to `NewEngine` (and `NewReplicaMetricsCollector`) will panic on any error code path that reaches these helpers — `internal/collector/replica_metrics.go#L50`, `internal/engines/saturation/engine.go#L373`
  > AGENTS.md: idiomatic Go — nil dereference causes a runtime panic. Either guard every call site consistently, or always inject a no-op recorder in tests.

- [ ] Inconsistent nil guard: `if e.recorder != nil` protects the scale-up/scale-down block inside `applySaturationDecisions` but the direct `e.recorder.Eventf` calls above it (for `K8SEventOptimizationFailed`) and inside `recordOptimizationFailedEvent` have no equivalent guard — `internal/engines/saturation/engine.go#L1328`
  > AGENTS.md: idiomatic Go — a field always set by the constructor should not be nil-checked in one place but not others.

- [ ] Scale-up/scale-down/resource-constrained events in `applySaturationDecisions` use the stale map pointer `va` rather than `&updateVa` (the freshly fetched version). Other `Eventf` calls in the same function already use `&updateVa` — `internal/engines/saturation/engine.go#L1326`
  > K8s best practice: events should reference the current resource version; using a stale pointer produces events associated with an outdated object.

**Test Coverage**
- [ ] `K8SEventResourceConstrained` (`WasLimited=true`) code path has no test at all — `internal/engines/saturation/engine.go#L1336`
  > A test should inject a mock GPU limiter that sets `WasLimited = true`, pass a `record.NewFakeRecorder`, and assert the `ResourceConstrained` warning event appears in the channel.

- [ ] The two new Ginkgo tests "should record K8SEventScaledUp/ScaledDown" comment out their assertions with "events may not be recorded" — they pass a `fakeRecorder` but never drain its channel — `internal/engines/saturation/engine_test.go#L650`
  > These tests verify no error occurs, not that the events were actually emitted. They should assert `Eventually(fakeRecorder.Events)` contains the expected event string.

- [ ] `recordOptimizationFailedEvent` call paths (error branches in `optimizeV1` and `optimizeV2`) are untested — `internal/engines/saturation/engine.go#L440`
  > No test triggers a data-preparation failure with a real recorder injected and then checks the `K8SEventOptimizationFailed` event was emitted.

- [ ] `TestEventRecordingOnScaleToZero` has a fallthrough default that treats a missing event as acceptable ("likely due to missing metrics source") — `internal/engines/scalefromzero/engine_test.go#L1033`
  > The test should set up the engine so the scale-to-zero code path is reached deterministically, then assert the event. Otherwise it gives false confidence.

**Security**
- [ ] Internal error strings (including potential internal IPs, pod names, and auth-failure details) are forwarded verbatim into K8s events via `msg+": "+err.Error()` — readable by anyone with `get events` RBAC in the namespace — `internal/engines/saturation/engine.go#L440`, `internal/engines/saturation/engine.go#L1174`
  > Use the static `msg` in the event and keep the full `err` in the structured log only (which already happens on the adjacent `logger.Error` call).

- [ ] `MetricsUnavailable` events fire on *every* optimization cycle when `replicaMetrics` is empty (level-triggered) — a model legitimately scaled to zero emits a `Warning` event every 30 s, flooding the event stream and obscuring real signals — `internal/collector/replica_metrics.go#L68`
  > Only emit the event on a state *transition* (edge-triggered): track whether metrics were available in the previous cycle and only fire when transitioning from available → unavailable.

shuynh2017:
> ### Code Review
> **Go Code Quality**
> 
> * [x]  `CollectReplicaMetrics` exported doc comment does not start with the identifier name — `internal/collector/replica_metrics.go#L55`
>   > AGENTS.md: "Every exported name should have a doc comment" / "Start comments with the name being described." Replace `// This function wraps collectReplicaMetrics to make sure to always record K8S events if needed.` with `// CollectReplicaMetrics wraps collectReplicaMetrics and records a K8s warning event...`
> * [x]  Nil panic risk in `recordMetricsUnavailableEvent` and `recordOptimizationFailedEvent`: both call `recorder.Eventf` without a nil guard; tests that pass `nil` to `NewEngine` (and `NewReplicaMetricsCollector`) will panic on any error code path that reaches these helpers — `internal/collector/replica_metrics.go#L50`, `internal/engines/saturation/engine.go#L373`
>   > AGENTS.md: idiomatic Go — nil dereference causes a runtime panic. Either guard every call site consistently, or always inject a no-op recorder in tests.
> * [x]  Inconsistent nil guard: `if e.recorder != nil` protects the scale-up/scale-down block inside `applySaturationDecisions` but the direct `e.recorder.Eventf` calls above it (for `K8SEventOptimizationFailed`) and inside `recordOptimizationFailedEvent` have no equivalent guard — `internal/engines/saturation/engine.go#L1328`
>   > AGENTS.md: idiomatic Go — a field always set by the constructor should not be nil-checked in one place but not others.
> * [x]  Scale-up/scale-down/resource-constrained events in `applySaturationDecisions` use the stale map pointer `va` rather than `&updateVa` (the freshly fetched version). Other `Eventf` calls in the same function already use `&updateVa` — `internal/engines/saturation/engine.go#L1326`
>   > K8s best practice: events should reference the current resource version; using a stale pointer produces events associated with an outdated object.
> 
> **Test Coverage**
> 
> * [x]  `K8SEventResourceConstrained` (`WasLimited=true`) code path has no test at all — `internal/engines/saturation/engine.go#L1336`
>   > A test should inject a mock GPU limiter that sets `WasLimited = true`, pass a `record.NewFakeRecorder`, and assert the `ResourceConstrained` warning event appears in the channel.
> * [x]  The two new Ginkgo tests "should record K8SEventScaledUp/ScaledDown" comment out their assertions with "events may not be recorded" — they pass a `fakeRecorder` but never drain its channel — `internal/engines/saturation/engine_test.go#L650`
>   > These tests verify no error occurs, not that the events were actually emitted. They should assert `Eventually(fakeRecorder.Events)` contains the expected event string.
> * [ ]  `recordOptimizationFailedEvent` call paths (error branches in `optimizeV1` and `optimizeV2`) are untested — `internal/engines/saturation/engine.go#L440`
>   > No test triggers a data-preparation failure with a real recorder injected and then checks the `K8SEventOptimizationFailed` event was emitted.
> * [x]  `TestEventRecordingOnScaleToZero` has a fallthrough default that treats a missing event as acceptable ("likely due to missing metrics source") — `internal/engines/scalefromzero/engine_test.go#L1033`
>   > The test should set up the engine so the scale-to-zero code path is reached deterministically, then assert the event. Otherwise it gives false confidence.
> 
> **Security**
> 
> * [x]  Internal error strings (including potential internal IPs, pod names, and auth-failure details) are forwarded verbatim into K8s events via `msg+": "+err.Error()` — readable by anyone with `get events` RBAC in the namespace — `internal/engines/saturation/engine.go#L440`, `internal/engines/saturation/engine.go#L1174`
>   > Use the static `msg` in the event and keep the full `err` in the structured log only (which already happens on the adjacent `logger.Error` call).
> * [x]  `MetricsUnavailable` events fire on _every_ optimization cycle when `replicaMetrics` is empty (level-triggered) — a model legitimately scaled to zero emits a `Warning` event every 30 s, flooding the event stream and obscuring real signals — `internal/collector/replica_metrics.go#L68`
>   > Only emit the event on a state _transition_ (edge-triggered): track whether metrics were available in the previous cycle and only fire when transitioning from available → unavailable.
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)

internal/engines/saturation/engine.go
```diff
@@ -392,6 +392,49 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
...
 
+func (e *Engine) recordOptimizationFailedEvent(
+	variantAutoscalings []llmdVariantAutoscalingV1alpha1.VariantAutoscaling,
+	reason string,
+) {
+	if e.Recorder == nil {
+		return
+	}
+	for _, va := range variantAutoscalings {
+		e.Recorder.Eventf(&va, corev1.EventTypeWarning, constants.K8SEventOptimizationFailed, reason)
```
ev-shindin: `reason` is passed as the `Eventf` format string — a `%` in it would corrupt the message. Use `Event` (no format args here)

internal/engines/saturation/engine.go
```diff
@@ -392,6 +392,49 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
...
+	action interfaces.SaturationAction,
+	targetReplicas int,
+	reason string,
+) {
+	if e.Recorder == nil {
+		return
+	}
+	switch action {
+	case interfaces.ActionScaleUp:
+		e.Recorder.Eventf(va, corev1.EventTypeNormal, constants.K8SEventScaledUp, reason)
```
ev-shindin: Same as line 403 - use `Event`

internal/engines/saturation/engine.go
```diff
@@ -392,6 +392,49 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
...
+	case interfaces.ActionScaleDown:
+		if targetReplicas == 0 {
+			e.Recorder.Eventf(va, corev1.EventTypeNormal, constants.K8SEventScaledToZero, reason)
```
ev-shindin: Same as line 403 - use `Event`

internal/engines/saturation/engine.go
```diff
@@ -392,6 +392,49 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
...
+		} else {
+			e.Recorder.Eventf(va, corev1.EventTypeNormal, constants.K8SEventScaledDown, reason)
```
ev-shindin: Same as line 403 - use `Event`

internal/collector/replica_metrics.go
```diff
@@ -63,31 +64,53 @@ import (
...
-// CollectReplicaMetrics collects per-replica metrics for all replicas of a model.
+func (c *ReplicaMetricsCollector) recordMetricsUnavailableEvent(
+	variantAutoscalings map[string]*llmdVariantAutoscalingV1alpha1.VariantAutoscaling,
+	reason string,
+) {
+	if c.recorder == nil {
+		return
+	}
+	for _, va := range variantAutoscalings {
+		c.recorder.Eventf(va, corev1.EventTypeWarning, constants.K8SEventMetricsUnavailable, reason)
```
ev-shindin: `reason` is passed as the `Eventf` format string — a `%` in it would corrupt the message. Use `Event` (no format args here)

internal/engines/saturation/engine.go
```diff
@@ -1362,11 +1414,19 @@ func (e *Engine) applySaturationDecisions(
...
+			// K8s best practice: events should reference the current resource version
+			e.recordOptimizationFailedEvent([]llmdVariantAutoscalingV1alpha1.VariantAutoscaling{updateVa}, msg)
+			logger.Error(err, msg, "variant", updateVa.Name)
 		} else {
 			// Only log detail if we had a decision or periodically (to avoid spamming logs on every loop for no-ops)
 			if hasDecision {
+				// Emit Kubernetes event for observability
+				e.recordScalingEvent(&updateVa, decision.Action, decision.TargetReplicas, decision.Reason)
+				if decision.WasLimited {
+					e.recordResourceConstrainedEvent(&updateVa, decision.Reason)
```
ev-shindin: This emits **two** events for a scaled-and-limited VA in one cycle, but #918 asks for "at most 1 event per VA per optimization cycle." Suggest either (a) emit only the `ResourceConstrained` warning when `decision.WasLimited` (skip the scaling event), or (b) confirm on the issue that scaling + warning is acceptable and adjust the criterion. There's no other per-VA-per-cycle guard, so this is the only place the limit is exceeded.

internal/engines/saturation/engine.go
```diff
@@ -446,11 +489,14 @@ func (e *Engine) optimizeV1(
 		data, err := e.prepareModelData(ctx, modelID, modelVAs, e.client)
 
 		if err != nil {
-			logger.Error(err, "Saturation data preparation failed", "modelID", modelID)
+			msg := "Saturation data preparation failed"
+			logger.Error(err, msg, "modelID", modelID)
+			e.recordOptimizationFailedEvent(modelVAs, msg)
```
ev-shindin, line 582: `OptimizationFailed` is recorded from six places, two of which loop over all of a model's VAs. Could a single VA receive more than one `OptimizationFailed` in one cycle (e.g. a model-level failure followed by a per-VA failure)? If the control flow doesn't already make these mutually exclusive, a brief guard or an early return would keep it to one. (K8s aggregates identical messages, but differing messages won't dedup.)

ev-shindin (changes requested):
Thanks for this — all six event types from #918 are emitted, attached to the VA (visible via `kubectl describe va`), and well covered by the three new event test files. A few things before merge:

1. **`Eventf` is used with the dynamic message in the *format* position** (6 call sites). `Eventf`'s 4th argument is a `fmt.Sprintf` format string, so any `%` in `reason` mangles the message (`%!X(MISSING)`). Today's reasons are pre-formatted so probably `%`-free, but it's a latent injection bug and an anti-pattern — and the correct pattern already exists in this file: `emitAcceleratorNotResolvedEvent` uses `Event` (literal message), not `Eventf`. Switch the six new calls to `Event` (inline suggestions below).
2. **Rate-limit "≤1 event per VA per cycle" isn't enforced** — a scaled-and-limited VA emits two events (`engine.go:1425-1427`). Either add a per-VA guard or confirm with the issue that two (scaling + warning) is acceptable.
3. Minor: `OptimizationFailed` is emitted from six sites — worth confirming they're mutually exclusive per VA per cycle.

internal/engines/saturation/engine.go
shuynh2017: Scaling and warning is acceptable. In this case, WVA manager consistently logs and records events for both, e.g.:
```
if d.WasLimited {
					logger.Info("Decision was limited by GPU availability",
						"variant", d.VariantName,
						"originalTarget", d.OriginalTargetReplicas,
						"limitedTarget", d.TargetReplicas,
						"limitedBy", d.LimitedBy)
				}
```
```
if hasDecision {
				// Emit Kubernetes event for observability
				e.recordScalingEvent(&updateVa, decision.Action, decision.TargetReplicas, decision.Reason)
				if decision.WasLimited {
					e.recordResourceConstrainedEvent(&updateVa, decision.Reason)
				}

				logger.Info("Successfully emitted metrics",
					"variant", updateVa.Name,
					"target", targetReplicas,
					"accelerator", acceleratorName)
			
}
```
- Updated https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1107/#issue-4401467846, https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/918#issue-4118894111

internal/engines/saturation/engine.go
shuynh2017: fixed.

internal/engines/saturation/engine.go
shuynh2017: fixed.

internal/engines/saturation/engine.go
shuynh2017: fixed.

internal/engines/saturation/engine.go
shuynh2017: fixed.

internal/collector/replica_metrics.go
shuynh2017: fixed.

internal/engines/saturation/engine.go
shuynh2017: Address "at most 1 event per VA per optimization cycle." by:
- centralize all calls to `e.Recorder.Event` by wrapping it with `func (e *Engine) recordEvent(` in `engine.go` which only calls `e.Recorder.Event` if there hasn't an event issued for the VA in this optimized cycle (with exception of `ResourceConstrained ` mentioned above).

shuynh2017:
@ev-shindin thank you again for the time for excellent review. Pls review again.

ev-shindin (commented):
Thanks — the `Eventf` → `Event` switch (printf-injection fixed) and the per-cycle `vaEventTracker` dedup both look good, and I verified the safe bits: reset per cycle, Go 1.25 (no loop-var aliasing), `updateVa` is loop-local (no all-but-first drop), single-threaded optimize (no map race).

One remaining issue: **the dedup is keyed by `*VariantAutoscaling` (pointer), not VA identity.** The same logical VA is held by different pointers at the different emit sites in one cycle — `vaMap` stores `&activeVAs[i]`, `recordScalingEvent` gets a DeepCopy's address, `recordOptimizationFailedEvent` gets a range-copy `&va`, and the collector uses its own pointer. So "one event per VA" holds *within* a site but not *across* sites (e.g. a collector `MetricsUnavailable` and an engine scaling event for the same VA would be keyed differently and both emit). It mostly survives today only because the failure/success paths rarely overlap — i.e. it relies on call-site exclusivity, the loop-local `updateVa`, and single-threadedness, any of which a refactor could break silently.

Fix is trivial and consistent with existing code — key by namespace/name, which the repo already does for `vaMap` (`utils.GetNamespacedKey(...)`, engine.go:381). Suggestions (by Claude) below. (Also: the collector currently *checks* the tracker but never *stores* into it, so `MetricsUnavailable` never participates in the ≤1 guarantee — the collector suggestion adds the store.)
## `internal/engines/saturation/engine.go`

### Line 127 — declaration → string key
````
```suggestion
	vaEventTracker map[string]bool
```
````

### Line 355 — init → string key
````
```suggestion
	e.vaEventTracker = make(map[string]bool)
```
````

### Lines 479–487 — `recordEvent` dedup: key by identity
````
```suggestion
	key := utils.GetNamespacedKey(va.Namespace, va.Name)
	if e.vaEventTracker != nil {
		if _, ok := e.vaEventTracker[key]; ok { // ensures only one event is recorded per VA
			return
		}
	}
	e.Recorder.Event(va, eventType, reason, message)
	if e.vaEventTracker != nil {
		e.vaEventTracker[key] = true
	}
```
````
(`utils` is already imported here — it's used at line 381.)

---

## `internal/collector/replica_metrics.go`

### `recordMetricsUnavailableEvent` signature — string key
On the `vaEventTracker map[*llmdVariantAutoscalingV1alpha1.VariantAutoscaling]bool,` parameter line:
````
```suggestion
	vaEventTracker map[string]bool,
```
````

### `CollectReplicaMetrics` signature — string key
Same change on the `vaEventTracker` parameter of `CollectReplicaMetrics` (the other `map[*…VariantAutoscaling]bool` param):
````
```suggestion
	vaEventTracker map[string]bool,
```
````

### The emit loop — key by identity **and** store after emitting
````
```suggestion
	for _, va := range variantAutoscalings {
		key := utils.GetNamespacedKey(va.Namespace, va.Name)
		if vaEventTracker != nil {
			if _, ok := vaEventTracker[key]; ok { // ensures only one event is recorded per VA
				continue
			}
		}
		c.recorder.Event(va, corev1.EventTypeWarning, constants.K8SEventMetricsUnavailable, reason)
		if vaEventTracker != nil {
			vaEventTracker[key] = true
		}
	}
```

ev-shindin (approved):
Thank you @shuynh2017

PR #1113, opened 2026-05-10
engines/saturation: generic multi-analyzer pipeline with any-up/all-down combine

deanlorenz:
## Summary

- Replaces the hardcoded saturation-only path in `runAnalyzersAndScore` with a generic map-based loop over `config.Analyzers`. Each enabled analyzer is looked up in `Engine.analyzers` (`name → interfaces.Analyzer`) and called via the common `Analyze()` interface.
- Combines results in dimensionless utilisation space (any-up / all-down) so analyzers using different capacity units (tokens, tok/s, …) can be compared safely.
- Adds `RegisterAnalyzer(name, interfaces.Analyzer)` so callers (e.g. `main.go`) can inject plugin analyzers without engine.go knowing the concrete type.
- Saturation always executes regardless of its `enabled` flag — its `VariantCapacities` carry `Cost` and `AcceleratorName` required by the optimizer.

## Changed files

| File | Change |
|------|--------|
| `internal/engines/saturation/engine.go` | Add `analyzers` map to `Engine` struct; seed with saturation in `NewEngine`; add `RegisterAnalyzer` |
| `internal/engines/saturation/engine_v2.go` | Add `combineAnalyzerResults`, `enabledAnalyzerResult`, `sumTotalCapacity`; rewrite `runAnalyzersAndScore` |
| `internal/engines/saturation/engine_combine_test.go` | New: 31 Ginkgo specs covering any-up, all-down, cold-start, score, saturation-disabled cases |
| `docs/saturation-scaling-config.md` | Multi-Analyzer Pipeline section: config schema, combine algorithm, `AnalyzerScoreConfig` reference |
| `docs/user-guide/saturation-analyzer.md` | Cross-reference to multi-analyzer pipeline docs |

## Test plan

- [x] `go test ./internal/engines/saturation/...` — all existing tests pass, 31 new specs in `engine_combine_test.go` pass
- [x] `go test ./internal/... ./pkg/... ./cmd/...` — full unit suite green
- [x] `go build ./...` — clean build

### DIFF
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -226,6 +226,12 @@ func NewEngine(client client.Client, scheme *runtime.Scheme, recorder record.Eve
 	return &engine
 }
 
+// RegisterAnalyzer adds an external analyzer implementation to the engine's
+// analyzer registry. Must be called before StartOptimizeLoop.
+func (e *Engine) RegisterAnalyzer(name string, a interfaces.Analyzer) {
+	e.analyzers[name] = a
+}
+
 // StartOptimizeLoop starts the optimization loop for the saturation engine.
 // It runs until the context is cancelled.
 func (e *Engine) StartOptimizeLoop(ctx context.Context) {
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -126,6 +126,7 @@ type Engine struct {
 	metricsRegistry *source.SourceRegistry
 
 	// saturationV2Analyzer is the V2 token-based saturation analyzer (initialized once).
+	// Also registered in analyzers under interfaces.SaturationAnalyzerName.
 	saturationV2Analyzer *saturation_v2.SaturationAnalyzer
 
 	// queueingModelAnalyzer is the queueing model-based analyzer (initialized once).
@@ -134,6 +135,11 @@ type Engine struct {
 
 	// capacityStore is shared with the V2 analyzer for caching capacity knowledge.
 	capacityStore *saturation_v2.CapacityKnowledgeStore
+
+	// analyzers is the generic name-to-implementation registry used by
+	// runAnalyzersAndScore. New analyzers are registered here; the engine loop
+	// calls each enabled analyzer by name without knowing its concrete type.
+	analyzers map[string]interfaces.Analyzer
 
 	// optimizer is the V2 scaling optimizer that produces VariantDecisions from
 	// AnalyzerResults. Selected per-cycle based on enableLimiter config:
@@ -167,6 +173,7 @@ func NewEngine(client client.Client, scheme *runtime.Scheme, recorder record.Eve
 	gpuLimiter := pipeline.NewDefaultLimiter("gpu-limiter", gpuInventory, gpuAlgorithm)
 
 	capacityStore := saturation_v2.NewCapacityKnowledgeStore()
+	satV2 := saturation_v2.NewSaturationAnalyzer(capacityStore)
 
 	// Initialize with default optimizer. The actual optimizer is selected
 	// per-cycle in optimize() based on dynamic config (enableLimiter flag
@@ -182,11 +189,14 @@ func NewEngine(client client.Client, scheme *runtime.Scheme, recorder record.Eve
 		ScaleToZeroEnforcer:     pipeline.NewEnforcer(requestCountFunc),
 		GPULimiter:              gpuLimiter,
 		metricsRegistry:         metricsRegistry,
-		saturationV2Analyzer:    saturation_v2.NewSaturationAnalyzer(capacityStore),
+		saturationV2Analyzer:    satV2,
 		queueingModelAnalyzer:   queueingmodel.NewQueueingModelAnalyzer(),
 		capacityStore:           capacityStore,
 		optimizer:               scalingOptimizer,
 		v1AnalyzerFactory:       defaultV1AnalyzerFactory,
+		analyzers: map[string]interfaces.Analyzer{
+			interfaces.SaturationAnalyzerName: satV2,
+		},
 	}
 
 	engine.executor = executor.NewPollingExecutor(executor.PollingConfig{
--- a/internal/engines/saturation/engine_combine_test.go
+++ b/internal/engines/saturation/engine_combine_test.go
@@ -1,0 +1,229 @@
+package saturation
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/interfaces"
+)
+
+// makeResult builds a minimal AnalyzerResult for combine tests.
+// vcs is a list of (replicaCount, perReplicaCapacity) pairs; RC and SC are model-level signals.
+func makeResult(RC, SC float64, vcs ...float64) *interfaces.AnalyzerResult {
+	var caps []interfaces.VariantCapacity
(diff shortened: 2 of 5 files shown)


internal/engines/saturation/engine_v2.go
```diff
@@ -73,8 +74,123 @@ func (e *Engine) runV2AnalysisOnly(
...
+			validCount++
+			excessFracs = append(excessFracs, er.result.RequiredCapacity/t)
+			slackFracs = append(slackFracs, er.result.SpareCapacity/t)
+		} else if er.result.RequiredCapacity > 0 {
+			// Cold start: demand detected but no replicas running yet.
+			scaleUpColdStart = true
+			validCount++
+		}
+		// t == 0 and RC == 0: no signal from this analyzer; skip.
+		totalWeighted += er.result.RequiredCapacity * er.score
```
ev-shindin, line 140: I think here you should normalize RequiredCapacity w.r.t. to the TotalCapacity of the analyzer as well

internal/engines/saturation/engine_v2.go
```diff
@@ -84,9 +200,9 @@ func (e *Engine) runAnalyzersAndScore(
 	scaleTargets map[string]scaletarget.ScaleTargetAccessor,
 	variantAutoscalings map[string]*llmdVariantAutoscalingV1alpha1.VariantAutoscaling,
 ) (*interfaces.AnalyzerResult, error) {
-	// Resolve per-analyzer threshold overrides before running the analyzer.
-	// The saturation analyzer reads thresholds from the config, so we apply
-	// per-analyzer overrides to the config's top-level fields.
+	logger := ctrl.LoggerFrom(ctx)
+
+	// Apply per-analyzer saturation threshold overrides before running the analyzer.
 	for _, aw := range config.Analyzers {
```
ev-shindin, line 206: `AnalyzerScoreConfig` exposes `ScaleUpThreshold` and `ScaleDownBoundary` for **all** analyzers, `Validate()` checks the pair-relationship for **all** analyzers (`saturation_scaling.go:262-266`), and the doc table says these fields "override global" without qualifier — but the engine only applies them to saturation.

For other analyzers (throughput, SLO, …) the per-entry threshold is silently dropped.

internal/engines/saturation/engine.go
```diff
@@ -216,6 +226,12 @@ func NewEngine(client client.Client, scheme *runtime.Scheme, recorder record.Eve
 	return &engine
 }
 
+// RegisterAnalyzer adds an external analyzer implementation to the engine's
+// analyzer registry. Must be called before StartOptimizeLoop.
+func (e *Engine) RegisterAnalyzer(name string, a interfaces.Analyzer) {
```
ev-shindin, line 231: The map is written here and read concurrently in `runAnalyzersAndScore` (`engine_v2.go:230-234`). The doc comment states a contract ("before StartOptimizeLoop"), but nothing enforces it — a misbehaving caller registering at runtime triggers a data race that won't reliably crash, just silently corrupt the map.

ev-shindin (changes requested):
Hi @deanlorenz . Here the major issues. There also minor ones I will add later.

internal/engines/saturation/engine_v2.go
```diff
@@ -73,8 +74,123 @@ func (e *Engine) runV2AnalysisOnly(
...
+			if sl < minSlack {
+				minSlack = sl
+			}
+		}
+	}
+
+	switch {
+	case scaleUpColdStart:
+		// satTotal == 0: denormalisation is not possible.
+		// Forward saturation's cold-start RC directly; fall back to 1.0 so the
```
ev-shindin, line 170: The comment ("fall back to 1.0 so the optimizer adds at least one replica") relies on the optimizer's `costAwareScaleUp` ceiling: any positive RC → ≥1 replica. That's true today but couples the cold-start contract to the optimizer's rounding behavior. If the optimizer ever changes (e.g., dividing RC by PerReplicaCapacity then flooring), `1.0` would round to zero.
Either:
- Use a named constant `coldStartReplicaSentinel = 1.0` with a comment that documents *exactly* what it means semantically.
- Compute a unit-correct sentinel: `math.Max(1.0, minPerReplicaCapacity)`

internal/engines/saturation/engine_v2.go
```diff
@@ -73,8 +74,123 @@ func (e *Engine) runV2AnalysisOnly(
...
+		combined.RequiredCapacity = maxExcess * satTotal
+		if allDown && minSlack < math.MaxFloat64 {
+			combined.SpareCapacity = minSlack * satTotal
+		} else {
+			combined.SpareCapacity = 0
+		}
+	}
+
+	combined.Score = priority * totalWeighted
+	return &combined
```
ev-shindin, line 188: `combineAnalyzerResults` makes the load-bearing scaling decision but emits nothing. With multiple analyzers, operators will need to know *which* analyzer triggered scale-up or *which* blocked scale-down. Suggest a single debug log at the end.

internal/engines/saturation/engine_v2.go
```diff
@@ -73,8 +74,123 @@ func (e *Engine) runV2AnalysisOnly(
...
+//
+// Any-up:  scale up when any enabled analyzer has utilExcess > 0.
+// All-down: scale down only when every analyzer with valid data has utilSlack > 0.
+// Results are denormalised back into saturation's capacity units for the optimizer.
+func combineAnalyzerResults(
+	satResult *interfaces.AnalyzerResult,
+	results []enabledAnalyzerResult,
+	priority float64,
+) *interfaces.AnalyzerResult {
+	combined := *satResult // copy; VariantCapacities from saturation are always the base
```
ev-shindin, line 111: `combined.VariantCapacities` is now the same backing slice as `satResult.VariantCapacities`; `combined.RoleCapacities` is the same map. The function never mutates them, so this is safe — but a future maintainer mutating `combined.VariantCapacities[i]` would silently corrupt the saturation result. Either:

- Add a `// shallow copy; do not mutate VariantCapacities/RoleCapacities` comment.
- Deep-copy on entry (cheap — small slices).

internal/engines/saturation/engine_v2.go
```diff
@@ -73,8 +74,123 @@ func (e *Engine) runV2AnalysisOnly(
 	return result, nil
 }
 
-// runAnalyzersAndScore runs the V2 saturation analyzer, then computes the
-// weighted composite score from enabled analyzers and model priority.
+// enabledAnalyzerResult pairs an enabled analyzer's result with its configured score weight.
+type enabledAnalyzerResult struct {
+	result *interfaces.AnalyzerResult
+	score  float64
```
ev-shindin, line 80: `AnalyzerScoreConfig.Score` is named "Score" in the YAML (matches the user-facing API), but inside Go it acts as a weight in the urgency-score computation. The field `score` here ends up next to `AnalyzerResult.Score` (composite urgency), which is genuinely confusing. Rename the field `weight`. I would also rename it in the config file.

internal/engines/saturation/engine_combine_test.go
ev-shindin, line 1: No test for per-analyzer threshold override application

ev-shindin (changes requested):
There are twofollow-up issues I think we need to open:
1. `Validate()` (`internal/config/saturation_scaling.go`)  doesn't reject duplicate analyzer names.
2. Queueing-model analyzer is not registered in `e.analyzers`

deanlorenz:
Closed. 
Reimplemented as #1225, #1228, #1246, #1266

Issue #920, opened 2026-03-23
Operational Grafana Dashboard


ev-shindin:
Add an operational Grafana dashboard for day-to-day WVA monitoring.

**Panels:**

1. **Replica Overview** — desired vs current replicas per variant
2. **Saturation Utilization** — `wva_saturation_utilization` trend per variant
3. **Capacity Breakdown** — `wva_kv_cache_tokens_used` vs `wva_kv_cache_tokens_total`
4. **Scaling Decision Rate** — `rate(wva_replica_scaling_total)` by reason (stacked bar)
5. **Error Rate** — `rate(wva_errors_total)` by component
6. **Optimization Loop Duration** — histogram quantiles of `wva_optimization_duration_seconds`
7. **GPU Availability** — `wva_available_gpus` by accelerator type
8. **Limiter Impact** — `rate(wva_decisions_limited_total)` over time
9. **Configuration State** —  optimization interval only (thresholds & active analyzer deferred to follow-up). This has been descoped to show only optimization interval.

**Implementation:**
- Add `deploy/grafana/wva-operational-dashboard.json`
- Ship as optional Grafana ConfigMap in Helm chart
- Use hardcoded `"type": "prometheus"` datasource (not template variables — they don't resolve in file-provisioned dashboards)

**Acceptance Criteria:**
- [ ] Dashboard JSON deployable via `deploy/grafana/` or Helm ConfigMap
- [ ] All 9 panels render correctly with populated metrics
- [ ] Dashboard documented in user guide

shuynh2017:
working on this ...

PR #1114, opened 2026-05-10
feat: Observability - Operational Grafana Dashboard

shuynh2017:
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/920

Add an operational Grafana dashboard for day-to-day WVA monitoring.

## Data Source
image

## Dashboard List
image

## Namespace Label Filter
image

## Namespace Filter
image

## Variant Filter
image

## Dashboard Panels
image

### WVA Errors Panel
**Description**: number of errors in the last hour
image

### Metric Collection Errors Panel
**Description**: number of errors in the last hour
image

### Metrics Collection Duration Panel
**Description**:  Duration of metrics collection operations by query type in seconds
image

### Saturation Utilization Panel
**Description**: Per-variant utilization ratio (0.0-1.0) from saturation analysis. V1 path: mean of per-replica KV-cache-usage fractions (matches the per-replica threshold V1 checks). V2 path: TotalDemand / TotalCapacity from the analyzer result. Numerically equivalent for uniform-capacity replicas; V2 is capacity-weighted for mixed-capacity cases.
image

### Capacity Breakdown Panel
**Description**: KV cache token usage vs capacity. wva_kv_cache_tokens_used: Total KV cache tokens currently in use across all replicas of a variant (sum of vLLM TokensInUse). wva_kv_cache_tokens_capacity: Total KV cache token capacity across all replicas of a variant (sum of vLLM TotalKvCapacityTokens).
image

### Scaling Decision Rate
**Description**:  Rate of scaling decisions by reason. Metric is wva_replica_scaling_total.
image

### Replica Overview Panel
**Description**: Replica count comparison. wva_current_replicas: Current number of replicas for each variant. wva_desired_replicas: Desired number of replicas for each variant.
image

### Limiter Impact Panel
**Description**: "Average number of scaling decisions limited by constraints per optimization cycle (30 seconds, averaged over 1 hour). Shows how often limiters prevent scaling due to insufficient capacity. For example, if there were 36 limited decisions total in the last hour, and there are 120 optimization cycles per hour (3600 seconds ÷ 30 seconds), then the average is 36 ÷ 120 = 0.3 decisions per cycle.",

image

### Optimization Duration Panel
**Description**: Duration of optimization loop cycles in seconds

image

## Requirements
**Panels:**

- [x] **Replica Overview** — desired vs current replicas per variant
- [x] **Saturation Utilization** — `wva_saturation_utilization` trend per variant
- [x] **Capacity Breakdown** — `wva_kv_cache_tokens_used` vs `wva_kv_cache_tokens_total`
- [x] **Scaling Decision Rate** — `rate(wva_replica_scaling_total)` by reason (stacked bar)
  ~~- This metric is in the code but not recorded anywhere.~~
- [x] **Error Rate** — `rate(wva_errors_total)` by component
- [x] **Optimization Loop Duration** — histogram quantiles of `wva_optimization_duration_seconds`
- [x] **GPU Availability** — `wva_available_gpus` by accelerator type
- [x] **Limiter Impact** — `rate(wva_decisions_limited_total)` over time
- [x] **Configuration State** — current thresholds and active analyzer
  - Since the configuration can be for VA specific. The suggestion is to only show the optimization interval. This has been descoped to show only optimization interval.

**Implementation:**
- [x] Add `deploy/grafana/wva-operational-dashboard.json` // `operational-dashboard.json` to be consistent with `benchmark-dashboard.json`.
- [x] Ship as optional Grafana ConfigMap in Helm chart // only install if `DEPLOY_OPERATIONAL_DASHBOARD=true`
- [x] Use hardcoded `"type": "prometheus"` datasource (not template variables — they don't resolve in file-provisioned dashboards)

**Documentation**
- [x] Guide for importing the operational dashboard
- [x] Troubleshooting section: "no data" scenarios, namespace label behavior, ServiceMonitor configuration.
  - [x] "No data" 

**Implementation:**
- [x] Add `docs/user-guide/monitoring.md`. 
  ~~- Temporarily store it under `docs/developer-guide/monitoring.md` for now. It's not clear where to store it under `llm-d`. ~~
- Reference from existing user guide documentation

**Acceptance Criteria:**
- [x] Dashboard JSON deployable via `deploy/grafana/` or Helm ConfigMap
- [x] All 9 panels render correctly with populated metrics
- [x] Dashboard documented in user guide https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1132
- [x] E2E Test coverage
  - [x] Grafana is installed and accessible with valid credentials.
  - [x] Prometheus data source exists, configured with appropriate TLS settings, and is accessible by Grafana.
  - [x] Dashboard called 'WVA Operational' exists.
  - [x] Panels to exist.


shuynh2017:
@ev-shindin @lionelvillard I showed the dashboard to Greg, and am tracking additional ideas for metrics/observability so there may be more work after this initial implementation.

shuynh2017:
@ev-shindin @lionelvillard @mamy-CS, pls review. Thanks.

shuynh2017:
@gyliu513, can you pls help to review/comment? Thanks.

deploy/kind-emulator/README.md
```diff
@@ -93,6 +94,7 @@ make create-kind-cluster KIND_ARGS="-t mix -n 4 -g 2"
 ```bash
 cd /path/to/repo
 export ENVIRONMENT=kind-emulator
+export DEPLOY_OPERATIONAL_DASHBOARD=false  # Optional: deploy Grafana dashboard
```
gyliu513: Should this be false or true? I can see this section is `Deploy WVA + monitoring only (no llm-d):`, here should be true?

deploy/kind-emulator/README.md
```diff
@@ -68,9 +68,10 @@ export KIND_IMAGE_PLATFORM=linux/amd64      # Single platform for kind load (avo
 **Deployment flags**:
 
 ```bash
-export DEPLOY_PROMETHEUS=true         # Deploy Prometheus stack
-export DEPLOY_WVA=true                # Deploy WVA controller
-export DEPLOY_PROMETHEUS_ADAPTER=true # Deploy Prometheus Adapter
+export DEPLOY_PROMETHEUS=true               # Deploy Prometheus stack
+export DEPLOY_OPERATIONAL_DASHBOARD=false   # Deploy Grafana and operational dashboard
```
gyliu513: suggest `true` here, as this is telling end user this is a flag, and setting to true can deploy but not false. This also align with other flags.

deploy/kubernetes/README.md
```diff
@@ -110,9 +110,10 @@ export ACCELERATOR_TYPE="A100"              # GPU type (auto-detected)
 **Deployment flags** (`deploy/install.sh`) — VA/HPA are managed separately; llm-d is deployed via `deploy/install-epp.sh` or the [llm-d guides](https://github.com/llm-d/llm-d/tree/main/guides/optimized-baseline):
 
 ```bash
-export DEPLOY_PROMETHEUS=true         # Deploy kube-prometheus-stack
-export DEPLOY_WVA=true                # Deploy WVA controller
-export DEPLOY_PROMETHEUS_ADAPTER=true # Deploy Prometheus Adapter
+export DEPLOY_PROMETHEUS=true               # Deploy kube-prometheus-stack
+export DEPLOY_OPERATIONAL_DASHBOARD=false   # Deploy kube-prometheus-stack-grafana
```
gyliu513:
```suggestion
export DEPLOY_OPERATIONAL_DASHBOARD=true   # Deploy kube-prometheus-stack-grafana
```

deploy/kubernetes/README.md
```diff
@@ -149,6 +150,7 @@ make deploy-wva-on-k8s   # install.sh (WVA + monitoring + scaler + LWS)
 ```bash
 export DEPLOY_WVA=true
 export DEPLOY_PROMETHEUS=true
+export DEPLOY_OPERATIONAL_DASHBOARD=false  # Grafana and operational dashboard
```
gyliu513: The default value of `DEPLOY_OPERATIONAL_DASHBOARD` is false, so no need to add it here if we do not want to enable it.

How about add a new example to enable DEPLOY_OPERATIONAL_DASHBOARD?

deploy/README.md
```diff
@@ -200,8 +200,9 @@ export WVA_IMAGE_REPO="ghcr.io/llm-d/llm-d-workload-variant-autoscaler"
 export WVA_IMAGE_TAG="latest"
 
 # Optional
-export DEPLOY_WVA=false            # Monitoring + scaler only
+export DEPLOY_WVA=false                     # Monitoring + scaler only
 export DEPLOY_PROMETHEUS=false
+export DEPLOY_OPERATIONAL_DASHBOARD=false   # Deploy Grafana and operational dashboard
```
gyliu513: How about set it as `true` here as the comment is `Deploy Grafana and operational dashboard`

deploy/README.md
```diff
@@ -230,6 +231,7 @@ make deploy-e2e-infra ENVIRONMENT=kind-emulator IMG=localhost/llm-d-workload-var
 ```bash
 export DEPLOY_WVA=true
 export DEPLOY_PROMETHEUS=true
+export DEPLOY_OPERATIONAL_DASHBOARD=false
```
gyliu513: `true` here? As this example is `WVA + monitoring only (no llm-d)`

docs/developer-guide/monitoring.md
```diff
@@ -0,0 +1,96 @@
...
+export DEPLOY_OPERATIONAL_DASHBOARD=true
+make deploy-wva-on-k8s
+```
+
+## Access Operational Dashboard
+Once the operational dashboard is enabled, Grafana is installed, configured, and ready to display the dashboard. Here are the next steps:
+
+- Forward Grafana port so the dashboard can be accessed locally:
+  ```
+  $ kubectl port-forward -n workload-variant-autoscaler-monitoring svc/kube-prometheus-stack-grafana 3000:80 &
```
gyliu513: for developer guide, suggest remove `&` to print out logs in console

docs/developer-guide/monitoring.md
```diff
@@ -0,0 +1,96 @@
...
+  ```
+- Get Grafana `admin` password:
+  ```
+  $ kubectl get secret -n workload-variant-autoscaler-monitoring   kube-prometheus-stack-grafana   -o jsonpath="{.data.admin-password}" | base64 -d;echo
+  Z9FEW12xG2k2tTZJVML75Kd80qi2oI0nJBsCjv7q
```
gyliu513: this is harmless, but suggest remove the output here.

docs/user-guide/monitoring.md
```diff
@@ -0,0 +1,96 @@
...
+- Forward Grafana port so the dashboard can be accessed locally:
+  ```
+  $ kubectl port-forward -n workload-variant-autoscaler-monitoring svc/kube-prometheus-stack-grafana 3000:80 &
+  ```
+- Get Grafana `admin` password:
+  ```
+  $ kubectl get secret -n workload-variant-autoscaler-monitoring   kube-prometheus-stack-grafana   -o jsonpath="{.data.admin-password}" | base64 -d;echo
+  Z9FEW12xG2k2tTZJVML75Kd80qi2oI0nJBsCjv7q
+  ```
+- Point browser to `http://localhost:3000/`, login with username `admin` and the password obtained in previous step.
```
gyliu513, line 24: I think grafana default password is `password`, did you change password for grafana?

docs/developer-guide/monitoring.md
```diff
@@ -0,0 +1,96 @@
...
+
+
+## Troubleshooting
+### services "kube-prometheus-stack-grafana" not found
+  - Make sure to install WVA cluster with `DEPLOY_OPERATIONAL_DASHBOARD=true` as this would install Grafana.
+  
+### No Data
+  - Check the datasource by browse to "Connections/Data sources", you should see a Prometheus data source `https://kube-prometheus-stack-prometheus.workload-variant-autoscaler-monitoring.svc.cluster.local:9090`. Click on `Test` button to test the data source.
+  - Check data source which is defined in `kube-prometheus-stack-grafana-datasource` configmap:
+    ```
```
gyliu513: if you want to include the output, suggest adding `console` to annotation. Ditto for others

```console
$ oc get cm kube-prometheus-stack-grafana-datasource -n workload-
    variant-autoscaler-monitoring -o yaml
    apiVersion: v1
    data:
      datasource.yaml: |-
        apiVersion: 1
        datasources:
        - name: "Prometheus"
          type: prometheus
          uid: prometheus
          url: http://kube-prometheus-stack-prometheus.workload-variant-autoscaler-monitoring:9090/
          access: proxy
          isDefault: true
          jsonData:
            httpMethod: POST
            timeInterval: 30s
        - name: "Alertmanager"
          type: alertmanager
          uid: alertmanager
          url: http://kube-prometheus-stack-alertmanager.workload-variant-autoscaler-monitoring:9093/
          access: proxy
    ...
```

test/e2e/operational_dashboard_test.go
```diff
@@ -0,0 +1,1091 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "grafana-curl-test",
+					Namespace: cfg.MonitoringNS,
+				},
+				Spec: corev1.PodSpec{
+					RestartPolicy: corev1.RestartPolicyNever,
+					Containers: []corev1.Container{
+						{
+							Name:    "curl",
+							Image:   "curlimages/curl:latest",
```
gyliu513: `curlimages/curl:latest` is a bare docker.io image. AGENTS.md explicitly forbids docker.io images in e2e tests. Must use a fully-qualified registry, e.g. `quay.io/curl/curl:latest` or pin a digest from a trusted registry.

ditto for others.

gyliu513 (commented):
Thanks @shuynh2017 !

deploy/kind-emulator/README.md
shuynh2017: agree to set to true to install by default

deploy/kind-emulator/README.md
shuynh2017: agree to set to true to install by default

deploy/kubernetes/README.md
shuynh2017: agree to set to true to install by default

deploy/kubernetes/README.md
shuynh2017: agree to set to true to install by default

deploy/README.md
shuynh2017: agree to set to true to install by default

deploy/README.md
shuynh2017: agree to set to true to install by default

docs/developer-guide/monitoring.md
shuynh2017: removed `&`

docs/developer-guide/monitoring.md
shuynh2017: removed output

docs/user-guide/monitoring.md
shuynh2017: Standalone Grafana installation defaults its password to admin, this specific Prometheus community bundle overrides it to protect unconfigured stacks (I didn't have to do any)

docs/developer-guide/monitoring.md
shuynh2017: added. Thank you.

test/e2e/operational_dashboard_test.go
shuynh2017: updated. Thanks.

shuynh2017:
@gyliu513, thanks for taking time to review. Good to hear you are now the point person for observability. Pls review again.

deploy/grafana/operational-dashboard.json
```diff
@@ -0,0 +1,1517 @@
...
+      },
+      "pluginVersion": "13.0.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "expr": "max by (namespace, variant_name, accelerator_type) (wva_saturation_utilization)",
```
ev-shindin: With `honorLabels:false` (the default here), the variant's namespace lands in `exported_namespace`; `namespace` is the controller's pod namespace. In general I suggest to create $namespace variable, so that it can be used with either value of `honorLabels:

deploy/grafana/operational-dashboard.json
```diff
@@ -0,0 +1,1517 @@
...
+      },
+      "pluginVersion": "13.0.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "expr": "max by (namespace, variant_name) (wva_kv_cache_tokens_used)",
```
ev-shindin: same as line 927

deploy/grafana/operational-dashboard.json
```diff
@@ -0,0 +1,1517 @@
...
+          "range": true,
+          "refId": "A"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "expr": "max by (namespace, variant_name) (wva_kv_cache_tokens_capacity)",
```
ev-shindin: same as line 927

deploy/grafana/operational-dashboard.json
```diff
@@ -0,0 +1,1517 @@
...
+          "range": true,
+          "refId": "A"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "expr": "max by (namespace, variant_name) (wva_desired_replicas)",
```
ev-shindin: same as line 927

deploy/grafana/operational-dashboard.json
```diff
@@ -0,0 +1,1517 @@
...
+      },
+      "pluginVersion": "13.0.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "expr": "max by (namespace, variant_name) (wva_current_replicas)",
```
ev-shindin: same as line 927

deploy/grafana/operational-dashboard.json
```diff
@@ -0,0 +1,1517 @@
...
+      },
+      "pluginVersion": "13.0.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "expr": "sum(rate(wva_decisions_limited_total[1h])) by (namespace, variant_name, limiter_name) * 30",
```
ev-shindin: same as line 927,
Also * 30` hardcodes a 30s optimization interval while the dashboard itself displays the *configurable* interval. Multiply by the actual interval instead

deploy/grafana/operational-dashboard.json
```diff
@@ -0,0 +1,1517 @@
...
+      },
+      "pluginVersion": "13.0.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "editorMode": "code",
+          "expr": "floor(sum(increase(wva_errors_total{component=\"collector\"}[1h])) or vector(0))",
```
ev-shindin: These panels are 6 and 4 separate queries, one per `component="…"` / `query_type="…"`. A new component/query_type silently won't appear. Consider a single auto-discovering query, e.g.:
`floor(sum by (component) (increase(wva_errors_total[1h])) or vector(0))`
Keeps the `or vector(0)` no-data handling, future-proofs new components, and shrinks the panel. (Non-blocking.)

deploy/grafana/operational-dashboard.json
```diff
@@ -0,0 +1,1517 @@
...
+    }
+  ],
+  "refresh": "",
+  "schemaVersion": 39,
+  "tags": [
+    "wva",
+    "operational"
+  ],
+  "templating": {
+    "list": []
```
ev-shindin: Add template variables 
Not a one-click suggestion because the panels also need matching filters, but this is the shape:
```json
    "list": [
      { "name": "namespace", "type": "query", "label": "Namespace",
        "datasource": { "type": "prometheus", "uid": "prometheus" },
        "query": { "query": "label_values(wva_saturation_utilization, exported_namespace)", "refId": "ns" },
        "refresh": 2, "includeAll": true, "multi": true },
      { "name": "variant", "type": "query", "label": "Variant",
        "datasource": { "type": "prometheus", "uid": "prometheus" },
        "query": { "query": "label_values(wva_saturation_utilization{exported_namespace=~\"$namespace\"}, variant_name)", "refId": "var" },
        "refresh": 2, "includeAll": true, "multi": true }
    ]
```
Then add `{exported_namespace=~"$namespace", variant_name=~"$variant"}` to the per-variant panel queries so the dropdowns actually filter.

docs/user-guide/monitoring.md
ev-shindin, line 1: This is an operator/user guide (importing a dashboard, troubleshooting), so it belongs under `docs/user-guide/` rather than `docs/developer-guide/`. (You flagged the uncertainty in the PR body — worth resolving before merge.)

docs/user-guide/monitoring.md
```diff
@@ -0,0 +1,95 @@
...
+- Copy and paste the content of `deploy/grafana/operational-dashboard.json`.
+- Name the new dashboard such as `My WVA Operational Dashboard`.
+- Your new dashboard now is the same as `WVA Operational Dashboard` except that you can edit and save.
+
+
+## Troubleshooting
+### services "kube-prometheus-stack-grafana" not found
+  - Make sure to install WVA cluster with `DEPLOY_OPERATIONAL_DASHBOARD=true` as this would install Grafana.
+  
+### No Data
```
ev-shindin, line 75: The PR description says this section covers "namespace label behavior," but it only covers the datasource. Please add the `namespace` vs `exported_namespace` explanation (the exact failure mode users hit): with the default `honorLabels:false`, WVA's `namespace` label is exposed as `exported_namespace`; the dashboard's per-variant panels use `exported_namespace` accordingly, and `namespace` shows the controller's own namespace.

ev-shindin (changes requested):
Thanks — this is a genuinely useful operational dashboard and the e2e coverage (Grafana install, datasource health, dashboard + panel existence) is excellent. Two correctness/UX issues and two in-scope acceptance gaps I'd want resolved before merge; the rest is polish.

1. **Namespace label is inconsistent across panels (most show the wrong namespace).** "Pods with Metrics" groups by `exported_namespace`, but 6 per-variant panels group by `namespace`. Both metrics set a code-side `namespace` label, and the WVA ServiceMonitor doesn't set `honorLabels`, so Prometheus renames the metric's `namespace` → `exported_namespace` and sets `namespace` = the *WVA controller's* pod namespace. So the 6 `namespace` panels group every variant under `workload-variant-autoscaler-system` instead of the model's namespace. **`exported_namespace` is the correct one** (matches the working Pods panel) — inline suggestions fix the 6. *(Alternative: set `honorLabels: true` on the ServiceMonitor and flip the Pods panel to `namespace` instead — but pick one; the current mix is wrong under any single scrape config.)*
2. **No template variables → unusable at scale.** `"templating": { "list": [] }` means every per-variant panel renders all series for all variants/namespaces at once. An operational dashboard needs `$namespace`/`$variant` dropdowns. Snippet + follow-up below.
3. **Two acceptance items of #920 are under-delivered — complete or descope (these are in scope of *this* PR, not a follow-up):**
   - **Scaling Decision Rate** — #920 lists `rate(wva_replica_scaling_total) by reason` as a required panel; it's dropped because `wva_replica_scaling_total` is defined but never recorded. Wire up the metric (one `.Inc()` at the decision-apply site) and add the panel, **or** descope it with the issue owner — it's currently an unchecked `[ ]`.
   - **Configuration State** — #920 asks for "current thresholds **and active analyzer**," but the panel ships only `optimization_interval`. The threshold metrics already exist (`wva_config_kv_spare_threshold`, `wva_config_queue_spare_threshold`) as does `wva_optimizer_active`/`wva_config_info`. The PR's "config can be VA-specific, so only show the interval" note is a reasonable basis for narrowing — but it needs explicit sign-off rather than a silent half-`[x]`.

**Scope note (please don't expand this PR):** in reviewing metric coverage I noted several *operationally valuable* metrics WVA already exports but that are **not in #920** — the scaling decision-drivers (`required_capacity`, `spare_capacity`), `enforcer_modifications_total`, `desired_ratio`, an optimization-loop stall/liveness panel, and controller-runtime health. These are genuine enhancements but out of scope here; I'll open a separate follow-up issue ("operational dashboard v2 — decision-driver & controller-health panels") so they don't bloat #1114. Keeping this PR scoped to #920.

Nice touches: `or vector(0)` on the error stat panels (shows 0, not "No data"), optional install via `DEPLOY_OPERATIONAL_DASHBOARD=true`, datasource `uid:"prometheus"` matching the provisioned datasource, and gofmt-clean e2e with no `docker.io` images.

shuynh2017:
>    * **Configuration State** — #920 asks for "current thresholds **and active analyzer**," but the panel ships only `optimization_interval`. The threshold metrics already exist (`wva_config_kv_spare_threshold`, `wva_config_queue_spare_threshold`) as does `wva_optimizer_active`/`wva_config_info`. The PR's "config can be VA-specific, so only show the interval" note is a reasonable basis for narrowing — but it needs explicit sign-off rather than a silent half-`[x]`.
> 
@ev-shindin, pls sign-off this narrowing. Thanks.

deploy/grafana/operational-dashboard.json
shuynh2017: fixed

deploy/grafana/operational-dashboard.json
shuynh2017: fixed

deploy/grafana/operational-dashboard.json
shuynh2017: fixed

deploy/grafana/operational-dashboard.json
shuynh2017: fixed

deploy/grafana/operational-dashboard.json
shuynh2017: fixed

deploy/grafana/operational-dashboard.json
shuynh2017: fixed

deploy/grafana/operational-dashboard.json
shuynh2017: fixed

deploy/grafana/operational-dashboard.json
shuynh2017: done

docs/user-guide/monitoring.md
shuynh2017: done

docs/user-guide/monitoring.md
shuynh2017: done

shuynh2017:
@ev-shindin thank you very much for your insights on the dashboard. Your suggestions on filters, showing components flexibly, and so on, really make the dashboard useful. It took a bit longer for me to make the changes since I'd like to see all the data showing after adding filters. Pls review again.

internal/engines/saturation/engine.go
```diff
@@ -1491,6 +1491,9 @@ func (e *Engine) applySaturationDecisions(
 		}
 
 		if hasDecision {
+			if err := e.metricsEmitter.EmitReplicaScalingMetrics(ctx, &updateVa, string(decision.Action), reason); err != nil {
```
ev-shindin: `decision.Action` is the bounded enum `scale-up` / `scale-down` / **`no-change`** (`interfaces/saturation_analyzer.go:322-324`), and this fires on every `hasDecision` cycle. A steady-state variant therefore increments `wva_replica_scaling_total{direction="no-change"}` **every `OptimizationInterval`**, so the "Scaling Decision Rate" panel is dominated by no-ops and the `direction` label no longer matches its documented `up/down` semantics (`constants/metrics.go`). Guard to actual scaling actions:

Also `reason` is passed straight into the `reason` **counter label** (`EmitReplicaScalingMetrics` → `LabelReason`). If `decision.Reason` can carry dynamic text (thresholds, counts, variant names), this is a Prometheus cardinality blow-up — one series per unique string — and the panel's `by (reason)` becomes unusable. Please confirm `decision.Reason` is a fixed set of short reason codes; if it can be free-form, map it to a bounded code for the label and keep the verbose text for the log/event only.

ev-shindin (changes requested):
Big improvement — thanks for working through the earlier round. Verified addressed: per-variant panels now group/filter on `exported_namespace` (+ the `$namespace_label` variable), template variables (`$namespace`/`$variant`) added, Limiter Impact uses `* max(wva_config_optimization_interval_seconds)`, error panels auto-discover via `sum by (component)`/`by (query_type)`, the doc moved to `docs/user-guide/`, and the e2e image is now `quay.io/curl/curl:latest`.

One blocking issue and one to confirm, both in the new metric emission (`internal/engines/saturation/engine.go`); inline. Non-blocking: (a) the `$namespace_label` variable changes panel *grouping* (`by ($namespace_label)`) but the selectors are hardcoded to `exported_namespace=~"$namespace"` — if an operator flips it to `namespace`, grouping and filter disagree; either interpolate the variable in the selector too or document that it only affects grouping. (b) Configuration State ships interval-only — happy to sign off on narrowing it to `optimization_interval`, provided the #920 checklist is updated to reflect the descope rather than left as "delivered."

internal/engines/saturation/engine.go
shuynh2017: Thanks again for the review:
- Fixed the `exported_namespace=~"$namespace"` issue.
- Updated https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/920 checklist is updated to reflect the descope of  Configuration State.
- Added guard for `no-change` decision.
- Mapped `decision.Reason` which can be unbounded to a set of bounded values. (Will remember to watch out for this when working with Prometheus metrics)

ev-shindin (dismissed):
Thanks @shuynh2017 ! `sanitizeReason` looks a bit fragile. Please:
1.Add TODO comment to eefactor code with better solution
2. Open follo-to issue to address this later

shuynh2017:
> Thanks @shuynh2017 ! `sanitizeReason` looks a bit fragile. Please: 1.Add TODO comment to eefactor code with better solution 2. Open follo-to issue to address this later

1- TODO added
2- Follow up issue: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1262

ev-shindin (approved):
Thanks @shuynh2017 !

PR #1115, opened 2026-05-11
ci: fix llm-d-infra reusable workflow references (.yml → .yaml)

ev-shindin:
## Summary

All seven workflows in this repo that call reusable workflows from `llm-d/llm-d-infra` started failing on/around May 6, 2026 with:

```
error parsing called workflow ".github/workflows/ci-signed-commits.yaml"
-> "llm-d/llm-d-infra/.github/workflows/reusable-signed-commits.yml@main"
: failed to fetch workflow: workflow was not found.
```

Root cause: the upstream reusable workflow files in `llm-d/llm-d-infra/.github/workflows/` are named `reusable-*.yaml` (verified via the directory listing on `main`), but every caller in this repo references them as `reusable-*.yml`. The files appear to have been renamed from `.yml` to `.yaml` around May 6, breaking every caller simultaneously. The "Check Signed Commits" check is the most visible failure since it gates merges on every PR; the other six workflows fail silently on their own schedules/triggers.

## Files changed

| File | Reference (before → after) |
|---|---|
| `.github/workflows/ci-signed-commits.yaml` | `reusable-signed-commits.yml` → `.yaml` |
| `.github/workflows/non-main-gatekeeper.yml` | `reusable-non-main-gatekeeper.yml` → `.yaml` |
| `.github/workflows/prow-github.yml` | `reusable-prow-commands.yml` → `.yaml` |
| `.github/workflows/prow-pr-automerge.yml` | `reusable-prow-automerge.yml` → `.yaml` |
| `.github/workflows/prow-pr-remove-lgtm.yml` | `reusable-prow-remove-lgtm.yml` → `.yaml` |
| `.github/workflows/stale.yaml` | `reusable-stale.yml` → `.yaml` |
| `.github/workflows/unstale.yaml` | `reusable-unstale.yml` → `.yaml` |

7 files changed, 7 insertions, 7 deletions.

## Test plan

- [ ] Verify the "Check Signed Commits" job runs successfully on this PR (it's gated on every PR; if the reference resolves, it should report a pass/fail rather than the "Invalid workflow file" parse error).
- [ ] Verify the `Non-Main Gatekeeper` check fires correctly on PRs targeting non-main branches.
- [ ] Spot-check: the `Prow Commands` (`/lgtm`, `/approve`, etc.) workflow responds to a comment on this PR.
- [ ] Confirm no remaining `.yml@main` references to `llm-d/llm-d-infra` workflows:
  ```
  grep -rn "llm-d/llm-d-infra/.github/workflows/.*\.yml@" .github/workflows/
  ```
  (should return nothing)

## Notes

This is purely a caller-side path fix. No logic changes; no upstream coordination required. The fix is consistent with other consumer repos that have presumably already migrated to the `.yaml` extension.

### DIFF
--- a/.github/workflows/ci-signed-commits.yaml
+++ b/.github/workflows/ci-signed-commits.yaml
@@ -3,7 +3,7 @@ on: pull_request_target
 
 jobs:
   signed-commits:
-    uses: llm-d/llm-d-infra/.github/workflows/reusable-signed-commits.yml@main
+    uses: llm-d/llm-d-infra/.github/workflows/reusable-signed-commits.yaml@main
     permissions:
       contents: read
       pull-requests: write
--- a/.github/workflows/non-main-gatekeeper.yml
+++ b/.github/workflows/non-main-gatekeeper.yml
@@ -5,4 +5,4 @@ on:
 
 jobs:
   gatekeeper:
-    uses: llm-d/llm-d-infra/.github/workflows/reusable-non-main-gatekeeper.yml@main
+    uses: llm-d/llm-d-infra/.github/workflows/reusable-non-main-gatekeeper.yaml@main
--- a/.github/workflows/prow-github.yml
+++ b/.github/workflows/prow-github.yml
@@ -9,4 +9,4 @@ permissions:
 
 jobs:
   prow:
-    uses: llm-d/llm-d-infra/.github/workflows/reusable-prow-commands.yml@main
+    uses: llm-d/llm-d-infra/.github/workflows/reusable-prow-commands.yaml@main
--- a/.github/workflows/prow-pr-automerge.yml
+++ b/.github/workflows/prow-pr-automerge.yml
@@ -5,4 +5,4 @@ on:
 
 jobs:
   auto-merge:
-    uses: llm-d/llm-d-infra/.github/workflows/reusable-prow-automerge.yml@main
+    uses: llm-d/llm-d-infra/.github/workflows/reusable-prow-automerge.yaml@main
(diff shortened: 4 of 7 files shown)


ev-shindin:
### Note for maintainers — the failing `Check Signed Commits` on this PR is the bug, not a real failure

The "Invalid workflow file" error on this PR's `Check Signed Commits` job points to the **old** `.yml` reference, even though this PR's branch already has the fixed `.yaml`:

```
error parsing called workflow ".github/workflows/ci-signed-commits.yaml"
-> "llm-d/llm-d-infra/.github/workflows/reusable-signed-commits.yml@main"
: failed to fetch workflow: workflow was not found.
```

That's because the workflow uses `on: pull_request_target`, and GitHub evaluates `pull_request_target` workflows against the **base branch's** version of the file — not the PR head's. The file on `main` still has the broken `.yml` reference, so every PR (including this one) keeps hitting the same error until the fix lands in `main`.

This is the standard "CI fix that can't fix itself" Catch-22.

**To unblock:** please either

1. **Bypass the failing `Check Signed Commits` check** and merge this PR (intended use of admin override for CI-self-fix PRs), or
2. **Push the 7-line patch directly to `main`** — same diff as this PR, no review needed for an emergency CI repair.

Either path will turn the check green on all subsequent PRs.

For long-term coordination: the underlying cause is the upstream `llm-d/llm-d-infra` rename of `reusable-*.yml` → `reusable-*.yaml` around May 6. If the maintainers there add `.yml` symlinks or restore the original names, all consumer repos would unblock simultaneously without needing per-consumer patches like this one. Worth raising upstream regardless.

@asm582 can you please take a care?

mamy-CS (approved):
Thank you for fixing this @ev-shindin, /lgtm /approve

PR #1116, opened 2026-05-11
feat: support multiple saturation scaling stacks per namespace via VA annotation

### LINKED ISSUE #880: Support Multiple Stacks per Namespace
### Overview
Enable support for multiple stacks per Kubernetes namespace. A stack is defined as a combination of VA (Variant Autoscaler) and `wva-saturation-scaling-config` configmap. Other WVA configmaps are at the cluster level and should remain shared across all stacks. 

### Motivation
Currently, the system supports a single stack per namespace. Allowing multiple stacks per namespace will improve flexibility for distributed inference workloads that require independent autoscaling configurations within the same namespace.

vivekk16:
## Summary

Closes #880

Enable a single WVA controller instance to manage multiple independent saturation scaling configurations (VA + saturation scaling ConfigMap) within the same Kubernetes namespace. Previously, all VAs in a namespace shared a single `wva-saturation-scaling-config` ConfigMap, making it impossible to run independent autoscaling configurations for different deployments of the same model side by side.

### Motivation

In distributed inference workloads it is common to have multiple deployments of the same model in one namespace — for example, a production stack alongside an experimental stack, or separate prefill and decode pipelines each requiring
distinct saturation thresholds. Without per-stack saturation config, operators were forced to split workloads across namespaces solely to get different thresholds, adding operational overhead.

### Approach

This implementation uses an **annotation on the VA resource** (`wva.llmd.ai/saturation-config`) to reference a specific saturation ConfigMap, rather than modifying the VA CRD spec. This was a deliberate choice informed by
the [VA deprecation proposal](https://github.com/lionelvillard/llm-d-workload-variant-autoscaler/blob/a18bca72/docs/proposals/deprecate-va-crd.md) — since the VA CRD may be deprecated in favor of HPA/ScaledObject annotations, adding new CRD fields would create unnecessary migration burden. Annotations are portable: when the team moves forward with HPA/ScaledObject-based config, the same annotation key can be adopted with minimal change.

### Key changes

- **VA annotation**: `wva.llmd.ai/saturation-config` specifies which saturation ConfigMap a VA should use (defaults to the well-known name if unset)
- **3-tier config resolution**: VA annotation-referenced ConfigMap →namespace-local default ConfigMap → global ConfigMap. Within a ConfigMap, `modelID#namespace` key → `default` key
- **Independent grouping**: VAs are now grouped by `modelID|namespace|configRef`, so two VAs for the same model with different annotations are optimized independently
- **Custom ConfigMap discovery**: Custom saturation ConfigMaps only need the standard `app.kubernetes.io/name: workload-variant-autoscaler` label — no additional labels required
- **Error handling**: If a VA references a ConfigMap that doesn't exist, the controller logs a warning, emits a Kubernetes `Warning` Event (`SaturationConfigNotFound`), and skips the model — no silent fallback to an
  unintended config
- **RBAC**: Broadened `get` permissions for ConfigMaps to allow reading custom-named ConfigMaps; `update` remains restricted to well-known names

### Files changed (15 files, +830 / -65)

| Area | Files | What changed |
|------|-------|--------------|
| API | `variantautoscaling_types.go`, `_test.go` | New annotation constant and `SaturationConfigRef()` getter |
| Config | `config.go`, `loader.go` | Scoped config storage (`namespace/cmName`), `resolveSaturationConfigForRef`, `HasScopedSaturationConfig` |
| Controller | `configmap_reconciler.go`, `configmap_bootstrap.go`, `predicates.go`, tests | Custom ConfigMap reconciliation, bootstrap discovery, single-label predicate filtering |
| Engine | `engine.go` | configRef-aware config resolution in both V1 and V2 optimizers, Warning Event emission |
| Utils | `variant.go`, `_test.go` | Extended grouping key to include `configRef` |
| Constants | `labels.go` | Re-export annotation key from API package |
| Helm | `rbac/role.yaml` | Unrestricted `get` for ConfigMaps |

### DIFF
--- a/api/v1alpha1/variantautoscaling_types.go
+++ b/api/v1alpha1/variantautoscaling_types.go
@@ -190,3 +190,20 @@ func (va *VariantAutoscaling) GetScaleTargetName() string {
 func (va *VariantAutoscaling) GetScaleTargetKind() string {
 	return va.Spec.ScaleTargetRef.Kind
 }
+
+// SaturationConfigAnnotationKey is the annotation key on VariantAutoscaling resources
+// that specifies which saturation scaling ConfigMap to use. When set, the VA uses
+// the named ConfigMap instead of the default well-known ConfigMap.
+// The referenced ConfigMap must exist in the same namespace as the VA and have
+// the label app.kubernetes.io/name: workload-variant-autoscaler.
+const SaturationConfigAnnotationKey = "wva.llmd.ai/saturation-config"
+
+// SaturationConfigRef returns the name of the saturation scaling ConfigMap
+// this VA should use, as specified by the wva.llmd.ai/saturation-config annotation.
+// Returns empty string if not set, meaning the default well-known ConfigMap should be used.
+func (va *VariantAutoscaling) SaturationConfigRef() string {
+	if va.Annotations != nil {
+		return va.Annotations[SaturationConfigAnnotationKey]
+	}
+	return ""
+}
--- a/api/v1alpha1/variantautoscaling_types_test.go
+++ b/api/v1alpha1/variantautoscaling_types_test.go
@@ -222,6 +222,55 @@ func jsonContainsKey(b []byte, key string) bool {
 	return ok
 }
 
+func TestSaturationConfigRef(t *testing.T) {
+	tests := []struct {
+		name        string
+		annotations map[string]string
+		expected    string
+	}{
+		{
+			name:        "no annotations returns empty",
+			annotations: nil,
+			expected:    "",
+		},
+		{
+			name:        "empty annotations returns empty",
+			annotations: map[string]string{},
+			expected:    "",
+		},
+		{
+			name: "annotation present returns value",
+			annotations: map[string]string{
+				"wva.llmd.ai/saturation-config": "my-custom-config",
+			},
+			expected: "my-custom-config",
+		},
+		{
+			name: "other annotations present but not saturation config",
+			annotations: map[string]string{
+				"some-other-annotation": "value",
+			},
+			expected: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			va := &VariantAutoscaling{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "test-va",
+					Namespace:   "default",
+					Annotations: tt.annotations,
+				},
+			}
+			result := va.SaturationConfigRef()
+			if result != tt.expected {
+				t.Errorf("SaturationConfigRef() = %q, want %q", result, tt.expected)
+			}
+		})
+	}
+}
+
 func TestMinMaxReplicasJSON(t *testing.T) {
 	minVal := int32(2)
 	va := &VariantAutoscaling{
--- a/charts/workload-variant-autoscaler/templates/rbac/role.yaml
+++ b/charts/workload-variant-autoscaler/templates/rbac/role.yaml
@@ -14,23 +14,29 @@ rules:
   verbs:
   - list
   - watch
-  # Note: This broad permission is required for namespace-local ConfigMap overrides.
-  # The controller filters by well-known names (wva-saturation-scaling-config, wva-model-scale-to-zero-config)
-  # in its predicate logic, providing effective access control.
+  # Note: This broad list/watch/get permission is required for namespace-local ConfigMap overrides
+  # and custom saturation ConfigMaps (multi-stack). The controller filters by the
+  # app.kubernetes.io/name label (cache) and well-known names (predicate), providing
+  # effective access control. The cache only watches labelled ConfigMaps.
 - apiGroups:
   - ""
   resources:
   - configmaps
+  verbs:
+  - get
   {{- if not .Values.wva.namespaceScoped }}
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
   resourceNames:
   - {{ include "workload-variant-autoscaler.fullname" . }}-variantautoscaling-config
   - {{ include "workload-variant-autoscaler.fullname" . }}-wva-saturation-scaling-config
   - wva-model-scale-to-zero-config
   - wva-queueing-model-config
-{{- end }}
   verbs:
-  - get
   - update
+{{- end }}
 - apiGroups:
   - ""
   resources:
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"maps"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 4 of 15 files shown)


asm582:
Thanks for this PR. This design opens the idea of using different thresholds for VA optimization by WVA, but it does not include prioritization. 

> or example, a production stack alongside an experimental stack, or separate prefill and decode pipelines each requiring distinct saturation thresholds

From the above use case to be clear, there will not be any prioritization of the selected VA for optimization.

vivekk16:
@asm582 Correct. There is no prioritization of VAs for optimization. Each VA is treated equally and independently by the controller. VAs that reference different saturation ConfigMaps are simply grouped and optimized separately with their own thresholds, but no VA takes precedence over another. The controller processes all VAs in the namespace without any ordering or priority distinction (e.g., production vs. experimental).

If prioritization becomes a requirement in the future, it would be a separate enhancement.

ev-shindin:
@vivekk16 To be honest I don't undersatnd why we need this. Original design was to have a config map with per-model saturation config, so that user can configure different vaues for the different models... Is this gone during refactoring? I need to review this.
Please do not merge this PR before my review.
/hold

vivekk16:
@ev-shindin 

This change enables multiple stacks per namespace. A "stack" is a VA (or group of VAs for the same model) paired with its own `wva-saturation-scaling-config` ConfigMap. Previously the system only supported one saturation scaling ConfigMap per namespace, so all VAs shared the same saturation config, meaning one stack per namespace. Now, a VA can reference a specific saturation ConfigMap via the wva.llmd.ai/saturation-config annotation, so multiple independent stacks can coexist in the same namespace with their own scaling configurations. If no annotation is set, behavior is unchanged and VAs fall back to the default well-known ConfigMap. 

The per-model config within each ConfigMap (`default + modelID#namespace keys`) is preserved exactly as before.

Use cases: (1) Disaggregated serving where prefill and decode have different saturation profiles and need independent scaling configs. (2) Multi-tenancy where teams sharing a namespace can own their saturation configs independently. (3) Safe rollout of new thresholds on one stack while others keep the proven config.

Let me know if you have any specific question.

ev-shindin:
Thanks @vivekk16 — clean implementation. Two parts to my review:

**Part 1 — verified: per-model configuration is preserved.** I was concerned a single-ConfigMap-per-namespace refactor had eliminated the existing per-model override capability. It hasn't. `internal/engines/saturation/engine.go:91-105` is **byte-identical** before and after this PR. The `{modelID}#{namespace}` key lookup, the `base.Merge(override)` field-level inheritance, the `ApplyDefaults()` fallback, and the five Ginkgo test cases in `engine_v2_test.go:186-261` all continue to work. Operators using the existing per-model entries keep working without changes. That concern is cleared.

**Part 2 — bigger question: is the per-VA annotation the right key for the new feature?**

EPP's unit of saturation is the **InferencePool**, verifiable from the EPP daemon's `--pool-name` / `--pool-namespace` startup flags. Routing decisions happen per-pool. The pool itself is keyed by `(name, namespace, spec.selector)` — the selector is a label selector that picks pods. The "model_name → pool" mapping is a strong convention but not CRD-enforced.

WVA already discovers InferencePools natively: `internal/controller/inferencepool_reconciler.go` maintains a datastore of pools, and the VA→Pool linkage is automatic via pod-label matching (the Deployment's labels match the pool's selector). This linkage is documented in `docs/developer-guide/troubleshooting.md:25-29`.

So the principled key for per-stack saturation config is **the InferencePool the VA serves**, which WVA already knows — not a separate annotation that has to be carried per-VA.

Consider the two scenarios the PR description cites:

**Scenario A — "production alongside experimental, same model in one namespace"** breaks down into two cases:
- *If they share an InferencePool:* EPP routes between all matching pods as one unit regardless of which Deployment owns them. WVA giving them different thresholds means the new replica from a per-VA scale-up receives the other VA's traffic anyway → autoscaler decisions inconsistent with the routing layer.
- *If they have separate InferencePools:* WVA already knows which pool each VA belongs to via the existing reconciler. Per-pool config (a ConfigMap keyed by `{poolName}#{namespace}`, say) would key off the actual routing boundary instead of relying on operators to annotate each VA correctly.

Either way, the cleaner solution today is either renaming the model (`granite-13b` vs `granite-13b-exp`) to give them separate pools and using the existing `{modelID}#{namespace}` key, or keying new config off the pool, not the VA.

**Scenario B — "separate prefill/decode pipelines"** is already addressed by role-aware aggregation (the V1 P/D path merged earlier). If different thresholds per role are wanted, the principled primitive is a per-role override key (e.g. `{modelID}#{namespace}#{role}`) in the existing ConfigMap, not a whole new ConfigMap-selection tier.

The per-VA annotation as designed has one concrete failure mode worth calling out: two VAs whose deployments belong to the same InferencePool can be given different `wva.llmd.ai/saturation-config` annotations, and WVA will optimize them with divergent thresholds even though EPP treats them as one pool. The variant-grouping key change (now includes `configRef`) explicitly opts out of pooling their demand. That's the autoscaler contradicting how the routing layer behaves — and every downstream consumer (KEDA, dashboards, future predictor work) now has to learn a `configRef` axis that doesn't exist anywhere else in the stack.

**Asking:**

1. Can you point to a concrete production setup where two VAs *must* share `(model_name, namespace, role)` but need different thresholds? The PR description doesn't cite one, and the two scenarios it does cite have simpler existing solutions.
2. If a real use case does emerge, would keying off the InferencePool (which WVA already discovers) rather than a per-VA annotation be acceptable? That would avoid the autoscaler-router inconsistency and re-use machinery WVA has.

Happy to be wrong about Part 2 if there's a concrete scenario I'm missing — keeping `/hold` until we resolve which scenario this is for.

ev-shindin:
> Thanks for this PR. This design opens the idea of using different thresholds for VA optimization by WVA, but it does not include prioritization.
> 
> > or example, a production stack alongside an experimental stack, or separate prefill and decode pipelines each requiring distinct saturation thresholds
> 
> From the above use case to be clear, there will not be any prioritization of the selected VA for optimization.



> Thanks for this PR. This design opens the idea of using different thresholds for VA optimization by WVA, but it does not include prioritization.
> 
> > or example, a production stack alongside an experimental stack, or separate prefill and decode pipelines each requiring distinct saturation thresholds
> 
> From the above use case to be clear, there will not be any prioritization of the selected VA for optimization.

@asm we already have prioritization mechanics via `variant_cost`. Why we need different mechanism?

ev-shindin:
IMO the cleanest vesion is to produce saturation thresholds as Prometheus metrics from the EPP side, so that operator will not need to care on WVA <-> EPP synchronization. The spart capacity (for v1) or scale-up/down thresholds (for v2) is intentionally per inference pool and should be configured by operator.

asm582:
> > Thanks for this PR. This design opens the idea of using different thresholds for VA optimization by WVA, but it does not include prioritization.
> > > or example, a production stack alongside an experimental stack, or separate prefill and decode pipelines each requiring distinct saturation thresholds
> > 
> > 
> > From the above use case to be clear, there will not be any prioritization of the selected VA for optimization.
> 
> > Thanks for this PR. This design opens the idea of using different thresholds for VA optimization by WVA, but it does not include prioritization.
> > > or example, a production stack alongside an experimental stack, or separate prefill and decode pipelines each requiring distinct saturation thresholds
> > 
> > 
> > From the above use case to be clear, there will not be any prioritization of the selected VA for optimization.
> 
> @asm we already have prioritization mechanics via `variant_cost`. Why we need different mechanism?

Wrong github id tagged, I did not mention that we need a different mechanism. I did point out that there will be no prioritization. Cost is orthogonal.

ev-shindin:
> I did not mention that we need a different mechanism. I did point out that there will be no prioritization. Cost is orthogonal.

We indeed use `variant_cost` for prioritization in v2 (and in v1).  If we are going to change this we need to outline what are relations between `variant_cost` and `priority` w.r.t scaling algorithms.

vivekk16:
Closing this in favor of VA deprecation and introduction of new ScalingPolicy CRD.

PR #1123, opened 2026-05-12
feat: Phase 1 — dual-mode annotation-based variant discovery

lionelvillard:
#1130 Phase 1 

## Summary

Implements Phase 1 (dual-mode) of the VA deprecation proposal (`docs/proposals/deprecate-va-crd.md`), plus the implementation plan and a small refactor.

**docs/plans/va-deprecation.md** — implementation plan for all three phases (dual-mode, deprecation, removal), organized as small semantically-separated tasks.

**Phase 1: annotation-based discovery alongside the existing VA CRD**

- `internal/annotations/` — new package with annotation key constants (`llm-d.ai/managed`, `model-id`, `variant-cost`, `synthetic`) and `Parse`/`IsManaged` helpers.
- `internal/utils/variant_fromannotations.go` — `VariantAutoscalingFromScaledObject`, `VariantAutoscalingFromHPA`, `IsSynthetic`: synthesize in-memory `VariantAutoscaling` objects from annotated objects so the downstream pipeline (collector / analyzer / optimizer / metrics) is unchanged.
- `internal/utils/variant.go` — `readyVariantAutoscalings` merges CRD-sourced and annotation-sourced VAs; new `annotationSourcedVariants` lists HPAs (always) and ScaledObjects (KEDA checked via `NoMatchError`). CRD wins on conflict.
- `internal/engines/saturation/engine.go` — `applySaturationDecisions`: synthetic VAs skip the API-server fetch and `DecisionCache`/`DecisionTrigger` (no CRD status to patch). Metric emission via `act.EmitMetrics` is preserved — KEDA/HPA still read `wva_desired_replicas`.
- `internal/controller/variantautoscaling_controller.go` — adds `kedaEnabled` field, `handleAnnotatedScalerEvent` for namespace tracking, HPA watch (always) and ScaledObject watch (if `kedaEnabled`), RBAC markers for `keda.sh/scaledobjects` and `autoscaling/horizontalpodautoscalers`.
- `cmd/main.go` — registers KEDA scheme unconditionally; detects and logs KEDA presence; passes `kedaEnabled` to the reconciler.
- `config/samples/keda/annotations/`, `config/samples/hpa/annotations/` — new kustomize overlays: annotated ScaledObject and HPA consuming `wva_desired_replicas` without a VA CRD.

**Refactor: CRD detection helpers**

- `internal/utils/crd/` — `CheckCRDInstalled` generic helper + `CheckKEDACRD` and `CheckLeaderWorkerSetCRD` wrappers extracted from `cmd/main.go`, removing duplicated discovery-client boilerplate.

## Test plan

- [x] `make test` — all 29 unit/integration packages pass
- [ ] `make test-e2e-smoke` on a kind cluster with KEDA installed: annotated ScaledObject (no VA) produces `wva_desired_replicas` metric and KEDA scales the Deployment
- [ ] same smoke test with KEDA absent: HPA annotation path still produces the metric
- [ ] existing VA-based e2e tests remain green (CRD path unchanged)

### DIFF
--- a/internal/controller/hpa_reconciler.go
+++ b/internal/controller/hpa_reconciler.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/annotations"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/datastore"
+)
+
+// HPAReconciler tracks namespaces for annotation-based WVA discovery via HPA objects.
+// Its sole job is to call NamespaceTrack / NamespaceUntrack so the engine's polling
+// loop can scope its List calls to namespaces that contain managed scalers.
+type HPAReconciler struct {
+	client.Client
+	Datastore datastore.Datastore
+}
+
+// +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs=get;list;watch
+// Note: Kubernetes RBAC does not support annotation-based selectors, so this grant is
+// necessarily cluster-wide. Access is read-only (get;list;watch). AnnotatedScalerPredicate
+// limits event processing to managed objects (see TODO #1134 for scoped List calls).
+
+// Reconcile tracks or untracks the namespace of an HPA bearing llm-d.ai/managed: "true".
+func (r *HPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	hpa := &autoscalingv2.HorizontalPodAutoscaler{}
+	if err := r.Get(ctx, req.NamespacedName, hpa); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if !hpa.DeletionTimestamp.IsZero() || !annotations.IsManaged(hpa) {
+		r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+	} else {
+		r.Datastore.NamespaceTrack("AnnotatedScaler", req.Name, req.Namespace)
+	}
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager registers HPAReconciler with the controller manager.
+func (r *HPAReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&autoscalingv2.HorizontalPodAutoscaler{},
+			builder.WithPredicates(AnnotatedScalerPredicate()),
+		).
+		Named("hpa").
+		Complete(r)
+}
--- a/internal/controller/scaledobject_reconciler.go
+++ b/internal/controller/scaledobject_reconciler.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/annotations"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/datastore"
+)
+
+// ScaledObjectReconciler tracks namespaces for annotation-based WVA discovery via KEDA
+// ScaledObject objects. Its sole job is to call NamespaceTrack / NamespaceUntrack so
+// the engine's polling loop can scope its List calls to namespaces that contain managed
+// scalers. Registered only when the KEDA CRD is detected at startup.
+type ScaledObjectReconciler struct {
+	client.Client
+	Datastore datastore.Datastore
+}
+
+// +kubebuilder:rbac:groups=keda.sh,resources=scaledobjects,verbs=get;list;watch
+// Note: Kubernetes RBAC does not support annotation-based selectors, so this grant is
+// necessarily cluster-wide. Access is read-only (get;list;watch). AnnotatedScalerPredicate
+// limits event processing to managed objects (see TODO #1134 for scoped List calls).
+
+// Reconcile tracks or untracks the namespace of a ScaledObject bearing llm-d.ai/managed: "true".
+func (r *ScaledObjectReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	so := &kedav1alpha1.ScaledObject{}
+	if err := r.Get(ctx, req.NamespacedName, so); err != nil {
+		if apierrors.IsNotFound(err) {
+			r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if !so.DeletionTimestamp.IsZero() || !annotations.IsManaged(so) {
+		r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+	} else {
+		r.Datastore.NamespaceTrack("AnnotatedScaler", req.Name, req.Namespace)
+	}
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager registers ScaledObjectReconciler with the controller manager.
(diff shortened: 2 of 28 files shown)


asm582:
### Code Review

**Go Code Quality**
- [x] `annotationSourcedVariants` lists all HPAs/ScaledObjects cluster-wide without a label selector — both `k8sClient.List(ctx, &hpaList)` and `k8sClient.List(ctx, &soList)` are unfiltered; in clusters with hundreds of HPAs, this fetches every HPA on every engine tick just to check one annotation. Consider adding a label selector or field selector to reduce cache/API pressure. — `internal/utils/variant.go#L225`
  > AGENTS.md: "Keep packages focused and cohesive" / performance consideration
  > Fixed: added `TODO(#1134)` comments; scoped filtering tracked in issue #1134

- [x] `VariantAutoscalingFromHPA` uses `&defaultMin` instead of `ptr.To(int32(1))` — the project already imports `k8s.io/utils/ptr` and the test file uses `ptr.To`. This is a minor consistency issue but aligns with idiomatic patterns. — `internal/utils/variant_fromannotations.go#L96`
  > AGENTS.md: idiomatic Go / consistency with existing codebase

- [x] `handleAnnotatedScalerEvent` doesn't untrack objects when the `llm-d.ai/managed` annotation is removed — if an update removes the annotation, the object won't be untracked (it's not being deleted, and `IsManaged` returns false, so neither branch executes). Consider adding an explicit untrack when `IsManaged` returns false. — `internal/controller/variantautoscaling_controller.go#L371`
  > Rule: controller-runtime patterns — watch handlers should handle all state transitions

- [x] No predicate filter on the HPA/ScaledObject watches — unlike the Deployment watch which uses `ScaleTargetPredicate()`, the HPA and ScaledObject watches have no predicate. Every HPA/ScaledObject event in the cluster triggers `handleAnnotatedScalerEvent`, even if unrelated to WVA. Add a predicate that checks for the `llm-d.ai/managed` annotation. — `internal/controller/variantautoscaling_controller.go#L400`
  > AGENTS.md: performance / unnecessary handler invocations

- [x] `readyVariantAutoscalings` deduplication key uses only `scaleTargetRef.Name` — the `crdTargets` map is keyed on `namespace/scaleTargetRef.name`, but two different kinds (e.g., Deployment vs LeaderWorkerSet) with the same name in the same namespace would collide. Use `namespace/kind/name` as the key. — `internal/utils/variant.go#L208`
  > Rule: correctness — potential false deduplication

**Test Coverage**
- [x] `annotationSourcedVariants` has no unit test — this is the critical function that lists and filters annotated objects with `NoMatchError` handling. Add a test with a fake client covering: HPAs only, ScaledObjects only, mixed, KEDA not installed, and deduplication. — `internal/utils/variant.go`

- [x] `readyVariantAutoscalings` merge path has no unit test — the CRD-wins-on-conflict deduplication and the non-fatal error handling are untested. — `internal/utils/variant.go`

- [x] `handleAnnotatedScalerEvent` has no unit test — the handler's namespace track/untrack behavior should be tested, especially edge cases (annotation removed, deletion in progress). — `internal/controller/variantautoscaling_controller.go`

- [x] `CheckCRDInstalled` has no unit test — the refactored CRD detection helper is the sole discovery mechanism for KEDA and LWS. A test with a fake discovery client would catch regressions. — `internal/utils/crd/crd.go`

- [x] No negative-cost validation test — `Parse` accepts negative costs (e.g., `"-5.0"`) since it only calls `strconv.ParseFloat`. If negative costs are invalid, add validation; if valid, add a test documenting this. — `internal/annotations/annotations_test.go`
  > Fixed: added validation in `Parse()` and a test case

**Security**
- [x] New RBAC grants `get;list;watch` on all HPAs cluster-wide — while read-only, it exceeds least privilege. The controller only needs HPAs with `llm-d.ai/managed: "true"`. Kubernetes doesn't support label-based RBAC, so this may be unavoidable, but document the justification. — `internal/controller/variantautoscaling_controller.go#L105`

- [x] `unsafeSsl: "true"` in sample ScaledObject — the KEDA trigger sample disables TLS verification. Users will copy-paste this. Add a comment warning that this should be `false` in production, or use a CA bundle. — `config/samples/keda/annotations/scaledobject.yaml#L67`

**Library Reuse**
- [x] `VariantAutoscalingFromHPA` uses `&defaultMin` pattern instead of `ptr.To(int32(1))` from `k8s.io/utils/ptr` (already in go.mod and used in the test file for this same package). — `internal/utils/variant_fromannotations.go#L96`

---

**Architecture notes (informational):**
- ✅ Clean separation of annotation schema into `internal/annotations/`
- ✅ Synthetic VA pattern preserves the downstream pipeline unchanged
- ✅ CRD detection refactor removes ~65 lines of duplicated boilerplate
- ✅ Thorough implementation plan in `docs/plans/va-deprecation.md`
- ⚠️ **Scalability**: unfiltered `List` calls in `annotationSourcedVariants` run on every engine tick. In large clusters with many HPAs, consider caching annotation-sourced VAs and invalidating on watch events.

internal/controller/hpa_reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if !hpa.DeletionTimestamp.IsZero() || !annotations.IsManaged(hpa) {
+		r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+	} else {
+		r.Datastore.NamespaceTrack("AnnotatedScaler", req.Name, req.Namespace)
+	}
```
asm582, line 60: What would be the behavior if two WVA instances are deployed in different namespaces in the cluster?

internal/controller/scaledobject_reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if !so.DeletionTimestamp.IsZero() || !annotations.IsManaged(so) {
+		r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+	} else {
+		r.Datastore.NamespaceTrack("AnnotatedScaler", req.Name, req.Namespace)
+	}
+	return ctrl.Result{}, nil
+}
```
asm582, line 63: Same comment as above for multi-controller install.

internal/utils/variant.go
```diff
@@ -195,9 +203,112 @@ func readyVariantAutoscalings(ctx context.Context, k8sClient client.Client) ([]w
...
+		"totalCount", len(readyVAs))
+
 	return readyVAs, nil
 }
 
+// annotationSourcedVariants lists HPAs and KEDA ScaledObjects bearing llm-d.ai/managed: "true"
+// and synthesizes in-memory VariantAutoscaling objects from them. ScaledObject discovery is
+// skipped gracefully when the KEDA CRD is not installed. When both an HPA and a ScaledObject
+// target the same scale target, the ScaledObject entry wins.
+func annotationSourcedVariants(ctx context.Context, k8sClient client.Client) ([]wvav1alpha1.VariantAutoscaling, error) {
```
asm582, line 245: Nit: comment mentions about precidence, should the func name be `annotationSourcedVariantsWithPrecedence`

asm582 (dismissed):

mamy-CS:
Does `CONTROLLER_INSTANCE` isolation apply to annotation-sourced variants? I don't see the path.

internal/utils/variant.go
```diff
@@ -195,9 +203,112 @@ func readyVariantAutoscalings(ctx context.Context, k8sClient client.Client) ([]w
 	logger.V(logging.DEBUG).Info("Found VariantAutoscaling resources ready for optimization",
 		"count", len(readyVAs),
 		"controllerInstance", controllerInstance)
+
+	// Merge annotation-sourced variants (dual-mode: CRD wins on conflict).
+	annotated, err := annotationSourcedVariants(ctx, k8sClient)
```
mamy-CS, line 208: What happens if `annotationSourcedVariants` returns an error after HPAs were collected but `ScaledObject` List failed?

internal/utils/variant.go
```diff
@@ -195,9 +203,112 @@ func readyVariantAutoscalings(ctx context.Context, k8sClient client.Client) ([]w
...
+
+	// KEDA ScaledObjects — may not be installed; handle gracefully.
+	// ScaledObject takes precedence over HPA for the same scale target.
+	// TODO(#1134): scope to tracked namespaces only (client.InNamespace per ds.ListTrackedNamespaces())
+	// to avoid iterating the full cluster cache on every engine tick.
+	var soList kedav1alpha1.ScaledObjectList
+	if err := k8sClient.List(ctx, &soList); err != nil {
+		if apimeta.IsNoMatchError(err) {
+			logger.V(logging.DEBUG).Info("KEDA ScaledObject CRD not available, skipping annotation discovery for ScaledObjects")
+		} else {
```
mamy-CS, line 280: Worth confirming here if callers intentionally discard this partial slice on error, or adjusting the merge path to retain the HPAs.

internal/engines/saturation/engine.go
mamy-CS, line 1323: `EmitMetrics` still runs for synthetic VAs, this is the path external autoscalers rely on. Please confirm no other metric paths are required for annotation-only variants.

internal/controller/hpa_reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/datastore"
+)
+
+// HPAReconciler tracks namespaces for annotation-based WVA discovery via HPA objects.
+// Its sole job is to call NamespaceTrack / NamespaceUntrack so the engine's polling
+// loop can scope its List calls to namespaces that contain managed scalers.
+type HPAReconciler struct {
+	client.Client
+	Datastore datastore.Datastore
+}
```
mamy-CS, line 38: Worth documenting multi-install behavior (isolation vs shared cluster visibility) or link a follow-up.

internal/controller/scaledobject_reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/annotations"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/datastore"
+)
+
+// ScaledObjectReconciler tracks namespaces for annotation-based WVA discovery via KEDA
+// ScaledObject objects. Its sole job is to call NamespaceTrack / NamespaceUntrack so
+// the engine's polling loop can scope its List calls to namespaces that contain managed
+// scalers. Registered only when the KEDA CRD is detected at startup.
```
mamy-CS, line 35: Same multi-install note as hpa reconciler, per-process tracking does not stop another controller’s engine from seeing the same annotated scaledobjects if listing stays global.

internal/controller/hpa_reconciler.go
lionelvillard: this should go in llm-d repo. Not specific to this PR.

internal/controller/hpa_reconciler.go
lionelvillard: the behavior is preserved (ie. same as VA behavior)

internal/controller/scaledobject_reconciler.go
lionelvillard: listing is scoped to a given namespace at the global level (in main). It applies to all reconcilers.

internal/controller/scaledobject_reconciler.go
lionelvillard: same answer

internal/engines/saturation/engine.go
lionelvillard: confirmed.

internal/utils/variant.go
lionelvillard: CRD wins.

internal/utils/variant.go
lionelvillard: this is not an exported function, it does not really matter.

internal/utils/variant.go
lionelvillard: this comment comes too late :-) Fixing the call site error handling.

mamy-CS (approved):
/lgtm /approve

PR #1127, opened 2026-05-12
checklists for analyzer in WVA

asm582:
High-level targets for the new analyzer to be graduated.

### DIFF
--- a/docs/developer-guide/analyzer-checklists.md
+++ b/docs/developer-guide/analyzer-checklists.md
@@ -1,20 +1,20 @@
-# Analyzer Graduation Criteria
+# Analyzer checklists
 
-This document defines the benchmark-based graduation criteria that provides data for consuming new scaling analyzer into WVA and its production-readiness. The new analyzer could be expected to work with existing analyzers in turn provide value to WVA by improving latency or cost or both. the **current default analyzer's** results recorded in [`docs/benchmark.md`](../benchmark.md) in general the expectation is that the new analyzer(s) should improve over the reported baselines for specific scenarion(s) that the analyzer targets.
+This document defines the benchmark-based checklists that will enable a analyzer to be graduated. The new analyzer could be expected to work with existing analyzers in turn provide value to WVA by improving latency or cost or both. the **current default analyzer's** results recorded in [`docs/benchmark.md`](../benchmark.md) in general the expectation is that the new analyzer(s) should improve over the reported baselines for specific scenarion(s) that the analyzer targets.
 
 ## Reference Workloads
 
-Every candidate analyzer in WVA must be periodically benchmarked against scenario(s) which it plan's to improve. Benchmarking should be done using llmd components by installing Gateway and the scheduler with flow controller enabled on GPU cluster. below are few sample current scenarios.
+Every candidate analyzer in WVA must be periodically benchmarked against scenario(s) which it plan's to improve. Benchmarking should be done using llmd components by installing Gateway and the llmd request scheduler plugins with flow controller enabled on GPU cluster. below are few sample current scenarios.
 
 | Scenario | Input Tokens | Output Tokens | Request Rate | Duration |
 |---|---|---|---|---|
-| **Prefill-heavy** | 4000 | 1000 | 20 RPS (Poisson) or depends on analyzer | depends on analyzer |
-| **Decode-heavy** | 1000 | 4000 | 20 RPS (Poisson) or depends on analyzer | depends on analyzer |
-| **Symmetrical** | 1000 | 1000 | 20 RPS (Poisson) or depends on analyzer | depends on analyzer |
+| **Prefill-heavy** | 4000 | 1000 |  depends on analyzer | depends on analyzer |
+| **Decode-heavy** | 1000 | 4000 |   depends on analyzer | depends on analyzer |
+| **Symmetrical** | 1000 | 1000 |    depends on analyzer | depends on analyzer |
 
 ## Reference Environment
 
-Benchmark results are comparable when run under identical conditions, we currently run benchmark on below profile:
+Benchmark results are comparable when run under identical conditions, we currently run benchmark on below profile, for comparison it is better to use the same hardware profile:
 
 - **Hardware**: NVIDIA H100 (OpenShift cluster)
 - **Model**: The model specified in `docs/benchmark.md` (currently Qwen/Qwen3-32B)
@@ -36,15 +36,6 @@ The following metrics are collected for each benchmark run:
 | **Avg KV cache utilization** | Mean KV cache pressure — how close to memory saturation |
 | **Avg queue depth** | Mean EPP queue depth — scaling responsiveness indicator |
 
-
-## Promotion to Default
-
-Graduation allows an analyzer to exist as a selectable option. The analyzer must demonstrate improvement on **one or all** of the following without regressing any:
-
-| Improvement Target | Threshold |
-|---|---|
-| P9x TTFT reduction | > lower than current default |
-| Cost reduction | > fewer avg replicas at equal or better latency |
 
 ## Recording Results
 
--- a/docs/developer-guide/analyzer-checklists.md
+++ b/docs/developer-guide/analyzer-checklists.md
@@ -61,4 +61,4 @@ All benchmark results must be added to [`docs/benchmark.md`](../benchmark.md) in
 | Cost (avg replicas × GPU/hr) | ... | ... |
 ```
 
-Include the analyzer's configuration parameters (thresholds, tuning knobs) alongside the results so that others can reproduce the run.
+Include the analyzer's configuration parameters (thresholds, tuning knobs) alongside the results so that others can reproduce the run with clear instructions in developer docs.
--- a/docs/developer-guide/analyzer-graduation-criteria.md
+++ b/docs/developer-guide/analyzer-graduation-criteria.md
@@ -1,27 +1,26 @@
 # Analyzer Graduation Criteria
 
-This document defines the benchmark-based graduation criteria for introducing a new scaling analyzer into WVA. All criteria are evaluated against the **current default analyzer's** results recorded in [`docs/benchmark.md`](../benchmark.md).
+This document defines the benchmark-based graduation criteria that provides data for consuming new scaling analyzer into WVA and its production-readiness. The new analyzer could be expected to work with existing analyzers in turn provide value to WVA by improving latency or cost or both. the **current default analyzer's** results recorded in [`docs/benchmark.md`](../benchmark.md) in general the expectation is that the new analyzer(s) should improve over the reported baselines for specific scenarion(s) that the analyzer targets.
 
 ## Reference Workloads
 
-Every candidate analyze in WVA must be benchmarked against all three canonical workload profiles or create new profiles relevant to analyzer's behavior. Benchmarking should be done using llmd components by installing Gateway and the scheduler on GPU cluster. below are the current scenarios.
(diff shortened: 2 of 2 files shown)


docs/developer-guide/analyzer-graduation-criteria.md
```diff
@@ -0,0 +1,97 @@
...
+
+This document defines the benchmark-based graduation criteria for introducing a new scaling analyzer into WVA. All criteria are evaluated against the **current default analyzer's** results recorded in [`docs/benchmark.md`](../benchmark.md).
+
+## Reference Workloads
+
+Every candidate analyze in WVA must be benchmarked against all three canonical workload profiles or create new profiles relevant to analyzer's behavior. Benchmarking should be done using llmd components by installing Gateway and the scheduler on GPU cluster. below are the current scenarios.
+
+| Scenario | Input Tokens | Output Tokens | Request Rate | Duration |
+|---|---|---|---|---|
+| **Prefill-heavy** | 4000 | 1000 | 20 RPS (Poisson) | depends on analyzer |
```
ev-shindin: IMO starting form workload of 20RPS with single replica is not reasonable. Its a spike scenarion that will be haevy to handle for any autoscaler. The scenarios should be more reasonable, and handling spikes is very advanced topic that requires special handling from the autoscaler.

ev-shindin (changes requested):

docs/developer-guide/analyzer-graduation-criteria.md
lionelvillard: right, so either start with X replicas or ignore the initial spike phase. From an autoscaler perspective, traffic is constant (in average) and we may want to say it explicitly (new column with scenario name?)

asm582: Thanks, I have updated the doc to reflect the same - traffic pattern should be representative of the value an analyzer provides.

docs/developer-guide/analyzer-checklists.md
```diff
@@ -0,0 +1,73 @@
...
+
+| Metric | What It Measures |
+|---|---|
+| **P9x TTFT** (ms) | Worst-case time to first token — user-perceived responsiveness |
+| **P9x ITL** (ms/token) | Worst-case inter-token latency — streaming quality |
+| **Error rate** | Fraction of requests that failed (timeouts, 5xx, etc.) |
+| **Avg replicas** | Mean replica count over the run — proxy for cost |
+| **Max replicas** | Peak replica count reached — headroom usage |
+| **Avg KV cache utilization** | Mean KV cache pressure — how close to memory saturation |
+| **Avg queue depth** | Mean EPP queue depth — scaling responsiveness indicator |
```
kahilam, line 37: Should we add "Avg pod startup (s)" to this metrics table? It's already tracked in all our existing benchmark results in benchmark.md.

docs/developer-guide/analyzer-checklists.md
```diff
@@ -0,0 +1,73 @@
...
+Graduation allows an analyzer to exist as a selectable option. The analyzer must demonstrate improvement on **one or all** of the following without regressing any:
+
+| Improvement Target | Threshold |
+|---|---|
+| P9x TTFT reduction | > lower than current default |
+| Cost reduction | > fewer avg replicas at equal or better latency |
+
+## Recording Results
+
+All benchmark results must be added to [`docs/benchmark.md`](../benchmark.md) in a new section following the existing format:
```
kahilam, line 42: Should we specify a minimum number of runs (e.g., 3) with averages reported to ensure reproducibility?

docs/developer-guide/analyzer-graduation-criteria.md
```diff
@@ -0,0 +1,73 @@
+# Analyzer Graduation Criteria
+
+This document defines the benchmark-based graduation criteria that provides data for consuming new scaling analyzer into WVA and its production-readiness. The new analyzer could be expected to work with existing analyzers in turn provide value to WVA by improving latency or cost or both. the **current default analyzer's** results recorded in [`docs/benchmark.md`](../benchmark.md) in general the expectation is that the new analyzer(s) should improve over the reported baselines for specific scenarion(s) that the analyzer targets.
```
deanlorenz: When a new analyzer is submitted, should the PR also fix the reference numbers in benchmark.md that define its CI gate? Otherwise, the bar rises automatically if sat_v2 results improve over time, which might not be the intent. Perhaps have an absolute target that we can refresh independently.

docs/developer-guide/analyzer-graduation-criteria.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Reference Workloads
+
+Every candidate analyzer in WVA must be periodically benchmarked against scenario(s) which it plan's to improve. Benchmarking should be done using llmd components by installing Gateway and the scheduler with flow controller enabled on GPU cluster. below are few sample current scenarios.
```
deanlorenz: Are these 3 scenarios intended to be exhaustive or illustrative. A ramp/staircase scenario (gradually increasing RPS) might be useful to add — TA should behave differently under load changes than at steady state.

docs/developer-guide/analyzer-graduation-criteria.md
```diff
@@ -0,0 +1,73 @@
...
+
+| Scenario | Input Tokens | Output Tokens | Request Rate | Duration |
+|---|---|---|---|---|
+| **Prefill-heavy** | 4000 | 1000 | 20 RPS (Poisson) or depends on analyzer | depends on analyzer |
+| **Decode-heavy** | 1000 | 4000 | 20 RPS (Poisson) or depends on analyzer | depends on analyzer |
+| **Symmetrical** | 1000 | 1000 | 20 RPS (Poisson) or depends on analyzer | depends on analyzer |
```
deanlorenz: If two analyzers run with different RPS, how do we compare them side by side? Maybe a rule that head-to-head comparison runs must use identical settings, even if the analyzer-specific scenario uses different ones?

docs/developer-guide/analyzer-graduation-criteria.md
```diff
@@ -0,0 +1,73 @@
...
+| **Error rate** | Fraction of requests that failed (timeouts, 5xx, etc.) |
+| **Avg replicas** | Mean replica count over the run — proxy for cost |
+| **Max replicas** | Peak replica count reached — headroom usage |
+| **Avg KV cache utilization** | Mean KV cache pressure — how close to memory saturation |
+| **Avg queue depth** | Mean EPP queue depth — scaling responsiveness indicator |
+
+
+## Promotion to Default
+
+Graduation allows an analyzer to exist as a selectable option. The analyzer must demonstrate improvement on **one or all** of the following without regressing any:
```
deanlorenz: Agree. Worth adding the complementary requirement: when the new analyzer is not selected, existing benchmark results should be unchanged. Makes it easier to adopt the analyzer incrementally without worrying about side effects.

docs/developer-guide/analyzer-graduation-criteria.md
```diff
@@ -0,0 +1,73 @@
...
+
+| Improvement Target | Threshold |
+|---|---|
+| P9x TTFT reduction | > lower than current default |
+| Cost reduction | > fewer avg replicas at equal or better latency |
```
deanlorenz: The thresholds here are relative to whatever the default produces at run time. Fixing the numbers explicitly at submission time (and updating them deliberately when needed) might make the gate more stable. 

Worth noting that the new analyzer may be *added* rather than replace. 

Also: for WVA-vs-HPA or WVA-vs-KEDA comparisons, would those live in a separate WVA-level comparison guide rather than per-analyzer? The analyzer criteria could stay focused on "does this improve WVA?" and the broader guide would track how WVA compares to alternatives. New analyzers would add a section there if they shift the comparison.

docs/developer-guide/analyzer-checklists.md
deanlorenz, line 42: Should these scenarios eventually be part of an automated CI run? Manually recorded results are good for documentation but wouldn't catch regressions in future PRs touching the same code path.

docs/developer-guide/analyzer-graduation-criteria.md
asm582: To me, they are illustrative at this point. We can change later once we have concrete community requirements.

docs/developer-guide/analyzer-checklists.md
asm582: Good point, sure.

deanlorenz:
## General comments:

1.  I think this is important and broader than "graduation". I see most of this as continuous rather than a one-time event. We should have performance-oriented Analyzer Benchmarks as part of our CI. Some target level of cost vs SLO can be set as CI gate (ie not comparative but success/failure).

2.  We already have more discussions about good scenario the demonstrate value, use-cases, edge case, ... -- we probably need to have a separate doc to capture scenario taxonomy, cold-start handling, minimum run durations (mean results are useless otherwise), reproducibility requirements, minimal number of runs to get significant results, ... 

## More "graduation" targets:

1. A new analyzer should add a new user guide with a specific configuration and the expected results. We should probably verify we have this for our existing WVA well-lit paths, as a reference for new contributions. 

2. If the new analyzer targets a use case not well covered by existing scenarios, it should add a new CI scenario to showcase the advantage. That will give us a path to extend our CI suite and a natural way to demo the value proposition. If we have a blog/guide the compares different AS configurations then this would add to it.

docs/developer-guide/analyzer-graduation-criteria.md
asm582: Good point, the statement is about showing improvement by running a target workload on a GPU cluster with the llmd system. The output of this run should be in a format similar to the one we use in benchmark.md. I think this statement is not about continuously running benchmarks via CI, but a point-in-time eval when a PR is landed. I will make this clear in the doc. Let me know if you think otherwise.

docs/developer-guide/analyzer-graduation-criteria.md
asm582: Agree, we can start with identical settings. RPS, as we discussed, is tricky, as prompt submission style and length matter too.

docs/developer-guide/analyzer-graduation-criteria.md
asm582: The concept of analyzer is local to WVA, so maybe we can start by comparing analyzers within WVA?

docs/developer-guide/analyzer-checklists.md
asm582: GPUs are scarce and expensive at this point, but yes, we can agree on which scenarios we run, let's say nightly, to catch any regression. This doc is not about CI, though.

deanlorenz: I understand these benchmarks are more comprehensive than CI. However, conceptually, we need a "value-base" regression for WVA. Maybe not CI, maybe only once per release, maybe once per blog. A new analyzer is a good trigger.

deanlorenz (approved):
LGTM. A few forward-looking notes — these aren't blockers, but worth tracking either as an appendix to this doc or in a follow-up issue.

**Scenario list is illustrative — where does it grow?** The three scenarios are a good starting point, but as the community adds use cases (ramp/staircase, cold-start, burst, …) we'll want a living home for this list so there's a clear place to extend it over time.

**Baseline drift and re-evaluation.** As scenarios are added and the stack evolves (model upgrades, EPP changes, new releases), existing benchmark numbers go stale. Worth capturing a cadence — even if the answer for now is "once per release" — so we don't end up with sat_v2 numbers from 6 months ago. A new analyzer submission is a natural trigger: re-run the reference analyzer under identical settings, and confirm no regression when the new analyzer is disabled.

Issue #1003, opened 2026-04-14
Limiter chain: compose multiple limiters with intersection semantics


ev-shindin:
### Problem

The current `Engine.optimizer` selection (`internal/engines/saturation/engine.go:269-275`) chooses **one** optimizer based on `enableLimiter`:

```go
if enableLimiter {
    e.optimizer = pipeline.NewGreedyByScoreOptimizer()
} else {
    e.optimizer = pipeline.NewCostAwareOptimizer()
}
```

There's only ever a single `Limiter` instance, with a single `Inventory`. Operators cannot combine multiple constraints — for example:

- "Limited by physical cluster inventory **AND** by per-namespace quota **AND** by cluster-wide quota"
- "Limited by namespace inventory **AND** by per-namespace quota"

With issues #1001 and #1002 introducing `NamespaceInventory` and `QuotaInventory`, operators will want to apply multiple of these simultaneously. The current single-limiter model forces them to pick one.

### Proposed solution

Add a `ChainLimiter` that composes multiple `Limiter` instances and applies them in sequence. Each decision must satisfy all limiters — a decision is constrained to the **minimum** of all `TargetReplicas` values produced by the chain.

#### Approach: sequential application with intersection

Apply each limiter in order. After each limiter runs, the next limiter sees the (possibly reduced) `TargetReplicas` from the previous step. Because limiters only ever **reduce** `TargetReplicas` (never increase), this naturally produces the intersection.

#### Why sequential works correctly

Each `Limit()` call reads `TargetReplicas` from the decision and may reduce it. The next limiter in the chain reads the already-reduced value. The final `TargetReplicas` is therefore `min(L1, L2, ..., Ln)` of what each limiter would have allowed individually — exactly the intersection semantics we want.

#### Decision step provenance

Each limiter already appends to `DecisionSteps`. The chain doesn't add its own step — the per-limiter steps tell the full story. Example output for a decision constrained by namespace quota:

```
- step: saturation-analyzer    target: 8 (scale-up from 4)
- step: cluster-inventory      target: 8 (allowed)
- step: namespace-quota        target: 6 (capped from 8 by namespace quota llm-d-prod/H100)
- step: cluster-quota          target: 6 (allowed)
```

### Configuration

```yaml
# wva-limiter-config ConfigMap
apiVersion: v1
kind: ConfigMap
metadata:
  name: wva-limiter-config
  namespace: workload-variant-autoscaler-system
data:
  default: |
[...]

PR #1129, opened 2026-05-13
Feat: quota-based limiter (per-GPU-type quotas, cluster/namespace scope)

### LINKED ISSUE #1002: Quota-based limiter (per-GPU-type quotas at cluster and/or namespace scope)
### Problem

The current `TypeInventory` derives GPU limits exclusively from cluster node discovery. Operators have no way to:

- Set explicit per-GPU-type caps independent of physical inventory (e.g., "even though the cluster has 32 H100s, WVA may only allocate 16 to inference workloads")
- Apply per-namespace quotas (e.g., "team-a may use up to 8 H100s, team-b up to 4")
- Combine cluster-wide and per-namespace quotas (e.g., "team-a ≤ 8, team-b ≤ 4, total inference ≤ 16")
- Govern allocation for cost reasons (chargeback, reserved capacity for non-inference workloads, contractual limits)

This is a gap for clusters where WVA shares GPUs with other workloads or where finance/governance constraints apply.

### Proposed solution

Add a `QuotaLimiter` that applies user-declared quotas independent of physical inventory. It implements the existing `Limiter` interface (or `Inventory` if we want chain-friendly composition — see design discussion below).

Quotas can be defined at two scopes:

- **Cluster scope:** cap on total GPUs of a given type across all namespaces
- **Namespace scope:** cap on GPUs of a given type within one namespace

Both scopes can be active simultaneously and are applied independently — a decision must satisfy both.

### Configuration

```yaml
# wva-limiter-config ConfigMap
apiVersion: v1
kind: ConfigMap
metadata:
  name: wva-limiter-config
  namespace: workload-variant-autoscaler-system
data:
  default: |
    limiters:
      # Cluster-wide quota: total inference H100s capped at 16
      - name: cluster-quota
        type: quota
        scope: cluster
        quotas:
          H100: 16
          A100: 8

      # Per-namespace quota: each namespace gets its own cap, keyed by literal namespace name
      - name: namespace-quota
        type: quota
        scope: namespace
        # Namespaces listed in `exclude` bypass this limiter entirely
        # (no constraint applied). Useful for system/privileged namespaces.
        exclude:
          - llm-d-system
          - kube-system
        quotas:
          llm-d-prod:
            H100: 12
            A100: 4
          llm-d-dev:
            H100: 4
            A100: 2
          llm-d-priority:
            H100: 8
            A100: -1            # -1 = unlimited for this type in this namespace
          # Special key "default" matches any namespace not listed above.
          # Omit "default" to deny unlisted namespaces (zero quota).
          default:
            H100: 2
```

#### Two ways to express "unlimited"

This limiter supports two complementary mechanisms for skipping constraints:

| Mechanism | Scope | Use case |
|-----------|-------|----------|
| `exclude: [namespace]` | Whole namespace bypasses **this limiter** entirely | System/privileged namespaces; no constraint of any kind from this limiter |
| `-1` value in `quotas` | Single GPU type unlimited within one namespace | Cap H100 but allow unlimited A100 in the same namespace |

`-1` follows Kubernetes conventions (e.g., `terminationGracePeriodSeconds: -1`).

#### Namespace lookup rules

1. If the namespace appears in `exclude` → bypass this limiter (no constraint applied)
2. Exact namespace name match in `quotas` wins (e.g., `llm-d-prod`)
3. If no exact match and `default` is present, use it as the fallback for any unlisted namespace
4. If no exact match and no `default` key, the namespace is treated as having **zero quota** for that limiter — variants in unlisted namespaces will be fully constrained

For each per-type entry:
- Positive integer → cap at that value
- `-1` → unlimited for this GPU type in this namespace
- `0` or missing → zero quota (denied)

> The reserved key name `default` cannot be used as a real namespace name in this config. If a real namespace happens to be named `default` (the Kubernetes default namespace), add it to `exclude` or use a more specific limiter strategy.

This lets operators express common policies cleanly:
- **Strict allowlist** — list every approved namespace, omit `default` → unlisted namespaces get 0 GPUs
- **Permissive with caps** — list special namespaces, set `default` to a small fallback → unlisted namespaces get a sane default
- **Pure default** — only the `default` key with one entry → all namespaces get the same cap
- **Privileged bypass** — list system namespaces in `exclude` → they're never constrained by this limiter
- **Mixed unlimited** — use `-1` per type for namespaces that should be capped on some GPUs but not others

When both quotas are present (combined with sub-issue #3 limiter chain), a decision must satisfy ALL of them. Example: a `team-a` variant requesting 6 H100s passes if cluster-quota has ≥6 H100s remaining AND `team-a` namespace-quota has ≥6 H100s remaining.

### Interaction with cluster inventory

When `QuotaLimiter` is used **alongside** `TypeInventory` (via the chain in sub-issue #3), allocations are constrained by `min(physical inventory, quota)`. This is the safe default — operators cannot accidentally configure a quota larger than physical capacity.

When `QuotaLimiter` is used **instead of** `TypeInventory`, no physical bound is enforced. This is appropriate for clusters where node discovery doesn't work (e.g., shared infrastructure where WVA can't list nodes) and the operator manages capacity externally.

### Validation

ConfigMap parser must validate:
- Quotas ≥ -1 (only `-1` is allowed as negative; means unlimited for that type)
- Accelerator type names are non-empty strings (no further validation — accelerator names are operator-defined)
- Namespace names exist in the cluster (warning only — namespaces may be created later)
- `exclude` entries are non-empty strings; warn if a namespace appears in both `exclude` and `quotas` (use `exclude` and ignore the quota entry, with a warning logged)

### Acceptance criteria

- New `QuotaInventory` struct implementing `Inventory` interface
- New `quotaAllocator` implementing `ResourceAllocator`
- Support for both `cluster` and `namespace` scope, configurable independently
- ConfigMap parser with validation
- Unit tests covering: cluster scope, namespace scope, both scopes combined, exhaustion, missing namespace
- Documentation explaining quota semantics and interaction with cluster inventory
- `DecisionStep` includes `"limited by quota[scope=cluster, type=H100]"` or `"limited by quota[scope=namespace, ns=team-a, type=H100]"`

ev-shindin:
## Summary

Adds a quota limiter that enforces operator-declared GPU caps independent of
physical inventory, and wires it as a startup-selectable replacement for the
existing physical-inventory limiter. Operators choose one mode at deploy time:

- `--limiter-type=inventory` (default) — today's behavior; `TypeInventory`
  discovers physical GPUs via the GPU operator.
- `--limiter-type=quota --quota-config-file=/path/to/quotas.yaml` — pure
  operator-declared caps; **no Node API access** in the controller.

The two modes are mutually exclusive in this PR. Composing physical and
quota bounds (`min(physical, quota)`) is the limiter chain's job and is
out of scope here — tracked in sub-issue #1003.

### Configuration

```yaml
limiters:
  - name: cluster-quota
    type: quota
    scope: cluster
    quotas:
      H100: 16
      A100: 8
      L40S: -1   # unlimited
  - name: namespace-quota
    type: quota
    scope: namespace
    exclude:
      - kube-system
      - llm-d-system
    namespaceQuotas:
      team-a:
        H100: 8
      team-priority:
        H100: -1
      default:
        H100: 2   # per-namespace fallback for unlisted namespaces
```

### Special values

| Value | Meaning |
|-------|---------|
| Positive integer | Cap in GPUs |
| `-1` | Unlimited — allocator passes through, no usage recorded |
| `0` or missing entry | Denied (no allocation) |

### Namespace lookup rules (`namespace` scope)

For namespace `N` and accelerator type `T`:

1. `N` in `exclude` → **pass-through**; the limiter records no usage so other
   chain members still apply.
2. `namespaceQuotas[N][T]` exists → that cap.
3. Else `namespaceQuotas["default"][T]` exists → that cap as the
   **per-namespace** default (Kubernetes `LimitRange` convention — each
   unlisted namespace gets its own budget at this level, not a shared pool).
4. Else → denied (0).

> ⚠️ The reserved key `default` collides with the real Kubernetes `default`
> namespace. To enforce a cap on workloads in the K8s `default` namespace,
> use `exclude` or omit the reserved key (strict allowlist).

## Operator UX

```bash
# Default (today's behavior, no flags needed):
manager

# Quota mode pointing at a YAML file:
manager --limiter-type=quota --quota-config-file=/etc/wva/quota.yaml
```

Also honored via env vars (`LIMITER_TYPE` / `QUOTA_CONFIG_FILE`) or the
unified config file. `config.Validate` rejects invalid combinations
(quota without file, unknown limiter type, empty entries) at startup
before any reconciler runs.

## Resource access in quota mode

Quota mode is fully Node-API-free. The limiter, inventory, allocator, and
factory paths never call `discovery.K8sWithGpuOperator.*` and never
instantiate a `corev1.NodeList` informer. The previously orthogonal
`collector.CollectInventoryK8S` call (gated by `WVA_LIMITED_MODE`) is now
**also** gated on `LimiterType == inventory` via the
`shouldCollectClusterInventory` predicate, so even the combination
`--limiter-type=quota` + `WVA_LIMITED_MODE=true` no longer lists Nodes.
`main.go` emits an informational log when this combination is detected.

| `--limiter-type` | `WVA_LIMITED_MODE` | Node API access |
|------------------|--------------------|------------------|
| `inventory` (default) | `false` (default) | Limiter path only (unchanged) |
| `inventory` | `true` | Limiter + `CollectInventoryK8S` (unchanged) |
| `quota` | `false` | **None** |
| `quota` | `true` | **None** (`CollectInventoryK8S` suppressed; startup log explains why) |

## DecisionStep observability

When `quotaAllocator.TryAllocate` caps a request (`granted < requested`), it
appends a `DecisionStep` via `VariantDecision.AddDecisionStep` formatted per
issue #1002:

- Cluster scope: `limited by quota[scope=cluster, type=H100]`
- Namespace scope: `limited by quota[scope=namespace, namespace=team-a, type=H100]`

Pass-through paths (excluded namespace, unlimited quota, unresolved
accelerator) do **not** record a step. Spec assertions use exact match, so
drift from the documented format fails CI.

## Validation

`QuotaLimiterEntries.Validate()` accumulates **all** errors with
`errors.Join`. Operators see every misconfiguration in a single pass:

- Empty / duplicate `name`.
- `type` other than `"quota"`.
- `scope` outside `{cluster, namespace}`.
- Cluster scope rejects `namespaceQuotas` and `exclude`.
- Namespace scope rejects `quotas`.
- Per-type values < `-1`.
- Empty accelerator type or namespace keys.
- **Warning (non-fatal)**: namespace appears in both `exclude` and
  `namespaceQuotas` — entry stays valid, `exclude` wins.

## Tests

All new tests use Ginkgo v2 + Gomega to match the project's existing suites.

| File | Specs | Coverage |
|------|-------|----------|
| `internal/config/quota_limiter_test.go` | 23 | `IsExcluded`, `QuotaForNamespace`, `Validate` (happy path + accumulating errors + table of error cases) |
| `internal/engines/pipeline/quota_inventory_test.go` | 27 | Cluster + namespace happy paths, exhaustion, missing-namespace, unknown-type, unlimited, exclude pass-through, default fallback, strict allowlist, combined cluster+namespace, DecisionStep emission contract (exact-match-pinned), `NamespaceAwareInventory` satisfaction, `GetResourcePools` for both scopes, totals, `Remaining` |
| `internal/engines/pipeline/composite_limiter_test.go` | 7 | Interface satisfaction, no-op paths, in-order invocation, most-restrictive-cap propagation, error fan-out, constituent introspection |
| `internal/engines/pipeline/limiter_factory_test.go` | 5 | Inventory mode, unknown-type rejection, single-entry quota → `DefaultLimiter`, multi-entry quota → `CompositeLimiter`, empty-entries error. Helpers use real YAML temp files via `GinkgoT().TempDir()` |
| `internal/engines/pipeline/default_limiter_namespace_test.go` | 3 | `NamespaceAwareInventory` receives `SetUsedByNamespace` with expected per-(ns,type) sums; missing-namespace/missing-type decisions skipped; plain `Inventory` implementations still work |
| `internal/engines/saturation/inventory_gate_test.go` | 4 | `shouldCollectClusterInventory` predicate |

## Test plan

- [x] `go build ./...` clean.
- [x] `go vet ./internal/{config,engines/{pipeline,saturation}}/...` clean.
- [x] `gofmt -d` clean on all touched files (LF-normalized against autocrlf working tree).
- [x] `go test ./internal/config/... ./internal/engines/pipeline/...` on Windows → pass.
- [x] `go test ./internal/config/... ./internal/engines/pipeline/... ./internal/engines/saturation/...` in WSL + envtest 1.34.0 → all green, **23/23 saturation Ginkgo specs pass** (including the new `shouldCollectClusterInventory` table).
- [x] DCO sign-off present on every commit.

## Documentation

`docs/developer-guide/quota-limiter.md` (new) covers:

- Design overview and scope (single-instance per scope; chain composition is #1003).
- Configuration schema with examples.
- Special values (`-1`, `0`, missing).
- Namespace lookup rules including the per-namespace-budget semantic and the
  `default` key collision callout.
- Validation rules + warning behavior.
- Pipeline integration (current contract: `DefaultLimiter.Limit` unconditionally
  calls `SetUsed` and conditionally calls `SetUsedByNamespace` via feature
  detection; each `QuotaInventory` ignores the method irrelevant to its scope).
- Startup wiring (flags, factory dispatch, multi-entry composite).
- Resource-access contract in quota mode (combination matrix above).
- `DecisionStep` trace format.
- Future work (limiter chain composition, reservation-style limiters).

## Follow-ups (separate PRs)

- **#1003** — Limiter chain composition. Replace the simple
  `CompositeLimiter` with a smarter chain that:
  - Composes physical (`TypeInventory`) and quota bounds as
    `min(physical, quota)` instead of the current mutually-exclusive choice.
  - Owns `DecisionStep` ordering / `LimitedBy` selection when multiple caps
    bind the same decision.
- Reservation-style limiters — the `type: "quota"` discriminator leaves
  room for additional limiter types (`reservation`, `priority`) under the
  same ConfigMap schema.

### DIFF
--- a/internal/engines/pipeline/default_limiter.go
+++ b/internal/engines/pipeline/default_limiter.go
@@ -76,13 +76,24 @@ func (l *DefaultLimiter) Limit(ctx context.Context, decisions []*interfaces.Vari
 	// Step 3: Create allocator with available resources
 	allocator := l.inventory.CreateAllocator(ctx)
 
+	// Snapshot TargetReplicas so updateDecisionMetadata can tell whether THIS
+	// limiter actually reduced a decision. In a CompositeLimiter each constituent
+	// runs Limit over the same slice; attributing on the sticky WasLimited flag
+	// alone would make every constituent that runs at/after the capping one
+	// re-emit the limited metric under its own name and overwrite LimitedBy.
+	// Comparing before/after scopes attribution to the limiter that did the cap.
+	targetBefore := make([]int, len(decisions))
+	for i, d := range decisions {
+		targetBefore[i] = d.TargetReplicas
+	}
+
 	// Step 4: Run allocation algorithm to distribute resources
 	if err := l.algorithm.Allocate(ctx, decisions, allocator); err != nil {
 		return fmt.Errorf("allocation algorithm failed: %w", err)
 	}
 
 	// Step 5: Update decision metadata
-	l.updateDecisionMetadata(decisions)
+	l.updateDecisionMetadata(decisions, targetBefore)
 
 	return nil
 }
@@ -140,19 +151,26 @@ func (l *DefaultLimiter) calculateUsedGPUsByNamespace(decisions []*interfaces.Va
 	return usedByNS
 }
 
-// updateDecisionMetadata sets LimitedBy and adds DecisionSteps.
-// Note: WasLimited is set by the algorithm during allocation.
-func (l *DefaultLimiter) updateDecisionMetadata(decisions []*interfaces.VariantDecision) {
-	for _, d := range decisions {
-		// If the algorithm marked the decision as limited, set LimitedBy
-		if d.WasLimited {
+// updateDecisionMetadata sets LimitedBy and adds DecisionSteps. targetBefore is
+// the per-decision TargetReplicas snapshot taken before this limiter's algorithm
+// ran, indexed to match decisions.
+//
+// Attribution (LimitedBy + the limited metric) is recorded only when THIS
+// limiter actually reduced the decision this pass (WasLimited is set AND the
+// target dropped). WasLimited is sticky across a CompositeLimiter's constituents,
+// so gating on it alone would double-count the metric and misattribute LimitedBy
+// to a constituent that merely ran after the one that capped.
+func (l *DefaultLimiter) updateDecisionMetadata(decisions []*interfaces.VariantDecision, targetBefore []int) {
+	for i, d := range decisions {
+		cappedHere := d.WasLimited && d.TargetReplicas < targetBefore[i]
+		if cappedHere {
 			d.LimitedBy = l.name
 			l.metricsEmitter.RecordDecisionsLimitedTotalMetric(d.VariantName, d.Namespace, d.LimitedBy)
 		}
 
 		// Add decision step for observability
 		reason := l.buildStepReason(d)
-		d.AddDecisionStep(l.name, reason, d.WasLimited)
+		d.AddDecisionStep(l.name, reason, cappedHere)
 	}
 }
 
--- a/internal/engines/pipeline/default_limiter.go
+++ b/internal/engines/pipeline/default_limiter.go
@@ -3,6 +3,8 @@ package pipeline
 import (
 	"context"
 	"fmt"
+	"maps"
+	"slices"
 
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/constants"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/interfaces"
@@ -63,6 +65,14 @@ func (l *DefaultLimiter) Limit(ctx context.Context, decisions []*interfaces.Vari
 	usedByType := l.calculateUsedGPUs(decisions)
 	l.inventory.SetUsed(usedByType)
 
+	// Step 2b: Feed namespace-aware inventories the per-namespace usage map.
+	// The per-type SetUsed above is always safe (namespace-only inventories
+	// treat it as a no-op); namespace-scoped inventories additionally need
+	// usage bucketed by namespace to enforce per-tenant caps correctly.
+	if nai, ok := l.inventory.(NamespaceAwareInventory); ok {
+		nai.SetUsedByNamespace(l.calculateUsedGPUsByNamespace(decisions))
+	}
+
 	// Step 3: Create allocator with available resources
 	allocator := l.inventory.CreateAllocator(ctx)
 
@@ -110,6 +120,26 @@ func (l *DefaultLimiter) calculateUsedGPUs(decisions []*interfaces.VariantDecisi
 	return usedByType
 }
 
+// calculateUsedGPUsByNamespace computes current GPU usage bucketed by
+// (namespace, accelerator type). Outer key: namespace; inner key: accelerator
+// type. Used to feed NamespaceAwareInventory implementations the per-namespace
+// usage map needed for per-tenant cap enforcement.
+func (l *DefaultLimiter) calculateUsedGPUsByNamespace(decisions []*interfaces.VariantDecision) map[string]map[string]int {
+	usedByNS := make(map[string]map[string]int)
+	for _, d := range decisions {
+		if d.AcceleratorName == "" || d.Namespace == "" {
+			continue
+		}
+		perType, ok := usedByNS[d.Namespace]
+		if !ok {
+			perType = make(map[string]int)
+			usedByNS[d.Namespace] = perType
+		}
+		perType[d.AcceleratorName] += d.CurrentReplicas * d.GPUsPerReplica
+	}
+	return usedByNS
+}
+
 // updateDecisionMetadata sets LimitedBy and adds DecisionSteps.
 // Note: WasLimited is set by the algorithm during allocation.
 func (l *DefaultLimiter) updateDecisionMetadata(decisions []*interfaces.VariantDecision) {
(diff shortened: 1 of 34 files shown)


internal/engines/pipeline/default_limiter.go
```diff
@@ -142,6 +170,11 @@ func (l *DefaultLimiter) buildStepReason(d *interfaces.VariantDecision) string {
 // ComputeConstraints refreshes the inventory and returns per-type resource availability.
 // This is the V2 path: expose constraints for the optimizer instead of modifying
 // decisions directly (which is what Limit() does for the V1 path).
+//
+// The V2 path is currently per-type only; namespace-aware constraints (e.g.,
+// from a namespace-scoped quota inventory) are not yet exposed via this method
+// because the optimizer's ResourceConstraints shape is per-type. Callers that
+// need namespace-aware constraints should use Limit() directly.
```
deanlorenz: Does this mean this does not work in v2 (which will become the default).
If yes, we should at least make very clear in the docs.

internal/engines/pipeline/quota_inventory.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+// clampPoolLimit converts a quota value to a ResourcePool.Limit field.
+// QuotaUnlimited (-1) becomes 0; this signals "no constraint" by convention
+// (since a Limit of 0 with no usage means TotalAvailable = 0, the optimizer
+// will not draw allocations from such a pool — and the per-allocator
+// TryAllocate path bypasses the limit check entirely for unlimited entries).
+func clampPoolLimit(limit int) int {
+	if limit < 0 {
+		return 0
```
deanlorenz: Does `0` mean no limit or `limit=0` (cannot use specific resource)?
If we allow a quota of `0` need to change this.

deanlorenz:
LGTM on the implementation — quota inventory, V1/V2 parity via `effectiveAvailable`, `mergeNamespaceConstraints`, `aggregateNamespacePools`, the `shouldCollectClusterInventory` gate, and the `CompositeLimiter` fan-out all look correct. Test coverage is solid.

Notes below are docs gaps and future-enhancement suggestions only.

---

**Missing user guide coverage**

The feature is documented in `docs/developer-guide/quota-limiter.md` but `docs/user-guide/` and `docs/developer-guide/configuration.md` (where new flags belong) aren't updated. At minimum `configuration.md` and the user guide should reference the new doc.

---

**Docs / usability**

**Quota reload lifecycle** — the developer guide doesn't mention that quota configuration is loaded once at startup and requires a controller restart to apply changes. Worth adding a note; operators who mount the quota file from a ConfigMap may not realize Kubernetes updates the file on disk while the running controller ignores it.

**`--limiter-type` is a global flag; per-entry `type:` field** — the `type: "quota"` field on each YAML entry is a discriminator reserved for future limiter types and must currently always be `"quota"`. The choice between inventory and quota is made globally via `--limiter-type`, not per entry. Worth a sentence in the developer guide so operators don't wonder whether `type: inventory` is valid inside the YAML.

**Fair-share interaction with namespace quotas** — the guide covers what namespace quotas enforce but not how they interact with the fair-share loop. The mean is `cluster GPUs / number of active models`, not quota-weighted; quotas are hard ceilings applied after. A quota-constrained model that takes less than its fair-share slot releases the remainder to subsequent rounds, not to other models in the same round. Fairness is also per-model, not per-namespace.

Worked example (cluster 8 GPUs; M1 wants 3/quota 2, M2 wants 4/quota 4, M3 wants 4/quota 4): round 1 mean ≈ 2.67 — M2 and M3 each get ≈ 2.67, M1 is capped at 2; round 2 distributes the remaining ≈ 0.67 to M2/M3. Final: M1=2, M2≈3, M3≈3. Outcome is correct; the path and the per-model fairness semantics are worth documenting.

---

**Nits**

- `composite_limiter.go` `Limit()` comment: "decisions slice is the source of usage truth (see `calculateUsedGPUs`), so each constituent sees the already-capped `TargetReplicas`" — `calculateUsedGPUs` uses `CurrentReplicas`, not `TargetReplicas`; the most-restrictive-wins property comes from `TryAllocate` receiving the mutated `TargetReplicas`. Behavior correct, comment misleading.

- `composite_limiter.go` struct comment: "partial commits do not leak" — true for per-allocator usage state; `TargetReplicas` mutations from earlier constituents do persist on error. Not a correctness issue (errors abort the cycle), but slightly imprecise.

- `quota_inventory.go` `quotaAllocator.Remaining()` (namespace scope): iterates `NamespaceQuotas` keys only; namespaces consuming via the `"default"` fallback aren't counted, so `Remaining()` can over-report. Reporting only, not allocation correctness — worth a note on the method.

- `config.go` `LimiterType` godoc and `quota-limiter.md` startup wiring table: "mutually exclusive in PR-1" — PR references in code and shipped docs don't age well; something like "mutually exclusive in the initial implementation (composing with physical inventory is tracked in #1003)" reads better after the PR closes.

deanlorenz (approved):
Reviewed rev 2 + follow-up commits. All original findings addressed; correctness fixes (cluster-unlimited V2, CompositeLimiter attribution, TotalUsed symmetry) look sound. LGTM.

Issue #1124, opened 2026-05-12
Move LWS e2e out of smoke; replace sustained load with small deterministic traffic


mamy-CS:
We have LWS coverage under smoke today (smoke_test.go). That’s heavy for PR smoke. It pulls in sustained-style load and needs enough cluster headroom to behave like a mini load test, which flakes or fails on smaller Kind/CI while still passing locally.

Proposal
- Remove LWS from the smoke suite, keep smoke fast and infra-light. Gate LWS behind full (or a dedicated label if needed) so PR smoke stays a tight correctness signal.
- Stop relying on sustained load for the LWS path, where we only need to prove scaling/saturation wiring. Prefer a short-lived Job (e.g., curl in a loop or a tiny script) with a fixed, small number of requests so we cross whatever saturation threshold we care about in a bounded, deterministic way, not “run load until something happens.”
- Scope should be focused on functionality, assert the things we care about (VA + HPA/LWS reconcile, scale signal, conditions, replica bounds) without benchmarking or long tail latency.

Discussion point
- Go through the gateway for end-to-end production like realistic path or keep it light with a narrow goal of VA + LWS + scaler reconcile and scale on signals we can drive another way (e.g. direct HTTP to the model service, fake metrics, or a tiny in-cluster client that only hits the decode service). The advantage here is that we are not paying gateway/EPP CPU and scheduling cost for a test that isn’t about routing.

asm582:
Our e2e should use the gateway and EPP to send the load.

PR #1142, opened 2026-05-13
chore: move LWS out of smoke, into its own

shuynh2017:
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1124
- Move it out of smoke.
- Will move into its own test in another PR.


mamy-CS (approved):
/lgtm /approve

PR #1144, opened 2026-05-14
bug-fix: observability - config info metrics

shuynh2017:
Fixes:
- Update to record global default config info. In doing so, move the recording of global default config into first thing in `optimize()` so that we won't skip.
- Reset config metric as there should only be one entry for each metric; otherwise, Grafana will show multiple entries for each of these metrics.
- Tested by manually updating the configs and observing the rendering in Grafana. There may be more similar fixes as Grafana dashboard is being build out.


shuynh2017:
still review

internal/engines/saturation/engine.go
```diff
@@ -242,6 +242,19 @@ func (e *Engine) recordActiveOptimizer() {
...
 
+func (e *Engine) recordDefaultConfigMetrics() {
+	metrics.SetConfigOptimizationInterval(float64(e.Config.OptimizationInterval().Seconds()))
+
+	globalSatCfgMap := e.Config.SaturationConfig()
+	// record global default config
+	if cfg, ok := globalSatCfgMap["default"]; ok {
+		cfg.ApplyDefaults()
+		metrics.SetConfigKvSpareThreshold(cfg.KvSpareTrigger)
+		metrics.SetConfigQueueSpareThreshold(cfg.QueueSpareTrigger)
```
ev-shindin, line 325: Non-blocking: `recordDefaultConfigMetrics` records the `default` saturation config, so the config-info / threshold metrics now reflect the **global default only** — per-model/namespace overrides aren't represented. Suggest a one-line note in the metric Help text or a comment here ("reflects the global default; per-model overrides not represented") so it isn't misread.

ev-shindin (dismissed):
LGTM Thanks @shuynh2017

internal/engines/saturation/engine.go
shuynh2017: updated help messages in `metrics.go`.

internal/engines/saturation/engine.go
```diff
@@ -315,6 +315,19 @@ func (e *Engine) recordActiveOptimizer() {
 	}
 }
 
+func (e *Engine) recordDefaultConfigMetrics() {
+	metrics.SetConfigOptimizationInterval(float64(e.Config.OptimizationInterval().Seconds()))
+
+	globalSatCfgMap := e.Config.SaturationConfig()
+	// record global default config
+	if cfg, ok := globalSatCfgMap["default"]; ok {
+		cfg.ApplyDefaults()
```
ev-shindin: ApplyDefaults()` mutates shared config state here. `Config.SaturationConfig()` returns `copySaturationConfig`, which (despite the "deep copy" comment) only copies struct values — the `Analyzers` slice header is shared with the stored global config. `ApplyDefaults` then writes `Analyzers[i].Score` / `.Enabled` in place, every optimize cycle, without holding `c.mu`. That races the configmap reconciler that updates the config under the lock; it's dormant today only because the stored config is already defaulted at parse time (so the loop writes nothing), which is fragile.

Suggest dropping `ApplyDefaults()` (the stored `default` config is already defaulted, so `KvSpareTrigger`/`QueueSpareTrigger` are correct without it), or deep-copying `Analyzers` before defaulting.

internal/engines/saturation/engine.go
```diff
@@ -315,6 +315,19 @@ func (e *Engine) recordActiveOptimizer() {
 	}
 }
 
+func (e *Engine) recordDefaultConfigMetrics() {
+	metrics.SetConfigOptimizationInterval(float64(e.Config.OptimizationInterval().Seconds()))
+
+	globalSatCfgMap := e.Config.SaturationConfig()
+	// record global default config
+	if cfg, ok := globalSatCfgMap["default"]; ok {
```
ev-shindin, line 323: This narrows config-metric recording to the `default` key. The previous per-model recording (the lines removed from `resolveSaturationConfig`) emitted for whatever configs were resolved; now `wva_config_kv_spare_threshold`, `wva_config_queue_spare_threshold`, and `wva_config_info` are skipped entirely when the global config has no `default` entry (e.g. an operator who configures only per-namespace / per-model overrides). Only the optimization-interval gauge would still be recorded. Worth confirming a `default` entry is a guaranteed invariant.

internal/engines/saturation/engine.go
ev-shindin, line 302: `recordDefaultConfigMetrics` now also calls `SetConfigOptimizationInterval` every cycle (and the setter `Reset()`s first), so this startup call is immediately superseded and now duplicates the responsibility in two places. Consider removing it so the interval gauge is owned in one place.

ev-shindin (commented):
LGTM. Please address `ApplyDefaults` issue, all others are optional

internal/engines/saturation/engine.go
shuynh2017: dropped

shuynh2017:
@ev-shindin, I've only addressed the non-optional due to the remaining time for this week. Thanks.

ev-shindin (approved):
Thanks @shuynh2017 !

PR #1147, opened 2026-05-14
bug-fix: observability - available gpus metric

shuynh2017:
Fixes:
Similar as https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1144, reset metrics as there should only be one entry for each accelerator type; otherwise, Grafana will show multiple entries for each of these metrics.
Tested by manually updating the configs and observing the rendering in Grafana. There may be more similar fixes as Grafana dashboard is being build out.


shuynh2017:
still review

internal/engines/saturation/engine.go
```diff
@@ -257,6 +257,11 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 
 	logger := ctrl.LoggerFrom(ctx)
 
+	// For each optimize cycle, reset available GPUs metrics. If controller doesn't do GPUs discovery
+	// in this cycle, then the metrics will not be available as intended. If controller does GPUs discovery,
+	// the number of available GPUs is fresh every cycle.
+	e.metricsEmitter.ResetAvailableGPUsMetric()
```
ev-shindin: `ResetAvailableGPUsMetric()` runs at the **start of `optimize()`** (line 263), but the re-record happens much later — node discovery runs during the limiter's `ComputeConstraints` in `optimizeV2`, emitting per-vendor across List calls. So between Reset and the (spread-out) re-records sits the whole analysis phase + network List calls. Two consequences:

1. **Scrape race:** a scrape landing in that window sees `available_gpus` empty/partial → transient dips in Grafana (a much larger window than #1144's microsecond Reset+Set).
2. **Vanishes on skipped/failed discovery:** discovery only runs on the limiter path and can fail (node-list unavailable, #1298). Reset-at-cycle-start wipes the metric and nothing re-records it → it disappears rather than holding last-known.

The comment shows (2) is intended, but for a slowly-changing inventory gauge, stable-last-known is usually friendlier than flapping-to-absent. **Suggestion:** call the reset immediately before the recording loop in the discovery path (after node listing completes) instead of at `optimize()` start — shrinks the empty-window to the in-memory emit loop, and a skipped/failed discovery then leaves prior values intact. Keep as-is only if "absent unless freshly discovered" is the deliberate semantic.

ev-shindin (dismissed):
LGTM

internal/engines/saturation/engine.go
shuynh2017: Thank you for showing the consequences. I intend to show "exactly what the engine sees" during the optimization. For example, if limiter is flipped from true to false, then metric/dashboard should show the manager not fetching GPUs; or if getting GPU failed, then dashboard should also show the manager couldn't get GPUs.

ev-shindin: Totally get the intent - surfacing "limiter off" / "fetch failed" on the dashboard is worth having. My concern is that encoding it as *the series disappearing* doesn't actually achieve that, and gives false signals:

- The reset is at the top of `optimize()`, but the re-record only happens later, inside the limiter (`engine.go:649`, V2 `:794`). A scrape between the two shows `available_gpus` empty **even on a cycle that did fetch GPUs** — so "couldn't get GPUs" fires spuriously and isn't distinguishable from "scraped mid-cycle."
- An early return (`:366`) or `len(allDecisions)==0` skips the limiter entirely, so a benign "nothing to do" cycle wipes the gauge and reads as a fetch failure.
- In Prometheus a vanished series renders as *No data*, which conflates "engine didn't fetch" with scraper-down / restart / leader change.

Could we instead express the state explicitly - e.g. a `wva_gpu_discovery_up` (1/0) or `_last_success_timestamp_seconds` gauge - and keep `available_gpus` as a stable capacity gauge? That gives "limiter off → 0" and "fetch failed → 0" unambiguously, without the scrape-window false negatives. If you'd rather keep the reset, moving it adjacent to the discovery re-record (so "reset with no following record" only happens on a real failed/skipped fetch) would at least remove the false signals.

internal/metrics/available_gpus_metrics_test.go
```diff
@@ -279,3 +279,85 @@ func TestRecordAvailableGPUsMetric_ZeroCount(t *testing.T) {
 		t.Errorf("Metric %s not found in gathered metrics", constants.WVAAvailableGpus)
 	}
 }
+
+func TestResetAvailableGPUsMetric(t *testing.T) {
```
ev-shindin: Non-blocking: Tests reset-in-isolation, so it passes regardless of the placement issue. Consider an `optimize()`-level test that the gauge is repopulated after the reset (and not left empty when the limiter isn't reached).

shuynh2017:
@ev-shindin, the idea of introducing `wva_gpu_discovery_up` and  keeping `available_gpus` as a stable capacity gauge is good. I will proceed.

internal/metrics/available_gpus_metrics_test.go
shuynh2017: With `wva_gpu_discovery_up `, the panels become:
- enableLimiter=true:
image

- then manually set enableLimiter=false, and wait for few seconds ... note the GPUs are still shown:
image

shuynh2017:
@ev-shindin, I've only addressed the non-optional due to the remaining time for this week. Thanks.

internal/engines/saturation/engine.go
```diff
@@ -638,6 +638,11 @@ func (e *Engine) optimizeV1(
 			globalSaturationConfig = cfg
 		}
 	}
+	if globalSaturationConfig.EnableLimiter {
```
ev-shindin: `SetGpuDiscoveryUp` is called only here in `optimizeV1`. `optimizeV2` (the `saturation` token-based analyzer) and `optimizeQueueingModel` never set it. A Prometheus `GaugeVec` creates series lazily on first `.Set()`, so on those paths the metric has **no series** → the Grafana "GPU Discovery Status" panel shows **"No data"** — while `available_gpus` *is* freshly recorded on V2 (`ComputeConstraints` → inventory `Refresh` → `Discover` → `RecordAvailableGPUsMetric`). The two metrics contradict.

ev-shindin: Set from `globalSaturationConfig.EnableLimiter`, never from whether `Discover()` succeeded. When `EnableLimiter=true` but the node `List` fails (API hiccup / RBAC), the limiter error is swallowed (`engine.go:657`, *"GPU limiter failed, proceeding with original decisions"*) and `available_gpus` goes stale — yet `discovery_up` stays `1`. That is exactly the false confidence the metric is meant to prevent (its own doc says "0 → last successful discovery").

deploy/grafana/operational-dashboard.json
ev-shindin, line 1: The "GPU Discovery Status" stat (`max(wva_gpu_discovery_up)`) renders "No data" for all V2 deployments (see my comment to `engine`). Also: the `wva_available_gpus` panel description reads **"shows mumber of currently available GPUs"** — "mumber" → "number".

ev-shindin (changes requested):
`discovery_up` is set in the engine optimize path from the `EnableLimiter` config flag, but `available_gpus` is recorded in the **discovery subsystem** (`internal/discovery/k8s_with_gpu_operator.go:116`, inside `listGPUNodes`/`Discover`, reached via the limiter's `TypeInventory` refresh). Setting the freshness signal from config in one of three optimize paths — instead of from the actual discovery call — is what causes both findings below.

**One fix resolves all gaps together:** delete `engine.go:641-645`; in `k8s_with_gpu_operator.go` call `metrics.SetGpuDiscoveryUp(1)` on the successful return of `listGPUNodes` (:120) and `metrics.SetGpuDiscoveryUp(0)` on each error return (:47/:62/:73). This ties the gauge to real discovery success, covers V1/V2/queueing-model uniformly (all refresh through `Discover`), writes it next to `available_gpus`, and makes the docs/dashboard accurate as written.

deploy/grafana/operational-dashboard.json
shuynh2017: Fixed mumber->number

shuynh2017:
"One fix resolves all gaps together:". This is much cleaner and still consistent with doc/dashboard. Done.

internal/engines/saturation/engine.go
shuynh2017: fixed as suggested.

shuynh2017: fixed as suggested.

ev-shindin:
@shuynh2017  please rebase

docs/developer-guide/prometheus.md
```diff
@@ -661,9 +661,34 @@ With WVA metrics, the value for the label `namespace` is the WVA controller name
   }
   ```
 
+### `wva_gpu_discovery_up`
+- **Type**: Gauge
+- **Description**: Indicates whether GPU discovery is on (1) or off (0). GPU discovery is enabled when the limiter configuration is active (e.g., `enableLimiter` is `true`). This metric helps operators understand whether WVA is actively discovering GPU resources.
```
ev-shindin, line 666: "GPU discovery is enabled when the limiter configuration is active (`enableLimiter` true)" - reword per above (`0` now means the last discovery **failed**, not "off").

internal/metrics/metrics.go
```diff
@@ -638,6 +654,19 @@ func (m *MetricsEmitter) RecordEnforcerMetric(policyType string) {
 	enforcerModificationsTotal.With(labels).Inc()
 }
 
+// SetGpuDiscoveryUp sets the GPU discovery status metric.
+// status should be 1 when GPU discovery is enabled, 0 when it is disabled.
```
ev-shindin, line 658: "status should be 1 when GPU discovery is enabled, 0 when it is disabled" — change "enabled/disabled" to "last call succeeded / returned an error".

deploy/grafana/operational-dashboard.json
```diff
@@ -177,7 +177,72 @@
         "type": "prometheus",
         "uid": "$datasource_uid"
       },
-      "description": "Number of available GPUs by accelerator type. WVA only discovers GPUs when a configuration such as \"enableLimiter\" is \"true\". Only available in clusters such as OpenShift where WVA can iterate over node objects. There is no exclusions such as tainted nodes or GPUs operating in different modes such as MIG. Metric is wva_available_gpus.",
+      "description": "Indicates whether GPU discovery is on or off. WVA only discovers GPUs when a configuration such as \"enableLimiter\" is \"true\". Related metric is wva_available_gpus. Metric is wva_gpu_discovery_up.",
```
ev-shindin, line 180: "Indicates whether GPU discovery is on or off" — reword per above; note the panel shows **No data** when discovery never ran (absent series), which is distinct from `0` (failed).

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -37,14 +37,24 @@ func NewK8sWithGpuOperator(client client.Client) *K8sWithGpuOperator {
 // This is the single internal node-listing primitive; public methods Discover,
 // discoverNodeGPUTypes, and DiscoverNodes project from its result.
 func (d *K8sWithGpuOperator) listGPUNodes(ctx context.Context) (map[string]NodeInfo, error) {
+	var err error
+	defer func() {
```
ev-shindin, line 41: The `defer`  that sets the gauge is never observed by a test: `gpu_discovery_up_test.go` tests only the `SetGpuDiscoveryUp` setter, and `k8s_with_gpu_operator_test.go` drives `listGPUNodes` but never gathers the gauge. Reverting any `err = fmt.Errorf(...)` to a bare `return nil, fmt.Errorf(...)` (so `err` stays nil → sets `1` on error), or dropping the defer, would pass every test. Add discovery-package tests: success → gauge==1; `Client.List` error (via `WithInterceptorFuncs`) → gauge==0; and `available_gpus` retains its last value after a failure while `discovery_up==0`.

internal/metrics/gpu_discovery_up_test.go
```diff
@@ -0,0 +1,199 @@
+/*
```
ev-shindin, line 1: Please remove license header

ev-shindin (approved):
LGTM. there are some optional improvements.

PR #1150, opened 2026-05-15
feat(deploy): Kustomize-first WVA controller install, deprecate Helm, fix e2e/CI

mamy-CS:
Replaces all Helm-based WVA controller installs with kustomize. The deploy scripts, CI workflows, and e2e tests now all go through Kustomize overlays. Helm stays for upstream installs (LWS, KEDA, kube-prometheus-stack, Prometheus Adapter), those aren't our charts.

Note: The Helm chart still exists (deprecation + deletion will follow)

### DIFF
--- a/.github/workflows/ci-benchmark.yaml
+++ b/.github/workflows/ci-benchmark.yaml
@@ -293,7 +293,7 @@ jobs:
           done
 
       - name: Apply latest CRDs
-        run: kubectl apply -f charts/workload-variant-autoscaler/crds/
+        run: kubectl apply -f config/crd/bases/
 
       - name: Deploy WVA and llm-d infrastructure
         env:
@@ -329,14 +329,13 @@ jobs:
           LLMD_WAIT_FOR_ESSENTIAL_LLM_D_ONLY: "true"
           E2E_DEPLOY_WAIT_TIMEOUT: "300s"
         run: |
-          # Split deploy: WVA base (install.sh) → llm-d (install-llmd-infra.sh) → chart capacity tuning (helm upgrade).
-          ./deploy/install.sh --release-name "$WVA_RELEASE_NAME" --environment openshift
+          # Split deploy: WVA base (install.sh) → llm-d (install-llmd-infra.sh) → capacity tuning via ConfigMap patch.
+          ./deploy/install.sh --environment openshift
           ./deploy/install-llmd-infra.sh -e openshift
-          helm upgrade "$WVA_RELEASE_NAME" ./charts/workload-variant-autoscaler -n "$WVA_NS" --reuse-values \
-            --set wva.capacityScaling.default.kvCacheThreshold="${KV_CACHE_THRESHOLD}" \
-            --set wva.capacityScaling.default.queueLengthThreshold="${QUEUE_LENGTH_THRESHOLD}" \
-            --set wva.capacityScaling.default.kvSpareTrigger="${KV_SPARE_TRIGGER}" \
-            --set wva.capacityScaling.default.queueSpareTrigger="${QUEUE_SPARE_TRIGGER}"
+          kubectl patch configmap workload-variant-autoscaler-saturation-scaling-config \
+            -n "$WVA_NAMESPACE" --type=merge \
+            -p "$(printf '{"data":{"default":"kvCacheThreshold: %s\\nqueueLengthThreshold: %s\\nkvSpareTrigger: %s\\nqueueSpareTrigger: %s\\n"}}' \
+              "${KV_CACHE_THRESHOLD}" "${QUEUE_LENGTH_THRESHOLD}" "${KV_SPARE_TRIGGER}" "${QUEUE_SPARE_TRIGGER}")"
 
       - name: Label namespaces for OpenShift monitoring
         run: |
@@ -956,7 +955,7 @@ jobs:
       - name: Cleanup infrastructure
         if: always()
         run: |
-          helm uninstall "$WVA_RELEASE_NAME" -n "$WVA_NAMESPACE" --ignore-not-found --wait --timeout 60s || true
+          kubectl delete clusterrole,clusterrolebinding -l app.kubernetes.io/name=workload-variant-autoscaler --ignore-not-found || true
           for release in $(helm list -n "$LLMD_NAMESPACE" -q 2>/dev/null); do
             helm uninstall "$release" -n "$LLMD_NAMESPACE" --ignore-not-found --wait --timeout 60s || true
           done
--- a/.github/workflows/ci-e2e-openshift.yaml
+++ b/.github/workflows/ci-e2e-openshift.yaml
@@ -686,9 +686,8 @@ jobs:
       - name: Apply latest CRDs
         run: |
           echo "Applying latest VariantAutoscaling CRD..."
-          # Helm doesn't auto-update CRDs, so we need to apply them manually
-          # to ensure the cluster has the latest schema (including scaleTargetRef)
-          kubectl apply -f charts/workload-variant-autoscaler/crds/
+          # Apply CRDs from the Kustomize source of truth (config/crd/bases/)
+          kubectl apply -f config/crd/bases/
 
       - name: Deploy WVA and llm-d infrastructure
         env:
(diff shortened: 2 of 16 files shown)


.github/workflows/ci-benchmark.yaml
```diff
@@ -329,14 +329,13 @@ jobs:
...
-          ./deploy/install.sh --release-name "$WVA_RELEASE_NAME" --environment openshift
+          # Split deploy: WVA base (install.sh) → llm-d (install-llmd-infra.sh) → capacity tuning via ConfigMap patch.
+          ./deploy/install.sh --environment openshift
           ./deploy/install-llmd-infra.sh -e openshift
-          helm upgrade "$WVA_RELEASE_NAME" ./charts/workload-variant-autoscaler -n "$WVA_NS" --reuse-values \
-            --set wva.capacityScaling.default.kvCacheThreshold="${KV_CACHE_THRESHOLD}" \
-            --set wva.capacityScaling.default.queueLengthThreshold="${QUEUE_LENGTH_THRESHOLD}" \
-            --set wva.capacityScaling.default.kvSpareTrigger="${KV_SPARE_TRIGGER}" \
-            --set wva.capacityScaling.default.queueSpareTrigger="${QUEUE_SPARE_TRIGGER}"
+          kubectl patch configmap workload-variant-autoscaler-saturation-scaling-config \
```
lionelvillard, line 335: Is this still needed? If yes, let's make sure to track its removal.

.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -747,13 +747,14 @@ jobs:
...
-          ./deploy/install.sh --release-name "$WVA_RELEASE_NAME" --environment openshift
+          ./deploy/install.sh --environment openshift
           # Gateway/EPP/ModelService + WVA poolGroup alignment (see deploy/install-llmd-infra.sh).
           ./deploy/install-llmd-infra.sh -e openshift
-          # Capacity thresholds are not passed from install.sh — tune after install for CI simulators.
-          helm upgrade "$WVA_RELEASE_NAME" ./charts/workload-variant-autoscaler -n "$WVA_NS" --reuse-values \
-            --set wva.capacityScaling.default.kvSpareTrigger="${KV_SPARE_TRIGGER}" \
-            --set wva.capacityScaling.default.queueSpareTrigger="${QUEUE_SPARE_TRIGGER}"
+          # Capacity thresholds are tuned after install via ConfigMap patch for CI simulators.
+          kubectl patch configmap workload-variant-autoscaler-saturation-scaling-config \
```
lionelvillard, line 753: same comment as above.

.github/workflows/ci-pr-checks.yaml
```diff
@@ -342,8 +342,8 @@ jobs:
           # Emulated Kind: chart decode may be removed post-deploy (LLMD_REMOVE_EMULATED_DECODE_DEPLOYMENTS, default true).
           # Pin llm-d release for deterministic CI behavior
           LLM_D_RELEASE: "v0.6.0"
-          # Dual-controller smoke installs a second Helm release; chart path must be explicit in CI.
-          WVA_E2E_CHART_PATH: ${{ github.workspace }}/charts/workload-variant-autoscaler
+          # Dual-controller smoke installs a secondary controller via Kustomize; overlay path must be explicit in CI.
+          WVA_E2E_SECONDARY_OVERLAY_PATH: ${{ github.workspace }}/config/e2e/secondary-controller
```
lionelvillard: I suggest to move all e2e related YAML files under `test/e2e`. Cleaner separation IMO.

deploy/lib/infra_llmd.sh
```diff
@@ -482,13 +476,13 @@ deploy_llm_d_infrastructure() {
         detect_inference_pool_api_group
         if [ -n "$DETECTED_POOL_GROUP" ]; then
             log_info "Detected InferencePool API group: $DETECTED_POOL_GROUP; upgrading WVA to watch it (scale-from-zero)"
-            if helm upgrade "$WVA_RELEASE_NAME" ${WVA_PROJECT}/charts/workload-variant-autoscaler \
-                -n "$WVA_NS" --reuse-values \
-                --set wva.namespaceScoped="${NAMESPACE_SCOPED:-true}" \
-                --set wva.poolGroup="$DETECTED_POOL_GROUP" --wait --timeout=60s; then
-                log_success "WVA upgraded with wva.poolGroup=$DETECTED_POOL_GROUP"
+            if kubectl patch deployment workload-variant-autoscaler-controller-manager \
```
lionelvillard, line 414: same comment here, patching WVA controller should not be done here, except when testing the well-lit path.

deploy/lib/infra_wva.sh
```diff
@@ -59,44 +59,47 @@ set_wva_logging_level() {
...
+        kustomize_overlay="$(cd "$WVA_PROJECT/config/openshift" && pwd)"
+    else
+        kustomize_overlay="$(cd "$WVA_PROJECT/config/default" && pwd)"
+    fi
+
+    # Build a throw-away overlay that pins the image without modifying tracked files.
+    local tmp_overlay
+    tmp_overlay=$(mktemp -d)
+
+    cat > "$tmp_overlay/kustomization.yaml" <<EOF
```
lionelvillard, line 84: an alternative is to reuse a predefined kustomization file and inject image repo/name using `kustomize edit set image`.

deploy/lib/infra_wva.sh
```diff
@@ -59,44 +59,47 @@ set_wva_logging_level() {
...
+- $kustomize_overlay
+images:
+- name: controller
+  newName: $WVA_IMAGE_REPO
+  newTag: "$WVA_IMAGE_TAG"
+EOF
+
+    # When namespace-scoped, restrict the controller to its own namespace only.
+    if [ "$NAMESPACE_SCOPED" = "true" ]; then
+        cat >> "$tmp_overlay/kustomization.yaml" <<'PATCH'
```
lionelvillard, line 99: looks like this kustomization file does not need to be generated.

test/e2e/smoke_test.go
```diff
@@ -392,40 +391,48 @@ var _ = Describe("Smoke Tests - Infrastructure Readiness", Label("smoke", "full"
 				Expect(err).NotTo(HaveOccurred(), "Failed to create secondary workload namespace")
 			}
 
-			By("Installing secondary namespace-scoped controller via Helm")
+			By("Installing secondary namespace-scoped controller via Kustomize")
```
lionelvillard, line 395: This approach is debatable, using helm/kustomize to install a controller. I would do it in Go. Something to consider as a follow-up, not in the PR.

.github/workflows/ci-benchmark.yaml
mamy-CS: Yes, the simulator produces modest kv-cache/queue metrics so defaults don't trigger scale-up. Will track removal when simulator metrics improve

.github/workflows/ci-e2e-openshift.yaml
mamy-CS: check comment above

.github/workflows/ci-pr-checks.yaml
mamy-CS: That dir is a Kustomize overlay, not a test artifact. I think keeping it under config/ alongside the other overlays would be more consistent

deploy/lib/infra_llmd.sh
mamy-CS: Agreed, tracked as part of the install-llmd-infra.sh deprecation track

test/e2e/smoke_test.go
mamy-CS: Agree, good long term idea

deploy/lib/infra_wva.sh
mamy-CS: you are right, thanks

.github/workflows/ci-pr-checks.yaml
lionelvillard: why is it called `config/e2e` then?

lionelvillard: `config` is user-facing, e2e tests are not.

deploy/lib/infra_wva.sh
mamy-CS: Yes was debating on that but kustomize edit requires the standalone binary, which isn't currently in `REQUIRED_TOOLS `for the deploy path. Happy to do it as a follow-up when we wire that up properly. For now I think the benefit is not much. Right now the deploy path only needs kubectl for Kustomize operations

.github/workflows/ci-pr-checks.yaml
mamy-CS: good point naming could be beter, ok then moving to test/e2e/testdata/ something like that

mamy-CS:
openshift ci passes (https://github.com/llm-d/llm-d-workload-variant-autoscaler/actions/runs/25940204409/job/76255941380?pr=1150), the opeshift failure for the comment trigger path is for ok-to-test trigger always runs main's workflow. Needs to be merged for the issue trigger path to pass.

.github/workflows/ci-benchmark.yaml
lionelvillard: The patch, if needed, should be done in Go.

config/manager/manager.yaml
```diff
@@ -134,7 +133,7 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 128Mi
+            memory: 512Mi
```
lionelvillard, line 136: what triggered this change?

mamy-CS: our helm was setting 512. The kubebuilder generated had the default 128. I noticed the controller started hitting OOM under load using kustomize and I matched it with the helm memory limit.

lionelvillard (approved):

Issue #1156, opened 2026-05-18
Helm chart defaults va.accelerator to H100 (GPU model), breaking METRICSREADY on deploy


DolevAdas:
charts/workload-variant-autoscaler/values.yaml:105, charts/workload-variant-autoscaler/templates/variantautoscaling.yaml:12

Description:
The default value for va.accelerator in the Helm chart is H100:
This value is stamped directly onto the VariantAutoscaling label:
`inference.optimization/acceleratorName: {{ .Values.va.accelerator }}`

The WVA controller expects vendor names (nvidia, amd, cpu) in this label, not GPU model names like H100. As a result, any deployment using the chart defaults — or passing ACCELERATOR_TYPE=nvidia through the deploy scripts — will produce a VA with acceleratorName: H100, which the controller doesn't recognize. The VA will immediately show METRICSREADY: False and the controller logs will print "Skipping status update for VA without accelerator info".

Additionally, deploy_wva_controller() in deploy/lib/infra_wva.sh never passes va.accelerator to helm upgrade, so even if ACCELERATOR_TYPE is set in the environment, it has no effect on the VA label.

PR #1157, opened 2026-05-18
Helm chart default accelerator changed from H100 to nvidia

DolevAdas:
Issue #1156 
The default value for va.accelerator in the Helm chart is H100:
This value is stamped directly onto the VariantAutoscaling label:

The WVA controller expects vendor names (nvidia, amd, cpu) in this label, not GPU model names like H100. As a result, any deployment using the chart defaults — or passing ACCELERATOR_TYPE=nvidia through the deploy scripts — will produce a VA with acceleratorName: H100, which the controller doesn't recognize. The VA will immediately show METRICSREADY: False and the controller logs will print "Skipping status update for VA without accelerator info".

Additionally, deploy_wva_controller() in deploy/lib/infra_wva.sh never passes va.accelerator to helm upgrade, so even if ACCELERATOR_TYPE is set in the environment, it has no effect on the VA label.

Fix:
 default accelerator changed from H100 to nvidia, and the comment now documents valid values (nvidia, amd, cpu)
added ${ACCELERATOR_TYPE:+--set va.accelerator=$ACCELERATOR_TYPE} to the helm call so the environment variable is forwarded when set

### DIFF
--- a/charts/workload-variant-autoscaler/values.yaml
+++ b/charts/workload-variant-autoscaler/values.yaml
@@ -100,9 +100,9 @@ llmd:
 va:
   enabled: true
   # accelerator: Optional. If not specified, it will be auto-discovered
-  # from scale target. If specified, it will be used as fall-back value if it can't 
-  # be discovered.
-  accelerator: H100
+  # from scale target. If specified, it will be used as fall-back value if it can't
+  # be discovered. Valid values: nvidia, amd, cpu
+  accelerator: nvidia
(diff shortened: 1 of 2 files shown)


lionelvillard:
thanks @DolevAdas for your PR! 

Helm chart is deprecated in favor of Kustomize (see `config`). The VariantAutoscaling CRD is also deprecated. I'm closing this PR as it does not apply anymore.

Issue #1160, opened 2026-05-18
Makefile passes NAMESPACE= but install scripts use WVA_NS=, silently ignoring namespace override


DolevAdas:
Makefile:159, 175 vs deploy/install.sh:30, deploy/install-llmd-infra.sh:40

Description:
The deploy-wva-on-k8s and deploy-wva-on-openshift Makefile targets pass the namespace via NAMESPACE
`NAMESPACE=$(or $(NAMESPACE),workload-variant-autoscaler-system) ... ./deploy/install.sh`
But install.sh (and install-llmd-infra.sh) reads the namespace from WVA_NS:
`WVA_NS=${WVA_NS:-"workload-variant-autoscaler-system"}`

There is no mapping from NAMESPACE to WVA_NS anywhere in the deploy scripts. The NAMESPACE environment variable set by the Makefile is silently ignored, and WVA always deploys to workload-variant-autoscaler-system regardless of what the user passes.

PR #1161, opened 2026-05-18
replaced NAMESPACE with WVA_NS

DolevAdas:
Issue #1160 
Lines 158–160: deploy-wva-on-openshift — replaced NAMESPACE with WVA_NS
Lines 174–176: deploy-wva-on-k8s — same replacement
WVA_NS now flows correctly into install.sh and install-llmd-infra.sh where it's actually read.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ CLUSTER_GPUS ?= 4
 KUBECONFIG ?= $(HOME)/.kube/config
 K8S_VERSION ?= v1.32.0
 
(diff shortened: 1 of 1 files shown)


mamy-CS (changes requested):
Thank you for the pr @DolevAdas, can we align this for `undeploy-wva-on-k8s` and `undeploy-wva-on-openshift `as well. We can also add` WVA_NS` at the top of the make file as a default for clarity.

PR #1162, opened 2026-05-18
docs(proposals): add quota-based limiter design proposal (#1002)

ev-shindin:
Adds the design proposal for issue #1002 covering operator-declared per-GPU-type quota caps at cluster and namespace scope. Evaluates alternatives (classic ResourceQuota, DRA, ClusterResourceQuota, Kueue, admission webhook, MIG, HPA maxReplicas) and recommends a WVA-internal QuotaLimiter (option A) plus a complementary ResourceQuotaReader (A2) shipped as a fast-follow.

### DIFF
--- a/docs/proposals/design-1002-quota-limiter.md
+++ b/docs/proposals/design-1002-quota-limiter.md
@@ -4,10 +4,10 @@
 
 The current `TypeInventory` derives accelerator caps exclusively from cluster node discovery. Operators have no way to:
 
-- Reserve burst headroom — e.g., a cluster has 32× H100 but inference may only use 16, leaving 16 for training.
-- Divide capacity across tenants/teams — e.g., `team-a ≤ 8× H100`, `team-b ≤ 4× H100`, no cross-team borrowing.
+- Cap WVA's own usage to reserve headroom for other workloads — e.g., a cluster has 32× H100 but inference may use at most 16, leaving 16 for training. The cap *is* the reservation; there is no separate reservation primitive.
+- Divide capacity across tenants/teams, with the Kubernetes namespace as the team boundary — e.g., `team-a ≤ 8× H100`, `team-b ≤ 4× H100`, no cross-team borrowing.
 - Cap GPU types independently — e.g., `H100 ≤ 16`, `A100 unlimited`, in the same cluster.
-- Enforce contractual / chargeback / governance limits that are decoupled from physical inventory.
+- Enforce contractual / governance limits decoupled from physical inventory. (Chargeback/billing is orthogonal and out of scope — see Non-goals.)
 - Run WVA in clusters where the operator's ServiceAccount cannot list `nodes` cluster-wide (managed K8s with locked-down RBAC, multi-tenant clusters).
 
 The acceptance criteria from #1002 frame this as "operator-declared GPU caps at cluster and/or namespace scope, enforced by the WVA scaling pipeline."
@@ -19,8 +19,8 @@ This proposal evaluates the alternatives — including native Kubernetes primiti
 **Goals.**
 - Cap GPUs per accelerator type at cluster and/or namespace scope, declared by the operator.
 - Enforce the cap at scaling-decision time so WVA never *requests* more replicas than allowed.
-- Compose cleanly with the existing `TypeInventory` so physical capacity remains an upper bound (sub-issue #1003).
-- Support an explicit "unlimited" sentinel and an "exclude this namespace" affordance for system tenants.
+- Compose cleanly with the existing `TypeInventory` so physical capacity remains an upper bound (sub-issue #1003). `TypeInventory` is re-read from node discovery each cycle, so `min(physical, quota)` tracks node add/drain automatically — the quota only ever lowers the live physical bound.
+- Support an explicit "unlimited" sentinel and an "exclude this namespace" affordance for system tenants. An excluded namespace is *not capped*, but its GPU usage still counts against physical capacity (exclusion removes the cap, not the accounting), so the cluster aggregate cannot overcommit.
 - Work on any conformant Kubernetes distribution — no OpenShift-only or vendor-only dependencies.
 - Provide a clear DecisionStep trace when the cap binds.
 
@@ -31,11 +31,13 @@ This proposal evaluates the alternatives — including native Kubernetes primiti
 - Cluster autoscaler-level decisions. WVA assumes nodes already exist; provisioning is upstream.
 - Multi-cluster federation. One controller, one cluster.
 
-## 3. Alternatives considered
+## 3. Options considered
+
+Options **A** and **A2** below are the **proposed design** (recommended in §5); **B–H** were considered and are rejected or judged orthogonal.
 
 ### A. WVA-internal `QuotaLimiter` (operator-declared per-GPU-type caps)
 
-Operator-declared YAML/ConfigMap, parsed into `QuotaLimiterEntries`, enforced via a new `QuotaInventory` (implements `Inventory` + `NamespaceAwareInventory`) inside the existing limiter pipeline.
+Operator-declared YAML/ConfigMap, parsed into `QuotaLimiterEntries`, enforced via a new `QuotaInventory` (implements `Inventory` + `NamespaceAwareInventory`) inside the existing limiter pipeline. This is **not** a duplicate of `ResourceQuota`: it expresses per-GPU-*model* caps (`H100` vs `A100`) enforced at *decision* time, which `ResourceQuota` cannot (a single `requests.nvidia.com/gpu` counter, enforced at admission). Where a `ResourceQuota` already exists, WVA reads it via A2 rather than redeclaring.
 
 | Dimension | Detail |
 |---|---|
@@ -65,7 +67,7 @@ Distinct from option B below. Here WVA *actively reads* `kind: ResourceQuota` ob
 
 The key idea: A2 separates *policy* (where the cap is declared — by admins in `ResourceQuota`) from *enforcement* (where it binds — in WVA's decision loop). This sidesteps the load-bearing problem with option B (admission-time enforcement leaving WVA blind to the cap) while preserving the K8s-native UX.
 
-A2 does not solve per-GPU-model granularity — it inherits the single-counter limitation of `requests.nvidia.com/gpu`. But for the platform-admin persona who already governs the cluster with `ResourceQuota` and either runs a homogeneous GPU pool or doesn't need per-model splits, A2 is the right answer.
+A2 does not solve per-GPU-model granularity — it inherits the single-counter limitation of `requests.nvidia.com/gpu`. But for the platform-admin persona who already governs the cluster with `ResourceQuota` and either runs a homogeneous GPU pool or doesn't need per-model splits, A2 is the right answer. (#1227 implements exactly this reader's operator-free-capacity kernel; the convergence path is to land it as the A2 `ConstraintProvider` here, composing via the limiter chain, rather than as a parallel limiter that replaces node discovery.)
 
 ### B. Classic Kubernetes `ResourceQuota` alone (admin-managed, WVA-unaware)
 
@@ -84,7 +86,7 @@ spec:
 
 | Strengths | Weaknesses |
 |---|---|
-| K8s-native; familiar to every operator. | **No per-GPU-model granularity.** `nvidia.com/gpu` is a single extended-resource counter — H100 and A100 count against the same bucket. `scopeSelector` cannot filter by node label (e.g., `nvidia.com/gpu.product=H100`); the supported scope operands are limited to `PriorityClass`/`BestEffort`/`NotBestEffort`/`Terminating`/`NotTerminating`/`CrossNamespacePodAffinity`. The NVIDIA k8s-device-plugin does not advertise per-model extended resources for full GPUs ([NVIDIA/k8s-device-plugin#424](https://github.com/NVIDIA/k8s-device-plugin/issues/424), open since 2023). The AMD and Intel device plugins follow the same single-counter pattern. **MIG and time-slicing are exceptions** — `nvidia.com/mig-1g.5gb`, `nvidia.com/gpu.shared`, etc. are distinct extended resources and *can* be quota'd separately — but those are slice-size or sharing-mode discriminators, not GPU-model discriminators. |
+| K8s-native; familiar to every operator. | **No per-GPU-model granularity.** `nvidia.com/gpu` is a single extended-resource counter — H100 and A100 count against the same bucket. `scopeSelector` cannot filter by node label (e.g., `nvidia.com/gpu.product=H100`); the supported scope operands are limited to `PriorityClass`/`BestEffort`/`NotBestEffort`/`Terminating`/`NotTerminating`/`CrossNamespacePodAffinity`. The NVIDIA k8s-device-plugin does not advertise per-model extended resources for full GPUs ([NVIDIA/k8s-device-plugin#424](https://github.com/NVIDIA/k8s-device-plugin/issues/424), open since 2023 — a design for per-SKU resource naming is in progress, but the single-counter gap stands today). The AMD and Intel device plugins follow the same single-counter pattern. **MIG and time-slicing are exceptions** — `nvidia.com/mig-1g.5gb`, `nvidia.com/gpu.shared`, etc. are distinct extended resources and *can* be quota'd separately — but those are slice-size or sharing-mode discriminators, not GPU-model discriminators. |
 | Enforced by the API server, can't be bypassed. | **Per-namespace only.** No cluster-wide aggregate (see option C). |
 | Already integrates with `LimitRange` for defaults. | **Enforcement is at pod admission, not at scaling decision.** WVA would scale up, K8s would reject the new pods, WVA would see a stuck rollout. Late binding. |
 | Works on every conformant Kubernetes. | No "exclude this namespace from the cap" affordance. |
@@ -115,7 +117,7 @@ Per-model quotas drop out naturally if the admin defines a `DeviceClass` per mod
 | Strengths | Weaknesses |
 |---|---|
 | **Per-GPU-model quotas are first-class.** Authoring a DeviceClass per model is the recommended NVIDIA DRA pattern. | **Requires Kubernetes ≥ 1.34** for GA. Earlier clusters get nothing or an alpha-quality experience. WVA targets a broader version floor. |
-| K8s-native; long-term direction of the ecosystem. | **DRA quota enforcement is itself still maturing.** [KEP #4840 "Improved Quota mechanism for DRA resources"](https://github.com/kubernetes/enhancements/issues/4840) tracks unresolved admission-vs-allocation timing issues that can over-reject under contention. Targeted alpha 1.32 → beta 1.33 → stable 1.34, but the implementation is still TBD as of writing. |
+| K8s-native; long-term direction of the ecosystem. | **The *improved* DRA quota is still maturing.** The basic per-`DeviceClass` device-*count* quota (`<class>.deviceclass.resource.k8s.io/devices`) ships with DRA GA in 1.34. But the allocation-time quota — capping e.g. total GPU *memory* rather than device count, and avoiding over-rejection under contention — is [KEP #4840](https://github.com/kubernetes/enhancements/issues/4840), still in progress as of mid-2026. |
 | Same enforcement-via-admission-controller story as classic `ResourceQuota`. | **Same late-binding problem as classic ResourceQuota.** Enforced at pod admission, not at WVA's decision time. WVA would still scale up and see stuck rollouts. |
 | Works alongside the existing device plugin via the Extended-Resource ↔ DRA bridge (alpha). | Operator burden: author DeviceClasses, run a DRA driver, manage `ResourceClaim` lifecycle. |
 | | Same observability gap as B — quota violations appear as pod-admission failures, not in WVA's decision trace. |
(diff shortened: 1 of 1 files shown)


docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+**Non-goals.**
+- Job admission, queueing, or fair-share scheduling. WVA scales long-running Deployments / LWS, not batch jobs.
+- Per-pod resource accounting, billing, or chargeback ledger. Other tools own that.
+- Replacing or duplicating Kubernetes `ResourceQuota` — we will read from it if available (option A2, §3).
+- Cluster autoscaler-level decisions. WVA assumes nodes already exist; provisioning is upstream.
+- Multi-cluster federation. One controller, one cluster.
+
+## 3. Alternatives considered
+
+### A. WVA-internal `QuotaLimiter` (operator-declared per-GPU-type caps)
```
asm582, line 38: Option A is introducing the WVA-proprietary quota system instead of reuse. Why is reuse not the first choice?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+
+**Verdict:** Wrong platform binding. We want WVA to work on any conformant K8s.
+
+### D. Kueue
+
+[Kueue](https://kueue.sigs.k8s.io/) is the SIG-scheduling Kubernetes-native job queueing system. `ClusterQueue` + `LocalQueue` + cohort-based fair sharing. `ResourceFlavor` objects can key on arbitrary node labels including `nvidia.com/gpu.product`, so per-GPU-model quotas (e.g., 8× H100 and 4× A100 in one ClusterQueue) are expressible — this is the most widely deployed workaround for the B/B2 gaps today on clusters that haven't yet adopted DRA.
+
+| Strengths | Weaknesses |
+|---|---|
+| **Per-GPU-model quotas via `ResourceFlavor` work today** on any K8s version, by selecting on the device plugin's node labels. | **Wrong scaling model.** Kueue queues `Job`/`Workload` resources that run to completion. WVA continuously rebalances long-running `Deployment`/`LWS` replicas. Adapting Kueue to that flow is a significant lift. |
```
asm582: Kueue supports deployment/statefulset/LWS: https://kueue.sigs.k8s.io/docs/tasks/run/deployment/

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+| Cohort fair-share and borrowing — far richer than what #1002 asks for. | Adds a new controller, new CRDs, new admission webhook. Operational overhead. |
+| Designed for AI/ML workloads. | Doesn't fit the "WVA decides scale, K8s schedules pod" loop WVA is built on. WVA would have to translate scaling decisions into Workload submissions, then Kueue decides admission timing. Latency and observability suffer. |
+
+**Verdict:** Out of scope for #1002. Right tool for batch-style ML jobs; wrong tool for online-inference horizontal scaling. Worth acknowledging because Kueue's `ResourceFlavor` is the closest existing K8s ecosystem solution to per-GPU-type quotas — if an org already runs Kueue, they may not need #1002 at all.
```
asm582: Users are using kueue with Dynamo to serve models.

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+
+| Strengths | Weaknesses |
+|---|---|
+| Cluster-wide aggregate, label-selected. | **OpenShift-only.** Not available on vanilla K8s, GKE, EKS, AKS. |
+| Same per-GPU-type limitation as `ResourceQuota`. | Same late-binding (admission-time) enforcement. |
+| Same operator UX as `ResourceQuota`. | Same no-exclude. |
+
+**Verdict:** Wrong platform binding. We want WVA to work on any conformant K8s.
+
+### D. Kueue
```
asm582, line 143: For a cluster with a Kueue deployment, there should be a generic interface to read quotas from Kueue.

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+| Enforced by the API server, can't be bypassed. | **Per-namespace only.** No cluster-wide aggregate (see option C). |
+| Already integrates with `LimitRange` for defaults. | **Enforcement is at pod admission, not at scaling decision.** WVA would scale up, K8s would reject the new pods, WVA would see a stuck rollout. Late binding. |
+| Works on every conformant Kubernetes. | No "exclude this namespace from the cap" affordance. |
+| Operators can already use it today, no WVA change needed. | Doesn't surface in WVA's decision trace; operators see admission failures, not capped scaling decisions. |
+
+**Verdict:** Necessary but not sufficient. Useful as a K8s-layer hard ceiling for safety, but cannot express per-GPU-model caps that #1002 requires.
```
asm582, line 99: How are conflicts with the resource quota and the WVA quota handled?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
+# Issue #1002 — Quota-based limiter: design proposal
+
+## 1. Problem statement
+
+The current `TypeInventory` derives accelerator caps exclusively from cluster node discovery. Operators have no way to:
+
+- Reserve burst headroom — e.g., a cluster has 32× H100 but inference may only use 16, leaving 16 for training.
```
asm582: Reservation is mentioned here. How are reservations and quotas handled?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+- Divide capacity across tenants/teams — e.g., `team-a ≤ 8× H100`, `team-b ≤ 4× H100`, no cross-team borrowing.
```
asm582: Why would team: namespace mapping not work here? And reuse the Kubernetes namespace?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+- Cap GPU types independently — e.g., `H100 ≤ 16`, `A100 unlimited`, in the same cluster.
+- Enforce contractual / chargeback / governance limits that are decoupled from physical inventory.
```
asm582: Chargeback and quota are orthogonal to me.

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+- Enforce the cap at scaling-decision time so WVA never *requests* more replicas than allowed.
+- Compose cleanly with the existing `TypeInventory` so physical capacity remains an upper bound (sub-issue #1003).
+- Support an explicit "unlimited" sentinel and an "exclude this namespace" affordance for system tenants.
+- Work on any conformant Kubernetes distribution — no OpenShift-only or vendor-only dependencies.
+- Provide a clear DecisionStep trace when the cap binds.
+
+**Non-goals.**
+- Job admission, queueing, or fair-share scheduling. WVA scales long-running Deployments / LWS, not batch jobs.
+- Per-pod resource accounting, billing, or chargeback ledger. Other tools own that.
+- Replacing or duplicating Kubernetes `ResourceQuota` — we will read from it if available (option A2, §3).
```
asm582, line 30: Option A duplicates ResourceQuota?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+- Run WVA in clusters where the operator's ServiceAccount cannot list `nodes` cluster-wide (managed K8s with locked-down RBAC, multi-tenant clusters).
+
+The acceptance criteria from #1002 frame this as "operator-declared GPU caps at cluster and/or namespace scope, enforced by the WVA scaling pipeline."
+
+This proposal evaluates the alternatives — including native Kubernetes primitives — before recommending an implementation path.
+
+## 2. Goals and non-goals
+
+**Goals.**
+- Cap GPUs per accelerator type at cluster and/or namespace scope, declared by the operator.
```
asm582, line 20: Since Kueue is mentioned in this proposal, https://kueue.sigs.k8s.io/docs/concepts/local_queue/  and https://kueue.sigs.k8s.io/docs/concepts/cluster_queue/ will provide per-gpu-type granularity, which should work on any K8S version. Why not reuse?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+- Enforce the cap at scaling-decision time so WVA never *requests* more replicas than allowed.
+- Compose cleanly with the existing `TypeInventory` so physical capacity remains an upper bound (sub-issue #1003).
```
asm582: Physical capacity will change when nodes are added or deleted in the cluster. How will this upper bound work?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+- Support an explicit "unlimited" sentinel and an "exclude this namespace" affordance for system tenants.
```
asm582: Please clarify: when a namespace is excluded, does its GPU usage still count against the cluster aggregate quota, or is it entirely invisible to the quota system?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+- Work on any conformant Kubernetes distribution — no OpenShift-only or vendor-only dependencies.
+- Provide a clear DecisionStep trace when the cap binds.
```
asm582, line 25: Can you show an example DecisionStep trace for each binding scenario — quota binds, physical inventory binds, and both bind simultaneously?

docs/proposals/design-1002-quota-limiter.md
asm582, line 38: Why is the recommended option (A) listed under 'Alternatives Considered' rather than as the proposed solution? Can the document be restructured so the recommended design has its own section, and 'Alternatives Considered' contains only the options that were rejected?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+| Zero new operators / CRDs | ✅ | ✅ | ✅ | ❌ (DRA driver) | ✅ | ❌ | ❌ | ✅ |
+| Surfaces in WVA decision trace | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ |
+| K8s-native operator UX | ❌ (new YAML) | ✅ | ✅ | ✅ | ✅ (OCP) | partial | n/a | ✅ |
+| Cross-variant budget sharing | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | depends | ❌ |
+| Exclude/bypass list | ✅ | ❌ | ❌ | ❌ | ❌ | partial (priority classes) | depends | n/a |
+| Quota enforcement maturity | shipped | not yet | mature | **WIP — KEP #4840** | mature | mature | varies | mature |
+
+## 5. Recommendation
+
+**Adopt option A (WVA-internal `QuotaLimiter`) and option A2 (`ResourceQuotaReader`).** A is the primary resolution of #1002. A2 is a complementary mechanism filed as its own smaller issue and shipped as a fast-follow PR.
```
asm582: Can you provide a per-criterion breakdown showing exactly which of the four acceptance criteria each alternative satisfies or fails, and why rather than asserting A is the only complete solution

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+
+| Strengths | Weaknesses |
+|---|---|
+| K8s-idiomatic — admission is *the* extension point for "should this be allowed?" | New deployable, new TLS cert chain, new failure mode (webhook unavailable ⇒ scaling broken). |
+| Decouples policy from scaler. | Information asymmetry — the webhook has to re-discover quota state per-request. Slow path. |
+| Easy to layer multiple policies. | Hard to give operators a coherent error: WVA's logs say "scaled to N", webhook says "rejected", state diverges. |
+
+**Verdict:** Higher operational complexity than embedding the check in WVA, with no clear advantage when WVA is the single source of scaling decisions.
+
+### G. GPU operator-level partitioning (NVIDIA MIG, time-slicing)
```
asm582, line 171: If G is orthogonal, as mentioned in the proposal, why not remove this section?

asm582 (changes requested):
Address review

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+
+The acceptance criteria from #1002 frame this as "operator-declared GPU caps at cluster and/or namespace scope, enforced by the WVA scaling pipeline."
+
+This proposal evaluates the alternatives — including native Kubernetes primitives — before recommending an implementation path.
+
+## 2. Goals and non-goals
+
+**Goals.**
+- Cap GPUs per accelerator type at cluster and/or namespace scope, declared by the operator.
+- Enforce the cap at scaling-decision time so WVA never *requests* more replicas than allowed.
```
asm582: What does 'never requests more replicas than allowed' mean? WVA will never set spec.replicas above the quota, or that pods will never be pending due to quota exhaustion? These are different guarantees. Which one does Option A actually provide?"

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+| Enforced by the API server, can't be bypassed. | **Per-namespace only.** No cluster-wide aggregate (see option C). |
+| Already integrates with `LimitRange` for defaults. | **Enforcement is at pod admission, not at scaling decision.** WVA would scale up, K8s would reject the new pods, WVA would see a stuck rollout. Late binding. |
+| Works on every conformant Kubernetes. | No "exclude this namespace from the cap" affordance. |
+| Operators can already use it today, no WVA change needed. | Doesn't surface in WVA's decision trace; operators see admission failures, not capped scaling decisions. |
+
+**Verdict:** Necessary but not sufficient. Useful as a K8s-layer hard ceiling for safety, but cannot express per-GPU-model caps that #1002 requires.
+
+### B2. Kubernetes Dynamic Resource Allocation (DRA)
```
asm582, line 101: B2 is explicitly called out as a future-phase concern and not a viable candidate; should it be removed from Alternatives Considered and replaced with a brief note in the recommendation section?

docs/proposals/design-1002-quota-limiter.md
asm582, line 20: Can you please add concrete examples showing the quota config and the resulting behavior for cluster-only, namespace-only, and cluster+namespace declared simultaneously, including what happens when a namespace quota exceeds the cluster quota?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+| K8s-native operator UX | ❌ (new YAML) | ✅ | ✅ | ✅ | ✅ (OCP) | partial | n/a | ✅ |
+| Cross-variant budget sharing | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | depends | ❌ |
+| Exclude/bypass list | ✅ | ❌ | ❌ | ❌ | ❌ | partial (priority classes) | depends | n/a |
+| Quota enforcement maturity | shipped | not yet | mature | **WIP — KEP #4840** | mature | mature | varies | mature |
+
+## 5. Recommendation
+
+**Adopt option A (WVA-internal `QuotaLimiter`) and option A2 (`ResourceQuotaReader`).** A is the primary resolution of #1002. A2 is a complementary mechanism filed as its own smaller issue and shipped as a fast-follow PR.
+
+### Why A is the resolution of #1002
```
asm582, line 208: Does Option A update and re-read quota state between workload reconciliations within the same cycle, or does it read quota once at the start of the cycle and use that snapshot for all workload decisions?

docs/proposals/design-1002-quota-limiter.md
```diff
@@ -0,0 +1,271 @@
...
+- **F (admission webhook)** adds operational complexity without a clear advantage over an in-process limiter.
+- **G (MIG/time-slicing)** and **H (HPA `maxReplicas`)** are orthogonal — they compose with A rather than substituting for it.
+
+The load-bearing properties that make A the right primary mechanism for #1002 are:
+
+1. **Per-GPU-type granularity** on any conformant K8s version, without an OS-vendor binding or a K8s-version floor.
+2. **Decision-time enforcement** — WVA never asks for more replicas than allowed, so operators see capped scaling decisions in the controller's trace rather than `Pending` pods stuck at admission.
+3. **`DecisionStep` observability** — the limiter emits `limited by quota[scope=…, type=…]` directly into WVA's per-cycle trace, fitting the observability work in #913 / #915.
+4. **Composable** — A implements the existing `Inventory` interface, so the limiter chain (#1003) will compose it with `TypeInventory` and any future K8s-native readers via `min(physical, wva-quota, k8s-native-quota)` with no schema change.
```
asm582: Given that A and A2 use incompatible granularities per-GPU-type vs per-extended-resource, can you please add a dedicated section specifying exactly how they compose, or justify why they should not coexist in the same cluster?

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Reframed to *start simple and universal, then layer reuse* (above). Reuse is first **where a source exists** (A2 / `KueueReader`); A is the zero-dependency baseline for everywhere else. (§5.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Agreed, added `KueueReader` to §5. For the cluster a `KueueReader` serves we should read the existing `ClusterQueue`/`ResourceFlavor`, don't redeclare.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Agreed, added `KueueReader` to §5. For the cluster a `KueueReader` serves we should read the existing `ClusterQueue`/`ResourceFlavor`, don't redeclare.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Agreed, added `KueueReader` to §5. For the cluster a `KueueReader` serves we should read the existing `ClusterQueue`/`ResourceFlavor`, don't redeclare.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: No conflict — the chain takes `min(physical, wva-quota, resource-quota, …)`; the tightest binds and the `DecisionStep` names the source. (§5 pt 4.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: "Reserve" isn't a separate primitive — the cap *is* the reservation (cap inference at 16 of 32, leaving 16 for training). (§1.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Yes — the namespace *is* the team boundary, so a per-namespace cap = a per-team cap, and we reuse it. The only gap is splitting *by GPU model within* the namespace — that's what A adds. (§1.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Agreed; chargeback is a non-goal. Dropped from the motivation.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: No — `ResourceQuota` is a single `requests.nvidia.com/gpu` counter enforced at admission; A adds per-GPU-*model* caps at *decision* time, which `ResourceQuota` can't express. Where one exists, A2 reads it. (§3.A.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Reused via the reader where Kueue is present; A covers the clusters without it. Both compose.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: `TypeInventory` is re-read from node discovery each cycle, so `min(physical, quota)` tracks node add/drain automatically; the quota only lowers the live bound. (§2.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: An excluded namespace is *not capped* but its usage *still counts* against physical capacity (exclusion removes the cap, not the accounting) → the cluster aggregate can't overcommit. (§2.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Added three — quota binds, physical binds, both bind. (§5.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Restructured — §3 retitled "Options considered," A/A2 flagged as the proposed design up front; recommendation in §5. (§3.)

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Fair point — replaced the assertion with the grid.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Trimmed G (and E, same reason) to a two-line note and omitted both from the matrix. They're kept only to *disclaim the common conflation* — operators routinely ask "doesn't MIG/HNC already cap this?" — and a one-liner pre-empts it. Fine to drop entirely if you'd rather.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: The first one, now stated precisely in §2: WVA never sets a desired replica count whose cumulative per-type GPU request exceeds the cap, so **WVA's own decisions never create quota-exhaustion `Pending` pods**. It is *not* a promise of zero `Pending` from non-quota causes (another tenant, physical exhaustion below quota) — `min(physical, quota)` bounds those but can't eliminate them. Option A provides the first guarantee.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Kept as a full option with an added rationale line: it's the K8s-native per-type path reviewers always ask about and the basis for the future DRA `ConstraintProvider` in §5 (and the per-criterion grid). Dropping it to a footnote invites "why didn't you consider DRA?". The verdict already scopes it as not-the-current mechanism. Happy to shorten if you prefer.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Added **§6.3 Worked examples** — YAML for all four shapes + resulting behavior. The namespace>cluster headline: `min` always binds, so a namespace cap above the cluster cap is clamped to the cluster aggregate (the looser cap can't raise the tighter), and `ScalingPolicy` admission emits a validation warning that the namespace cap exceeds the ceiling.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: **Snapshot once at cycle start** (documented in §3.A): WVA reads the cap once per optimize cycle and decrements it in memory across models, so all models in a cycle share one consistent budget (no double-spend of the last GPU). It does **not** re-read from the API between per-model decisions mid-cycle (racy/order-dependent); the next cycle re-reads fresh. Same per-cycle snapshot discipline `TypeInventory` already uses.

docs/proposals/design-1002-quota-limiter.md
ev-shindin: Added a dedicated **"How A and A2 compose"** subsection. They're different granularities but not in conflict: A2's counter is an *aggregate ceiling* on the sum across types, A's caps are *per-type ceilings*, and both hold simultaneously via the chain (`min(physical, A[type])` per type, the per-ns sum additionally clamped by A2). Concrete: `{H100:8, A100:4}` with a `ResourceQuota` of `10` total → 8 H100 + ≤2 A100. **And per your note, admission-vs-decision is now first-class:** the reason they compose is the *enforcement layer*, not the granularity — B has the *same* single-counter granularity as A2 yet can't join the `min()` because it binds at admission, so the two enforcement points diverge. That distinction (not an original #1002 criterion) is what decides whether a cap can enter WVA's decision loop at all — the §4 "decision time" row, and the cause of the anticipated-capacity trap (§3.D).

ev-shindin:
> Address review

Thanks @asm582 — addressed all of these in the revision. Headline: I've framed it as **start simple and universal, then layer reuse** — A is the zero-dependency baseline (per-GPU-type, decision-time, any cluster), and the K8s-native readers (A2 = your #1227's kernel; a `KueueReader` next, DRA later) are additive reuse composed via `min(physical, …)`. The Kueue verdict is corrected (it manages Deployment/STS/LWS), and the surviving point is the enforcement layer — Kueue gates at admission, which triggers the anticipated-capacity trap unless WVA reads the quota and enforces at decision time.

PR #1163, opened 2026-05-18
docs(deploy): deprecate Helm chart, promote Kustomize as primary install

mamy-CS:
Mark Helm chart as deprecated and update docs and also pin kustomize image tag to v0.7.0 (current release)

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -2,7 +2,7 @@
 
 Complete guide for deploying the Workload-Variant-Autoscaler (WVA) on Kubernetes, OpenShift, and Kind clusters.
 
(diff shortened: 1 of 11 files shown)


deploy/README.md
lionelvillard, line 1: I would add a warning in top of the file saying this documentation is developer-oriented. For user-oriented doc, add a link to https://github.com/llm-d/llm-d/blob/main/guides/workload-autoscaling/README.wva.md (until we have better link)

lionelvillard (approved):
thanks @mamy-CS . Added a minor comment

deploy/README.md
mamy-CS: sounds good

Issue #1139, opened 2026-05-13
Run experiments with static replicas


asm582:
For `Model: Qwen/Qwen3-32B` the average replicas for prefill, decode, and symmetrical is less than 2 replicas. It is worth running an experiment with 2 constant replicas and reporting the performance. 

The only change required for this setup is to scale down WVA pods with the command:

`oc scale --replicas=0 <wva-pod-name> -n <wva-installed-namespace>`

After WVA pods are scale down, run make benchmark-run command on the target namespace.

kahilam:
@asm582 , I added static replica results here in.md file for comparison for prefill, decode,symmetrical.
https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1171

asm582:
@kahilam can we close this issue?

PR #1171, opened 2026-05-19
docs(benchmark): add Qwen3-32B static 2-replica results

kahilam:
## Summary

- Add benchmark results for **Qwen/Qwen3-32B** with **static 2 replicas** (WVA disabled, HPA deleted) for Prefill Heavy, Decode Heavy, and Symmetrical scenarios at 600s duration.

Ref: #1139

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -70,6 +70,25 @@ Summary of WVA benchmark runs with configuration details.
 | Avg pod startup (s) | 115 | 106 | 109 | 110 | _TBD_ |
 | Cost (avg replicas × GPU/hr) | _TBD_ | 1.77 | 1.73 | 1.73 | _TBD_ |
 
+### Prefill Heavy — Qwen/Qwen3-32B (Static 2 Replicas, 600s)
+
+**llm-d Release:** main (includes v0.7.0 WVA)
+**Model:** Qwen/Qwen3-32B
+**Workload:** 4000 prompt tokens, 1000 output tokens, 20 RPS, 600s duration
+**Setup:** WVA disabled, HPA deleted, 2 constant replicas ([#1139](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1139))
+
+| Metric | Run 1 |
+|--------|-------|
+| P99 TTFT (ms) | 534,272 |
+| P99 ITL (ms/token) | 55.96 |
+| Avg replicas | 2.00 |
+| Max replicas | 2 |
+| Avg KV cache utilization | 51.0% |
+| Avg queue depth (EPP) | 170.2 |
+| Error count | 0 |
+| Incomplete count | 511 |
+| Avg pod startup (s) | 98 |
+
 ### Prefill Heavy — Qwen/Qwen3-0.6B (600s)
 
 **llm-d Release:** v0.6.0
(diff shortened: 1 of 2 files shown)


docs/benchmark.md
```diff
@@ -70,6 +70,25 @@ Summary of WVA benchmark runs with configuration details.
...
+**Workload:** 4000 prompt tokens, 1000 output tokens, 20 RPS, 600s duration
+**Setup:** WVA disabled, HPA deleted, 2 constant replicas ([#1139](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1139))
+
+| Metric | Run 1 | Run 2 | Run 3 | Avg |
+|--------|-------|-------|-------|-----|
+| P99 TTFT (ms) | 534,272 | 546,613 | 553,154 | 544,680 |
+| P99 ITL (ms/token) | 55.96 | 58.10 | 58.13 | 57.40 |
+| Avg replicas | 2.00 | 2.00 | 2.00 | 2.00 |
+| Max replicas | 2 | 2 | 2 | 2 |
+| Avg KV cache utilization | 51.0% | 84.3% | 86.6% | 74.0% |
```
asm582: Why is the average KV cache different for static replicas?

asm582 (changes requested):
Address review

docs/benchmark.md
kahilam: > Why is the average KV cache different for static replicas?

@asm582 The ~51% KV cache is reproducible across 3 independent runs that I did today. EPP consistently routes ~85% of traffic to one pod and ~15% to the other. This appears to be the prefix-cache scorer building affinity with one pod. Is this expected EPP behavior, or should there be a rebalancing mechanism when KV cache utilization is significantly uneven across pods?

kahilam: Run 1 had both pods start at the exact same time (0s gap). When two pods cold-start simultaneously, EPP's prefix-cache scorer picks one pod for the first few requests, builds affinity to it, and keeps routing most traffic there, resulting in one pod at ~85% and the other at ~15% (averaging ~51%).

Runs 2 and 3 were executed by an **automation script t**hat does a full teardown/standup cycle before each run. The standup deploys 1 replica first, and only after standup completes (~10-14 min later) does the script scale to 2. This meant pod 1 was already warm and registered with EPP before pod 2 joined, so EPP distributed traffic evenly across both pods (~85% each).
When both pods start at the same time, EPP sends most traffic to one pod. When they start one after the other, EPP splits traffic evenly.

asm582: @kahilam, based on your description, the experiment is incorrect; we should run traffic only when both pods are ready to receive it. Can we redo this?

docs/benchmark.md
```diff
@@ -155,15 +155,13 @@ Summary of WVA benchmark runs with configuration details.
...
-| P99 ITL (ms/token) | 139.80 | 113.25 | 113.59 | 122.21 |
-| Avg replicas | 2.00 | 1.90 | 1.90 | 1.93 |
+| P99 TTFT (ms) | 357,610 | 356,292 | 355,795 | 356,566 |
+| P99 ITL (ms/token) | 113.4 | 114.9 | 113.2 | 113.8 |
+| Avg replicas | 2.00 | 2.00 | 2.00 | 2.00 |
 | Max replicas | 2 | 2 | 2 | 2 |
-| Avg KV cache utilization | 87.5% | 45.7% | 44.3% | 59.2% |
-| Avg queue depth (EPP) | 67.6 | 75.3 | 73.4 | 72.1 |
+| Avg KV cache utilization | 67.0% | 66.3% | 67.0% | 66.8% |
 | Error count | 0 | 0 | 0 | 0 |
```
asm582, line 163: Do we know why the error count is zero?

asm582 (approved):

PR #1173, opened 2026-05-19
doc only pr positive path filter

mamy-CS:


lionelvillard (approved):

lionelvillard:
using `predicate-quantifier: 'every'` may have been a better solution.

PR #1186, opened 2026-05-22
feat(deploy): refactor llm-d infra deploy

### LINKED ISSUE #1101: Deprecate install-llmd-infra script and adopt llm-d upstream installer
deploy/install-llmd-infra.sh is documented as a transitional shim (clone llm-d, helmfile for gateway/EPP/ModelService, patches, WVA poolGroup alignment). Long term, llm-d should own inference-stack installation (helmfile / CLI / official guides), and this repo should limit itself to WVA + monitoring + scaler backend plus thin integration hooks where unavoidable. This reduces duplicated logic and drift between this repo and llm-d.

Acceptance criteria
- [ ] Default CI paths do not depend on vendoring full llm-d install logic in this repo for the happy path.
- [ ] make deploy-e2e-infra / documented dev flows either call upstream install or a single maintained wrapper.
- [ ] OpenShift + Kind e2e remain green (smoke at minimum; document full coverage).
- [ ] README clearly states install order: base infra (deploy/install.sh) vs inference stack (upstream vs deprecated shim during transition).

mamy-CS:
This PR replaces the helmfile-based `install-llmd-infra.sh` with a focused `install-epp.sh` that installs only what WVA actually needs: the GAIE standalone EPP chart and its tokenreview RBAC. Everything else (model server, gateway) is the llm-d project's responsibility.

As part of this, also fixed scale-from-zero, which was silently broken, the tests were enabled but `ENABLE_SCALE_TO_ZERO` was never wired through to the WVA ConfigMap patch or the EPP flowControl feature gate, so the engine never saw any pending requests. Both are now correctly propagated end-to-end in kind CI and OpenShift CI.

Ful tests pass locally:
```
[AfterSuite] PASSED [0.057 seconds]

Ran 43 of 53 Specs in 1086.995 seconds
SUCCESS! -- 43 Passed | 0 Failed | 0 Pending | 10 Skipped
--- PASS: TestE2E (1087.00s)
PASS
ok      github.com/llm-d/llm-d-workload-variant-autoscaler/test/e2e     1088.259s

```


.github/workflows/ci-benchmark.yaml
```diff
@@ -323,15 +326,75 @@ jobs:
...
+          GAIE_VERSION: ${{ env.GAIE_VERSION }}
         run: |
-          # Split deploy: WVA base (install.sh) → llm-d (install-llmd-infra.sh) → capacity tuning via ConfigMap patch.
+          # WVA base: monitoring + controller + scaler backend.
           ./deploy/install.sh --environment openshift
-          ./deploy/install-llmd-infra.sh -e openshift
+
+          # Create llm-d namespace and HuggingFace token secret.
+          kubectl create namespace "$LLMD_NAMESPACE" --dry-run=client -o yaml | kubectl apply -f -
+          kubectl create secret generic llm-d-hf-token \
```
asm582, line 336: llmd should create this secret while deploying modelservers, reuse that?

.github/workflows/ci-benchmark.yaml
```diff
@@ -323,15 +326,75 @@ jobs:
...
+            -n "$LLMD_NAMESPACE" --dry-run=client -o yaml | kubectl apply -f -
+
+          # OpenShift: install GAIE CRDs before standalone chart.
+          kubectl apply -k "https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd/?ref=${GAIE_VERSION}" || true
+
+          # Model server via Kustomize with benchmark tuning.
+          yq ".spec.replicas = 1" -i llm-d/guides/optimized-baseline/modelserver/gpu/vllm/base/patch-vllm.yaml
+          kubectl apply -k llm-d/guides/optimized-baseline/modelserver/gpu/vllm/base -n "$LLMD_NAMESPACE"
+
+          # Post-apply: patch model ID and vLLM performance args for benchmark.
```
asm582, line 347: Why does this PR tune perf params for benchmark?

.github/workflows/ci-benchmark.yaml
```diff
@@ -323,15 +326,75 @@ jobs:
...
+            apiGroup: rbac.authorization.k8s.io
+            kind: ClusterRole
+            name: optimized-baseline-epp-tokenreview-creator
+          subjects:
+          - kind: ServiceAccount
+            name: optimized-baseline-epp
+            namespace: $LLMD_NAMESPACE
+          RBAC_EOF
+
+          # Tune capacity thresholds for benchmark scenario.
```
asm582, line 397: why is tuning required for vanilla install?

.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -695,63 +678,68 @@ jobs:
...
-          # OpenShift user-workload Prometheus (not kube-prometheus-stack in a custom namespace).
           MONITORING_NAMESPACE: openshift-user-workload-monitoring
-          # User-workload Prometheus cannot use the controller-manager SA token for /metrics bearer auth.
           WVA_METRICS_SECURE: "false"
-          # Lower saturation thresholds for simulator mode — the simulator's
-          # KV-cache and queue metrics are modest, so default thresholds
-          # (kvSpareTrigger=0.1, queueSpareTrigger=3) are too high to trigger
-          # scale-up reliably. These values trigger when kvUsage > 0.30 or
-          # queueLength > 0.5, which the simulator produces under load.
+          ENABLE_SCALE_TO_ZERO: "true"
```
asm582, line 689: Enabling scale to zero has dependency on HPA/KEDA.

.github/workflows/ci-e2e-openshift.yaml
```diff
@@ -695,63 +678,68 @@ jobs:
...
+            -n "$LLMD_NAMESPACE" --dry-run=client -o yaml | kubectl apply -f -
+
+          # OpenShift: install GAIE CRDs before standalone chart.
+          kubectl apply -k "https://github.com/kubernetes-sigs/gateway-api-inference-extension/config/crd/?ref=${GAIE_VERSION}" || true
+
+          # Model server via Kustomize (reduce replicas for CI).
+          yq ".spec.replicas = 1" -i llm-d/guides/optimized-baseline/modelserver/gpu/vllm/base/patch-vllm.yaml
+          kubectl apply -k llm-d/guides/optimized-baseline/modelserver/gpu/vllm/base -n "$LLMD_NAMESPACE"
+
+          # Post-apply: patch model ID and vLLM batch size.
```
asm582, line 724: why do we need to change batch size?

deploy/lib/epp-flow-control.values.yaml
```diff
@@ -0,0 +1,27 @@
...
+      featureGates:
+      - flowControl
+      plugins:
+      - type: queue-scorer
+      - type: kv-cache-utilization-scorer
+      - type: prefix-cache-scorer
+      - type: no-hit-lru-scorer
+      schedulingProfiles:
+      - name: default
+        plugins:
```
asm582, line 26: Where did we get the list of plugins from?

deploy/install-epp.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+# EPP infrastructure setup for all environments (kind-emulator, kubernetes, openshift).
+# Installs Gateway API CRDs, GAIE CRDs, and the GAIE standalone chart (EPP + InferencePool).
+# For llm-d model serving, see the llm-d project guides at https://github.com/llm-d/llm-d.
+#
+# Usage:
+#   LLM_D_RELEASE=v0.8.0 GAIE_VERSION=v1.5.0 LLMD_NS=llm-d-sim ./deploy/install-epp.sh
+#
+
+set -e
```
asm582, line 11: I am not sure if we need this shell script, can we inline these vars?

asm582 (changes requested):
PR is going in the right direction. Please address comments.

.github/workflows/ci-benchmark.yaml
mamy-CS: Again, not a behavioral change. There are no behavioral changes on this pr, this pr is pure deployment refactor. Previously infra_llmd.sh consumed thos params and applied the exact same kubectl patch commands on the model server deployment. Those env vars were already set in ci-benchmark.yaml. I just removed the abstraction layer, making the same patches visible inline. Benchmark has always required these params.

.github/workflows/ci-benchmark.yaml
mamy-CS: Same comment as above, no behavioural change. It is what already existed, just bubbled out from the script.

.github/workflows/ci-e2e-openshift.yaml
mamy-CS: Good point yes but already satisfied.

.github/workflows/ci-e2e-openshift.yaml
mamy-CS: Same comment as above

.github/workflows/ci-benchmark.yaml
mamy-CS: Not a behavioral change. install-llmd-infra.sh created llm-d-hf-token explicitly, same as us now. llm-d's kustomize overlays reference the secret but never create it. They expect it to be pre-created (with a real token in production, dummy for emulated CI). That design predates this PR. I just made the step inline since llmd infra is gone instead of hiding it in the script.

mamy-CS: Actually, llm-d explicitly treats llm-d-hf-token as a user prerequisite, see here [helpers/hf-token.md](https://github.com/llm-d/llm-d/blob/main/helpers/hf-token.md)

deploy/lib/epp-flow-control.values.yaml
mamy-CS: it comes verbatim from upstream, https://github.com/llm-d/llm-d/blob/main/guides/optimized-baseline/router/optimized-baseline.values.yaml
Only addition is the flow control gate

.github/workflows/ci-benchmark.yaml
asm582: > ci-benchmark.yaml.

All benchmarks are run via llm-d-benchmark. Who is using this yaml?

.github/workflows/ci-e2e-openshift.yaml
asm582: How does enabling HPA features gate on OpenShift for scale to zero? Should there be a check when one enables scaleToZero?

deploy/install-epp.sh
mamy-CS: Maybe reasonable but it is the user-facing entrypoint for epp (to sepereate concern) documented in deploy/README.md same as install.sh being the wva entry.

.github/workflows/ci-e2e-openshift.yaml
asm582: I remember you mentioned OpenShift is using llm-d-sim, are these params for sim?

mamy-CS:
> PR is going in the right direction. Please address comments.

Addressed, thanks

deploy/lib/epp-flow-control.values.yaml
asm582: While we can add, I am not sure if these make sense for llm-d-sim, we should add that comment explicitly.

deploy/install-epp.sh
asm582: Why do we care about the user-facing ePP entry point for WVA? We should just have one script that installs WVA and what WVA absolutely needs.

asm582:
Hope to get more improvements on subsequent PRs by referring to this issue: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1187

.github/workflows/ci-benchmark.yaml
mamy-CS: That's just the ci workflow that was added for benchmark. Maybe needs updating? Please review and update

.github/workflows/ci-e2e-openshift.yaml
mamy-CS: openshift ci will be turned off soon

deploy/lib/epp-flow-control.values.yaml
mamy-CS: Thanks the same plugins were always configured via optimized-baseline.values.yaml, applied whether or not it's a simulator run, comment added for clarity

.github/workflows/ci-e2e-openshift.yaml
mamy-CS: you are right there, currently that runs with keda but needs to be addressed if someone wants native hpa.

deploy/install-epp.sh
mamy-CS: That's what we had before, one big install script. epp should be separate IMO.

mamy-CS: The separation is intentional. install.sh is scoped to WVA controller + monitoring + scaler backend. EPP is llm-d infrastructure versioned against LLM_D_RELEASE and GAIE_VERSION, not WVA releases

asm582:
OpenShift test failing is a git workflow issue as informed.

/lgtm
/approve

PR #1189, opened 2026-05-24
perf: scope annotationSourcedVariants List calls to tracked namespaces

### LINKED ISSUE #1134: perf: scope annotationSourcedVariants List calls to tracked namespaces
## Problem

`annotationSourcedVariants` in `internal/utils/variant.go` issues two unfiltered `List` calls on every engine tick:

```go
k8sClient.List(ctx, &hpaList)          // fetches all HPAs cluster-wide
k8sClient.List(ctx, &soList)           // fetches all ScaledObjects cluster-wide
```

These are cache-local reads (no API round-trip via the controller-runtime informer cache), but in large clusters with many unrelated HPAs and ScaledObjects they waste CPU iterating and filtering objects that will never carry `llm-d.ai/managed: "true"`.

Identified in PR #1123 review comment: https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1123#issuecomment-4433913451

## Proposed Fix

Scope the lists to namespaces already known to contain managed objects, using `Datastore.ListTrackedNamespaces()` (already exists on the interface). Issue `client.InNamespace(ns)` per tracked namespace and union results. Fall back to cluster-wide when the tracked-namespaces list is empty (first tick before any watch event fires).

This requires threading `trackedNamespaces []string` through the call chain:

1. `annotationSourcedVariants(ctx, client, trackedNamespaces []string)`
2. `readyVariantAutoscalings(ctx, client, trackedNamespaces []string)`
3. `filterVariantsByScaleTargetAccessor(ctx, client, trackedNamespaces []string, ...)`
4. `ActiveVariantAutoscaling(ctx, client, trackedNamespaces []string)` — public API change
5. `InactiveVariantAutoscaling(ctx, client, trackedNamespaces []string)` — same
6. Engine call site (`engine.go:259`) — pass `e.ds.ListTrackedNamespaces()`, requires adding `ds datastore.Datastore` to the `Engine` struct

## Acceptance Criteria

- [ ] `annotationSourcedVariants` lists only per-namespace when tracked namespaces are known
- [ ] Falls back to cluster-wide on empty tracked-namespaces list
- [ ] `make test` passes

jia-gao:
## Summary

Resolves the `TODO(#1134)` markers introduced in #1123. `annotationSourcedVariants` ran two unfiltered cluster-wide List calls (HPAs + KEDA ScaledObjects) on every engine tick (every 30s for saturation, every 100ms for scale-from-zero), then filtered in-process by the `llm-d.ai/managed` annotation. In clusters with many unrelated HPAs and ScaledObjects this wasted CPU iterating objects that will never carry the annotation.

Closes #1134.

## Change

Scope the per-tick **HPA** List calls to namespaces tracked by `HPAReconciler`. The ScaledObject List stays cluster-wide so the late-KEDA-install case keeps working (see "What's intentionally not scoped" below).

The gate that switches scoped discovery on lives on the datastore:

- New `ResourceTypeAnnotatedHPA` / `ResourceTypeAnnotatedScaledObject` constants. `HPAReconciler` and `ScaledObjectReconciler` each record tracks under their own type, so a managed HPA and a managed ScaledObject sharing `metadata.name` in one namespace cannot collide.
- New `Datastore.MarkAnnotatedHPAsSynced()` / `AnnotatedHPANamespaces()`. The return value is tri-state:
  - `nil` → HPA cache sync not yet complete; the caller falls back to a single cluster-wide HPA list so the startup window doesn't silently drop managed HPAs in unreconciled namespaces.
  - `[]string{}` (empty non-nil) → synced, but no managed HPAs exist anywhere; the caller skips the HPA list entirely instead of cluster-wide scanning for nothing.
  - non-empty `[]string` → scope HPA discovery to these namespaces.
- A `RunnableFunc` in `main.go` runs after the HPA / ScaledObject reconcilers are registered:
  1. `mgr.GetCache().WaitForCacheSync(ctx)` — wait for informer caches to populate.
  2. `controller.BootstrapAnnotatedHPATracking(ctx, mgr.GetClient(), ds)` — enumerate the synced cache directly and pre-populate the tracker, so the gate doesn't open while the reconciler workqueue is still draining its initial events.
  3. `ds.MarkAnnotatedHPAsSynced()` — open the gate.

`utils.annotationSourcedVariants` now takes `hpaScopedNamespaces []string` with those tri-state semantics. The HPA loop iterates per-namespace scopes (or one cluster-wide scope, or zero scopes) via the new `hpaListScopes` helper. The ScaledObject loop is a single unconditional cluster-wide call.

Wiring from the engines:

- `saturation.Engine` gains a `ds datastore.Datastore` field; `NewEngine` now takes it explicitly. The `optimize()` call site passes `e.ds.AnnotatedHPANamespaces()`. `cmd/main.go` constructs the saturation engine with the same `ds` already passed to scale-from-zero.
- `scalefromzero.Engine` already holds `Datastore`; its `optimize()` call site passes `e.Datastore.AnnotatedHPANamespaces()`.

## What's intentionally not scoped (yet)

**ScaledObject List stays cluster-wide.** The ScaledObjectReconciler is only registered when KEDA's CRD is present at WVA startup. If a cluster boots WVA without KEDA and KEDA is installed later, ScaledObjects in non-HPA namespaces would never be tracked — the previous cluster-wide List discovered them via controller-runtime's lazy informer creation, and scoping the SO list would silently regress that case. A follow-up issue can add a separate sync gate for SO scoping (the SO reconciler already tracks under `ResourceTypeAnnotatedScaledObject` for forward compatibility); the helpers extend cleanly when the time comes.

## Known behaviour

Annotation-sourced HPAs in a newly tracked namespace may take up to one optimization cycle to be picked up after creation. `HPAReconciler.NamespaceTrack` runs asynchronously off the informer workqueue, so a managed HPA can briefly be visible in the informer cache while the namespace isn't yet in `AnnotatedHPANamespaces()`. The next engine tick picks it up (≤30s for the saturation loop, ≤100ms for scale-from-zero). The cluster-wide fallback would close this micro-window but at the cost of giving back the per-tick optimization on every startup; the tradeoff seemed worth flagging here so reviewers can disagree.

## Tests

Added unit tests:

- **`internal/datastore/datastore_test.go`**
  - `TestAnnotatedHPANamespacesGatedOnSync` — pre-sync returns `nil`; post-sync returns the tracked slice; idempotent re-marking.
  - `TestAnnotatedHPANamespacesEmptyAfterSync` — synced + zero managed HPAs returns `[]string{}` (non-nil) so the caller can skip the HPA list entirely instead of falling back to cluster-wide.
  - `TestAnnotatedHPANamespacesFiltersByResourceType` — VA-tracked / InferencePool-tracked / SO-tracked namespaces must not bleed into the HPA query.
  - `TestAnnotatedHPANamespacesAndScaledObjectsAreSeparate` — HPA and ScaledObject sharing `metadata.name` in one namespace are tracked independently; untracking one does not remove the namespace if the other still exists.
- **`internal/controller/annotated_scaler_bootstrap_test.go`**
  - Pre-population enumerates managed HPAs across namespaces; unmanaged and deleting HPAs are skipped; ScaledObject lists are never issued (interceptor-asserted); HPA list errors propagate.
- **`internal/utils/variant_test.go`** — existing scope tests rewritten to assert HPA scoping honours `hpaScopedNamespaces` while the ScaledObject list stays cluster-wide; added a case for the tri-state empty-non-nil path (zero HPA Lists issued, SO List still runs).
- **`internal/controller/handler_test.go`** — pre-populated tracks now use `ResourceTypeAnnotatedHPA` / `ResourceTypeAnnotatedScaledObject` directly.

## Verification

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean
- [x] `go test ./internal/...` (all packages green, including envtest-backed saturation/controller/actuator suites)

## Commit-by-commit story

For ease of review, the PR is structured as five commits that each respond to a focused concern from preliminary Codex review passes:

1. `20c1c0c` — initial scoping: thread `trackedNamespaces` through the discovery call chain.
2. `3c77b9d` — gate scoping on `mgr.GetCache().WaitForCacheSync` so the startup window doesn't drop unreconciled namespaces.
3. `4430c22` — pre-populate the tracker from the synced cache before opening the gate, since `WaitForCacheSync` doesn't guarantee the reconciler workqueue has drained.
4. `e46ee25` — filter by resource type (don't forward VA/InferencePool-tracked namespaces) and kind-qualify the tracker key (HPA vs ScaledObject can't collide).
5. `8a43ce3` — separate `ResourceTypeAnnotatedHPA` vs `ResourceTypeAnnotatedScaledObject`; scope only HPAs (SO stays cluster-wide to preserve the late-KEDA-install case); fix nil-vs-empty distinction so synced + zero HPAs skips the list entirely.

Squash-merging is fine; I kept the commits separate so the review history is legible.

## Notes for reviewers

- The CRD-sourced VA list (`readyVariantAutoscalings`) is unchanged — that path already filters via the `controller-instance` label when configured and isn't the hot loop's bottleneck.
- Out of scope (happy to file follow-ups if useful):
  - SO scoping with its own sync gate (see "What's intentionally not scoped" above).
  - Caching annotated objects on watch events to skip the per-tick List entirely.
  - Adding a `controller-runtime` predicate filter on the HPA / ScaledObject watches (Abhishek raised this in the #1123 review). The current `AnnotatedScalerPredicate` covers event filtering; the optimization opportunity is on the List path.


jia-gao:
Gentle ping — this resolves the `TODO(#1134)` markers left in #1123 (closes #1134) and has been green since late May. @lionelvillard @mamy-CS @asm582 since you were on the #1123 review — would appreciate a look when you have a cycle. Scoped change; the ScaledObject path is intentionally left cluster-wide (rationale in the description) so the late-KEDA-install case keeps working.

asm582:
Thanks for this PR, @jia-gao. Since this is a perf PR, do we have numbers on the CPU cycles saved with your changes?

jia-gao:
Good call — added a benchmark so this isn't hand-wavy (`BenchmarkAnnotationSourcedVariants` in `internal/utils/variant_bench_test.go`, just pushed). Reproduce with:

```
go test ./internal/utils/ -run '^$' -bench AnnotationSourced -benchmem
```

Same cluster, same 10 managed variants returned by both arms — only the HPA List scoping differs (`nil` gate = cluster-wide/before, tracked namespaces = scoped/after). The ScaledObject list is identical and empty in both, so the delta isolates the HPA-scoping change.

| cluster HPAs | before (cluster-wide) | after (scoped) | speedup |
|---|---|---|---|
| 100 | 301 µs, 350 KB, 809 allocs | 73 µs, 61 KB, 538 allocs | 4.1× |
| 1,000 | 3.72 ms, 3.48 MB, 4.4k allocs | 127 µs, 61 KB, 538 allocs | 29× |
| 5,000 | 13.8 ms, 19.5 MB, 20k allocs | 273 µs, 61 KB, 538 allocs | 50× |

The point isn't the ratio — it's that the scoped path is **flat** (~61 KB / 538 allocs regardless of cluster size) because per-tick cost drops from O(all HPAs in cluster) to O(managed HPAs). Since `annotationSourcedVariants` runs in the scale-from-zero loop every 100 ms, on a 5k-HPA cluster that's ~14% of a core spent here continuously (13.8 ms × 10/s) before → ~0.3% after; on a small cluster it's a modest 4×.

Caveat for fairness: the controller-runtime fake client has no namespace index, so it overcharges the scoped arm's iteration — a real informer cache makes the scoped path even cheaper. So these understate the wall-clock saving on a live cluster while reporting allocations faithfully.

PR #1190, opened 2026-05-24
feat: add wva_saturation_metrics_up freshness gauge

### LINKED ISSUE #1082: Observability: Add per-VA freshness `up` gauge for saturation metrics

jia-gao:
## Summary

Add `wva_saturation_metrics_up{variant_name, namespace}` — a per-VA freshness signal for the five PR #933 saturation/capacity gauges so dashboards and alerts can distinguish *"the system says utilization is 0.42"* from *"the analyzer was aware of this VA last cycle but produced no fresh decision so 0.42 is the stalest sample"* without relying on Prometheus' 5-minute implicit staleness marker.

Closes #1082.

## Change

In `applySaturationDecisions`, both branches of the `hasDecision` switch now emit the freshness gauge:

```go
if hasDecision {
    act.RecordSaturationMetrics(ctx, decision)
    act.RecordSaturationFreshness(ctx, decision.VariantName, decision.Namespace, true)
} else {
    act.RecordSaturationFreshness(ctx, va.Name, va.Namespace, false)
}
```

The existing TODO comment in `applySaturationDecisions` (which named #1082 as the tracker) is rewritten to describe the implemented behaviour.

- **Constant** in `internal/constants/metrics.go`: `WVASaturationMetricsUp = "wva_saturation_metrics_up"`.
- **Gauge registration** in `internal/metrics/metrics.go`: declared, initialized, registered alongside the other PR #933 gauges. Label set is the smallest cardinality common subset (`variant_name`, `namespace` + optional `controller_instance`); see *Label set rationale* below.
- **Emitter** `MetricsEmitter.RecordSaturationFreshness` in the same file.
- **Actuator wrapper** mirrors the existing `Actuator.RecordSaturationMetrics` style.

## Scope: per-VA, not controller-wide

This gauge is intentionally scoped to **per-VA staleness only**: did the analyzer process this specific variant in the most recent cycle? Controller-wide failures (`optimize()` returning before `applySaturationDecisions`, e.g. `ActiveVariantAutoscaling` errors, limited-mode inventory collection failures) are already observable via the existing `wva_optimization_duration_seconds{status="error"}` histogram. Conflating the two into one gauge would mean either:

- clearing the freshness gauge at the start of each cycle (introduces a scrape race — a healthy controller mid-cycle looks `absent()` if Prometheus scrapes between the reset and the re-emission); or
- bookkeeping a "seen variants" set and explicitly stale-marking on failure (worth its own design discussion if maintainers want it).

Keeping the gauge per-VA matches the issue body's intent (*"Setting to 1.0 whenever RecordSaturationMetrics is called … Setting to 0.0 for variants the analyzer is aware of but did not produce a fresh decision for in this cycle"*) and leaves controller-wide health to the metric already designed for it.

Recommended alert pattern:

```
# Per-VA staleness — the gauge this PR adds:
wva_saturation_metrics_up == 0

# Controller-wide failure — existing metric:
increase(wva_optimization_duration_seconds_count{status="error"}[5m]) > 0
```

## Label set rationale

`{variant_name, namespace}` — the smallest cardinality common subset across the five gauges this is meant to guard. Per the issue body:

> The `accelerator_type` and `analyzer_version` labels on individual gauges are not needed for freshness — **staleness is a per-VA property, not per-label-permutation**.

This also keeps cardinality cheap (2 labels, +1 if `CONTROLLER_INSTANCE` is set) so the gauge is safe at thousands of VAs.

One consequence worth naming explicitly: if a variant changes accelerator or analyzer mode mid-life, the *previous* `wva_saturation_utilization{accelerator_type="A100"}` series stays in Prometheus until the implicit staleness marker fires, and the freshness gauge cannot mark it stale because freshness has no `accelerator_type` label. That's a **series-retirement** concern, not a **label-alignment** concern — the right tool is the sibling `DeleteSaturationMetrics` work (per `issue-wire-delete-saturation-metrics.md` referenced in the issue) which will remove retired series at VA-deletion / label-change time. Expanding the freshness labels to mirror the saturation gauges would multiply cardinality and still wouldn't help with retired series (we don't emit freshness for the old labelset either).

## Scope explicitly NOT in this PR

Per-VA deletion of the gauge series on VA deletion is **not** wired up here. The five existing PR #933 gauges have the same leak-on-VA-deletion behaviour today; the sibling delete-metrics work referenced in the issue will introduce `DeleteSaturationMetrics` and handle all six series at once. If maintainers prefer that I bundle it into this PR, happy to expand scope — let me know.

## Tests

`internal/metrics/metrics_test.go` gains a new `Describe("RecordSaturationFreshness", …)` block covering:

- Fresh (`1.0`) and stale (`0.0`) emission for two distinct VAs.
- Fresh → stale transition on the same series (the whole point of the gauge: the value must drop to `0.0` on a stale cycle, not retain the previous `1.0`).
- Label-set minimality — only `{variant_name, namespace}` appear on the series; the test will fail loud if anyone adds a label by accident.
- `controller_instance` label injection when `CONTROLLER_INSTANCE` is set.

`resetMetrics` is updated to clear the new var so test isolation holds.

## Verification

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean
- [x] `go test ./internal/...` (all packages green, including envtest-backed saturation suites)

## Notes for reviewers

- Naming follows Prometheus' `up`-style convention. The issue lists alternatives (`wva_variant_metrics_fresh`, `wva_saturation_freshness`); happy to change if there's a project convention I missed.
- Scope is the five PR #933 gauges. If this freshness pattern proves useful, generalising to `wva_desired_replicas` etc. is a clean follow-up.

### DIFF
--- a/internal/constants/metrics.go
+++ b/internal/constants/metrics.go
@@ -186,6 +186,17 @@ const (
 	// Labels: variant_name, namespace, model_name
 	WVAKvCacheTokensCapacity = "wva_kv_cache_tokens_capacity"
 
+	// WVASaturationMetricsUp is a per-VA freshness signal for the five
+	// saturation/capacity gauges above. Set to 1.0 in cycles where the
+	// optimizer produced a fresh decision for the variant (i.e. the other
+	// gauges were just refreshed), and 0.0 in cycles where the analyzer was
+	// aware of the variant but no fresh decision was emitted. Lets
+	// dashboards distinguish "the system says utilization is X" from "the
+	// system has not updated utilization in N minutes and X is the stalest
+	// sample" without relying on Prometheus' 5-minute staleness marker.
+	// Labels: variant_name, namespace
+	WVASaturationMetricsUp = "wva_saturation_metrics_up"
+
 	// WVAPodMappingMissTotal is a counter that tracks pods whose metrics could not be
 	// attributed to a managed scaler (neither the llm-d.ai/variant label nor the
 	// pod locator resolved them). Makes the otherwise-silent skip visible.
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -71,6 +71,7 @@ var (
 	requiredCapacity      *prometheus.GaugeVec
 	kvCacheTokensUsed     *prometheus.GaugeVec
 	kvCacheTokensCapacity *prometheus.GaugeVec
+	saturationMetricsUp   *prometheus.GaugeVec
 
 	// controllerInstance stores the optional controller instance identifier.
 	// When set, it's added as a label to all emitted metrics.
@@ -111,6 +112,10 @@ func InitMetrics(registry prometheus.Registerer) error {
 	// requiredCapacityLabels: satModelLabels + "unit" to disambiguate V1 (binary 0/1)
 	// vs V2 (continuous token demand) values of the wva_required_capacity gauge.
 	requiredCapacityLabels := []string{constants.LabelVariantName, constants.LabelNamespace, constants.LabelModelName, constants.LabelUnit}
+	// satFreshnessLabels: smallest cardinality shared across the five
+	// saturation/capacity gauges. Freshness is a per-VA property, so
+	// model_name / accelerator_type / unit don't need to be on this series.
+	satFreshnessLabels := []string{constants.LabelVariantName, constants.LabelNamespace}
 
 	if controllerInstance != "" {
 		baseLabels = append(baseLabels, constants.LabelControllerInstance)
@@ -119,6 +124,7 @@ func InitMetrics(registry prometheus.Registerer) error {
 		satAccelLabels = append(satAccelLabels, constants.LabelControllerInstance)
 		satModelLabels = append(satModelLabels, constants.LabelControllerInstance)
 		requiredCapacityLabels = append(requiredCapacityLabels, constants.LabelControllerInstance)
+		satFreshnessLabels = append(satFreshnessLabels, constants.LabelControllerInstance)
 	}
 
 	replicaScalingTotal = prometheus.NewCounterVec(
@@ -190,6 +196,13 @@ func InitMetrics(registry prometheus.Registerer) error {
 			Help: "Total KV cache token capacity across all replicas of a variant (sum of vLLM TotalKvCapacityTokens).",
 		},
 		satModelLabels,
+	)
+	saturationMetricsUp = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: constants.WVASaturationMetricsUp,
+			Help: "Per-VA freshness signal for the saturation/capacity gauges: 1.0 in cycles where the optimizer produced a fresh decision for the variant, 0.0 in cycles where the analyzer was aware of the variant but did not refresh those gauges. Pairs with wva_saturation_utilization, wva_spare_capacity, wva_required_capacity, wva_kv_cache_tokens_used, and wva_kv_cache_tokens_capacity so dashboards can gate alerts on this gauge instead of relying on Prometheus' implicit staleness marker.",
+		},
+		satFreshnessLabels,
 	)
 
 	optimizationDurationLabels := []string{constants.LabelStatus}
@@ -440,6 +453,9 @@ func InitMetrics(registry prometheus.Registerer) error {
 	}
 	if err := registry.Register(kvCacheTokensCapacity); err != nil {
 		return fmt.Errorf("failed to register kvCacheTokensCapacity metric: %w", err)
+	}
+	if err := registry.Register(saturationMetricsUp); err != nil {
+		return fmt.Errorf("failed to register saturationMetricsUp metric: %w", err)
 	}
 
 	return nil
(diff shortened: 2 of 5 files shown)


jia-gao:
Rebased onto latest `main` and resolved the conflicts (`constants/metrics.go`, `saturation/engine.go`, `metrics_test.go`) — all keep-both except the engine guard: main now gates the accelerator-dimensioned saturation series on `IsAcceleratorResolved`, so I split that out and keep `wva_saturation_metrics_up` firing on `hasDecision` alone (it carries only `{variant_name, namespace}`, no `accelerator_type`). Local `build`/`vet`/`gofmt` clean and the metrics + saturation suites pass. Removing the stale label — this is active.

internal/constants/metrics.go
```diff
@@ -186,6 +186,17 @@ const (
...
+	// WVASaturationMetricsUp is a per-VA freshness signal for the five
+	// saturation/capacity gauges above. Set to 1.0 in cycles where the
+	// optimizer produced a fresh decision for the variant (i.e. the other
+	// gauges were just refreshed), and 0.0 in cycles where the analyzer was
+	// aware of the variant but no fresh decision was emitted. Lets
+	// dashboards distinguish "the system says utilization is X" from "the
+	// system has not updated utilization in N minutes and X is the stalest
+	// sample" without relying on Prometheus' 5-minute staleness marker.
+	// Labels: variant_name, namespace
+	WVASaturationMetricsUp = "wva_saturation_metrics_up"
```
ev-shindin, line 198: Prometheus uses `up` for target scrape health (injected by the server). A custom `..._up` freshness gauge is a mild convention clash — some operators read `_up` as "is the target/scrape healthy" rather than "were these gauges refreshed." It's clearly namespaced (`wva_saturation_metrics_*`), so not wrong; just consider `_fresh` / `_freshness` for less ambiguity. Minor.

internal/metrics/metrics.go
```diff
@@ -191,6 +197,13 @@ func InitMetrics(registry prometheus.Registerer) error {
 		},
 		satModelLabels,
 	)
+	saturationMetricsUp = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: constants.WVASaturationMetricsUp,
```
ev-shindin, line 202: There's already `wva_metrics_freshness_status{variant_name, status}` (constants/metrics.go) for **input** freshness — whether the scraped vLLM metrics are fresh/stale/missing. This new `wva_saturation_metrics_up` is **output** freshness — whether the optimizer refreshed the saturation/capacity gauges this cycle. They're genuinely different signals, but the names are close enough that a dashboard author will conflate them. Suggest the Help text (or docs) explicitly cross-reference and distinguish the two ("input-scrape freshness" vs "saturation-gauge refresh"), and confirm there's no overlap with the intended use of the existing metric.

internal/metrics/metrics.go
```diff
@@ -853,3 +869,36 @@ func (m *MetricsEmitter) RecordSaturationMetrics(
...
+// cycle (not gated on hasDecision), so it must tolerate tests that haven't
+// called InitMetrics — nil-guarded the same way as SetMetricsFreshnessStatus.
+// In production InitMetrics runs at startup from main.go and the guard is a
+// no-op fast path.
+func (m *MetricsEmitter) RecordSaturationFreshness(ctx context.Context, variantName, namespace string, fresh bool) {
+	if saturationMetricsUp == nil {
+		return
+	}
+	labels := prometheus.Labels{
+		constants.LabelVariantName: variantName,
```
ev-shindin, line 892: Nit: The new gauge uses `{variant_name, namespace}`; the existing `wva_metrics_freshness_status` uses `{variant_name, status}` (no namespace). Minor inconsistency between two freshness-ish metrics — not blocking, but aligning labels (or noting why they differ) helps joined dashboard queries.

ev-shindin (approved):
LGTM Clean new metric, good tests, correct to skip a reset. The one thing I'd encourage is clarifying (in Help/docs) how `wva_saturation_metrics_up` (output-gauge freshness) relates to the existing `wva_metrics_freshness_status` (input-scrape freshness) so the two aren't conflated. Approving.

jia-gao:
Thanks for the approve, @ev-shindin! This repo merges via Prow/Tide, which keys off the `/lgtm` + `/approve` label commands rather than the GitHub "Approve" button — the PR currently has neither label, so the auto-merge cron isn't picking it up. Since you're in both OWNERS lists, would you mind dropping `/lgtm` and `/approve` here when you get a sec? Should auto-merge after that. 🙏

ev-shindin:
Merging please look on my last comments (follow-up PR be welcome)

Issue #1130, opened 2026-05-13
[Autoscaling] Deprecate & Remove the VariantAutoscaling CRD


lionelvillard:
## Context

WVA currently requires operators to create a `VariantAutoscaling` (VA) CRD per model variant. The CRD duplicates information already present on the scaling target (Deployment, ScaledObject, HPA) and makes WVA a mandatory integration point.

The internal pipeline — Prometheus collection, saturation/queueing analyzers, cost-aware optimization, exposure of `wva_desired_replicas` — does **not** change. Only discovery changes.

This epic tracks the full implementation of [docs/plans/va-deprecation.md](docs/plans/va-deprecation.md) across three phases:

- **Phase 1** — Dual-mode: add annotation-based discovery without breaking existing VA CRD users.
- **Phase 2** — Deprecation: mark the VA CRD deprecated, point users at annotations (docs-only migration; no CLI).
- **Phase 3** — Removal: delete CRD manifests, reconciler, RBAC, and public API types. Keep `VariantAutoscaling` as an **internal struct** to minimize churn in the collector/analyzer/optimizer.

Watch surfaces in Phase 1: both **KEDA ScaledObject** and **HPA** (any object bearing `llm-d.ai/managed: "true"` with a `scaleTargetRef`).

---

## Phase 1 — Dual-Mode Discovery

Add annotation-based discovery that runs alongside the existing CRD reconciler. Both paths feed the same internal pipeline.

- [x] **1.1** Annotation schema package — create `internal/annotations/annotations.go` with annotation constants (`llm-d.ai/managed`, `llm-d.ai/model-id`, `llm-d.ai/variant-cost`) and a `Parse()` helper that validates required fields and returns a `SyntheticVA` struct.
- [x] **1.2** Synthesize `VariantAutoscaling` from annotations — add `internal/utils/variant_fromannotations.go` with a function that builds an in-memory `*wvav1alpha1.VariantAutoscaling` from a `ScaledObject` or `HPA`, including synthesized `OptimizationReady=true` status conditions and a tag to distinguish synthetic from CRD-sourced objects.
- [x] **1.3** Extend discovery functions — add `ActiveVariantsFromAnnotations(ctx, client)` in `internal/utils/variant.go` that lists `ScaledObjectList` and `HorizontalPodAutoscalerList`, and extend `ActiveVariantAutoscaling` / `InactiveVariantAutoscaling` to concatenate both sources with CRD winning on tie.
[...]

PR #1194, opened 2026-05-26
docs(proposals): add ScalingPolicy CRD config UX proposal

ev-shindin:
Introduces a single typed CRD as the source of truth for WVA scaling policy, consolidating configuration currently spread across four ConfigMaps, the VariantAutoscaling CRD spec, annotations, and ~25 environment variables. Complements the companion VA CRD deprecation proposal, which covers discovery but not policy.


docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
...
+
+An operator asking "what policy applies to model `granite-13b` in namespace `prod`?" must inspect a global ConfigMap, a possible namespace-local override of the same ConfigMap, a separate scale-to-zero ConfigMap, the queueing-model ConfigMap, annotations on the `ScaledObject`, fields on the VA CRD, and environment variables on the controller. No single `kubectl` command returns the effective policy.
+
+Validation timing is also inconsistent: VA CRD spec fields are admission-validated by OpenAPI; ConfigMap fields are validated by WVA on parse (errors surface in controller logs); annotations are validated by the controller after creation (no admission gate). Typos surface at different points through different channels.
```
lionelvillard: it is technically possible to validate configmap fields and annotations at admission time. This statement is not a convincing argument IMO.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
+# Proposal: ScalingPolicy CRD for WVA configuration
+
+---
+
+## Problem Statement
+
+WVA configuration is spread across six layers: four ConfigMaps (`wva-variantautoscaling-config`, `wva-saturation-scaling-config`, `wva-queueing-model-config`, `wva-model-scale-to-zero-config`), the `VariantAutoscaling` CRD spec, annotations on `ScaledObject`/`HPA`/Namespace, ~25 environment variables, Helm values, and implicit surfaces like hardcoded `ServiceMonitor` names. Each surface was added for a good reason at the time. The net result is roughly 60 distinct configuration knobs scattered across six locations, with no single way to discover, validate, or override them.
+
+An operator asking "what policy applies to model `granite-13b` in namespace `prod`?" must inspect a global ConfigMap, a possible namespace-local override of the same ConfigMap, a separate scale-to-zero ConfigMap, the queueing-model ConfigMap, annotations on the `ScaledObject`, fields on the VA CRD, and environment variables on the controller. No single `kubectl` command returns the effective policy.
```
lionelvillard: Agree, this is the real problem.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
...
+
+Discovery is namespace-owned because the operator deploying a model knows which model it is. Threshold tuning is namespace-owned because each team knows their own workload. Quota is cluster-admin-owned because allocation governance flows from above. Infrastructure is install-time because it doesn't change per-workload.
+
+The `ScalingPolicy` CRD is the only new surface this proposal introduces.
+
+---
+
+## Goals
+
+1. Single typed CRD as the source of truth for scaling policy
```
lionelvillard: this is not a goal. It's a potential solution.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
...
+
+---
+
+## Design Philosophy
+
+Three concerns, three surfaces. Each surface has one purpose, one owner, one precedence rule.
+
+| Concern | Surface | Owner |
+|---|---|---|
+| Discovery (which workloads WVA manages) and per-workload identity | Annotations on `ScaledObject`/`HPA` | Namespace owner |
```
lionelvillard: Instead of `Namespace owner` I would probably use `tenant`. Then `tenant` needs to be defined.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
...
+| Scaling policy: thresholds, analyzers, scale-to-zero, queueing-model | `ScalingPolicy` CRD (this proposal) | Namespace owner |
+| Quota: per-namespace GPU caps | `ScalingPolicy` cluster default only | Cluster admin |
+| Infrastructure: Prometheus URL, TLS, leader election, log level | Env vars + Helm values | Cluster admin (install-time) |
```
lionelvillard: Replace Helm by Kustomize

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
...
+
+Discovery is namespace-owned because the operator deploying a model knows which model it is. Threshold tuning is namespace-owned because each team knows their own workload. Quota is cluster-admin-owned because allocation governance flows from above. Infrastructure is install-time because it doesn't change per-workload.
```
lionelvillard: What's a team? A tenant? A namespace-owner? Please clarify the different personas used in this proposal.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
...
+The `ScalingPolicy` CRD is the only new surface this proposal introduces.
+
+---
+
+## Goals
+
+1. Single typed CRD as the source of truth for scaling policy
+2. Admission-time validation (OpenAPI + CEL) for every policy field
+3. Native K8s RBAC enforces the governance split between thresholds (namespace owner) and quota (cluster admin)
+4. One precedence rule across all policy fields: per-pool → per-namespace default → cluster default
```
lionelvillard: I'm not sure that's a goal. What's the use case for spreading the configuration across multiple levels?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
...
+spec:
+  poolRef:
+    name: granite-premium-pool        # InferencePool in the same namespace
+  saturation:
+    scaleUpThreshold: 0.95
+  scaleToZero:
+    enabled: false
+  priority: 2.0
+status:
+  modelID: ibm/granite-13b            # derived by controller
```
lionelvillard: I'm not convinced this is simpler than having annotation on each HPA/KEDA objects. There are several levels of indirection from HPA to ScalingPolicy, HPA -> ScaleTargetRef -> InferencePool -> ScalingPolicy, to traverse for the user to build a mental model.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+The controller discovers which `InferencePool` a workload's pods belong to by watching `InferencePool` objects and matching their `spec.selector` against the workload's pod template labels. Operators do not author the pool-to-workload mapping; it is derived from the inference-extension API.
+
+### Three-Tier Resolution
+
+For each workload, the controller resolves a policy via three deterministic lookups, in fixed order:
```
lionelvillard: I wonder if these tree-tier resolution is strictly necessary. Is it?

docs/proposals/design-config-ux.md
vivekk16, line 1: The proposal defers `spec.role` support on the assumption that per-role `InferencePools` are coming, but KServe's `LLMInferenceService` today creates **one `InferencePool`** with **two independently scaled workloads** (prefill and decode) under it, each with its own `ScalingSpec` and separate VA CR. 

Prefill is compute-bound and decode is memory-bandwidth-bound; they need different saturation thresholds, scale-to-zero behavior, and priority. 

With `poolRef` as the only match key, there's no way to express per-role policy for a single-pool P/D stack, which is how KServe ships P/D today. We definitely need an optional `spec.role` field (`prefill` | `decode`, omitted = applies to both) in the initial schema. It's a small additive field that becomes naturally redundant if per-role pools land later, but without it the CRD can't support the primary consumer's current architecture.

To make this concrete, here are the multi-stack-per-namespace cases I evaluated against the proposal:

| # | Scenario | Stacks | Pools | Covered by `poolRef`? |
|---|----------|--------|-------|-----------------------|
| 1 | **P/D disaggregation**: prefill (bursty, compute-heavy) and decode (steady, memory-bound) for the same model | 2 workloads | 1 pool | **No**: both roles share one pool, can't set different thresholds |
| 2 | **Canary deployment**: v1 (proven, tight thresholds) alongside v2 (testing, relaxed thresholds) | 2 stacks | 2 pools | Yes |
| 3 | **Shared namespace**: two teams with different SLAs deploying the same model in a namespace they can't subdivide | 2 stacks | 2 pools | Yes |
| 4 | **A/B testing**: same model with different vLLM configs (e.g. speculative decoding on/off) that change the resource profile | 2 stacks | 2 pools | Yes |

Cases 2 through 4 work because each stack has its own `InferencePool`, so `poolRef` naturally distinguishes them. Case 1 is the only one that breaks, but it's also the most important one, since P/D disaggregation is a core llm-d feature and the architecture KServe ships today.

Given that P/D disaggregation is actively shipping in KServe and is central to the llm-d value proposition, `spec.role` needs to be part of the initial schema. We can't ship a configuration CRD that has a blind spot for the most important scaling differentiation in the system.

cc @pierDipi @Gregory-Pereira @shuynh2017

docs/proposals/design-config-ux.md
ev-shindin: The `HPA → ScaleTargetRef → InferencePool → ScalingPolicy` chain only matters at author time. At read time the controller writes the resolved policy back onto the workload as annotations (`scaling.llm-d.ai/effective-policy` + the three source objects)  `kubectl describe scaledobject` (or `wva-config explain <workload>`) shows the merged result and its sources, so the user never traverses the chain manually. The annotation alternative removes that indirection but reintroduces per-workload edits for any cross-cutting change ("priority=2 for all prod models") — the Knative 30-annotation regret captured in alternative #4. Net: indirection at author time, zero at read time, and cross-cutting changes stay one edit.

docs/proposals/design-config-ux.md
ev-shindin: It's really two essential tiers + one convenience tier. The cluster default is structurally required — it's the only admission-legal home for `spec.quota` and the mandatory bootstrap object. Per-pool is the feature itself. Those two map to the "two-layer dominant shape". The middle namespace-default tier is a DRY convenience — set a value once across all a tenant's pools instead of repeating it per pool — and is strictly optional in resolution (absent → falls through to the next tier), so it adds no cost when unused. I prefer to keep this for better UX.

docs/proposals/design-config-ux.md
ev-shindin: I've added an optional `spec.role` (`prefill | decode`; omitted = applies to both) to the initial schema, matched against the workload's `llm-d.ai/role` annotation. It's additive, unblocks single-pool P/D now, and becomes naturally redundant if per-role pools land later. The per-pool uniqueness rule is now keyed on the (`poolRef.name`, `role`) tuple so prefill/decode policies can coexist for one pool, and the per-pool resolution tier matches `spec.role` against the workload's role (absent = both). Updated in the "Per-Pool Keying, Plus an Optional Per-Role Field" section.

BTW I don't think that we have support for per-role thresholds in our analytics, so we will need to add this.

cc @pierDipi @Gregory-Pereira @shuynh2017

docs/proposals/design-config-ux.md
ev-shindin: Updated (goal -> outcome) **Use case for levels**: a cluster admin sets platform-wide threshold defaults + quota once; a tenant overrides thresholds for their namespace without touching quota; an individual pool gets fine-tuned.

docs/proposals/design-config-ux.md
ev-shindin: Clarified — see the new Personas section: **cluster admin** (platform-wide, system namespace) and **tenant** (owns a namespace and its workloads).

docs/proposals/design-config-ux.md
ev-shindin: Done

docs/proposals/design-config-ux.md
ev-shindin: Updated

docs/proposals/design-config-ux.md
ev-shindin: Agree. Updated the paragraph.

docs/proposals/design-config-ux.md
ev-shindin: Updated

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Design Philosophy
+
+Three concerns, three surfaces. Each surface has one purpose, one owner, one precedence rule.
+
+### Personas
+
+This proposal uses two personas consistently:
+
+- **Cluster admin** — operates the WVA install and the system namespace; owns platform-wide defaults and quota.
```
lionelvillard, line 31: WVA supports namespaces-scoped installation. What's the impact on this proposal?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+Fields merge from cluster default → namespace default → per-pool override. Scalars: higher tier wins if set. Maps: overlay per key. Lists with a natural identifier (`spec.analyzers[*]` keyed by `name`): merge by name — lower-tier entries inherited unless the higher tier overrides them, higher-tier entries added when no lower-tier match exists. Lists without an identifier: replace wholesale.
+
+The resolved policy is published back onto the managed workload as annotations — `scaling.llm-d.ai/effective-policy` (the merged values) and `scaling.llm-d.ai/policy-sources` (the three contributing `ScalingPolicy` objects). Annotations rather than a status field, because `ScaledObject` is KEDA's CRD and `HPA` is a built-in type: a structural CRD prunes unknown status fields and a built-in type has a fixed schema, so a third-party controller cannot add `status.effectivePolicy` to either. `kubectl describe scaledobject foo` shows the annotations directly, and `wva-config explain <workload>` pretty-prints the merged policy and its sources — no manual traversal of the `HPA → … → ScalingPolicy` chain.
```
lionelvillard: Can you provide an example of an HPA object with the`scaling.llm-d.ai/effective-policy` (the merged values) and `scaling.llm-d.ai/policy-sources` (the three contributing `ScalingPolicy` objects) annotations.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
+# Proposal: ScalingPolicy CRD for WVA configuration
```
asm582, line 1: Thanks for this proposal. Can we start small and reuse an already explored pattern? We learn from llmd-router which uses https://github.com/llm-d/llm-d-router/blob/a7f83f85d087fe0f33bf49e3ef0be5ef3c4b4ba7/apix/config/v1alpha1/endpointpickerconfig_types.go EndpointPickerConfig pattern that helps configure EPP instance with plugins.

asm582 (commented):
Address review

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+The resolved policy is published back onto the managed workload as annotations — `scaling.llm-d.ai/effective-policy` (the merged values) and `scaling.llm-d.ai/policy-sources` (the three contributing `ScalingPolicy` objects). Annotations rather than a status field, because `ScaledObject` is KEDA's CRD and `HPA` is a built-in type: a structural CRD prunes unknown status fields and a built-in type has a fixed schema, so a third-party controller cannot add `status.effectivePolicy` to either. `kubectl describe scaledobject foo` shows the annotations directly, and `wva-config explain <workload>` pretty-prints the merged policy and its sources — no manual traversal of the `HPA → … → ScalingPolicy` chain.
+
+### Quota Lives Only at the Cluster Default
+
+Quota is governed top-down: cluster admins set per-namespace GPU caps; tenants cannot relax them. The CRD schema reflects this directly:
```
asm582: Quota governance and scaling policy are different intents being forced into the same kind; this violates Kubernetes API conventions.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+## Non-Goals
+
+- Replacing infrastructure config (Prometheus URL, TLS, leader election) — those stay in env vars / Kustomize
+- Replacing variant-discovery annotations from the VA CRD deprecation
+- Removing ConfigMaps in one shot — a non-breaking phased rollout is part of the plan
+- Multi-cluster federation, custom analyzer authoring UX, observability tuning — out of scope
+
+---
+
+## Proposed Solution
```
asm582, line 69: The proposal should highlight how the configuration stays in sync with the EPP and  WVA scaling policy.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+apiVersion: scaling.llm-d.ai/v1alpha1
+kind: ScalingPolicy
+metadata:
+  name: default
+  namespace: workload-variant-autoscaler-system
+spec:
+  saturation: { ... }                 # threshold defaults
+  quota:
+    default:                          # for namespaces not listed below
+      perGPU: { H100: 8 }
```
asm582: Apart from being quota a different concern, in the same API why not reuse core kubernetes quotas: https://kubernetes.io/docs/concepts/policy/resource-quotas/#types-of-resource-quota

docs/proposals/design-config-ux.md
asm582: Modifying the referred object here, violates Kubernetes API conventions

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+  namespace: production
+spec:
+  poolRef:
+    name: granite-premium-pool        # InferencePool in the same namespace
+  saturation:
+    scaleUpThreshold: 0.95
+  scaleToZero:
+    enabled: false
+  priority: 2.0
+status:
```
asm582: what are the status conditions for validity?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+---
+
+## Proposed Solution
+
+### One Namespaced CRD: `ScalingPolicy`
+
+A single namespaced CRD under `scaling.llm-d.ai/v1alpha1` carries every policy concern. The match key is `spec.poolRef` — a `LocalObjectReference` to the `InferencePool` (from the Gateway API inference extension) whose pods the policy applies to. WVA always runs alongside llm-d, which ships the inference extension, so `InferencePool` is always installed and watchable.
+
+Tenants never type model identifiers into a policy spec. The controller derives `status.modelID` from the referenced pool and exposes it via a `+kubebuilder:printcolumn` so `kubectl get scalingpolicy -n production` shows both the matched pool and the model.
```
asm582: For clarity, can we define tenants, In a Model-as-a-Service platform admin team is supposed to spawn modelservers, correct?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
+# Proposal: ScalingPolicy CRD for WVA configuration
+
+---
+
+## Problem Statement
+
+WVA configuration is spread across six layers: four ConfigMaps (`wva-variantautoscaling-config`, `wva-saturation-scaling-config`, `wva-queueing-model-config`, `wva-model-scale-to-zero-config`), the `VariantAutoscaling` CRD spec, annotations on `ScaledObject`/`HPA`/Namespace, ~25 environment variables, Kustomize overlays, and implicit surfaces like hardcoded `ServiceMonitor` names. Each surface was added for a good reason at the time. The net result is roughly 60 distinct configuration knobs scattered across six locations, with no single way to discover, validate, or override them.
```
lionelvillard: These 60 knobs have different characteristics, for instance prometheus base URL configuration parameter granularity is per controller, whereas kv cache threshold is per variant (ie. hpa). There is no desire to be able to override prometheus base URL and a strong desire to override kv cache threshold at the variant level. It is not desired to have a single way to override *everything*. 

As for validation, this is also hard to have a single way. CRD validation is done differently than environments variables validation.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+An operator asking "what policy applies to model `granite-13b` in namespace `prod`?" must inspect a global ConfigMap, a possible namespace-local override of the same ConfigMap, a separate scale-to-zero ConfigMap, the queueing-model ConfigMap, annotations on the `ScaledObject`, fields on the VA CRD, and environment variables on the controller. No single `kubectl` command returns the effective policy.
```
lionelvillard: VA CRD mentions can be removed from this proposal IMO.

lionelvillard: > No single `kubectl` command returns the effective policy

What configuration parameters fit into the "policy" category? Certainly today there is no 60 policy parameters, right?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+Validation is also uneven — though admission *timing* is no longer the real differentiator. Since `ValidatingAdmissionPolicy` went GA (K8s 1.30), ConfigMaps and annotations *can* be admission-validated in-process with CEL, no webhook needed. What a typed CRD adds over that is structure: natively typed fields (int, enum, required, defaulted), `kubectl explain` discoverability, and OpenAPI + CEL co-located on the schema. A `ValidatingAdmissionPolicy` over ConfigMap `data`/annotations validates opaque strings against a shape defined elsewhere — admission checks without the typing, discoverability, or defaulting. Absent such a policy today, ConfigMap typos surface in controller logs and annotation typos after creation.
```
lionelvillard: I'm not sure what's the problem here. This paragraph describes the various validation mechanisms provided by k8s. By "Validation is also uneven", do you mean today there is a gap and not all configuration parameters are validated?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+Anticipated growth makes this worse, not better. Issue #1002 will add another ConfigMap surface for quotas. PR #1052 already added per-analyzer thresholds. The companion VA CRD deprecation pushes *more* config into annotations. Without a deliberate shape, surface count grows monotonically.
```
lionelvillard: Adding annotations does not add surfaces. Adding a new ConfigMap does add a surface.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+This proposal introduces a single typed CRD as the source of truth for scaling policy, while leaving discovery on annotations (per the VA CRD deprecation) and infrastructure config on env vars / Kustomize. Three concerns, three surfaces, no overlap.
```
lionelvillard: > This proposal introduces a single typed CRD as the source of truth for scaling policy

This implies variant-level overrides can only be achieved by modifying the ScalingPolicy CR attached to the variant or to create one if there is no such object. Why not allowing variant-level overrides at the HPA level directly?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Context: the VariantAutoscaling CRD is being deprecated
```
lionelvillard: The configuration problem pre-exists the VA CRD deprecation (surface count, uneven validation, etc...). This context is irrelevant (IMO).

docs/proposals/design-config-ux.md
lionelvillard: > Three concerns, three surfaces, no overlap

`Scope` is probably a better word than `concern`.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+The companion proposal `docs/proposals/deprecate-va-crd.md` removes the VA CRD as the user-facing API and moves variant *discovery* + per-workload metadata to annotations on `ScaledObject`/`HPA`. That proposal addresses discovery cleanly. It does not address *policy* — thresholds, analyzer weights, scale-to-zero, queueing-model parameters, quota — which today still live in ConfigMaps. This proposal picks up where the deprecation leaves off.
+
+---
+
+## Design Philosophy
+
+Three concerns, three surfaces. Each surface has one purpose, one owner, one precedence rule.
```
lionelvillard: There are potentially four levels of scope (or concerns): controller, namespace, inferencepool (unique to llm-d), and variant.

docs/proposals/design-config-ux.md
lionelvillard: In namespace-scoped mode, controller-scoped configuration parameters are owned by the tenant, not the cluster-admin.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+| Concern | Surface | Owner |
+|---|---|---|
+| Discovery (which workloads WVA manages) and per-workload identity | Annotations on `ScaledObject`/`HPA` | Tenant |
+| Scaling policy: thresholds, analyzers, scale-to-zero, queueing-model | `ScalingPolicy` CRD (this proposal) | Tenant |
+| Quota: per-namespace GPU caps | `ScalingPolicy` cluster default only | Cluster admin |
+| Infrastructure: Prometheus URL, TLS, leader election, log level | Env vars + Kustomize overlays | Cluster admin (install-time) |
+
+Discovery is tenant-owned because the tenant deploying a model knows which model it is. Threshold tuning is tenant-owned because each tenant knows their own workload. Quota is cluster-admin-owned because allocation governance flows from above. Infrastructure is install-time because it doesn't change per-workload.
+
+The `ScalingPolicy` CRD is the only new surface this proposal introduces.
```
lionelvillard: how is this related to Personas?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+---
+
+## Goals
+
+1. A single, discoverable, admission-validated source of truth for scaling policy — one place to read the effective policy for any workload
```
lionelvillard: Did you mean to create 2 bullets? Source of truth and effective policy are two different things and this proposal relies on 2 surfaces: CRD and annotations.

docs/proposals/design-config-ux.md
lionelvillard: This approach is not generic enough IMO. An inference pool can be composed of any type of variants, P variant, D variant, E variant, cheap variant, expensive variant, throughput optimized variant, etc... Adding `spec.role` is too restrictive and not future-proof.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+kind: ScalingPolicy
+metadata:
+  name: granite-premium-priority      # name describes purpose, not identity
+  namespace: production
+spec:
+  poolRef:
+    name: granite-premium-pool        # InferencePool in the same namespace
+  saturation:
+    scaleUpThreshold: 0.95
+  scaleToZero:
```
lionelvillard, line 85: how is this different than setting minReplica=0 (now that HPAScaletoZero feature gate is GA)?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+    enabled: false
+  priority: 2.0
+status:
+  modelID: ibm/granite-13b            # derived by controller
```
lionelvillard: I don't see how is `modelID` related to scalingpolicy. Can you explain?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+Tenants never type model identifiers into a policy spec. The controller derives `status.modelID` from the referenced pool and exposes it via a `+kubebuilder:printcolumn` so `kubectl get scalingpolicy -n production` shows both the matched pool and the model.
+
+```yaml
+apiVersion: scaling.llm-d.ai/v1alpha1
+kind: ScalingPolicy
+metadata:
+  name: granite-premium-priority      # name describes purpose, not identity
+  namespace: production
+spec:
+  poolRef:
```
lionelvillard: it should be `inferencePoolRef` to be more explicit.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+Fields merge from cluster default → namespace default → per-pool override. Scalars: higher tier wins if set. Maps: overlay per key. Lists with a natural identifier (`spec.analyzers[*]` keyed by `name`): merge by name — lower-tier entries inherited unless the higher tier overrides them, higher-tier entries added when no lower-tier match exists. Lists without an identifier: replace wholesale.
+
+The resolved policy is published back onto the managed workload as annotations — `scaling.llm-d.ai/effective-policy` (the merged values) and `scaling.llm-d.ai/policy-sources` (the three contributing `ScalingPolicy` objects). Annotations rather than a status field, because `ScaledObject` is KEDA's CRD and `HPA` is a built-in type: a structural CRD prunes unknown status fields and a built-in type has a fixed schema, so a third-party controller cannot add `status.effectivePolicy` to either. `kubectl describe scaledobject foo` shows the annotations directly, and `wva-config explain <workload>` pretty-prints the merged policy and its sources — no manual traversal of the `HPA → … → ScalingPolicy` chain.
+
+### Quota Lives Only at the Cluster Default
```
lionelvillard: I would remove Quota from this proposal because: 1. the quota proposal is still a draft and 2. It is not clear that quota should be part of the scaling policy.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,305 @@
...
+
+7. **Annotated `ResourceQuota` for per-GPU-type quotas.** Tempting (K8s-native, no new CRD), but K8s admission does not interpret the annotation. Multiple `ResourceQuota` objects on the same `requests.nvidia.com/gpu` counter compound as a logical AND, giving `min(hard_i)` instead of per-type caps. Per-GPU-type quotas at admission require DRA (K8s 1.34+, KEP #4840 still maturing). WVA can additively *read* `ResourceQuota` as a complementary input in a future PR (per design-1002's A2), but not as a substitute for `spec.quota`.
+
+8. **`spec.modelID` as the match key (with `metadata.name` as a sanitized DNS-safe slug).** Earlier draft of this proposal. Rejected once we accepted llm-d-as-stack: the `InferencePool` reference is always available, exact-match, and post-transition naturally encodes role. `status.modelID` is preferable to `spec.modelID` because the model identity is derivable, not an authoritative input.
+
+9. **Deploy-tool values (Kustomize overlays / Helm) as the policy surface.** Tightly couples deploy-time and runtime; requires a redeploy per policy change.
+
+See `design-config-ux-analysis.md` for the full strengths/weaknesses breakdown of each alternative.
```
lionelvillard, line 340: where is this file?

docs/proposals/design-config-ux.md
ev-shindin: Pushing back: in WVA variants aren't policy units. The engine scales the **pool**; the optimizer distributes replicas across variants by cost-efficiency/capacity/accelerator. So cheap/expensive/throughput/hardware are *optimizer inputs*, not policy axes — per-variant policies would fight the optimizer. The one axis needing different scaling behavior behind one pool is **role** (prefill vs decode), which the optimizer can't infer — and it's temporary (per-role pools make it redundant). Made this rationale explicit in "Per-Pool Keying". cc @vivekk16

docs/proposals/design-config-ux.md
ev-shindin: True for thresholds/scale-to-zero. **Quota is the exception:** a per-namespace WVA can't enforce a cluster cap (each sees only its namespace; sum of caps can exceed cluster capacity; independent instances overcommit), and caps are a cluster-admin decision, not a tenant's. So quota is intrinsically cluster-scoped — cluster-admin-owned, system-namespace, cluster-scoped controller. Added a "Quota & Deployment Scope" section + a Personas caveat.

docs/proposals/design-config-ux.md
ev-shindin: Not an input — matching is by pool. `status.modelID` is derived from the pool and shown as a printcolumn (display-only). Clarified in the text.

docs/proposals/design-config-ux.md
ev-shindin: Done — renamed `spec.poolRef` → `spec.inferencePoolRef` throughout.

docs/proposals/design-config-ux.md
ev-shindin: Different layer: HPAScaleToZero/KEDA is the *mechanism* to reach 0; `spec.scaleToZero` is WVA's *idle policy* (no requests over a retention window) that produces the `desired-replicas=0` signal. Raw HPA scales to 0 only when the metric itself hits 0 — no "idle for N minutes" notion. Added a note.

docs/proposals/design-config-ux.md
ev-shindin: Updated

docs/proposals/design-config-ux.md
ev-shindin: Updated

docs/proposals/design-config-ux.md
ev-shindin: Mapped: controller↔cluster-default, namespace↔namespace-default, inferencepool↔per-pool. No fourth "variant" tier — WVA scales the pool and the optimizer resolves variant cost/hw/capacity itself. The only within-pool axis is **role**, a temporary bridge until per-role pools land. Stated the tiers as exactly three.

docs/proposals/design-config-ux.md
ev-shindin: Renamed to "configuration domain" (what *kind* of config), and reserved "tier" for the cluster/namespace/pool resolution levels so the two aren't conflated.

docs/proposals/design-config-ux.md
lionelvillard: > Pushing back: in WVA variants aren't policy units. The engine scales the pool;

That's not the point I'm trying to make, which is about extensibility without having to change the schema.

docs/proposals/design-config-ux.md
lionelvillard: k8s `ResourceQuota` is namespace-scoped. And we don't know yet how WVA is going to handle quota.

docs/proposals/design-config-ux.md
lionelvillard: can't WVA enable scale to zero when minReplica=0?

docs/proposals/design-config-ux.md
lionelvillard: Sorry I should have been clearer in my comment: what's the purpose of `modelId` in the status?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+- Removing ConfigMaps in one shot — a non-breaking phased rollout is part of the plan
+- Multi-cluster federation, custom analyzer authoring UX, observability tuning — out of scope
+
+---
+
+## Proposed Solution
+
+### One Namespaced CRD: `ScalingPolicy`
+
+A single namespaced CRD under `scaling.llm-d.ai/v1alpha1` carries every scaling-policy setting. The match key is `spec.inferencePoolRef` — a `LocalObjectReference` to the `InferencePool` (from the Gateway API inference extension) whose pods the policy applies to. WVA always runs alongside llm-d, which ships the inference extension, so `InferencePool` is always installed and watchable.
```
lionelvillard: > The match key is `spec.inferencePoolRef`

Have you considered the other way around, HPA -> ScalingPolicy? There is trade-off obviously. Having `inferencePoolRef` implies ScalingPolicy cannot be reused across pools, whereas HPA -> ScalingPolicy allows for defining sharable scaling profiles. On the downside, the one ScalingPolicy per pool rule is a bit harder to validate/enforce. 

> WVA always runs alongside llm-d, which ships the inference extension, so `InferencePool` is always installed and watchable.

this is a safe assumption distracting the reader from the main point of the design.

docs/proposals/design-config-ux.md
lionelvillard: Closing this as it's a distraction.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+
+Tenants never type model identifiers into a policy spec. `status.modelID` plays no part in resolution — matching is by pool. The controller derives it from the referenced pool and exposes it via a `+kubebuilder:printcolumn` (display-only) so `kubectl get scalingpolicy -n production` shows both the matched pool and the model.
```
lionelvillard: I don't see how this is related to scaling policy. This paragraph seems like a distraction than adding value to this design.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+
+> Because the model is derivable from the `InferencePool` this way, the per-workload `llm-d.ai/model-id` annotation is itself redundant — proposed for removal, with WVA reading the pool instead. That further shrinks the workload's footprint to the discovery label plus the few annotations that are *not* derivable (e.g. `variant-cost`).
```
lionelvillard: This feels out of place. Again this seems like a distraction. Looks like this could move under a new "Deprecation" section.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+```yaml
+apiVersion: scaling.llm-d.ai/v1alpha1
+kind: ScalingPolicy
+metadata:
+  name: granite-premium-priority      # name describes purpose, not identity
+  namespace: production
+spec:
+  inferencePoolRef:
+    name: granite-premium-pool        # InferencePool in the same namespace
+  saturation:
```
lionelvillard: This part lacks of details, I want to see more. That's where the scaling engine knobs are exposed. The user-facing documentation exposes concepts such as Analyzer, Optimizer, etc... Since the scaling engine is (at least there is an aspiration to make it extensible) extensible, the ScalingPolicy schema should support these extensions without having to change it.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+    scaleUpThreshold: 0.95
+  scaleToZero:
+    enabled: false
+  priority: 2.0
+status:
+  modelID: ibm/granite-13b            # derived by controller
+```
+
+The controller discovers which `InferencePool` a workload's pods belong to by watching `InferencePool` objects and matching their `spec.selector` against the workload's pod template labels. Tenants do not author the pool-to-workload mapping; it is derived from the inference-extension API.
```
lionelvillard: which controller? what is the inference-extension API ?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+
+**EPP / WVA boundary.** WVA and EPP share the `InferencePool` as their only common object: EPP owns scheduling/plugin config (`EndpointPickerConfig`), WVA owns scaling policy, and both merely *read* the pool — there is no two-way config to keep in sync. The `spec.analyzers[*]` list reuses the typed, name-keyed shape of EPP's `EndpointPickerConfig` plugin list.
```
lionelvillard: that feels out of place. 

> The `spec.analyzers[*]`

what is spec.analyzers?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+
+> **`scaleToZero` vs `minReplicas: 0`.** `HPAScaleToZero` / KEDA activation is the *mechanism* that lets a target reach zero; `spec.scaleToZero` is WVA's *idle policy* — no requests over a retention window — which produces the `wva_desired_replicas = 0` signal HPA/KEDA then act on. Raw HPA scales to zero only when the metric value itself hits zero; it has no "idle for N minutes" notion.
```
lionelvillard: This is a distraction (sorry my fault). I would remove this, does not add anything to this proposal.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Three-Tier Resolution
+
+For each workload, the controller resolves a policy via three deterministic lookups, in fixed order. These three are the **only** resolution tiers — there is no per-variant tier; variant cost/hardware/capacity differences are resolved by the optimizer, not by policy (see [Per-Pool Keying](#per-pool-keying-plus-an-optional-per-role-field)):
+
+1. **Per-pool override** — `ScalingPolicy` in the workload's namespace with `spec.inferencePoolRef.name` matching the workload's pool and `spec.role` matching the workload's role (or `spec.role` absent, which applies to both roles).
```
lionelvillard: >  and `spec.role` matching the workload's role 

If I understand this correctly it means that there are, potentially, two scaling policies per inference pool. In other words, one per variant (p and d). 

An alternative to this approach is to have one scaling policy per inference pool and to expose the *per role* configuration knobs supported by the v2 saturation analyzer (or any analyzers supporting roles). You give hints above (see `spec.analyzers[*]`, which I assume is not part of the EndpointPickerConfig) about the existence of per-analyzer knobs, so maybe that the right place.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+
+**Single configuration surface.** The aim is that, after the VA CRD deprecation, `ScalingPolicy` is the *one* place a user configures WVA scaling. The only WVA-specific things left on the `ScaledObject`/`HPA` are the discovery **label** (`llm-d.ai/managed`, a label not an annotation — per #1130) and the few per-workload annotations that aren't derivable (`variant-cost`, `role`) — not policy. (`model-id` is proposed for removal — derivable from the `InferencePool`.) New configuration, including quota's eventual surface, belongs in `ScalingPolicy`, not in another ConfigMap.
+
+---
+
+## Goals
+
+1. A single **authoring** source of truth for scaling policy — one typed `ScalingPolicy` CRD, not six surfaces
+2. A single **inspectable effective policy** per `(pool[, role])` — every field resolves through one deterministic precedence (cluster default → namespace default → per-pool), inspectable via `wva-config explain` / `kubectl get scalingpolicy`
```
lionelvillard: what is `wva-config`?

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,307 @@
...
+
+Discovery is tenant-owned because the tenant deploying a model knows which model it is. Threshold tuning is tenant-owned because each tenant knows their own workload. Quota is cluster-admin-owned because allocation governance flows from above. Infrastructure is install-time because it doesn't change per-workload.
+
+**Single configuration surface.** The aim is that, after the VA CRD deprecation, `ScalingPolicy` is the *one* place a user configures WVA scaling. The only WVA-specific things left on the `ScaledObject`/`HPA` are the discovery **label** (`llm-d.ai/managed`, a label not an annotation — per #1130) and the few per-workload annotations that aren't derivable (`variant-cost`, `role`) — not policy. (`model-id` is proposed for removal — derivable from the `InferencePool`.) New configuration, including quota's eventual surface, belongs in `ScalingPolicy`, not in another ConfigMap.
+
+---
+
+## Goals
+
+1. A single **authoring** source of truth for scaling policy — one typed `ScalingPolicy` CRD, not six surfaces
```
lionelvillard: `one typed `ScalingPolicy` CRD` is not a goal, it's a solution

docs/proposals/design-config-ux.md
lionelvillard: >  every field resolves through one deterministic precedence (cluster default → namespace default → per-pool), inspectable via `wva-config explain` / `kubectl get scalingpolicy`

this is not a goal, it's a solution

ev-shindin: A proposed CLI (`wva-config explain <pool>` / `wva-config migrate`). It was referenced before being introduced — removed from Goals; it's now defined where it's used (Proposed Solution / Migration).

docs/proposals/design-config-ux.md
ev-shindin: Updated

docs/proposals/design-config-ux.md
ev-shindin: Updated

docs/proposals/design-config-ux.md
ev-shindin: Adopted your alternative. Switched to **one `ScalingPolicy` per pool** and dropped `spec.role`. Per-role differences (saturation thresholds) are now expressed in `spec.analyzers[*].parameters` (the v2 saturation analyzer already computes per-role) — exactly your suggestion, and it rides the extensible schema from the L87 thread. The planned direction is a **separate `InferencePool` per role** (confirmed by @elevran), after which one policy per pool gives per-role for free and even the analyzer `perRole` block collapses. Uniqueness key drops to just `inferencePoolRef.name`.

docs/proposals/design-config-ux.md
ev-shindin: Removed

docs/proposals/design-config-ux.md
ev-shindin: You're right it was premature. `spec.analyzers` is now defined in the new Schema section, and the EPP/WVA-boundary note moved there (after the schema), not before it.

docs/proposals/design-config-ux.md
ev-shindin: Clarified inline: the **WVA controller**; the **Gateway API Inference Extension** (`gateway-api-inference-extension`), which defines `InferencePool`.

docs/proposals/design-config-ux.md
ev-shindin: Added a **"Schema: Scaling-Engine Configuration"** section that mirrors EPP's `EndpointPickerConfig`: the pluggable parts — `analyzers[]` and `limiters[]` — are `{type, name, parameters}` lists with `parameters` as `x-kubernetes-preserve-unknown-fields` (each plugin validates its own at load), so a new analyzer/limiter plugs in with **no CRD change**. The **optimizer is a single fixed stage**, not pluggable and not in the schema — its cost-aware/greedy mode is internal (derived from whether a limiter binds); nothing to configure now, left open for later. Stable fields (`priority`, `scaleToZero`) stay typed for `kubectl explain`/CEL; Goal 3 softened to "every *common* field."

docs/proposals/design-config-ux.md
ev-shindin: Done — moved to a new **"Relationship to the VA CRD Deprecation"** section (model-id removal + per-role-collapses-into-per-pool), out of the CRD description.

docs/proposals/design-config-ux.md
ev-shindin: Attached

docs/proposals/design-config-ux.md
ev-shindin: reworked

docs/proposals/design-config-ux.md
ev-shindin: Considered, but the match direction is deliberate. Reuse is the **namespace/cluster default tiers** — define common config once, it applies across pools; per-pool policies carry only the deltas — so sharable profiles need no per-workload reference. An `HPA → scalingPolicyRef` would put a WVA-specific field back on every workload, against the single-surface goal; and one-policy-per-pool is a trivial CEL uniqueness rule on `inferencePoolRef.name` (harder the other way). Sharing across an arbitrary *subset* of pools is the only gap — a `labelSelector` on the policy later, if it's ever needed. Made the "sharing = namespace/cluster default" point explicit in the doc.

docs/proposals/design-config-ux.md
ev-shindin: Agreed — that's display-only observability, not part of the design (and `model-id` is being removed anyway, per the VA-deprecation section). Removed the paragraph.

docs/proposals/design-config-ux.md
ev-shindin: Separate tproposal (#1162) on quota management already provided. I would like to have a single configuration surface and trying to avoid fragmrntation

docs/proposals/design-config-ux.md
ev-shindin: Kept quota on a **single surface**: declared directly on the cluster-default `ScalingPolicy` (`spec.quota`) — **no interim ConfigMap** (shipping one would reintroduce the fragmentation this proposal removes). The schema and cluster-scoped enforcement live in the dedicated quota proposal (#1162); I made the cluster-default **optional** for threshold-only installs. For the record: quota *is* required; the disagreement was the surface, not the need. The workload surface shrinks accordingly — only the discovery **label** (`llm-d.ai/managed`, a label not an annotation, per #1130) and `variant-cost` stay on HPA/KEDA; `model-id` is proposed for removal (derivable from the `InferencePool`). Quota is intrinsically cluster-scoped (a per-namespace WVA can't enforce a cluster cap; caps are a cluster-admin decision), which is why it lives at the cluster-default tier.

docs/proposals/design-config-ux.md
ev-shindin: Agreed - updated

docs/proposals/design-config-ux.md
ev-shindin: Moved to dedicated section

docs/proposals/design-config-ux.md
ev-shindin: Phase 1 already starts small (CRD alongside ConfigMaps, no removal). I'll reuse the `EndpointPickerConfig` typed, name-keyed list shape for `spec.analyzers[*]` (noted in the doc). What it doesn't cover is multi-tenant *resolution* (per-pool→namespace→cluster precedence) + ownership — hence the tiered CRD.

docs/proposals/design-config-ux.md
ev-shindin: Agreed the schema needs a separate discussion — it lives in the dedicated quota proposal (#1162). On the surface, though: quota lands **directly on the cluster-default `ScalingPolicy`**, not a standalone ConfigMap (that would be exactly the fragmentation you're flagging). Quota is a scaling *constraint* — the GPU budget the optimizer allocates within — so a single resolved surface is the right home; #1162 owns the schema and cluster-scoped enforcement.

docs/proposals/design-config-ux.md
ev-shindin: Shared anchor is the `InferencePool` — both observe the same object, no duplicated config to sync. EPP owns plugin/scheduling config; WVA owns scaling policy; both just read the pool. Added an "EPP/WVA boundary" note.

docs/proposals/design-config-ux.md
ev-shindin: Core `ResourceQuota` can't express per-GPU-*type* caps today (admission ignores annotations; multiple quotas on `requests.nvidia.com/gpu` AND to `min(hard)`, not per-type; per-type needs DRA 1.34+). WVA could *read* it later, not substitute. The core-quota-vs-`spec.quota` comparison lives in the dedicated quota proposal (#1162).

docs/proposals/design-config-ux.md
ev-shindin: "Tenant" = owner of a namespace and its model workloads, not an end customer. In MaaS that's the platform/serving team that spawns the model servers; they author workloads + scaling policy. Distinct from the **cluster admin** who owns the WVA install (and quota). Stated the MaaS mapping in Personas.

docs/proposals/design-config-ux.md
ev-shindin: Agrred - updated.

docs/proposals/design-config-ux.md
```diff
@@ -0,0 +1,343 @@
+# Proposal: ScalingPolicy CRD for WVA configuration
+
```
mamy-CS, line 2: This doc is very thorough for a proposal at this stage. The core design choice, namespaced CRD with three-tier name-based lookup, makes sense to me. The proposal mentions the solution for the authoring problem well, but I think it undersells the runtime problem. The three-tier merge rules (especially the list replace-wholesale behavior mentioned) and the implicit tier discrimination via field absence are I think, going to cause production incidents. These aren't implementation details, they're part of the UX contract. The proposal is asking operators to trust a merge algorithm it doesn't fully show.

The second weakness is forward dependencies. Several decisions lean on things that don't exist yet: spec.paused, per-role InferencePools, the quota schema in #1162. Each one is reasonable individually, but together they mean the proposal is partly a bet on futures. If any of those don't land as expected, the design has awkward residue. 

It would be good to have a merge-semantics example and the cross-object CEL uniqueness mechanism to be clarified. Overall, great direction @ev-shindin.

ev-shindin: **Thorough review — merge semantics / effective policy / cross-object uniqueness / forward deps:**
Thanks, really useful — and you caught a real bug. Three changes:
1. **Merge semantics shown.** Added a worked three-tier example (cluster ⊕ namespace ⊕ per-pool) showing scalar override, inheritance, and analyzer list-merge-by-name. Clarified the replace-wholesale rule too: the only list in the policy schema is `analyzers` (keyed by `name`), so it merges, never replaces — there's no replace-wholesale list today (the rule is a documented fallback for a future identifier-less list).
2. **Effective policy is now observable.** The controller publishes the resolved merge on `status.effectivePolicy` (+ `status.sources`), so `kubectl get scalingpolicy -o yaml` shows what's in force and which tiers produced it — operators aren't trusting an unseen algorithm.
3. **Cross-object uniqueness fixed — you're right, CEL can't do this.** `x-kubernetes-validations` is single-object. Reworked to controller-side conflict detection: deterministic winner (oldest, ties by name) + a `Conflict` status condition on the losers, no webhook. CEL now only does the single-object rules (cluster-default-only fields).
On forward dependencies: added a "Forward Dependencies" note — only per-role `InferencePool`s and the #1162 quota schema are referenced, both with graceful interim fallbacks (analyzer `parameters` for per-role; quota simply absent), so neither is load-bearing. (`spec.paused` was only an aside in the analysis doc, not a dependency — removed it.) Thanks again.

PR #1198, opened 2026-05-27
fix(collector): drop model_name filter from cache_config_info query

biranofer:
vllm:cache_config_info is an info-style metric. Its label set is derived from CacheConfig fields (num_gpu_blocks, block_size, cache_dtype, ...) plus the "engine" label — it does NOT carry model_name like vLLM's regular gauges and counters. Filtering it by model_name returns zero rows, which silently forces the saturation V2 analyzer into computeReplicaCapacityFallback (per-step batch budget × kvCacheThreshold ≈ 6.5K tokens/pod for typical vLLM defaults). For Llama-3.1-8B on H100 this under-states real KV memory capacity (~412K tokens/pod, k1 = blocks × block_size) by ~50×, causing runaway scale-up.

Drop the model_name matcher from the QueryCacheConfigInfo template and the corresponding ParamModelID from its Params list. The query is now scoped only by namespace; the collector already correlates the results to the active model's pods by instance key (the kv_cache_usage and queue_length queries remain model-scoped and discover the relevant instance set).

Add a NOTE comment on the query template documenting the rationale so a model_name filter doesn't get re-added by reflex.

Verified end-to-end: with the fix, totalSupply jumps from 13,106 (fallback) to 659,148 (real KV × kvCacheThreshold for 2 H100s) on a benchmark run that previously over-provisioned both variants to maxReplicas=10.

### DIFF
--- a/internal/collector/registration/saturation.go
+++ b/internal/collector/registration/saturation.go
@@ -53,11 +53,20 @@ func RegisterSaturationQueries(sourceRegistry *source.SourceRegistry) {
 	// Uses max to deduplicate when multiple series exist per instance with different label combinations
 	// Used by Saturation Analyzer V2 for token capacity computation
 	// Preserves instance (IP:port for multi-vLLM pods), pod (for pod lookup), llm_d_ai_variant (for direct pod-to-VA mapping), and config labels
+	//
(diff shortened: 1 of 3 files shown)


ev-shindin (approved):
Thank you @biranofer

PR #1199, opened 2026-05-27
chore: Install all resources in one namespace when running openshift E2E tests in CI

lionelvillard:
## Summary

- **Fix OpenShift E2E test failures** caused by the namespace-scoped WVA controller being unable to see \`VariantAutoscaling\` resources created by tests: collapse the two-namespace CI setup (\`LLMD_NAMESPACE\` + \`WVA_NAMESPACE\`) into a single namespace so the controller and all test-created objects share the same watched namespace.
- **Skip Gateway API / GAIE CRD installation on the shared OpenShift cluster** (CRDs are pre-installed); keep installation enabled for kind/kubernetes CI via a new \`SKIP_CLUSTER_CRDS\` flag in \`deploy/lib/infra_epp.sh\`.
- **Rename \`deploy/ci/prchecks/\` → \`deploy/ci-pr-checks/\`** for consistency with the sibling \`deploy/ci-e2e-openshift/\` directory; update all references in \`ci-pr-checks.yaml\`.

## Root cause (OpenShift E2E failures)

Failure: https://github.com/llm-d/llm-d-workload-variant-autoscaler/actions/runs/26514111690/job/78086776409

The namespace-scoped WVA controller overlay injects \`--watch-namespace=\$(POD_NAMESPACE)\`, restricting it to \`WVA_NAMESPACE\` only. The e2e tests create \`VariantAutoscaling\` resources in \`LLMD_NAMESPACE\` (a different namespace), so the controller logged "No active VariantAutoscalings found" indefinitely and the saturation V2 tests timed out after 180 s.

Fix: set \`LLMD_NAMESPACE\` to the same value as \`WVA_NAMESPACE\` in the OpenShift E2E workflow so all objects land in one namespace.

## Test plan

- [ ] Trigger `/retest` on this PR — saturation V2 and saturation analyzer path tests should pass
- [ ] Confirm controller log shows `Watching single namespace` and processes VAs created by tests
- [ ] Kind-based PR checks still install Gateway API + GAIE CRDs (default \`SKIP_CLUSTER_CRDS=false\`)
- [ ] OpenShift E2E skips CRD installation (\`SKIP_CLUSTER_CRDS=true\` passed to \`install-epp.sh\`)


asm582 (approved):
/lgtm

lionelvillard:
Bypassing merging rules as kind e2e suite has succeeded here: https://github.com/llm-d/llm-d-workload-variant-autoscaler/actions/runs/26532347721

PR #1196, opened 2026-05-26
chore: move most ci script to files

lionelvillard:
## Summary

Bash scripts embedded directly in GitHub Actions YAML cannot be modified by fork contributors without touching the workflow file itself — which forks cannot push to. Moving them to files in the repo makes them accessible and editable on forks, aligned with the pattern already used in `deploy/`.

- Extracted **2 scripts** from `ci-pr-checks.yaml` → `deploy/ci/prchecks/` (`install-kind.sh`, `build-wva-image-local.sh`, both shared between smoke and full jobs)
- Extracted **19 scripts** from `ci-e2e-openshift.yaml` → `deploy/ci-e2e-openshift/` (tool installation, GPU checks, deployment, verification, cleanup, etc.)

**Rule applied:** only `run:` blocks that do not contain `${{ }}` expressions directly in the shell code were extracted — steps that reference GitHub Actions context variables inline (e.g. `set-output`, `Validate PR head SHA`, `Post GPU status to PR`) are kept as-is. Single-command trivials (`make build`, `make test`, etc.) are also kept inline.

## Test plan

- [ ] CI passes on this PR (smoke e2e)
- [ ] `bash -n deploy/ci/prchecks/*.sh deploy/ci-e2e-openshift/*.sh` — all scripts pass syntax check
- [ ] Remaining inline `run: |` blocks in both workflows are intentional (context-var steps or single-command trivials)

PR #1204, opened 2026-05-27
feat: always run full e2e test on main repo PR

lionelvillard:
## Problem

For `issue_comment` events (the `/ok-to-test` trigger), GitHub Actions always loads the workflow file from the **default branch** — not from the PR branch. This is a GitHub security constraint that applies even for same-repo (non-fork) PRs.

This caused two issues:
1. Script paths in the checked-out PR code could diverge from paths referenced in the main-branch workflow (e.g. after #1196 renamed `deploy/ci-pr-checks/` → `deploy/ci/prchecks/`), breaking runs silently.
2. The PR head SHA captured at comment time could be stale if the branch was force-pushed before the run completed.

## Solution

For same-repo `pull_request` events, the workflow file is loaded from the PR branch itself, so script paths always match. Run the full Kind E2E suite automatically on every same-repo PR push, without requiring `/ok-to-test`.

Fork PRs are unchanged — they still require `/ok-to-test` from a maintainer.

## Changes

**`.github/workflows/ci-pr-checks.yaml`**

- `check-full-tests` job: added an early-return branch for `pull_request` events that outputs `run_full: true` for same-repo branches and `false` for forks.
- `e2e-tests-full` job condition: added `pull_request` to the allowed event names; added `needs.lint-and-test.result != 'failure'` guard so full E2E is skipped when lint/unit tests fail.

## Test plan

- [ ] Push a commit to a same-repo PR branch → `e2e-tests-full` runs automatically without `/ok-to-test`
- [ ] Fork PR still requires `/ok-to-test` to trigger full E2E
- [ ] Lint/test failure on a same-repo PR skips full E2E

### DIFF
--- a/.github/workflows/ci-pr-checks.yaml
+++ b/.github/workflows/ci-pr-checks.yaml
@@ -201,6 +201,14 @@ jobs:
               }
             }
 
+            // Same-repo pull_request: always run full tests
+            if (context.eventName === 'pull_request') {
+              const baseRepo = `${context.repo.owner}/${context.repo.repo}`;
+              const headRepo = context.payload.pull_request.head.repo?.full_name ?? baseRepo;
+              core.setOutput('run_full', headRepo === baseRepo ? 'true' : 'false');
+              return;
+            }
+
             // Check workflow_dispatch input
             const workflowInput = '${{ github.event.inputs.run_full_tests }}';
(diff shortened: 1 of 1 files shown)


asm582 (approved):
/lgtm

PR #1206, opened 2026-05-27
feat(config): add namespace-scoped/kubernetes overlay (1/4)

lionelvillard:
## Summary
- Adds `config/overlays/namespace-scoped/kubernetes/kustomization.yaml`, mirroring the existing `namespace-scoped/openshift` overlay without OpenShift-specific patches (Thanos querier, Service-CA, user-workload-monitoring binding).
- Wires the new overlay into the `kustomize-build` CI check in `.github/workflows/ci-pr-checks.yaml`.

This is **PR 1 of 4** for #1205. Subsequent PRs will:
- PR 2: extract dual-controller test into its own file, repoint `testdata/secondary-controller` to this overlay, add `test-e2e-multi-controller[-with-setup]` Make targets.
- PR 3: stub `ci-e2e-multi-controller.yaml` workflow so branch-protection can reference the job name.
- PR 4: wire the full implementation into the workflow and remove the RBAC patch-and-restore workarounds.

## Test plan
- [x] `bin/kustomize build config/overlays/namespace-scoped/kubernetes` succeeds locally
- [x] Output contains the expected `RoleBinding`s from the namespace-scoped component (no shared `ClusterRoleBinding` collisions when used alongside the cluster-scoped overlay)
- [ ] CI `kustomize-build` job passes

Refs #1205

Issue #1205, opened 2026-05-27
Dual-controller smoke test runs in wrong environment: add dedicated namespace-scoped kind CI job


lionelvillard:
## Problem

The `"Dual namespace-scoped controllers isolation"` context in `test/e2e/smoke_test.go` currently runs as part of the smoke suite inside the `e2e-tests-smoke` CI job, which deploys the primary WVA controller **cluster-scoped** via `config/overlays/cluster-scoped/kubernetes`.

When the test then installs a secondary controller via the `test/e2e/testdata/secondary-controller` overlay (which also uses `config/overlays/cluster-scoped/kubernetes` as its base), both kustomize overlays share `namePrefix: wva-`. This causes three cluster-scoped resources to be silently overwritten:
- `wva-manager-rolebinding` (ClusterRoleBinding)
- `wva-epp-metrics-reader-role-binding` (ClusterRoleBinding)
- `wva-metrics-auth-rolebinding` (ClusterRoleBinding)

The test works around this with fragile kubectl patch-and-restore commands in `BeforeAll`, acknowledged by a TODO comment in the code:

```go
// TODO: run dual-controller isolation in a dedicated fresh cluster rather than layering
// a namespace-scoped secondary controller on top of an existing cluster-scoped primary.
// The two modes are mutually exclusive by design: cluster-scoped ClusterRoleBindings
// (metrics-auth-rolebinding, manager-rolebinding, epp-metrics-reader-role-binding) are
// shared resources and get overwritten by each kustomize apply, requiring fragile
// patch-and-restore workarounds. A proper fix is a separate Kind cluster per scenario.
```

## Desired End State

A dedicated CI job (`e2e-multi-controller`) that:
1. Spins up a fresh kind cluster
2. Deploys **both** WVA controllers namespace-scoped from the start using a new `config/overlays/namespace-scoped/kubernetes` overlay
3. Runs only the multi-controller tests (extracted from `smoke_test.go` into their own file `test/e2e/multi_controller_test.go`)
4. Eliminates the RBAC patch-and-restore workarounds entirely — no shared ClusterRoleBinding names means no conflict

## Multi-PR plan

| PR | Scope |
|----|-------|
| PR 1 | Add `config/overlays/namespace-scoped/kubernetes/kustomization.yaml` (currently only openshift has a namespace-scoped overlay); update `kustomize-build` CI check |
[...]

PR #1207, opened 2026-05-27
test(e2e): extract dual-controller suite into multi_controller_test.go (2/4)

lionelvillard:
## Summary
- Moves the `Dual namespace-scoped controllers isolation` Context from `test/e2e/smoke_test.go` into a new `test/e2e/multi_controller_test.go` with `Label("multi-controller")`.
- Repoints `test/e2e/testdata/secondary-controller/kustomization.yaml` at the `namespace-scoped/kubernetes` overlay introduced in PR 1. The `--watch-namespace` patch is dropped because the `namespace-scoped` component already injects it.
- Adds `make test-e2e-multi-controller` and `make test-e2e-multi-controller-with-setup` targets.

The patch-and-restore RBAC workaround stays in place; PR 4 removes it once the dedicated CI job runs this suite in a fresh kind cluster.

This is **PR 2 of 4** for #1205. **Stacked on PR 1** (#1206) — the diff against `main` will include PR 1's commits until that one merges. Please review/merge #1206 first.

## Test plan
- [x] `go build ./...` succeeds
- [x] `bin/kustomize build test/e2e/testdata/secondary-controller` succeeds and produces a deployment with the watch-namespace arg from the namespace-scoped component
- [ ] `make test-e2e-smoke` runs without the multi-controller block
- [ ] `make test-e2e-multi-controller` (with infra deployed) runs the moved Context

Refs #1205


mamy-CS (approved):
/lgtm

PR #1145, opened 2026-05-14
Adding Kubernetes labels to Associated Prometheus Metrics and Pods to VA

dumb0002:
This PR replaces the complex ownership traversal approach (Pod → ReplicaSet → Deployment → VA) with a simple label-based mapping system. The label `llmd.ai/variant` is now required in the pod templates of the `ScaleTargetObjects`, which is propagated into Prometheus metrics and used for direct pod-to-VariantAutoscaling mapping.

### Motivation
The previous approach required 2-3 Kubernetes API calls per pod per metrics collection cycle to traverse the ownership chain. This was slow, complex, and error-prone. The new label-based approach eliminates all these API calls and simplifies the codebase significantly.

#### 1. Label Infrastructure
- **Added constants** (`internal/constants/labels.go`):
  - `VariantLabelKey = "llm-d.ai/variant"` - Kubernetes pod label
  - `VariantLabelPrometheusKey = "llm_d_ai_variant"` - Prometheus metric label

#### 2.  Removed ownership traversal
  - Deleted `internal/collector/source/pod_va_mapper.go` and its test 
  - Removed the podVAMapper field from ReplicaMetricsCollector; NewReplicaMetricsCollector no longer instantiates it. 

#### 3.  Kustomize ServiceMonitor (config/prometheus/vllm-servicemonitor.yaml)
  - Added a new kustomize-native ServiceMonitor for vLLM pods (the Helm chart is deprecated)
  - Wired into config/prometheus/kustomization.yaml
  - Contains the relabelings rule to propagate llm-d.ai/variant → llm_d_ai_variant
  
#### 4. Metrics Propagation
Updated ServiceMonitor configurations to propagate pod labels into Prometheus metrics:
- **Modified** `charts/workload-variant-autoscaler/templates/vllm-servicemonitor.yaml`
- **Modified** `test/e2e/fixtures/infra_builder.go`
- **Modified** `test/utils/resources/llmdsim.go`

Added `metricRelabelings` configuration:
```yaml
metricRelabelings:
- sourceLabels: [__meta_kubernetes_pod_label_llmd_ai_variant]
  targetLabel: llm_d_ai_variant
  action: replace
```

#### 5. PromQL Query Updates
Updated all PromQL queries to preserve the variant label in grouping clauses:
- **Modified** `internal/collector/registration/saturation.go` (6 queries)
- **Modified** `internal/collector/registration/throughput_analyzer.go` (3 queries)
[...]

Issue #1072, opened 2026-04-29
metrics: use Kubernetes labels to associate metrics to variants instead of pod_name traversal


lionelvillard:
## Summary

WVA currently maps Prometheus metrics to variants by extracting pod names from metric labels and traversing a multi-hop Kubernetes ownership chain (pod → ReplicaSet → Deployment/LWS → VariantAutoscaling). This is brittle. Instead, WVA should scope metrics to an InferencePool using the pool's existing label selector, and distinguish between variants using a new WVA-specific Kubernetes label stamped on pods.

## Current Behavior

When collecting replica metrics, WVA:

1. Extracts pod identity from Prometheus metric labels using a `pod` → `pod_name` fallback (see `internal/collector/replica_metrics.go`, lines ~153-377, repeated ~8 times across metric types).
2. Performs a 4-hop ownership chain traversal to map pod -> ReplicaSet -> Deployment/LWS -> VariantAutoscaling (`internal/collector/source/pod_va_mapper.go`, lines 73-129).
3. Silently drops the metric if any hop fails.

The existing warning in `internal/collector/replica_metrics.go` (line 496) already signals this fragility:

> "Pod has vLLM metrics but no dispatch rate — possible pod/pod_name label mismatch"

### Why this is brittle

- **Label inconsistency**: vLLM metrics use `pod`, scheduler metrics use `pod_name`; the fallback logic masks this mismatch rather than eliminating it.
- **Race conditions**: A pod extracted from a metric may have already been deleted or not yet indexed in the Kubernetes API.
- **Stale metrics**: Prometheus retains data for deleted pods; old metrics can collide with new pods during rollouts.
- **Fragile traversal**: Four hops mean four independent failure modes, each of which silently discards metrics.
- **Coupling to ownership topology**: Supporting new workload kinds requires modifying the traversal logic.

### InferencePool scoping — no new label needed

InferencePool already carries a `spec.selector` that identifies which pods belong to the pool. WVA already reads this selector (see `internal/utils/pool/pool.go:64-67` for v1, `88-91` for v1alpha2) and uses it to find the EPP during scale-from-zero. That same selector can be embedded directly in PromQL queries to scope metrics to the right pool without any ownership traversal.

### Variant distinction — one new label

A single new WVA-specific label on pod templates is sufficient to distinguish variants within an InferencePool:

| Label key | Value | Set by |
|---|---|---|
[...]

PR #1210, opened 2026-05-28
docs(proposals): add pod-to-VariantAutoscaling mapping proposal

ev-shindin:
## Summary

  Proposes an informer-backed pod → `VariantAutoscaling` cache inside WVA, replacing the contract introduced by #1145 where the tenant must add   `llm-d.ai/variant` to every pod template (twice for `LeaderWorkerSet`)  plus configure a ServiceMonitor/PodMonitor relabeling rule. The label  remains as an opt-in escape hatch for non-standard topologies; WVA  never writes to tenant-owned resources.

  The cache absorbs the 4-hop ownership walk once at watch-event time;  metric-path lookups become O(1). Cache misses surface as a   `wva_pod_va_cache_miss_total` counter and a `PodMappingMissing` status
  condition instead of silently degrading into  `computeReplicaCapacityFallback` (the same silent-fallback failure  mode #1198 fixed for `cache_config_info`).

  The cache value is an abstract "discovery key", so the proposal  composes with the in-flight VA CRD deprecation
  (`docs/proposals/deprecate-va-crd.md`) with no consumer changes when   the discovery anchor moves from VA onto annotations on   `ScaledObject`/`HPA`.

  Docs-only change (`docs/proposals/design-pod-va-mapping.md`).

  ## Related

  - Counter-proposal to the controller-stamps direction in #1072
  - Corrective follow-up to #1145 (preserves #1145's multi-vLLM-per-pod
    work + ServiceMonitor relabeling; demotes the required label to opt-in)
  - Composes with `docs/proposals/deprecate-va-crd.md` and
    `docs/proposals/design-config-ux.md` (#1194)
  - Pre-empts the silent-fallback class of bug exposed by #1198


dumb0002:
@ev-shindin (cc: @lionelvillard ), thanks for putting together this detailed proposal. Below are my comments & clarification questions:

 a) This proposal would require to add proper Pod informer to WVA, is it?  
            - _Watch cost for namespace-scoped vs cluster-scoped WVA controller_? 

 b) Memory cost: IMO this is a major issue and it needs for discussion/investigation.

ev-shindin:
> a) This proposal would require to add proper Pod informer to WVA, is it? - _Watch cost for namespace-scoped vs cluster-scoped WVA controller_?
> 
> b) Memory cost: IMO this is a major issue and it needs for discussion/investigation.

Thanks @dumb0002 . I addressed your concerns in the proposal revision. PTAL.

dumb0002:
@ev-shindin, thanks! The revised proposal answered all my previous comment. Overall the revised proposal LGTM, my only concern is the new complexity that it adds to the WVA controller: `the juice is worth the squeeze?` (cc: @lionelvillard )


Could we explore cheaper alternatives? For example: 

  **Option A** — Validating webhook / CEL admission policy on the VA. Reject a VA whose scaleTargetRef doesn't match any Deployment/LWS, or whose target's pod template is missing the label. Failure becomes loud at apply time, not at scrape time. The proposal mentions this and dismisses it ("doesn't remove the requirement"). But removing the requirement isn't actually the goal — the goal is preventing silent failures. A webhook does that without a Pod informer. 

  **Option B** — Reconcile-time check. When the VA reconciler runs, list pods matching scaleTargetRef's selector, verify they carry the label, and set PodMappingMissing if not. No informer, no cache, no event-driven complexity. The check runs every ~30s — slower to detect than an informer, but for a config-correctness could be acceptable. 
  
 **Option C** — Just better error logging. When computeReplicaCapacityFallback fires, log a structured warning with the exact pod name, the labels it carries, and what it expected. This is the cheapest possible fix and addresses the "silently wrong" complaint directly. No status condition, no metric, no informer.

lionelvillard:
> replacing the contract introduced by https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1145 where the tenant must add  llm-d.ai/variant to every pod template (twice for LeaderWorkerSet) plus configure a ServiceMonitor/PodMonitor relabeling rule.

This is common practice. Why replacing it?

ev-shindin:
> > replacing the contract introduced by #1145 where the tenant must add  llm-d.ai/variant to every pod template (twice for LeaderWorkerSet) plus configure a ServiceMonitor/PodMonitor relabeling rule.
> 
> This is common practice. Why replacing it?

You're right that labeling pods + a ServiceMonitor relabel rule is common in general — but it's specifically uncommon for an **autoscaler**. HPA, KEDA, VPA, the Cluster Autoscaler: you attach them to a workload via `scaleTargetRef` and they read the target; none require editing the workload's pod template to be scaled. WVA already takes a `scaleTargetRef` exactly like HPA, so the required label is the one thing that breaks the norm.

Concretely, the most common adoption path: I have a Deployment already running in production and I want WVA to scale it. With the required label I can't just point a `VariantAutoscaling` at it — I first have to edit `spec.template.metadata.labels`, which is a pod-template change, i.e. a **rolling restart of the whole workload** (plus a GitOps edit) *before autoscaling even starts working*. An autoscaler should attach to an existing workload **without touching it**; the label requirement makes WVA non-adoptable in place.

I'm not removing label-based identification — `llm-d.ai/variant` + the relabel rule stay as the **opt-in** path for custom kinds / exotic owner chains. What the *default* shouldn't require is a pod-template edit, or a hand-maintained join key whose value must equal the VA name (and whose referent VA-deprecation is removing). WVA already reads `scaleTargetRef` and the target's pod selector, so it can derive pod→VA itself at reconcile time — the same way HPA resolves the target's pods.

I revised the proposal around this (it now leads with "Requiring a pod-template label is not how autoscalers attach") and added a **"How This Resolves the Fragility"** section mapping each silent-failure mode to why it can't happen once the join key is gone. @lionelvillard PTAL.

ev-shindin:
> Could we explore cheaper alternatives? For example:

 Thanks @dumb0002 — fair challenge, and you were right that the always-on Pod informer was the heaviest part. I split the two goals, because A/B/C address one but not the other:

**(1) Make silent failures loud** — A/B/C all do this. I took the cheap win: **Option C** (a structured warning at the `computeReplicaCapacityFallback` boundary — pod, labels-seen, key-expected) is now part of the miss signal.

**(2) Remove the tenant authoring burden** (the config-ux persona; also the in-place-adoption / rolling-restart problem in my reply to @lionelvillard) — A/B/C all *keep* the required label + relabel rule and just make their absence loud; they fix the symptom, not the hand-maintained join key.

**Your Option B gets us both if we *derive* instead of *check*:** at reconcile WVA already lists the pods behind `scaleTargetRef`'s selector, so it resolves each pod→VA there (owner-ref confirmation so a stray pod sharing the selector isn't misattributed) and builds the map — **no informer, no always-on Pod cache, no cluster-wide memory**, one bounded LIST per reconcile. Resolution moves off the per-metric hot path (lookups are O(1)); it doesn't disappear, it just runs once per reconcile instead of per metric. I've pivoted the proposal to this (the "Watch Cost" section now reads "no new watch — bounded per-reconcile list").

On **Option A** (webhook/CEL): I kept it as Alternative #4 / defense-in-depth for the opt-in label path — CEL can check the label is *present* but not that its value equals a VA name (cross-object; CEL is single-object), it leaves the requirement (and the rollout-to-adopt cost) in place, and it validates the VA-name anchor that VA-deprecation removes.

Net on "juice worth the squeeze": the squeeze is now just a per-reconcile list + owner-ref resolve (no new watch, no extra memory), and the new **"How This Resolves the Fragility"** section shows the juice — each silent-failure mode removed at the root rather than detected after the fact. PTAL.

dumb0002:
@ev-shindin, thanks for revising the proposal - removing the Pod informer eliminates the most expensive structural commitment, and per-reconcile selector LIST is a much lightweight approach indeed. 

**Some additional comments:**
      a) For a Deployment-managed pod, owner refs go `Pod → ReplicaSet → Deployment`. So confirming the Deployment owns the pod requires either:
  - A LIST/GET on the ReplicaSet (one extra API call per pod), or
  - Trusting selector+namespace match without confirmation.

b) For LWS, owner chain is `Pod → StatefulSet (per group) → LeaderWorkerSet`. Same issue — confirming LWS ownership needs a StatefulSet read




/LGTM  (cc: @lionelvillard)

lionelvillard:
there is lot of noise in this proposal. Very difficult to understand what problem is being solved and what's the solution.

docs/proposals/design-pod-va-mapping.md
```diff
@@ -0,0 +1,55 @@
+# Proposal: Pod → `VariantAutoscaling` Mapping via Per-Cycle Derivation
+
+## User story
+
+A platform engineer has a `Deployment` of vLLM replicas (or a `LeaderWorkerSet`) **already running in production**. They want WVA to autoscale it, so they create a `VariantAutoscaling` pointing at the workload via `scaleTargetRef` — exactly how they'd attach an HPA — and expect it to work.
```
lionelvillard: `VariantAutoscaling` is deprecated. Can you replace it by "HPA".

Also creating an HPA is not what a platform engineer does. It will follow a recipe or a guide. In llm-d, it will apply an kustomization file.

docs/proposals/design-pod-va-mapping.md
```diff
@@ -0,0 +1,55 @@
...
+
+It doesn't. WVA reports a desired-replica count, but it's quantitatively wrong, with **no error surfaced**. Digging in, they discover they also had to:
```
lionelvillard: see comment above. They don't have to dig in.

docs/proposals/design-pod-va-mapping.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. add `llm-d.ai/variant: <VA-name>` to the workload's **pod template** — forcing a **rolling restart** of the whole workload (and a GitOps source edit) just to start autoscaling — and on a `LeaderWorkerSet`, to *both* `leaderTemplate` and `workerTemplate`; and
```
lionelvillard: >  forcing a **rolling restart** of the whole workload 

that's IMO the core of the problem.

docs/proposals/design-pod-va-mapping.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Solution
+
+WVA derives the mapping itself from the scale target it already reads, once per optimization cycle (the timer-driven loop that already scrapes pods, every `OptimizationInterval` — not the per-VA-event `Reconcile`) — **no tenant-authored join key, no pod mutation, no new watch.**
+
+- **Per-cycle derivation.** Once per cycle, for each managed `scaleTargetRef`, WVA lists the pods behind the target's selector and resolves each to that target's `VariantAutoscaling`. This is kind-aware for the standard kinds — a `Deployment`'s and a `LeaderWorkerSet`'s selectors and owner chains differ — with the opt-in label (below) covering kinds the selector can't resolve. Ownership is confirmed by walking the pod's owner chain back to the scale target (for a `Deployment`, pod → `ReplicaSet` → Deployment), so a stray pod merely sharing the selector isn't misattributed — the *depth* of that walk is an open question. Resolution moves from per-metric to once-per-cycle.
```
lionelvillard: > the *depth* of that walk is an open question
is it?

docs/proposals/design-pod-va-mapping.md
```diff
@@ -0,0 +1,55 @@
...
+- **No new watch.** The per-cycle list reuses the pod LIST the scraping source already issues (`pod_scraping_source.go` already `List`s the pods it scrapes), so it adds **no always-on Pod watch** and no cluster-scale Pod cache; WVA's controller already watches the `VariantAutoscaling` and its scale target, and neither watches `Pod`.
+- **O(1) hot path.** Each per-metric lookup becomes a single `Lookup(podKey) → VA` map read (`podKey` = `namespace/name`, or `:port` for multi-vLLM pods); no Kubernetes calls in the metrics path.
+- **Misses become a signal, not silence.** An unresolved metric increments `wva_pod_va_map_miss_total{reason}` and logs a structured warning; a VA whose pods exist but aren't resolving gets a `PodMappingMissing` status condition. The old silent wrong-capacity fallback becomes visible. (Most misses are the brief pod-creation-to-next-cycle window — transient and self-healing.)
+- **Opt-in label for the long tail.** `llm-d.ai/variant` stays as a *higher-precedence override* for custom workload kinds / non-standard topologies the selector can't resolve. WVA reads it, never writes it; the ServiceMonitor relabel rule stays for installs that keep it.
```
lionelvillard: > Opt-in label for the long tail

what does that mean? What's the tail here?

ev-shindin:
Solution is already integrated

PR #1214, opened 2026-05-28
chore: delete WVA Helm chart and helm-release workflow

### LINKED ISSUE #1088: Deprecate Helm charts; adopt Kustomize for install manifests
Track replacing Helm-based deployment with Kustomize-based layouts. Helm should be deprecated and eventually removed once parity and migration paths are defined.

mamy-CS:
The chart has been deprecated since we switched to Kustomize-first install.


asm582 (commented):
/lgtm

may be wait till llm-d-benchmark moves to Kustomize

PR #1216, opened 2026-05-28
test(e2e): replace load Job with fake-metrics in V1 saturation tests

mamy-CS:
Eliminates the real http load job by configuring the simulator with `--fake-metrics` at startup. Tests now skip when not running against the simulator, consistent with the V2 pattern.
Note - Openshift CI will eventually be removed, and nightly scaling minimal real world like scenerios will be added to nightly tests.


asm582 (approved):
/lgtm

lionelvillard (approved):

PR #1217, opened 2026-05-28
chore(test): bump llm-d-inference-sim to v0.9.0

mamy-CS:
v0.9.0 replaces the deprecated UDS tokenizer with vLLM Render, which is required when `--model` resolves to a real HuggingFace model. Switch the default `MODEL_ID` to a dummy name so the simulator uses the fast in-process regex tokenizer with no render sidecar or network dependency at startup. Sufficient for simulated e2e. 
Release notes: https://github.com/llm-d/llm-d-inference-sim/releases/tag/v0.9.0
Tokenization guide: https://github.com/llm-d/llm-d-inference-sim/blob/main/docs/tokenization.md


deploy/kind-emulator/install.sh
```diff
@@ -268,6 +274,46 @@ load_image() {
     log_success "Image '$full_image' pulled directly into KIND cluster '$CLUSTER_NAME' nodes"
 }
 
+# Pre-loads the llm-d-inference-sim image into the KIND cluster so tests that create
+# model service Deployments don't pull it cold and hit PodReadyTimeout.
+load_sim_image() {
```
lionelvillard, line 226: is there already a bash function loading images into KIND Cluster? If yes, then it should be reused.

mamy-CS: needed some refactor, didnt want to mix in this pr but ok

PR #1221, opened 2026-05-29
 chore(test): bump llm-d-inference-sim to v0.9.0

mamy-CS:
v0.9.0 replaces the deprecated UDS tokenizer with vLLM Render, which is required when `--model` resolves to a real HuggingFace model. Switch the default `MODEL_ID` to a dummy name so the simulator uses the fast in-process regex tokenizer with no render sidecar or network dependency at startup. Sufficient for simulated e2e. 
Release notes: https://github.com/llm-d/llm-d-inference-sim/releases/tag/v0.9.0
Tokenization guide: https://github.com/llm-d/llm-d-inference-sim/blob/main/docs/tokenization.md


test/utils/resources/llmdsim.go
```diff
@@ -41,20 +41,19 @@ func CreateLlmdSimDeployment(namespace, deployName, modelName, appLabel, port st
 					Containers: []corev1.Container{
 						{
 							Name:            appLabel,
-							Image:           "ghcr.io/llm-d/llm-d-inference-sim:v0.7.1",
+							Image:           "ghcr.io/llm-d/llm-d-inference-sim:v0.9.0",
```
asm582: Nit: Can we use a const for the image name and reuse it?

Makefile
```diff
@@ -14,7 +14,7 @@ CONTROLLER_NAMESPACE ?= workload-variant-autoscaler-system
 MONITORING_NAMESPACE ?= openshift-user-workload-monitoring
 LLMD_NAMESPACE       ?= llm-d-optimized-baseline
 GATEWAY_NAME         ?= # discovered automatically in e2es
-MODEL_ID             ?= unsloth/Meta-Llama-3.1-8B
+MODEL_ID             ?= test-sim-model
```
asm582: Follow the OpenAI model ID format even for dummy model

asm582 (changes requested):
Address review

test/utils/resources/llmdsim.go
mamy-CS: done

Makefile
mamy-CS: This is intentional, following `gpt-4` style for example so we dont hit the token issue in sim. Or you meant org/model style?

mamy-CS: OpenAI's own SDK tests use [gpt-5.4](https://github.com/openai/openai-python/blob/main/tests/test_client.py). `test-sim-model` is unambiguous about its purpose, but we can rename to maybe gpt-sim-1?

mamy-CS: Changed to e2ewva/dummy-model

asm582 (approved):
/lgtm

Issue #1001, opened 2026-04-14
Namespace-scoped GPU inventory and limiter


ev-shindin:
### Problem

The current `TypeInventory` (`internal/engines/pipeline/type_inventory.go`) tracks GPU inventory cluster-wide:

- `Discover()` aggregates all node GPUs into a single per-type map (e.g., `{"H100": 16, "A100": 8}`)
- `typeAllocator.TryAllocate()` debits from the cluster-wide pool regardless of which namespace the variant belongs to

In a multi-tenant cluster where namespaces are mapped to specific node pools (via taints/labels), this causes incorrect behavior:

- A variant in namespace `team-a` can consume GPU capacity that physically lives on `team-b`'s nodes
- The limiter approves allocations that the scheduler will then reject because no matching node exists
- One namespace can exhaust the inventory total, blocking scaling decisions for variants in other namespaces

### Proposed solution

Add a `NamespaceInventory` implementation of the existing `Inventory` interface that:

1. Resolves namespace → cluster nodes via a **node label selector per namespace** (idiomatic Kubernetes — nodes are cluster-scoped, but operators already use node labels + admission policies to pin namespaces to specific nodes)
2. Discovers GPU inventory **per namespace** by intersecting cluster node discovery with each namespace's selected nodes
3. Tracks `Limit`, `Used`, `Available` per `(namespace, accelerator type)` instead of just per accelerator type
4. Creates a `namespaceTypeAllocator` that routes allocations to the correct namespace pool based on `decision.Namespace`

#### Why label selectors, not explicit node lists

Nodes in Kubernetes are cluster-scoped, not namespaced. There is no native "this node belongs to namespace X" concept. Operators that want per-namespace node pools rely on:

- Node taints + automatic toleration injection per namespace (via admission webhooks like Kyverno or `PodNodeSelector`)
- Node labels + namespace-pinned `nodeSelector` injection
- Cloud-managed node pools with team-specific labels

In all of these patterns, **node labels are the ground truth** for which nodes belong to which team/namespace. Using label selectors here:
- Reuses the labels operators already maintain (no duplicated config)
- Survives node replacement (autoscaling, repairs) — the selector still matches
- Aligns with how the Kubernetes scheduler itself works

Explicit node lists were considered and rejected — they're brittle (break on node replacement) and duplicate information that should live on the nodes themselves.

#### Refresh semantics

[...]

PR #1222, opened 2026-05-30
Add namespace-scoped GPU inventory limiter (V1 saturation path)

jia-gao:
## Summary

Adds an opt-in **namespace-scoped GPU inventory limiter**. Part of #1001 (scoped to the V1 saturation path; V2 coverage is a tracked follow-up).

The GPU limiter tracks one cluster-wide pool per accelerator type. In a multi-tenant cluster where namespaces are pinned to node pools (taints+tolerations, `PodNodeSelector` admission, managed node pools), that lets a workload in one namespace consume capacity on another tenant's nodes: the limiter approves allocations the scheduler then can't place, and one namespace can exhaust the shared total and stall scaling for everyone else.

The new limiter partitions GPU capacity per namespace by intersecting discovered nodes with a per-namespace node label selector — reusing the node labels operators already maintain. It builds on the `NodeDiscovery` interface (`DiscoverNodes`, already in `internal/discovery`), so **no discovery changes are needed**.

## Design

- **`NamespaceInventory`** implements `Inventory`; **`namespaceTypeAllocator`** implements `ResourceAllocator`. Pools are keyed per `(namespace bucket, accelerator type)`. `GetResourcePools` aggregates to cluster-wide per-type pools for interface/reporting use.
- **`NamespaceLimiter`** (implements `Limiter`) owns per-bucket usage accounting, because the type-keyed `Inventory.SetUsed` contract can't carry the namespace dimension. Usage and allocation resolve namespaces to buckets through the **same resolver**, so default-fallback namespaces' usage aggregates under the shared default pool rather than being dropped.
- **Lookup rules:** `exclude` (bypass) → exact selector → `default` selector → zero inventory (deny). Excluded namespaces are unconstrained, but their existing usage is still charged against the pool they draw from so shared pools aren't overcommitted.
- **Each node is assigned to exactly one bucket** (named selectors take precedence over `default`; deterministic on overlap) so overlapping/misconfigured selectors never double-count the same physical GPUs.
- **Unknown accelerator names resolve per bucket:** when a namespace's pool is single-type, an unresolved accelerator resolves to it, so existing replicas debit the right pool even in an otherwise heterogeneous cluster.
- **Config:** optional `wva-limiter-config` ConfigMap, parsed once at bootstrap via `sigs.k8s.io/yaml` (so `metav1.LabelSelector` json tags apply). Read once at startup — changing it requires a controller restart. Absent config preserves the previous cluster-wide behavior exactly.
- **Observability:** `LimitedBy` is the stable `namespace-inventory` string (bounded metric-label cardinality); per-namespace/type detail goes in the `DecisionStep` reason (`limited by namespace-inventory[ns=…, type=…]`).

## Scope: V1 saturation path (V2 is a follow-up)

This enforces isolation on the **V1 saturation (`Limit`) path**. The V2 constraint path (`ResourceConstraints` / `GreedyByScoreOptimizer`) is **type-keyed and cannot express per-namespace partitioning** without a larger change to the constraint model and the fair-share optimizer. So on the V2 path the engine keeps the cluster-wide limiter and logs a **one-time warning** when namespace inventory is configured under an active V2 analyzer.

Per review discussion, this PR is intentionally scoped to V1; V2 coverage is tracked as a follow-up via a composable `NamespaceConstraintProvider` in the constraint chain (aligned with #1003 / #1002).

On V1 the namespace limiter **replaces** the cluster-wide `TypeInventory` rather than intersecting with it. This is deliberate: nodes are assigned one-bucket-each, so per-bucket sums never exceed physical capacity and no cluster-total cap is needed on top. One consequence: nodes matching no selector and no `default` contribute to no pool, so their capacity is intentionally unusable until they are labeled or a `default` selector is added.

## Known limitations (documented in the admin guide)

Accounting is **decision-based** (like the existing cluster-wide limiter): it subtracts GPUs used by the WVA-managed variants seen this cycle, not actual pod placement. Replicas with an unresolved accelerator in a multi-type pool aren't attributed. Disjoint single-type node pools per namespace avoid this. See `docs/user-guide/admin/namespace-inventory.md`.

## Tests

- `internal/engines/pipeline/namespace_inventory_test.go` — per-namespace pools, exhaustion isolation, default fallback, deny-when-unlisted, exclude bypass, **selector re-evaluation on node change**, cluster-wide pool aggregation, used-subtraction, heterogeneous unresolved → zero, **single-bucket assignment on overlapping selectors**.
- `internal/engines/pipeline/namespace_limiter_test.go` — pool capping, **cross-namespace isolation**, **default-bucket usage aggregation**, **excluded-namespace usage charged against shared pools**, exclude unconstrained, deny, **per-bucket accelerator resolution in heterogeneous clusters**, unresolved-in-multi-type-bucket skip, homogeneous resolution.
- `internal/config/limiter_test.go` — ConfigMap parsing (matchLabels + matchExpressions + exclude + default), empty/missing/invalid.
- Integration test — multi-namespace contention (envtest), covering the acceptance criterion.

## Verification

- [x] `go build ./...`, `go vet ./...`, `gofmt -l` clean
- [x] `go test ./internal/...` (incl. envtest-backed controller + saturation suites)
- [x] `golangci-lint run` — 0 issues

## Notes for reviewers

- Sample config: `config/samples/limiter-configmap.yaml`. Admin docs: `docs/user-guide/admin/namespace-inventory.md`.


asm582:
@jia-gao Thanks for this PR.  As I understand, we are trying to allocate a fair share of resources. Asking for more information:

- Why can't namespaces and nodepools have a 1:1 mapping?
- As per the implementation, some of the features are already available in https://github.com/kubernetes-sigs/kueue , can those be reused before we implement ?

jia-gao:
Thanks @asm582 — good questions. I spent some time with Kueue's docs before answering so this is grounded rather than a guess.

On **namespace ↔ node pool being 1:1**: that would totally work, and it's really just a special case of what's here (a selector matching a single pool). I reached for label selectors mainly because Kubernetes has no built-in "node pool" object, so pools usually already live as node labels (cloud nodegroup labels, GFD, team/taint conventions), and a namespace sometimes wants only a slice of a pool (a single GPU type). But if a strict 1:1-by-name mapping is cleaner for where you'd like this to go, I'm happy to simplify to that — it's a thin layer on the same code.

On **reusing Kueue**: fair point, and the overlap is real — `ResourceFlavor.spec.nodeLabels` partitions nodes much like these selectors, and `ClusterQueue` + `LocalQueue` already give per-namespace, per-flavor GPU quota. I'm honestly not sure whether this should lean on Kueue instead, and I'd value your read. A couple of distinctions I noticed while reading, just in case they're useful:

- Kueue seems to act at pod *admission* (over-quota pods suspend, and it leaves replica counts to HPA), whereas this limiter is trying to shape the autoscaling *decision* itself — capping the replica target before WVA acts on it. So they may be solving adjacent problems rather than the same one.
- Kueue's `nominalQuota` is admin-declared, while this reads physical GPU capacity from node discovery — so the numbers come from different places.

(I also came across kubernetes-sigs/kueue#6586 on quota tracking with overlapping nodes, which is close to a case this PR handles — mentioning it as a data point, not a knock on Kueue.)

So I'd defer to you and @ev-shindin on direction. A few paths I could see: (a) keep this as a lightweight option for clusters not running Kueue; (b) later add a Kueue-aware mode that reads `LocalQueue → ClusterQueue` quota when Kueue is present; (c) lean on Kueue more directly if that fits the project better. Happy to go whichever way you both think makes the most sense.

internal/engines/saturation/engine.go
```diff
@@ -499,15 +546,22 @@ func (e *Engine) optimizeV1(
 		}
 	}
 	if globalSaturationConfig.EnableLimiter && len(allDecisions) > 0 {
+		// Prefer the namespace-scoped limiter on the V1 path when configured;
+		// otherwise apply the cluster-wide GPU limiter.
+		limiter := e.GPULimiter
+		if e.nsLimiter != nil {
+			limiter = e.nsLimiter
```
ev-shindin: #1001 describes `NamespaceInventory` and `TypeInventory` *coexisting in a chain enforcing intersection constraints*. Here the namespace limiter **replaces** the cluster-wide one on V1 rather than composing with it. It's defensible — nodes are partitioned one-bucket-each, so bucket sums ≤ physical and there's no double-count — but two consequences are worth confirming as intentional: (1) there's no separate cluster-total cap layered on top, and (2) nodes matching no selector (and no `default`) contribute to no pool, so their capacity is silently unusable. Could you add a sentence (code comment + docs) stating the partition-instead-of-intersection choice is deliberate?

internal/engines/saturation/engine.go
```diff
@@ -675,6 +729,15 @@ func (e *Engine) optimizeV2(
 	// Stage 2: Compute GPU constraints and call optimizer
 	var constraints []*pipeline.ResourceConstraints
 	if _, ok := e.optimizer.(*pipeline.GreedyByScoreOptimizer); ok {
+		// Namespace-scoped inventory is not enforced on the V2 constraint path:
+		// ResourceConstraints/GreedyByScoreOptimizer are type-keyed and cannot
+		// partition GPUs per namespace. Constraints here are cluster-wide. Warn
+		// once so operators relying on per-namespace isolation aren't misled.
+		if e.nsLimiter != nil {
```
ev-shindin: On the V2 (token-based) analyzer, isolation is only warned about, not enforced. V1 is the zero-config default so the feature works out of the box, but V2 is where active development is heading (role-aware P/D, quota #1002), and a V2 deployment gets **no** per-namespace isolation. This is the main thing blocking a clean "Closes #1001": the issue's *Backward Compatibility* section implies V2 coverage too. Suggest either (a) scope this PR to V1 explicitly — retitle to *"part of #1001"* and open a V2 follow-up — or (b) hold "Closes #1001" until V2 lands. (Happy to file the follow-up.)

internal/engines/pipeline/namespace_limiter.go
```diff
@@ -0,0 +1,182 @@
...
+		bucket, excluded, hasPool := l.inventory.resolver.resolve(d.Namespace)
+		if excluded || !hasPool {
+			continue
+		}
+		if used[bucket] == nil {
+			used[bucket] = make(map[string]int)
+		}
+		used[bucket][d.AcceleratorName] += d.CurrentReplicas * d.GPUsPerReplica
+	}
+	return used
```
ev-shindin, line 136: Excluded namespaces both **bypass** the limiter (`namespace_inventory.go:351-354` returns the full request) and are **not charged** here, and accounting is decision-based (no `DiscoverUsage`). So an excluded tenant drawing from the shared `default` pool is never subtracted → that pool can hand out GPUs that are physically gone. This diverges from the cluster `TypeInventory` path, which charges **all** pod usage (incl. excluded namespaces) via namespace-agnostic `DiscoverUsage`. Please charge excluded-namespace usage against shared pools (count it, just don't cap it).

ev-shindin (changes requested):
Thanks for this — it's a careful, well-structured implementation and the **code quality is merge-grade** 👍.

My reservations aren't about the implementation — they're about whether this **closes #1001** as written. Acceptance-criteria status:

| Criterion | Status |
|---|---|
| `NamespaceInventory` + `namespaceTypeAllocator` | ✅ |
| ConfigMap parser (`LabelSelector` + `exclude`) | ✅ |
| Selector re-evaluation on `Refresh()` | ✅ |
| Unit tests (allocation, exhaustion, defaults, denial, bypass, re-eval) | ✅ |
| **Integration test — multi-namespace contention** | ⚠️ missing (only unit-level) |
| User documentation | ✅ |
| `DecisionStep` observability | ✅ |

Three things I'd like resolved before this is marked *Closes #1001* (details in the inline threads):

1. **V2 path is not enforced** (`engine.go`) — isolation applies to V1 only; V2 just logs a warning. V1 is the default so this works out of the box, but V2 is where active development is heading, and #1001's backward-compat note implies coverage there too. 
2. **Excluded namespaces can overcommit a shared pool** (`namespace_limiter.go`) — they bypass *and* aren't charged, diverging from the cluster path's `DiscoverUsage` accounting. Accept + document, or charge their usage. 
3. **Chain became "replace" on V1** (`engine.go`) — #1001 describes intersecting with `TypeInventory`; the engine replaces it. Defensible given node partitioning, but worth confirming as intentional. 

**Suggested path:** if the team is happy scoping this to V1 (reasonable — it's the default), retitle to *"part of #1001"* and let me open the V2 follow-up (I've drafted one around a composable `NamespaceConstraintProvider` in the constraint chain, aligned with #1003/#1002). Otherwise I'd hold "Closes #1001" until V2 lands. Either way, the integration test and the excluded-namespace question are the two I'd want answered before merge.

Requesting changes on those three points — happy to pair on the integration test or the V2 follow-up.

ev-shindin:
> * Why can't namespaces and nodepools have a 1:1 mapping?
> * As per the implementation, some of the features are already available in https://github.com/kubernetes-sigs/kueue , can those be reused before we implement ?

Adding context on both questions, since they both come down to **where** the cap is enforced — at pod *admission* (the scheduler/quota layer) vs. at WVA's scaling *decision*.

### Why not a strict 1:1 namespace ↔ nodepool mapping?

1:1 is actually supported — a named selector per namespace *is* a 1:1 mapping. But mandating only 1:1 has two problems:

- **Shared pools are a legitimate, common pattern.** Several namespaces sharing one pool (the `default` bucket here) is how operators avoid stranding GPUs in idle tenants' dedicated pools. Strict 1:1 forecloses that.
- **More fundamentally, node pinning alone doesn't fix the problem this PR targets — because pinning is admission-time and the bug is decision-time.** Taints/tolerations, `nodeSelector`, PodNodeSelector admission, managed nodepools — they all control *where a pod is allowed to land*. They do **not** tell WVA *how many replicas to request*. Today `TypeInventory` aggregates GPUs cluster-wide, so even with perfect 1:1 node pinning, WVA can decide to scale namespace-A's variant against an H100 count that includes namespace-B's nodes, request the replicas, and then the scheduler refuses to place them (they're pinned away) → a stuck rollout, and WVA is blind to why. The 1:1 *node* mapping is real, but WVA's *inventory* is still cluster-wide. This PR closes exactly that gap: it makes WVA's decision respect the per-namespace partition, so it never requests replicas that can't run.

So 1:1 isn't prohibited — it's the named-selector case. The PR generalizes it (shared/default pools) and, crucially, moves the partition into WVA's decision loop, which node pinning by itself can't do.

### Can Kueue be reused instead of implementing this?

Kueue is genuinely good here and worth reading where it's deployed — `ResourceFlavor` + `ClusterQueue` can express per-nodepool, per-GPU-type quota on any K8s version. But the same admission-vs-decision split applies, and it's the reason Kueue-as-the-mechanism doesn't substitute for this:

- **Kueue enforces at admission; WVA decides replica count.** Used as the enforcement layer, Kueue gates the replicas WVA over-requested into `Pending`. WVA then can't tell "running" from "gated-Pending" and folds them into anticipated supply — on the V2 saturation analyzer this is concrete: `saturation_v2/analyzer.go:97` counts `(ReplicaCount + PendingReplicas) × PerReplicaCapacity` and `:119` subtracts it from required capacity, so `requiredCapacity` is suppressed and a sibling variant that *could* serve the same pool is never scaled — the **anticipated-capacity trap**. The pool stays under-served while WVA believes it's converging. Kueue-as-mechanism relocates the problem to a gate WVA can't see; it doesn't remove it.
- **The reuse that *does* fit is reading Kueue at decision time.** A `ConstraintProvider`/reader that pulls Kueue's `ResourceFlavor`/`ClusterQueue` quota into WVA's decision loop gives the same per-namespace/per-type bound this PR provides, sourced from Kueue where it already runs — so WVA never requests what Kueue would gate. Where Kueue isn't deployed, the node-label inventory in this PR provides the partition with zero new operators. Both compose in the same decision-time chain (`min(physical, namespace-pool, …)`); they aren't either/or.

Net: I'd frame this as **start with a decision-time mechanism that works on any cluster (this PR), and layer Kueue/native sources in as decision-time *readers*** for orgs that run them — rather than adopting Kueue as the enforcement layer, which leaves WVA's decision blind to the cap and reintroduces the stuck-rollout / anticipated-capacity failure. That's the same direction the limiter-chain / quota work (#1003 / #1002) is taking, so a Kueue reader would slot in naturally as a follow-up.

jia-gao:
Thanks @ev-shindin — really generous review, and the admission-time vs decision-time framing nails it better than I put it. The "anticipated-capacity trap" (gated-`Pending` folding into anticipated supply at `analyzer.go:97/:119`) is exactly the failure I was gesturing at — agreed that's the crux, and that a Kueue *reader* in the decision chain is the right way to reuse Kueue rather than leaning on it as the enforcement layer.

**On scope:** agreed, let's scope this to V1. I'll retitle to *"part of #1001"*, and I'd gladly take you up on the V2 follow-up — your composable `NamespaceConstraintProvider` in the constraint chain (aligned with #1003/#1002) is a much better home for it than bolting per-namespace logic onto the type-keyed V2 path here. Please go ahead and open it; happy to help.

**On the three points:**

1. **V2** — covered by the scope decision above.
2. **Excluded namespaces overcommitting the shared pool** — agreed; I'll charge their usage against the pool (count it, don't cap it) so the shared `default` pool reflects GPUs that are physically gone, matching the cluster path's `DiscoverUsage` accounting.
3. **Replace vs intersect on V1** — intentional; I'll say so explicitly in both a code comment and the docs, including the consequence that nodes matching no selector and no `default` contribute to no pool (capacity intentionally unusable until they're labeled or a `default` is set).

I'll also add the **multi-namespace contention integration test** — may ping you to pair on the harness. Will push the updates and retitle shortly.

asm582:
@jia-gao For your use case,  is ResourceQuota already set on these namespaces?

internal/engines/saturation/engine.go
jia-gao: Done in 0e36004 — documented the deliberate replace-not-intersect choice both as a code comment at the V1 limiter site and in the admin guide, including the consequence you flagged: nodes matching no selector and no `default` contribute to no pool and are unusable until they're labeled or a `default` selector is added.

jia-gao:
Good question @asm582. In the multi-tenant setups this targets, `ResourceQuota` may or may not be set — but I don't think it closes the gap either way, for the same reason the Kueue thread landed where it did (the admission-vs-decision split @ev-shindin laid out):

- `ResourceQuota` caps a namespace's total `requests.nvidia.com/gpu` at **pod admission** — it rejects pods over quota. It doesn't tell WVA how many replicas to *request*, so WVA still sizes the scale-up against cluster-wide `TypeInventory`, asks for replicas the quota then rejects, and is blind to why the rollout stalls — the same stuck-rollout this PR avoids by partitioning at the decision layer.
- It's also a flat per-namespace GPU count, not tied to specific nodes/pools. It can't express "namespace A draws from *these* 8 H100 nodes," which is what node-pinned tenancy needs and what the node-label inventory captures.

Where a cluster *does* run `ResourceQuota` (or Kueue), the clean reuse is the same: read that quota at decision time as another bound in the limiter chain (`min(physical, namespace-pool, quota, …)`) rather than relying on admission to reject after the fact. Happy to note that as a follow-up alongside the Kueue reader.

internal/engines/saturation/engine.go
jia-gao: Scoped this PR to V1 (retitled to *part of #1001*); the V2 path stays warn-only here. Per your offer, the V2 follow-up via a composable `NamespaceConstraintProvider` in the constraint chain is the better home for it than bolting per-namespace logic onto the type-keyed V2 path — please go ahead and open it, happy to help.

internal/engines/pipeline/namespace_limiter.go
jia-gao: Done in 0e36004 — excluded namespaces' usage is now charged against the pool they draw from via a new `chargeBucket` resolver (count, don't cap), matching the cluster path's all-usage accounting. Added a unit test plus the multi-namespace contention integration test.

asm582:
What is the preferred deployment pattern for your use case? 
What if operators deploy WVA per namespace, using WVA_NODE_SELECTOR and a dedicated node pool?

jia-gao:
Good question. Honestly, running a WVA per namespace with `WVA_NODE_SELECTOR` is a clean setup, and if the pools are dedicated and don't overlap (one tenant, one pool) I don't think you need this PR at all — each instance only sees its own nodes, so the inventory's already scoped. No argument from me there.

The case I'm after is the single cluster-wide WVA on a shared cluster. The thing per-instance sharding can't really do there is shared pools: `WVA_NODE_SELECTOR` only carves out disjoint shards, but plenty of multi-tenant setups have a few namespaces sharing one pool (the `default` bucket here) so idle tenants don't strand GPUs. If you run one WVA per namespace with overlapping selectors, each sees the whole shared pool and allocates against all of it — they double-commit, because neither knows about the other. The in-process limiter pins every node to exactly one bucket, so the shared pool gets split cleanly; separate instances just can't coordinate that.

There's also the plain operational cost — a WVA per tenant is a Deployment, leader election, RBAC, and metrics wiring each, created and torn down as namespaces come and go, versus one controller that picks up a new tenant with a config edit and keeps a single cross-tenant view (which #1003/#1002 will want anyway).

So I don't see them as competing: per-namespace WVA is great for dedicated pools, this covers the single-controller / shared-pool case. I can add a short "when to use which" to the docs. Genuine question back, though: what deployment shape does the project want as the default? If it's per-namespace instances, that changes the math here and I'd rather get behind that than keep pushing. @ev-shindin, you framed the decision-time angle better than I did — does this line up with how you see the topology?

jia-gao:
@ev-shindin all three points are addressed in `0e36004`: the replace-vs-intersect choice is documented (code comment + admin guide, including the no-selector/no-`default` unusable-capacity consequence), excluded-namespace usage is now charged against the shared pool via a new `chargeBucket` resolver (count, don't cap) to match the cluster `DiscoverUsage` accounting, and the V2 path is scoped out — retitled to *part of #1001*, with the composable `NamespaceConstraintProvider` follow-up yours to open. Also added the multi-namespace contention integration test. Would you re-review to clear the changes-requested?

@asm582 also happy to hear your read on the per-namespace-WVA (`WVA_NODE_SELECTOR`) vs single-controller shared-pool topology question above — that's the main open design point from my side.

jia-gao:
Rebased onto latest `main` — the merge conflict is resolved and CI is green again. All three of your points are addressed in `0e36004`: the replace-vs-intersect choice is documented (code comment + admin guide, including the no-selector/no-`default` unusable-capacity consequence), excluded-namespace usage is charged against the shared pool via the `chargeBucket` resolver (count, don't cap) to match the cluster `DiscoverUsage` accounting, and the V2 path is scoped out (retitled to *part of #1001*, with the composable `NamespaceConstraintProvider` follow-up yours to open). Also added the multi-namespace contention integration test. Whenever you have a cycle, could you re-review to clear the changes-requested? Thanks for the approve on #1190 earlier.

ev-shindin:
Hi @jia-gao ! Can you please rebase it again? Also note on https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1441. Would you like to implement follow-up PR that will enable this feature for v2 path.? You can look to quota limiter: https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1129 as na example.

jia-gao:
Thanks @ev-shindin! Rebased onto latest main and force pushed.

Two things from the conflict resolution worth a look during review. In `config.go` I kept your `limiter` selector and put the namespace inventory config in a separate `nsInventory` field so the two don't collide. In `engine.go` I moved the V2 path over to `selectV2Optimizer` since the old `ComputeConstraints` signature is gone. I left a one time warning log that namespace scoping isn't enforced on V2 yet, so operators don't assume isolation they aren't getting.

On the follow up, yes I'll take it. Since #1441 has V1 going away, moving this onto V2 is what keeps the feature alive, so it makes sense to do now. From reading #1129 most of the framework is already generic, so I think it mostly comes down to making `NamespaceInventory` implement `NamespaceAwareInventory` and surfacing it as a constraint provider, with the per namespace caps coming from node intersection instead of quotas.

One question before I open it. My limiter is the physical inventory one, and #1129 leaves min(physical, quota) composition to #1003. Do you want the V2 work to land on its own, or should I coordinate it with #1003 so the two compose cleanly? Fine with me either way.

ev-shindin:
Thanks @jia-gao — rebase looks clean, and both conflict-resolution choices are the right call.
**On the V2 follow-up and #1003:** land it on its own, decoupled from #1003. The two are different layers:

- Your task is to make `NamespaceInventory` *a constraint provider* on V2 — implement `NamespaceAwareInventory` and surface it via `ComputeConstraints`, with per-namespace caps from node intersection. That's exactly what #1129 did for quotas, and you're right that the framework is already generic enough.
- #1003 (`ChainLimiter`, intersection/min semantics) is the layer that *composes* multiple providers. It can't chain a physical-namespace provider that doesn't exist yet, so your provider is a prerequisite input to #1003, not part of it. "On its own" is also the correct dependency order.
- #1441 makes this time-sensitive: V2 becomes the default at v0.9.0 and V1 is removed at v0.11.0, so getting the provider onto V2 independently keeps the feature alive without blocking on the larger chain work.

So: land the V2 provider solo, and let #1003 pick both providers up afterward. The only thing to coordinate now is the **interface contract**, so they compose cleanly later:

1. Implement `NamespaceAwareInventory` with the same closed-allowlist contract #1129 uses — a namespace present with an empty inner map means deny-all, an *excluded* namespace is omitted (= open / pass-through). Match that exactly or the chain's `min()` will misread your pools.
2. Keep the "type absent = deny" vs "unbounded" distinction aligned. You won't have a `-1`/unlimited value, but the absent-vs-uncapped semantics still have to line up with the quota provider's sentinel, or composition will mis-handle a type one provider caps and the other doesn't.
3. Heads-up from reviewing #1129: `DefaultLimiter.ComputeConstraints` has a fallback where, if every active namespace is excluded/non-listed, `NamespaceResourcePools` returns empty and `rc.Pools` falls back to the *static all-namespaces* `GetResourcePools()` aggregate rather than the active-set aggregate. If your provider reuses that path, it inherits the same fallback — worth confirming it doesn't feed a stale cluster cap into `min()` once #1003 chains physical + quota. Deriving the cluster aggregate from the *active* namespace set (as the non-fallback branch does) avoids it.

Happy to review the follow-up when you open it. If you'd rather sequence it *with* #1003 instead of ahead of it, I'm fine either way — but I don't think you need to block on it.

BTW: I will review current PR ASAP (may be today or at Sunday)

internal/engines/pipeline/namespace_limiter.go
```diff
@@ -0,0 +1,185 @@
...
+// namespace still debit that pool even when the cluster as a whole is
+// heterogeneous; otherwise their usage would go uncounted and a sibling
+// decision could over-allocate.
+func (l *NamespaceLimiter) resolveUnknownAccelerators(decisions []*domain.VariantDecision) {
+	for _, d := range decisions {
+		if constants.IsAcceleratorResolved(d.AcceleratorName) {
+			continue
+		}
+		bucket, excluded, hasPool := l.inventory.resolver.resolve(d.Namespace)
+		if excluded || !hasPool {
```
ev-shindin: The `chargeBucket` fix from `0e36004` closes the common case, but not this one: `resolveUnknownAccelerators` resolves via `resolver.resolve()`, which returns `("", true, false)` for an excluded namespace, so this guard skips it and the accelerator name is never filled in. `calculateUsedByBucket` then drops the still-unresolved decision, so `chargeBucket` - which *would* charge an excluded namespace to the shared `default` pool is never reached.

`resolve()` and `chargeBucket()` disagree on excluded namespaces: one says "no pool", the other says "charge to default". Trigger: an excluded system namespace sharing `default`, whose decision arrives with an unresolved accelerator name (the homogeneous-cluster case this function exists for) -> its running GPUs aren't debited -> a non-excluded tenant in `default` gets GPUs that are physically occupied. Same overcommit we flagged originally, just via the unresolved-accelerator path.

Fix: resolve unknown accelerators for excluded namespaces too (resolution only needs a bucket, which `chargeBucket` supplies), and add a test for an excluded, `default`-sharing namespace whose accelerator arrives unresolved.

internal/engines/pipeline/namespace_inventory.go
```diff
@@ -0,0 +1,405 @@
+/*
```
ev-shindin: AGENTS.md: *"Do not include license headers in source files."*

internal/config/config.go
```diff
@@ -505,6 +510,24 @@ func (c *Config) UpdateSaturationConfigForNamespace(namespace string, config map
 
 // UpdateScaleToZeroConfig updates the global scale-to-zero configuration.
 // Thread-safe. Takes a copy of the provided map to prevent external modifications.
+// LimiterConfig returns the namespace-scoped GPU inventory limiter
+// configuration. It is populated once at startup; the engine constructs the
+// namespace limiter from it during initialization, so changes require a
+// controller restart. Thread-safe.
+func (c *Config) LimiterConfig() LimiterConfig {
```
ev-shindin: `LimiterConfig()` / `UpdateLimiterConfig()` were inserted inside `UpdateScaleToZeroConfig`'s godoc. Move the two new methods out so each godoc attaches correctly.

internal/engines/saturation/engine.go
```diff
@@ -299,6 +326,28 @@ func (e *Engine) RegisterAnalyzer(name string, a domain.Analyzer) error {
 	return nil
 }
 
+// buildNamespaceLimiter constructs the namespace-scoped GPU limiter from the
+// limiter config, or returns (nil, nil) when no namespace-inventory limiter is
+// configured. Node label selectors are compiled here so an invalid selector is
+// reported once at startup rather than per cycle.
+func buildNamespaceLimiter(lc config.LimiterConfig, disc discovery.NodeDiscovery) (pipeline.Limiter, error) {
```
ev-shindin: Nit: `buildNamespaceLimiter`'s invalid-selector error (and the `NewEngine` "log + disable, don't block startup" fallback) has no test. Add one with a bad `MatchExpressions` operator asserting the wrapped error and a `nil` limiter.

internal/engines/pipeline/namespace_inventory.go
```diff
@@ -0,0 +1,405 @@
...
+
+// SetUsed satisfies the Inventory interface. NamespaceInventory requires
+// per-namespace usage, which the type-keyed map cannot express, so this is a
+// no-op: NamespaceLimiter supplies usage via SetUsedByBucket instead.
+func (i *NamespaceInventory) SetUsed(_ map[string]int) {}
+
+// SetUsedByBucket records GPU usage per (bucket, accelerator type). The keys
+// are bucket names (as returned by the resolver), so the caller must remap
+// workload namespaces onto their buckets before calling. Copies the input.
+func (i *NamespaceInventory) SetUsedByBucket(usedByBucketType map[string]map[string]int) {
```
ev-shindin, line 223: Nit: `SetUsedByBucket` hand-rolls a nested-map deep copy; `copyNestedIntMap` (`quota_inventory.go`) already does exactly this `map[string]map[string]int` copy, can you just reuse?

internal/engines/pipeline/namespace_inventory.go
```diff
@@ -0,0 +1,405 @@
...
+	if sel, has := i.selectors[DefaultSelectorKey]; has && sel.Matches(set) {
+		return DefaultSelectorKey, false, true
+	}
+	return "", false, false
+}
+
+// bucketAcceleratorTypes returns the accelerator types present in a bucket's
+// pool. Used by NamespaceLimiter to resolve unknown accelerator names against
+// the namespace's own pool rather than the cluster-wide aggregate.
+func (i *NamespaceInventory) bucketAcceleratorTypes(bucket string) []string {
```
ev-shindin, line 209: Nit: `bucketAcceleratorTypes` builds its keys slice with a manual loop; `slices.Collect(maps.Keys(...))` is the in-package pattern (`default_limiter.go:253`).

internal/engines/pipeline/namespace_inventory.go
```diff
@@ -0,0 +1,405 @@
...
+	// Pre-create every configured bucket so a named selector matching no nodes
+	// yields an explicit empty pool (its namespaces are denied scale-up) rather
+	// than silently inheriting another bucket's capacity.
+	limitByBucketType := make(map[string]map[string]int, len(i.selectors))
+	for bucket := range i.selectors {
+		limitByBucketType[bucket] = make(map[string]int)
+	}
+
+	// Named bucket keys in deterministic order for unambiguous node assignment.
+	namedBuckets := make([]string, 0, len(i.selectors))
```
ev-shindin: `namedBuckets` uses manual filter-append + `sort.Strings`; `slices.Sort` / `slices.Sorted(maps.Keys(...))` is the package convention and drops the `sort` import.

ev-shindin (changes requested):
Thanks @jia-gao — rebase and both conflict-resolution choices look right, and the core allocation logic is well tested. Requesting changes on two things: the excluded-namespace charge-back is only partially closed (re-opening our earlier thread), and the new files carry license headers AGENTS.md forbids. A few non-blocking nits inline.

jia-gao:
Thanks @ev-shindin, all four points are fixed.

The excluded namespace case you found was real. `resolveUnknownAccelerators` now takes its bucket from `chargeBucket` rather than `resolve`, so the two agree, and an excluded namespace sharing `default` gets resolved and charged. Added a test for exactly that case, and confirmed it fails against the old code. License headers are gone, the godoc split is fixed, and the nits are done: `copyNestedIntMap`, `slices.Collect`, `slices.Sorted`, plus tests for `buildNamespaceLimiter`'s error path.

One larger change while rebasing. #1486 made the inline `limiters:` list the sole source of limiter selection and removed `Config.limiter`, which made this PR's separate `wva-limiter-config` ConfigMap a second mechanism for the same thing. So I folded `namespace-inventory` into the inline schema as a limiter type with a `selectors` field, selected by `EffectiveLimiterMode` and built in `NewLimiterFromConfig`. It picks up live reload for free, and the separate ConfigMap, its parser, the bootstrap wiring, the sample, and the engine's `nsLimiter`/`buildNamespaceLimiter` are all deleted. A `quota` entry still wins over `namespace-inventory`, since composing declared caps with physical partitioning belongs to #1003.

The integration test now drives the real path, inline config to `EffectiveLimiterMode` to the factory, instead of a test-only builder. Let me know if you'd rather the schema change land separately.

internal/config/quota_limiter.go
```diff
@@ -76,9 +78,16 @@ type QuotaLimiterConfig struct {
...
+	// constraint applied). Meaningful when Scope == QuotaScopeNamespace, and
+	// for the namespace-inventory limiter.
 	Exclude []string `yaml:"exclude,omitempty" json:"exclude,omitempty"`
+
+	// Selectors applies when Type == LimiterTypeNamespaceInventory. Keys are
+	// namespace names (with the reserved key `default` matching any namespace
+	// not explicitly listed); values are node label selectors whose matching
+	// nodes form that namespace's GPU pool. Unlike quotas, which are declared
+	// caps, these partition physically discovered capacity.
+	Selectors map[string]metav1.LabelSelector `yaml:"selectors,omitempty" json:"selectors,omitempty"`
```
ev-shindin: `metav1.LabelSelector` has only `json` tags, and the ConfigMap is parsed with `gopkg.in/yaml.v3` (`configmap_helpers.go:40`), which ignores them. An empty selector becomes `labels.Everything()`, so the lexicographically-first bucket takes all GPUs and every other tenant gets zero - silently, and validation passes. Any operator copying the docs hits this. **Fix:** parse the `limiters` block with `sigs.k8s.io/yaml` + `UnmarshalStrict` (already vendored; honors `json` tags), and add a YAML round-trip test.

internal/config/saturation_scaling.go
```diff
@@ -478,9 +482,23 @@ func (c *SaturationScalingConfig) validateLimiters() error {
...
+			// meaningless here; exclude is shared and stays allowed.
+			if l.Scope != "" || len(l.ClusterQuotas) > 0 || len(l.NamespaceQuotas) > 0 {
+				return fmt.Errorf("limiters[%d] (type %q): must not set quota fields (scope/quotas/namespaceQuotas)", i, l.Type)
+			}
+			if len(l.Selectors) == 0 {
+				return fmt.Errorf("limiters[%d] (type %q): requires at least one entry in selectors", i, l.Type)
+			}
+			for ns, sel := range l.Selectors {
+				if _, err := metav1.LabelSelectorAsSelector(&sel); err != nil {
+					return fmt.Errorf("limiters[%d] (type %q): invalid node selector for namespace %q: %w", i, l.Type, ns, err)
```
ev-shindin: Validation only checks the selector *compiles*. An empty `LabelSelector{}` compiles to `labels.Everything()` (no error), so a named namespace with an empty selector grabs every node. Reject an empty/match-all selector.

internal/engines/pipeline/namespace_inventory.go
```diff
@@ -0,0 +1,375 @@
...
+// the decision; an unresolved type in a heterogeneous cluster yields zero
+// (the limiter resolves it for homogeneous clusters before allocation).
+func (a *namespaceTypeAllocator) TryAllocate(_ context.Context, decision *domain.VariantDecision, gpusRequested int) (int, error) {
+	if gpusRequested <= 0 {
+		return 0, nil
+	}
+
+	bucket, excluded, hasPool := a.resolver.resolve(decision.Namespace)
+	if excluded {
+		// Bypass: this limiter applies no constraint to excluded namespaces.
```
ev-shindin: The excluded branch returns the full request without decrementing `remainingByBucket`. Accounting only charges pre-cycle `CurrentReplicas`, so an excluded namespace that grows in the same batch as a capped tenant sharing `default` lets both take the same GPUs (one-cycle overcommit). **Fix:** debit the bucket here too; skip only the availability check. Add a growth test (existing tests use `Current==Target`).

internal/engines/saturation/engine.go
```diff
@@ -746,15 +752,25 @@ func (e *Engine) optimizeV1(
...
+			"limiter", limiter.Name(),
 			"decisionCount", len(allDecisions))
 
 		decisionPtrs := make([]*domain.VariantDecision, len(allDecisions))
 		for i := range allDecisions {
 			decisionPtrs[i] = &allDecisions[i]
 		}
 
-		if err := e.currentGPULimiter().Limit(ctx, decisionPtrs); err != nil {
+		if err := limiter.Limit(ctx, decisionPtrs); err != nil {
```
ev-shindin, line 816: On error the engine applies the unconstrained decisions; since `Refresh()` runs inside `Limit()`, a transient discovery failure silently disables isolation for the cycle. An isolation control should fail closed. (Pre-existing call site - may be out of scope, but worth a decision.)

internal/config/quota_limiter.go
```diff
@@ -144,6 +153,9 @@ func (q QuotaLimiterConfig) clone() QuotaLimiterConfig {
 	if q.Exclude != nil {
 		out.Exclude = slices.Clone(q.Exclude)
 	}
+	if q.Selectors != nil {
+		out.Selectors = maps.Clone(q.Selectors)
```
ev-shindin: `maps.Clone` shares each selector's `MatchLabels`/`MatchExpressions`, breaking the "deep copy" contract (`config.go`); mutating a nested field corrupts the store outside `c.mu`. **Fix:** `out.Selectors[ns] = *sel.DeepCopy()` per entry (as `NamespaceQuotas` already does).

internal/engines/pipeline/namespace_inventory.go
```diff
@@ -0,0 +1,375 @@
...
+
+// chargeBucket returns the bucket a namespace's existing GPU usage should be
+// charged against, ignoring exclusion. Excluded namespaces bypass the cap, but
+// their running replicas still occupy physical GPUs in the pool they draw from
+// (named bucket if the namespace has an explicit selector, else the shared
+// default), so their usage must debit that pool — otherwise the pool hands out
+// GPUs that are already gone, diverging from the cluster-wide path that charges
+// all usage. ok is false when no bucket applies (no matching selector and no
+// default), in which case the usage cannot be attributed to any pool.
+func (r bucketResolver) chargeBucket(namespace string) (bucket string, ok bool) {
```
ev-shindin, line 60: A namespace both excluded and named: uncapped (`resolve`) but charged to its own bucket (`chargeBucket`). No test, and validation doesn't reject the overlap. Add a test or reject it.

internal/config/quota_limiter.go
ev-shindin, line 1: Namespace-inventory reuses the quota plumbing - same `Limiters []QuotaLimiterConfig` list, same struct (now carrying `Selectors` beside quota-only fields), same `clone()`/validation, a parallel `EffectiveNamespaceInventoryEntry()`. That overloading, and the reuse being only *partly* extended, is the shared root cause of several findings above:
-  a `metav1.LabelSelector` in a yaml.v3-parsed struct.
- `Selectors` added to the shared quota `clone()`.
- Live-reload: `limiterSignature` (`engine.go:~418`, unchanged by this PR) still fingerprints `EffectiveQuotaEntries()` only, so editing a namespace-inventory entry's `Selectors`/`Exclude` skips the rebuild — contradicting the doc's no-restart claim. Fix: include `EffectiveNamespaceInventoryEntry()` in the signature (+ reload test).

Recommend: parse the `limiters` block with `sigs.k8s.io/yaml` + `UnmarshalStrict`, and give namespace-inventory its own config type with its own deep-copy/validation.

docs/user-guide/admin/namespace-inventory.md
```diff
@@ -0,0 +1,134 @@
...
+the limiter for the pipeline. This is deliberate: each node is assigned to
+exactly one bucket, so per-bucket capacity never exceeds physical capacity and a
+separate cluster-total cap on top would be redundant. One consequence is
+intentional — **nodes matching no selector and no `default` contribute to no
+pool**, so their GPUs are unusable until those nodes are labeled or a `default`
+selector is added.
+
+Excluded namespaces are unconstrained by this limiter (they are never capped),
+but their existing usage is still charged against the pool they draw from so a
+shared pool is not overcommitted. A namespace whose selector matches no GPU
```
ev-shindin, line 130: The limiter accounts as if pods run on their selector's nodes but doesn't pin them; without taints/PodNodeSelector, pods can land on another bucket's nodes and overcommit it. Document this as a prerequisite.

ev-shindin (changes requested):

ev-shindin (changes requested):
Thanks @jia-gao ! several items still need to be resolved. Also please resolve sign-off issue marked by DCO check.

jia-gao:
Thanks @ev-shindin, all eight fixed and DCO signed off.

The selector parsing one was the real find. Decoding our own documented YAML with yaml.v3 gives `MatchLabels = nil`, so every selector became match-all and the first bucket would take the whole cluster. Your root cause call was right: reusing the quota struct caused it. Namespace-inventory now has its own `NodeSelector` type with yaml tags and its own `NamespaceInventoryConfig` with its own validation and deep copy, plus YAML round-trip tests over the doc example.

I did not switch to `sigs.k8s.io/yaml` + `UnmarshalStrict` though. `SaturationScalingConfig` uses yaml tags throughout, so decoding it through a json-tag path would break every other field. The yaml-tagged selector type fixes the bug without touching the parser. Happy to add `UnmarshalStrict` separately if you still want it.

The rest: empty selectors and excluded/named overlap now rejected; excluded namespaces debit the bucket on allocation so in-batch growth can't overcommit; `limiterSignature` covers the namespace-inventory entry so selector and exclude edits rebuild; per-entry deep copy of the nested selector fields; docs call out node pinning as a prerequisite.

On fail-closed, I scoped it to this limiter: it holds scale-up at current replicas when refresh fails, since a transient discovery error shouldn't lift isolation. Left the cluster-wide and quota paths alone, a broader policy feels like its own change.

PR #1223, opened 2026-05-31
docs: add scaling optimizers developer guide

ev-shindin:
Document the two scaling optimizers in the V2 pipeline that turn analyzer capacity signals into per-variant replica targets:

- CostAwareOptimizer (cost-aware, unlimited mode): per-model, minimizes cost
- GreedyByScoreOptimizer (greedy-by-score, limited mode): fair-shares scarce GPUs across models, with role-aware allocation for P/D disaggregated models

Covers the ScalingOptimizer interface, inputs/outputs, optimizer selection via enableLimiter, both scale-up/scale-down algorithms, resource constraints, fair-share iteration, shared helpers, observability (wva_optimizer_active), and a guide for adding new optimizers. Indexed in docs/README.md.


lionelvillard:
Thanks @ev-shindin. We are also missing user-facing documentation: https://github.com/llm-d/llm-d/tree/main/docs/architecture/advanced/autoscaling. I don't think we need much, at least "priority score" should be defined and the Configuration section should have a subsection documenting this feature with an example.

PR #1224, opened 2026-06-01
feat(nightly): implement real-vLLM E2E V2 saturation tests for OCP nightly

mamy-CS:
Nightly e2e run on OCP cluster with KEDA + real vLLM + EPP gateway (scale up and down)
```
[AfterSuite] PASSED [0.918 seconds]

Ran 4 of 57 Specs in 1064.883 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 53 Skipped
--- PASS: TestE2E (1064.89s)
PASS
ok      github.com/llm-d/llm-d-workload-variant-autoscaler/test/e2e     1066.703s

```


docs/plans/nightly-real-vllm-e2e.md
```diff
@@ -0,0 +1,221 @@
...
+## Why not sustained load?
+
+Sustained load (continuous traffic for the test duration) is non-deterministic:
+vLLM throughput varies with GPU state, model warm-up, and scheduler behaviour. Saturation
+may appear at different times across runs, making timing-based assertions unreliable.
+
+The approach used here is a **controlled burst**: send enough concurrent requests to
+guarantee the queue fills immediately, verify WVA detects saturation and the scaler responds,
+then let the burst complete and verify scale-down. This is deterministic because:
+
```
asm582: Does this section make sense with EPP and flow controller enabled?

docs/plans/nightly-real-vllm-e2e.md
```diff
@@ -0,0 +1,221 @@
+# Nightly Real-vLLM E2E Test Plan
```
asm582: Why can't we deploy WVA with a precise-prefix path and simplify OpenShift testing?

asm582 (commented):
Added comments.

docs/plans/nightly-real-vllm-e2e.md
mamy-CS: Good point, I updated the diagram to show EPP + flow controller in the inference path. The controlled burst section now explicitly states the assumption. The flow controller does not throttle the burst below the saturation threshold. This holds empirically, I tested with 10 concurrent requests, and they reached vLLM and saturated the queue on the nightly cluster (Llama-3.1-8B, max-num-seqs=1). If the flow controller threshold is tightened in the future, the burst size (default 10) can be increased independently.

docs/plans/nightly-real-vllm-e2e.md
mamy-CS: yes, will be simplified after some bugs fixed upstream. Not a blocker here though.

asm582: ok lets just use that path to test OpenShift and call it done?

docs/plans/nightly-real-vllm-e2e.md
```diff
@@ -0,0 +1,224 @@
...
+## Test design
+
+### New Ginkgo label: `nightly`
+
+A new `Label("nightly")` is added alongside the existing `smoke` / `full` labels. Nightly
+tests may assume:
+- `USE_SIMULATOR=false` (real vLLM)
+- `ENVIRONMENT=openshift`
+- `GATEWAY_NAME` env var is set (by the reusable workflow) and the gateway is reachable from within the cluster
+- `max_num_seqs=1` (set via the upstream kustomize patch described below) so any second concurrent request immediately queues
```
asm582: If we are using a well-lit path guide, why are we tuning models?

docs/plans/nightly-real-vllm-e2e.md
```diff
@@ -0,0 +1,224 @@
...
+new VA. Gateway address is discovered from `GATEWAY_NAME` using the same cluster-lookup
+pattern as the existing e2e tests (not a hardcoded URL).
+
+**Shared burst mechanism**: submit a curl Job with `N = threshold + 2` concurrent slow
+requests (long `max_tokens`). With `max_num_seqs=1`, all but one queue immediately.
+The Job is cleaned up in `AfterEach` regardless of outcome.
+
+---
+
+#### Test 1 — V1 threshold-based analyzer (to be deprecated once V2 is fully validated)
```
asm582: Nit: deprecation message should be removed from this doc.

asm582 (changes requested):
Address comments

asm582:
@mamy-CS are we still working on this?

mamy-CS:
> 

yes, but the challenge remains GPU availability for the runs.

PR #1225, opened 2026-06-01
engines/saturation: multi-analyzer registry

deanlorenz:
## Summary

- Adds an analyzer registry on `Engine`: `analyzers []analyzerEntry` with `RegisterAnalyzer(name, interfaces.Analyzer)`. Saturation V2 is pre-registered at slot 0 by `NewEngine`; external analyzers (throughput, SLO, …) are registered from `cmd/main.go` before `StartOptimizeLoop`.
- The optimize loop iterates the registry snapshot and invokes `Analyze` on every registered non-saturation analyzer. On this branch saturation continues to drive every scaling decision; non-saturation results are intentionally discarded — combine semantics, score consumption, and per-analyzer threshold application land in follow-up PRs (`multi-analyzer-optimizer`, `multi-analyzer-threshold`).
- Race-safety: `StartOptimizeLoop` snapshots the registry to `analyzersSnapshot` and flips `started` before launching the goroutine; later `RegisterAnalyzer` calls panic with `RegisterAnalyzer called after StartOptimizeLoop`. Goroutine launch establishes happens-before, so the loop is race-free without runtime locking.
- Resilience: each registered non-saturation analyzer runs in isolation — errors are logged and discarded, panics are recovered. A faulty plugin analyzer cannot take down the optimize goroutine or block saturation.
- Programmer-error semantics: re-registering an existing name **panics**. The registry is not a hot-swap mechanism; a duplicate registration is a wiring mistake worth surfacing immediately.

## Context

This PR replaces the registration scope previously carried on #1113. Per the design discussion in `planning/PR1113-review.md`, that PR is being split into three independent multi-analyzer-support PRs that can land in any order:

1. **This PR** — analyzer registration (Item 3).
2. **`multi-analyzer-threshold`** — universal threshold post-step applied to every analyzer's result (Item 2).
3. **`multi-analyzer-optimizer`** — delete the engine-side combine; per-analyzer slice flows to the optimizers (Item 1).

#1113 will be closed once this one is reviewed and merged.

## Behavior change worth flagging

`RegisterAnalyzer` previously replaced-in-place on duplicate name; this PR makes it panic. No caller does duplicate registration today (saturation is the only registered analyzer on `main`), so blast radius is zero — but the change is intentional.

## Test plan

- [x] `make build` — clean
- [x] `golangci-lint run` — 0 issues
- [x] `make test` — full unit-test sweep green
- [x] `go test -race ./internal/engines/saturation/...` — clean
- [x] All 3 kustomize overlays build clean (cluster-scoped/k8s, namespace-scoped/openshift, namespace-scoped/k8s)
- [x] DCO sign-off on all 3 commits

Test additions in `internal/engines/saturation/engine_register_test.go`:
- T1 — saturation pre-registered at slot 0 by `NewEngine`
- T2 — `RegisterAnalyzer` appends in registration order
- T3 — duplicate-name registration panics with a clear message
- T4 — `RegisterAnalyzer` after `StartOptimizeLoop` panics with the documented message
- T5 — `StartOptimizeLoop` snapshots the registry and flips `started` before launching the executor
- T6 — one cycle calls `Analyze` on every registered non-saturation analyzer exactly once, in registration order
- T7 — a registered analyzer that returns an error does not abort the cycle
- T8 — a registered analyzer that panics is recovered; cycle continues
- T9 — under \`-race\`, the snapshot reader does not race with concurrent post-Start \`RegisterAnalyzer\` attempts (4 writers all panic; reader iterates 200×)

T10 (saturation's `*AnalyzerResult` is what flows to the optimizer regardless of what other analyzers return) is verified by inspection: `runAnalyzersAndScore` returns `baseResult` and `runRegisteredAnalyzers` discards every non-saturation result. Happy to add behavioral coverage if reviewers prefer.

### DIFF
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -149,7 +149,7 @@ type Engine struct {
 
 	// analyzers is the engine's analyzer registry, mutated only during setup
 	// (NewEngine + RegisterAnalyzer). After StartOptimizeLoop it is frozen —
-	// further RegisterAnalyzer calls panic. The optimize goroutine reads
+	// further RegisterAnalyzer calls return an error. The optimize goroutine reads
 	// analyzersSnapshot, never analyzers, so iteration is race-free without
 	// runtime locking.
 	analyzers []analyzerEntry
@@ -163,8 +163,8 @@ type Engine struct {
 	analyzersSnapshot []analyzerEntry
 
 	// started transitions to true in StartOptimizeLoop. Late RegisterAnalyzer
-	// calls panic so the contract "register before Start" is enforced rather
-	// than just documented.
+	// calls return an error so the contract "register before Start" is enforced
+	// rather than just documented.
 	started bool
 
 	// optimizer is the V2 scaling optimizer that produces VariantDecisions from
@@ -276,7 +276,7 @@ func (e *Engine) RegisterAnalyzer(name string, a interfaces.Analyzer) error {
 //
 // Before launching the goroutine, the registered analyzers are snapshotted
 // to a frozen slice that runAnalyzersAndScore iterates. The started flag is
-// flipped so subsequent RegisterAnalyzer calls panic. The snapshot is the
+// flipped so subsequent RegisterAnalyzer calls return an error. The snapshot is the
 // natural place to invoke any future per-analyzer Init(ctx) hook.
 func (e *Engine) StartOptimizeLoop(ctx context.Context) {
 	e.analyzersSnapshot = make([]analyzerEntry, len(e.analyzers))
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -52,6 +52,14 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/utils/scaletarget"
 )
 
+// analyzerEntry binds a registered analyzer to its name. The engine stores
+// these in registration order so runAnalyzersAndScore iterates analyzers
+// deterministically.
+type analyzerEntry struct {
+	name     string
+	analyzer interfaces.Analyzer
+}
+
 // v1Analyzer is the minimal surface of *saturation.Analyzer that optimizeV1
 // depends on. Defined here so tests can substitute a stub via Engine's
 // v1AnalyzerFactory field without exposing a public interface.
@@ -128,6 +136,7 @@ type Engine struct {
 	metricsRegistry *source.SourceRegistry
 
 	// saturationV2Analyzer is the V2 token-based saturation analyzer (initialized once).
+	// Also pre-registered in analyzers under interfaces.SaturationAnalyzerName.
 	saturationV2Analyzer *saturation_v2.SaturationAnalyzer
 
 	// queueingModelAnalyzer is the queueing model-based analyzer (initialized once).
@@ -136,6 +145,14 @@ type Engine struct {
 
 	// capacityStore is shared with the V2 analyzer for caching capacity knowledge.
 	capacityStore *saturation_v2.CapacityKnowledgeStore
+
+	// analyzers is the engine's analyzer registry, iterated in registration
+	// order by runAnalyzersAndScore. NewEngine pre-registers the V2 saturation
+	// analyzer; RegisterAnalyzer appends additional analyzers (e.g. throughput).
+	// Saturation always runs and drives scaling decisions; other registered
+	// analyzers are invoked but their results are not consumed yet — combine
+	// and per-analyzer threshold logic lands in follow-up PRs.
+	analyzers []analyzerEntry
 
 	// optimizer is the V2 scaling optimizer that produces VariantDecisions from
 	// AnalyzerResults. Selected per-cycle based on enableLimiter config:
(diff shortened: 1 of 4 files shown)


internal/engines/saturation/engine.go
```diff
@@ -220,9 +253,36 @@ func NewEngine(client client.Client, scheme *runtime.Scheme, recorder record.Eve
 	return &engine
 }
 
+// RegisterAnalyzer adds an external analyzer to the engine's analyzer
+// registry. Must be called before StartOptimizeLoop; calling it after
+// panics with a clear message so the "register before Start" contract is
+// enforced rather than silently corrupting concurrent state.
+// Re-registering an existing name also panics — the registry is not a
+// hot-swap mechanism. The analyzer is appended in registration order.
+func (e *Engine) RegisterAnalyzer(name string, a interfaces.Analyzer) {
```
ev-shindin: **`panic` vs `error` / naming.** `RegisterAnalyzer` panics on misuse, which diverges from the repo convention. Setup-time panic-on-programmer-error is idiomatic (cf. `prometheus.MustRegister`), but the plain name `RegisterAnalyzer` (no `Must`) signals error-return semantics by convention. Consider renaming to **`MustRegisterAnalyzer`** or returning an `error`.

internal/engines/saturation/engine_v2.go
```diff
@@ -99,14 +111,32 @@ func (e *Engine) runAnalyzersAndScore(
...
 	}
 
-	// Compute weighted score from enabled analyzers
+	// Build AnalyzerInput once; shared by all non-saturation analyzers.
+	input := interfaces.AnalyzerInput{
+		ModelID:        modelID,
+		Namespace:      namespace,
+		ReplicaMetrics: replicaMetrics,
+		VariantStates:  variantStates,
+		Config:         &config,
```
ev-shindin, line 133: Non-saturation analyzers receive `&config` *after* saturation's threshold overrides were applied to it. Harmless now (results discarded), but per-analyzer override resolution leaking saturation's values is a small smell — noted as tracked on another branch.

internal/engines/saturation/engine_v2.go
```diff
@@ -123,6 +153,52 @@ func (e *Engine) runAnalyzersAndScore(
...
+) {
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Error(fmt.Errorf("panic: %v", r),
+				"registered analyzer panicked; result discarded on this branch",
+				"name", entry.name, "modelID", modelID)
+		}
+	}()
+	if _, err := entry.analyzer.Analyze(ctx, input); err != nil {
+		logger.Error(err, "registered analyzer failed; result discarded on this branch",
```
ev-shindin: A persistently failing/panicking registered analyzer logs at `Error` *every* optimize cycle with no functional impact. Consider sampling or a lower level.

ev-shindin (commented):
Thanks @dean! Please address my comment

internal/engines/saturation/engine.go
deanlorenz: Good point — renamed to `MustRegisterAnalyzer` in the latest push. Matches the `prometheus.MustRegister` / `regexp.MustCompile` convention.

internal/engines/saturation/engine_v2.go
deanlorenz: Noted — this is from the precursor saturation-only threshold-override loop that mutates the local `config` copy before `AnalyzerInput` is built. Harmless on this branch (non-saturation results discarded). Added a code comment pointing at PR #1228 (`multi-analyzer-threshold`) where the clean fix lands: engine applies thresholds universally after each analyzer runs rather than pre-mutating the config.

internal/engines/saturation/engine_v2.go
deanlorenz: Good catch — downgraded to `V(DEBUG).Info` in the latest push. Plugin failure is non-fatal and shouldn't show up as an engine `Error` in operator logs on every cycle.

deanlorenz (commented):
Thanks for the review! Addressed in the latest push:
- `RegisterAnalyzer` → `MustRegisterAnalyzer` (comment 1)
- Plugin-failure logging downgraded from `Error` to `V(DEBUG).Info` (comment 3)
- Comment 2 (threshold override leak into non-saturation `AnalyzerInput`): harmless on this branch, added a code comment tracking it to PR #1228.

internal/engines/saturation/engine.go
deanlorenz: Changed `RegisterAnalyzer` to return `error` instead of panicking — reverts the `MustRegisterAnalyzer` rename from the previous push. Both misuse conditions (called after `StartOptimizeLoop`, duplicate name) now return descriptive errors rather than panicking.

ev-shindin: The only thing that remain to change: panic -> error in docstrings `engine.go` 3 places, `engine_register_test.go` - 1 place

deanlorenz: Fixed in 5c73ea5 — updated the three struct field comments (`analyzers`, `started`) and the `StartOptimizeLoop` docstring in `engine.go`, and the race-safety test comment in `engine_register_test.go`. The remaining `panic` references in the file are `NewEngine`'s nil-guard (unrelated) and `spyAnalyzer`'s panic field used to simulate panicking analyzers in the recovery test — those are intentional.

ev-shindin (approved):
Thanks @deanlorenz

PR #1227, opened 2026-06-02
Add kubernetes namespace based quota proposal

asm582:
Kubernetes-based namespace quotas where kube quotas equal capacity.

### DIFF
--- a/docs/proposals/namespace-quota-capacity-discovery.md
+++ b/docs/proposals/namespace-quota-capacity-discovery.md
@@ -1,0 +1,552 @@
+Namespace Quota Capacity Discovery
+
+**Authors:** Abhishek Malvankar
+**Status:** Provisional
+**Created:** 2026-06-02
+**Last Updated:** 2026-06-02
+
+---
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Notes and Constraints](#notes-and-constraints)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [Activation](#activation)
+  - [Accelerator Name Convention](#accelerator-name-convention)
+- [Test Plan](#test-plan)
+  - [Unit Tests](#unit-tests)
+  - [Integration Tests](#integration-tests)
+  - [E2E Tests](#e2e-tests)
+- [Production Readiness Review](#production-readiness-review)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+
+---
+
+## Summary
+
+Add `NamespaceQuotaDiscovery`, a new implementation of the existing `CapacityDiscovery`
+interface that derives GPU capacity from Kubernetes `ResourceQuota` objects rather than
+from GPU Operator node labels. When enabled, the greedy optimizer respects the quota
+hard limit as the total GPU budget for the namespace — no GPU Operator, no node-level
+access required.
+
+---
+
+## Motivation
+
+The default capacity discovery backend (`K8sWithGpuOperator`) requires the
+[NVIDIA GPU Operator](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/overview.html)
+to label nodes with GPU model and memory information. This creates a hard dependency that
+rules out a class of deployment environments:
+
+- **Managed cloud services** (GKE, EKS, AKS) often do not expose GPU Operator labels on
+  their managed node pools.
+- **Multi-tenant clusters** commonly restrict node listing to cluster administrators,
+  making per-node inventory inaccessible to namespace-scoped controllers.
+- **Quota-governed namespaces** (common in OpenShift and enterprise on-prem clusters)
+  already encode GPU capacity in `ResourceQuota.spec.hard` — this is the authoritative
+  source and scanning nodes would duplicate it.
+
+In each of these cases a namespace-scoped quota is both available and sufficient: the
+quota hard limit already equals the GPU capacity visible to workloads in the namespace.
+Reading it directly is simpler and more accurate than reverse-engineering capacity from
+node labels.
+
+### Goals
+
+1. Add a `NamespaceQuotaDiscovery` backend that reads GPU capacity from
+   `ResourceQuota.spec.hard` and current usage from `ResourceQuota.status.used`.
+2. Plug it into the existing `CapacityDiscovery` / `UsageDiscovery` interfaces with zero
+   changes to `TypeInventory`, `DefaultLimiter`, or `GreedyByScoreOptimizer`.
+3. Activate it via a single environment variable; the default path is unchanged.
+4. Work on any Kubernetes distribution — GKE, EKS, AKS, OpenShift, kind — without
+   requiring GPU Operator or node-level RBAC.
+
+### Non-Goals
+
+- Replacing `K8sWithGpuOperator` — it remains the default and is required for
+  GPU-model-aware scheduling (e.g., differentiating A100 from H100).
+- Cross-namespace capacity aggregation.
+- Dynamic reconfiguration without a controller restart.
+- Supporting quota scopes (e.g., `BestEffort`, `Terminating`) — only the namespace-wide
+  hard limit is read.
+
+---
+
+## Proposal
+
+### User Stories
+
+**Story 1 — Managed cloud operator**
+
+> As a platform engineer running WVA on GKE Autopilot, I cannot access GPU Operator
+> node labels, but my namespace has a `ResourceQuota` limiting `nvidia.com/gpu: 8`.
+> I want the greedy optimizer to treat 8 as the total GPU budget so it does not
+> over-schedule replicas beyond what the cluster can place.
+
+**Story 2 — Multi-tenant cluster operator**
+
+> As an operator on a shared OpenShift cluster, I have namespace-admin access but
+> cannot list nodes cluster-wide. My namespace quota is the authoritative GPU limit.
+> I want WVA to read that quota rather than failing with a permissions error.
+
+**Story 3 — Quota-first operations**
+
+> As a platform team that manages GPU capacity via quota rather than node labels,
+> I want a single source of truth: the quota. I do not want to maintain GPU Operator
+> labels in parallel.
+
+**Story 4 — Fair GPU rebalancing under shifting load**
+
+> As an ML platform engineer, I am serving model A and model B in the same namespace,
+> each initially at low request volume. Both start with 10% of the cluster's GPUs.
+>
+> Model A's request volume then increases 10×. A naive per-model autoscaler would try
+> to give it 100% of GPUs, but model B is still holding its 10%, so model A ends up
+> with 90%.
+>
+> Later, model B's request volume also increases 10×. Both models are now equally
+> resource-hungry. The correct allocation is 50/50. But a naive independent autoscaler
+> will not rebalance: model A still has a non-empty queue, so it sees no reason to
+> release GPUs. Model B cannot get more GPUs because model A holds them, so model B's
+> queue grows indefinitely.
+>
+> I want the autoscaler to detect this imbalance — using queue size or time-in-queue as
+> the signal — and converge to a 50/50 split. The quota is the hard ceiling on total
+> GPUs; the optimizer is responsible for dividing them fairly across models whose queues
+> are equally saturated.
+>
+> **Why 50/50?** Because both models have equally large queues relative to their current
+> allocation. The fair-share criterion is equal queue size (or equal time-in-queue for
+> models whose work units take different amounts of GPU time). A model holding more than
+> its fair share of GPUs while a peer is equally starved should give resources up, even
+> if it still has its own backlog to drain.
+
+### Notes and Constraints
+
+- **Accelerator key is the resource name, not the model name.** `ResourceQuota` stores
+  `nvidia.com/gpu: 8`, not `NVIDIA-A100-PCIE-80GB: 8`. `TypeInventory` pools will be
+  keyed by `nvidia.com/gpu`.
+
+  **`acceleratorName` on the VA spec is optional.** When unset and the quota defines
+  exactly one GPU vendor (the common single-tenant case), `DefaultLimiter` automatically
+  resolves the empty name to the only pool that exists — `nvidia.com/gpu`. Setting
+  `acceleratorName` explicitly is only required when the namespace quota covers more than
+  one GPU vendor (e.g., both `nvidia.com/gpu` and `amd.com/gpu`), because the limiter
+  cannot determine which pool to debit without a hint.
+
+- **Usage from `status.used`, not pod listing.** The Kubernetes quota admission
+  controller maintains `ResourceQuota.status.used` automatically and is the authoritative
+  source. Re-summing pod requests would be redundant and could diverge transiently.
+
+- **Single namespace.** The implementation reads quotas from the one namespace named in
+  `WVA_CAPACITY_NAMESPACE`. The controller may manage workloads across other namespaces
+  but capacity is governed by this one quota namespace.
+
+- **Multiple quotas are aggregated.** If a namespace has more than one `ResourceQuota`
+  object each quota is treated as a separate inventory entry (outer key in the
+  `Discover()` map). `TypeInventory.Refresh()` sums them as it does for multiple nodes —
+  no change to the aggregation logic is needed.
+
+### Risks and Mitigations
+
+| Risk | Likelihood | Impact | Mitigation |
+|---|---|---|---|
+| Quota larger than available node capacity causes pending pods | Medium | Medium | Document the requirement that quota hard limit must reflect actual node pool size. The optimizer acts conservatively: it schedules up to the quota, but Kubernetes will hold pods pending if nodes are full. |
+| `status.used` is stale immediately after quota creation | Low | Low | The admission controller updates `status.used` synchronously on pod admission; staleness is bounded by the API server response time, well within the 30-second optimization interval. |
+| Multi-vendor quota with unset `acceleratorName` | Low | Low | `acceleratorName` is optional in single-vendor namespaces — the limiter auto-resolves it. In multi-vendor namespaces it must be set; the optimizer skips unresolved variants and logs a warning rather than over-allocating. |
+| Mixed GPU types in one namespace share one pool | Low | Low | Documented as a known limitation. Use `K8sWithGpuOperator` for heterogeneous GPU namespaces. |
+
+---
+
+## Design Details
+
+### New File: `internal/discovery/namespace_quota.go`
+
+`NamespaceQuotaDiscovery` implements `FullDiscovery` by reading `ResourceQuota` objects.
+It is a leaf implementation — it introduces no new interfaces, types, or abstractions.
+
+```
+internal/discovery/
+  interface.go                 # unchanged
+  types.go                     # unchanged
+  k8s_with_gpu_operator.go     # unchanged
+  namespace_quota.go           # NEW
+  namespace_quota_test.go      # NEW
+```
+
+**`Discover(ctx) → map[quotaName]map[resourceName]AcceleratorModelInfo`**
+
+1. List `ResourceQuotaList` in `d.Namespace`.
+2. For each quota, scan `Spec.Hard` for GPU resource names (see table below).
+3. Return each quota as its own outer key so `TypeInventory.Refresh()` can aggregate
+   without any modification.
+
+**`DiscoverUsage(ctx) → map[resourceName]int`**
+
+1. List `ResourceQuotaList` in `d.Namespace`.
+2. Aggregate `Status.Used` values for GPU resource names across all quotas.
+3. Return the map. No pod listing. No inference.
+
+**`DiscoverNodes(ctx) → map[string]NodeInfo`**
+
+Returns an empty map. Quota-based discovery has no per-node information; callers that
+require node affinity must use `K8sWithGpuOperator`.
+
+**Interface compliance assertion:**
+
+```go
+var _ FullDiscovery = (*NamespaceQuotaDiscovery)(nil)
+```
+
+**GPU resource names recognized:**
+
+| Resource name | Vendor |
+|---|---|
+| `nvidia.com/gpu` | NVIDIA |
+| `requests.nvidia.com/gpu` | NVIDIA (prefixed form) |
+| `amd.com/gpu` | AMD |
+| `requests.amd.com/gpu` | AMD (prefixed form) |
+| `intel.com/gpu` | Intel |
+| `requests.intel.com/gpu` | Intel (prefixed form) |
+
+### API Changes
+
+No new CRDs, no new API types, no new flags in the saturation ConfigMap. The only user-
+facing surface is the environment variable and the `acceleratorName` convention on VA specs.
+
+### Activation
+
+`WVA_CAPACITY_NAMESPACE` is read once in `NewEngine()`. Setting it to a non-empty
+namespace switches the discovery backend for the lifetime of the controller process.
+
+```go
+// internal/engines/saturation/engine.go — NewEngine()
+var gpuDiscovery discovery.FullDiscovery
+if ns := os.Getenv("WVA_CAPACITY_NAMESPACE"); ns != "" {
+    gpuDiscovery = discovery.NewNamespaceQuotaDiscovery(client, ns)
+} else {
+    gpuDiscovery = discovery.NewK8sWithGpuOperator(client)
+}
+gpuInventory := pipeline.NewTypeInventoryWithUsage("cluster-gpu-inventory", gpuDiscovery)
+// remaining NewEngine() lines unchanged
+```
+
+This is the only change to `engine.go`.
+
+### Accelerator Name Convention
+
+| Discovery mode | `TypeInventory` pool key | VA `acceleratorName` | Required? |
+|---|---|---|---|
+| `K8sWithGpuOperator` | `A100`, `H100` (normalized from node label) | `A100` | Optional — auto-resolved in homogeneous clusters |
+| `NamespaceQuotaDiscovery`, single vendor | `nvidia.com/gpu` | omit or `nvidia.com/gpu` | **Optional** — limiter auto-resolves to the only pool |
+| `NamespaceQuotaDiscovery`, multi-vendor | `nvidia.com/gpu`, `amd.com/gpu` | `nvidia.com/gpu` or `amd.com/gpu` | **Required** — limiter cannot pick a pool without a hint |
+
+The difference in key convention from `K8sWithGpuOperator` is intentional.
+`NormalizeAcceleratorName` strips vendor and memory suffixes from full GPU model names
+(e.g. `NVIDIA-A100-PCIE-80GB` → `A100`); it is not applied to quota resource names
+because there is no suffix to strip. `nvidia.com/gpu` passes through unchanged.
+
+The auto-resolution behaviour is implemented in `DefaultLimiter.resolveUnknownAccelerators()`:
+when the inventory has exactly one pool and a decision's `AcceleratorName` is unset, the
+limiter fills it in before allocation. This means most single-vendor quota deployments
+require no `acceleratorName` at all on the VA spec.
+
+---
+
+## Test Plan
+
+### Unit Tests
+
+**Location:** `internal/discovery/namespace_quota_test.go`
+
+**Tool:** `controller-runtime/pkg/client/fake` — same pattern as
+`k8s_with_gpu_operator_test.go`. No cluster required.
+
+**Running:**
+```bash
+go test ./internal/discovery/... -run TestNamespaceQuota
+```
+
+**Note on fake client and `status.used`:** The fake client does not compute
+`status.used` from `spec.hard`. Build `ResourceQuota` objects with both fields populated
+explicitly and register them with `WithStatusSubresource()`.
+
+```go
+quota := &corev1.ResourceQuota{
+    ObjectMeta: metav1.ObjectMeta{Name: "gpu-quota", Namespace: "llm-workloads"},
+    Spec: corev1.ResourceQuotaSpec{
+        Hard: corev1.ResourceList{"nvidia.com/gpu": resource.MustParse("8")},
+    },
+    Status: corev1.ResourceQuotaStatus{
+        Used: corev1.ResourceList{"nvidia.com/gpu": resource.MustParse("3")},
+    },
+}
+fakeClient := fake.NewClientBuilder().
+    WithScheme(scheme).
+    WithStatusSubresource(quota).
+    WithRuntimeObjects(quota).
+    Build()
+```
+
+#### `Discover()` coverage
+
+| Test | Input | Expected |
+|---|---|---|
+| Single NVIDIA quota | `spec.hard["nvidia.com/gpu"]: 8` | `{"gpu-quota": {"nvidia.com/gpu": {Count:8}}}` |
+| Two quotas, same namespace | Quota A: 4 GPUs, Quota B: 4 GPUs | Two outer keys; `TypeInventory.Refresh()` aggregates to 8 |
+| No GPU entries in quota | Only CPU and memory limits | Empty map, no error |
+| `requests.nvidia.com/gpu` prefix | Prefixed hard limit: 4 | Counted as `nvidia.com/gpu: 4` |
+| Multi-vendor quota | `nvidia.com/gpu: 4`, `amd.com/gpu: 2` | Both keys present with correct counts |
+| Empty namespace | No `ResourceQuota` objects | Empty map, no error |
+
+For the two-quota case, instantiate a real `TypeInventory` (not a mock) and call
+`Refresh()` to confirm that cross-quota aggregation works end-to-end at the unit level.
+This is the only test that crosses a component boundary at unit level.
+
+#### `DiscoverUsage()` coverage
+
+| Test | Input | Expected |
+|---|---|---|
+| GPUs in use | `status.used["nvidia.com/gpu"]: 3` | `{"nvidia.com/gpu": 3}` |
+| No usage | `status.used` empty or absent | Empty map, no error |
+| Multi-vendor usage | Both `nvidia.com/gpu` and `amd.com/gpu` used | Both keys returned |
+| Multiple quotas | Two quotas each with partial usage | Values summed by resource name |
+
+### Integration Tests
+
+Integration tests validate the full constraint pipeline — from `NamespaceQuotaDiscovery`
+through `TypeInventory` → `DefaultLimiter.ComputeConstraints()` → `GreedyByScoreOptimizer`
+— using a fake client and in-process components. No cluster required.
+
+**Location:** `internal/engines/pipeline/` (alongside existing `type_inventory_test.go`)
+
+**Running:**
+```bash
+go test ./internal/engines/pipeline/... -run TestQuotaConstraints
+```
+
+#### Coverage
+
+| Test | What it validates |
+|---|---|
+| Quota ceiling respected | Two models requesting more GPUs than quota allows; optimizer total allocation ≤ quota hard limit |
+| Zero quota | `nvidia.com/gpu: 0`; optimizer produces zero scale-up decisions |
+| Usage already at limit | `status.used` equals `spec.hard`; optimizer sees zero available GPUs |
+| Usage partially consumed | Available = hard − used; optimizer scales up to the remaining headroom |
+| Quota discovery error | `Discover()` returns an error; `ComputeConstraints()` propagates it; engine falls back to unlimited |
+
+### E2E Tests
+
+E2E tests run against a real cluster with the controller deployed. GPU hardware is not
+required — `--fake-metrics` injects synthetic saturation signals.
+
+**Location:** `test/e2e/quota_limiter_test.go`
+
+**Labels:** `Label("quota-limiter", "full")` for all cases;
+add `Label("smoke")` to the happy-path case if it completes in under 2 minutes.
+
+**Running:**
+```bash
+# Full suite including quota limiter
+make test-e2e-full
+
+# Quota limiter tests only
+make test-e2e-full -- --label-filter="quota-limiter"
+```
+
+#### Coverage
+
+**Happy path — quota ceiling enforced:**
+1. Apply namespace with `ResourceQuota` limiting `nvidia.com/gpu: 4`.
+2. Deploy two `VariantAutoscaling` objects with `gpusPerReplica: 1` and no
+   `acceleratorName` set (auto-resolved to `nvidia.com/gpu` by the limiter), each
+   requesting replicas that would together exceed 4.
+3. Enable `--fake-metrics` to drive both to saturation.
+4. Assert `sum(TargetReplicas * gpusPerReplica) ≤ 4` across both VAs within one cycle.
+
+**Quota updated at runtime:**
+1. Start with `nvidia.com/gpu: 4` and two saturated VAs capped at 4 GPUs total.
+2. `kubectl patch resourcequota gpu-quota --patch '{"spec":{"hard":{"nvidia.com/gpu":"8"}}}'`
+3. Assert optimizer scales up to fill the new headroom within one optimization interval.
+
+**Zero quota:**
+1. Set `nvidia.com/gpu: 0` in the quota.
+2. Assert no scale-up decisions are produced.
+
+**Fallback to unlimited when quota discovery fails:**
+1. Delete the `ResourceQuota` object mid-run.
+2. Assert the controller logs a warning but continues producing decisions
+   (falls back to unlimited mode, not a crash).
+
+---
+
+## Production Readiness Review
+
+### Feature Enablement and Rollback
+
+- **Enabled by:** setting `WVA_CAPACITY_NAMESPACE` on the controller Deployment.
+- **Disabled by:** unsetting the variable and restarting the controller.
+- **Rollback:** remove the env var; the controller reverts to `K8sWithGpuOperator`
+  immediately on restart with no state to clean up.
+- **Side effects of disabling:** the optimizer reverts to node-based capacity. If node
+  capacity exceeds the quota that was previously enforced, the optimizer may produce
+  larger scale-up decisions. This is expected and safe.
+
+### Monitoring Requirements
+
+- Existing `wva_available_gpus` metric continues to be emitted (populated from
+  `TypeInventory.TotalAvailable()` regardless of discovery backend).
+- Log line at `Info` level on every `Discover()` call: `"quota-based capacity discovered"`
+  with `namespace`, `totalGPUs`, and `quotaCount` fields.
+- Log line at `Warning` level when `Discover()` returns an empty map for a non-empty
+  namespace (possible misconfiguration).
+
+### Dependencies
+
+- No new Go module dependencies. `ResourceQuota` is `core/v1`, already imported.
+- No new RBAC beyond what the controller already holds for `resourcequotas` (get, list,
+  watch). Verify the existing `config/rbac/role.yaml` includes these verbs; add them if
+  absent.
+
+### Scalability
+
+- `Discover()` issues one `ResourceQuotaList` API call per optimization cycle (default:
+  30 seconds). A namespace with N quotas returns N objects in a single list response.
+  This is strictly less expensive than the current `listGPUNodes()` which issues one
+  API call per GPU vendor.
+- No watch caching is added in this KEP; the existing client cache handles list responses.
+
+---
+
+## Gap Analysis: Does This KEP Solve Story 4?
+
+**Short answer: No — not on its own.**
+
+This KEP solves Stories 1–3 (discovering GPU capacity from namespace quotas instead of GPU
+Operator node labels). Story 4 requires a different and deeper change in the optimizer.
+This section documents the gap so a follow-up proposal can be scoped correctly.
+
+### Why Story 4 is not solved
+
+Tracing through the code with the concrete scenario (100 total GPUs, Model A: 90 replicas,
+Model B: 10 replicas, both queues now growing equally):
+
+1. **`computeCurrentGPUUsage`** sums current replicas for all models:
+   `{"nvidia.com/gpu": 100}` (90 from A + 10 from B).
+
+2. **`ComputeConstraints`** computes available = quota hard − used = 100 − 100 = **0**.
+
+3. **`mergeConstraints`** gives the optimizer `{"nvidia.com/gpu": 0}` available GPUs.
+
+4. Both models have `RequiredCapacity > 0` (each has a growing queue) so both enter
+   `fairShareScaleUp`.
+
+5. **`fairShareScaleUp` exits immediately** on the first check: `totalGPUs == 0 → break`.
+   Nothing is allocated or reallocated.
+
+6. `buildDecisionsWithOptimizer` produces targets equal to current replicas. Model A stays
+   at 90, Model B stays at 10. **The rebalancing never happens.**
+
+### Root cause
+
+`GreedyByScoreOptimizer` is an **additive** optimizer: it distributes GPUs that are not yet
+allocated. When the cluster is 100% utilized and both models are still saturated, available
+= 0 and the fair-share loop has nothing to hand out.
+
+The optimizer has no mechanism to **preempt** GPUs from a model that holds more than its
+fair share and **redistribute** them to a peer that is equally or more starved. Scale-down
+for a model only triggers when its `SpareCapacity > 0` (i.e., it has excess capacity
+relative to its own demand). A model holding 90% of GPUs but still facing a growing queue
+reports `SpareCapacity = 0` — it will never voluntarily release GPUs under the current
+logic.
+
+### What Story 4 actually requires
+
+A redistributive fair-share pass that runs before or instead of the current additive pass
+when the cluster is fully utilized:
+
+1. **Compute a target fair-share per model.**
+   When N models are all saturated, each is entitled to `total_GPUs / N`. Equal
+   queue size (or equal time-in-queue for models with different work-unit durations) is the
+   right fairness criterion — not equal request rate.
+
+2. **Force scale-down models above their fair share**, even if their own queue is non-empty.
+   Model A holds 90, fair share is 50: target = 50, regardless of its queue.
+
+3. **Allocate freed GPUs to models below their fair share.**
+   Model B holds 10, fair share is 50: target = 50, funded by Model A's release.
+
+4. **Use queue size / time-in-queue as the balancing signal**, not just `RequiredCapacity`.
+   Two models with equal queues should converge to equal GPU allocation. A model with a
+   larger queue per GPU should get a larger share.
+
+### What this KEP contributes toward Story 4
+
+This KEP is a **prerequisite** for Story 4, not a solution:
+
+- Without quota-based discovery, the optimizer cannot reliably know the total GPU ceiling
+  in managed cloud or quota-governed environments. Story 4's fair-share formula
+  (`total / N`) requires a trustworthy `total`.
+- Quota discovery provides that ceiling cleanly, without GPU Operator dependency.
+
+Story 4 should be addressed in a follow-up KEP scoped to the optimizer: introducing a
+redistributive fair-share mode that activates when available GPUs = 0 and multiple models
+are simultaneously saturated.
+
+---
+
+## Drawbacks
+
+- Introduces a second discovery backend that must be maintained alongside
+  `K8sWithGpuOperator`. The risk is low: both implement the same `FullDiscovery`
+  interface and the new backend is simpler (two API calls vs. three vendor-specific
+  node list calls).
+- In multi-vendor namespaces, operators must set `acceleratorName` on each VA spec or
+  decisions for that variant will be skipped. The failure mode is conservative (no
+  over-allocation) but requires operator attention. A log warning makes this diagnosable.
+  Single-vendor namespaces — the common case — are unaffected: the limiter auto-resolves
+  an unset `acceleratorName` to the only pool.
+
+---
+
+## Alternatives
+
+### Alternative 1 — New `ConstraintProvider` instead of `CapacityDiscovery`
+
+Implement a standalone `QuotaConstraintProvider` that implements `ConstraintProvider`
+directly (bypassing `TypeInventory`) and register it alongside the existing GPU limiter
+in `engine.go`.
+
+**Rejected because:** it duplicates the aggregation and pool-management logic already
+in `TypeInventory`. The `CapacityDiscovery` interface exists precisely to separate the
+"how do we learn about GPUs" concern from the "how do we track and allocate them" concern.
+Reusing that separation is the right call.
+
+### Alternative 2 — Pod-request summation for `DiscoverUsage()`
+
+List all pods in the namespace and sum their GPU resource requests, mirroring
+`K8sWithGpuOperator.DiscoverUsage()`.
+
+**Rejected because:** `ResourceQuota.status.used` is already the authoritative, consistent
+value maintained by the Kubernetes admission controller. Re-summing pod requests is
+redundant, requires a Pod list API call, and can diverge transiently (e.g. a pod that has
+been admitted but whose status has not yet propagated). Reading `status.used` is strictly
+simpler and more accurate.
+
+### Alternative 3 — Environment variable per GPU type
+
+Allow `WVA_QUOTA_CAPACITY_NVIDIA=8`, `WVA_QUOTA_CAPACITY_AMD=4` etc. as static overrides
+without reading the actual quota.
+
+**Rejected because:** it duplicates configuration already managed by the quota and
+creates a second source of truth that can drift. Reading the live quota is always fresh.
--- a/docs/proposals/namespace-quota-capacity-discovery.md
+++ b/docs/proposals/namespace-quota-capacity-discovery.md
@@ -75,7 +75,7 @@ subsequently increases 10×, the correct allocation is 50/50 — but a naive ind
 autoscaler will not rebalance: model A still has a queue, so it sees no reason to release
 GPUs. Solving this requires a redistributive optimizer that can force model A below its
 current allocation even while it is still saturated. That optimizer cannot be built without
-a reliable GPU ceiling to divide. This KEP provides that ceiling; the redistributive
+a reliable GPU ceiling to divide. This proposal provides that ceiling; the redistributive
 optimizer is addressed in a follow-up proposal.
 
 ### Goals
@@ -696,13 +696,13 @@ correct `namespace` and `totalGPUs` fields.
   30 seconds). A namespace with N quotas returns N objects in a single list response.
   This is strictly less expensive than the current `listGPUNodes()` which issues one
   API call per GPU vendor.
-- No watch caching is added in this KEP; the existing client cache handles list responses.
+- No watch caching is added in this proposal; the existing client cache handles list responses.
 
 ---
 
 ## Relationship to Fair-Share Rebalancing
 
-This KEP delivers quota-based capacity discovery. It is a prerequisite for fair-share
+This proposal delivers quota-based capacity discovery. It is a prerequisite for fair-share
 rebalancing across competing models but does not implement the rebalancing itself. This
 section documents why, so a follow-up proposal can be scoped correctly.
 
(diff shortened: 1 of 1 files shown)


atantawi (commented):
## Holistic review

### Sensible
- Pluggable into `FullDiscovery` cleanly; no changes to `TypeInventory`/`DefaultLimiter`/optimizer needed. The proposed 5-line `NewEngine()` edit and the referenced function names all map to the actual code.
- `NormalizeAcceleratorName("nvidia.com/gpu")` does pass through unchanged (no hyphen → returned as-is). The pool-key story holds.
- Reading `status.used` over re-summing pod requests is the right call — admission-controller-maintained.
- Strong, candid Gap Analysis on Story 4; the trace through `fairShareScaleUp` (the `totalGPUs == 0 → break`) is accurate.

### Real gaps

**1. Doesn't actually solve #205.** Issue #205 asks for GPU *discovery* on GKE/EKS/AKS via cloud labels (`cloud.google.com/gke-accelerator`, instance types, AKS SKUs). This proposal sidesteps that by reading quotas. A GKE user without a `ResourceQuota` is no better off after this lands. Either retitle the linkage, or extend scope.

**2. Multi-namespace controller + single quota namespace is incoherent.** Notes say "The controller may manage workloads across other namespaces but capacity is governed by this one quota namespace." If WVA scales deployments in namespaces A/B/C while reading quota from D, the ceiling is meaningless — pods in A/B/C don't consume D's quota. Either restrict to a single namespace, or read per-managed-namespace quotas.

**3. Quota > node capacity → silent over-allocation.** The risk table marks this Medium/Medium with mitigation "document the requirement." That isn't a mitigation. WVA will issue scale-up decisions believing 100 GPUs exist while only 40 do; the result is perpetually pending pods. At minimum, describe a cross-check (compare to node-discovered total when available, log/cap) or call it an explicit non-goal/follow-up.

**4. Story 4 doesn't belong here.** It's introduced as a User Story and then disowned in the Gap Analysis. Move it to motivation, or split into a separate KEP — leaving it as a goal-shaped user story is misleading about what this KEP delivers.

**5. RBAC handwaving.** PRR says "Verify existing `config/rbac/role.yaml` includes [resourcequotas] verbs; add them if absent." They're absent today. The KEP should commit: which file, namespaced `Role` vs `ClusterRole` (matters for Story 2's multi-tenant cluster), and the exact verbs.

**6. `status.used` semantics differ from the existing path, claimed equivalent.** `K8sWithGpuOperator.DiscoverUsage` sums requests from *scheduled* pods only. `ResourceQuota.status.used` includes pending pods *and* non-WVA workloads. Both defensible — but calling re-summing "redundant" understates the change. One-line acknowledgement is enough.

### Minor
- **Activation by env var with restart required** is inconsistent with the saturation ConfigMap hot-reload pattern. Defend or align.
- **Hybrid mode unconsidered** in Alternatives: `min(nodes, quota)` would catch gap #3 cheaply. Worth a paragraph rejecting or accepting it.
- **Multiple quotas with overlapping scopes** (e.g., a default quota + a `priorityClass`-scoped one) would double-count. The proposal disclaims scopes but doesn't address overlap among non-scoped quotas.
- **`DiscoverNodes` returning empty** is safe today (no internal consumers outside `discovery/`); fine as forward-looking, but worth saying "no current caller."
- **Naming axis** drifts: `K8sWithGpuOperator` names by label source, `NamespaceQuotaDiscovery` by mechanism. Not blocking; future `GKEDiscovery`/`EKSDiscovery` may want to slot in.

### Bottom line
The mechanism is well-grounded and the integration plan is minimal and correct. But the PR currently *claims* to solve #205 and *appears* to address Story 4 while delivering neither. Tighten scope to "quota-as-ceiling for quota-governed namespaces," fix the multi-namespace incoherence, commit to the RBAC change, and downgrade the over-quota risk to either a real mitigation or an explicit non-goal. After that, this is a reasonable Provisional KEP.

---
*Drafted with [Claude Code](https://claude.com/claude-code).*

asm582:
> ## Holistic review
> ### Sensible
> * Pluggable into `FullDiscovery` cleanly; no changes to `TypeInventory`/`DefaultLimiter`/optimizer needed. The proposed 5-line `NewEngine()` edit and the referenced function names all map to the actual code.
> * `NormalizeAcceleratorName("nvidia.com/gpu")` does pass through unchanged (no hyphen → returned as-is). The pool-key story holds.
> * Reading `status.used` over re-summing pod requests is the right call — admission-controller-maintained.
> * Strong, candid Gap Analysis on Story 4; the trace through `fairShareScaleUp` (the `totalGPUs == 0 → break`) is accurate.
> 
> ### Real gaps
> **1. Doesn't actually solve #205.** Issue #205 asks for GPU _discovery_ on GKE/EKS/AKS via cloud labels (`cloud.google.com/gke-accelerator`, instance types, AKS SKUs). This proposal sidesteps that by reading quotas. A GKE user without a `ResourceQuota` is no better off after this lands. Either retitle the linkage, or extend scope.
> 
> **2. Multi-namespace controller + single quota namespace is incoherent.** Notes say "The controller may manage workloads across other namespaces but capacity is governed by this one quota namespace." If WVA scales deployments in namespaces A/B/C while reading quota from D, the ceiling is meaningless — pods in A/B/C don't consume D's quota. Either restrict to a single namespace, or read per-managed-namespace quotas.
> 
> **3. Quota > node capacity → silent over-allocation.** The risk table marks this Medium/Medium with mitigation "document the requirement." That isn't a mitigation. WVA will issue scale-up decisions believing 100 GPUs exist while only 40 do; the result is perpetually pending pods. At minimum, describe a cross-check (compare to node-discovered total when available, log/cap) or call it an explicit non-goal/follow-up.
> 
> **4. Story 4 doesn't belong here.** It's introduced as a User Story and then disowned in the Gap Analysis. Move it to motivation, or split into a separate KEP — leaving it as a goal-shaped user story is misleading about what this KEP delivers.
> 
> **5. RBAC handwaving.** PRR says "Verify existing `config/rbac/role.yaml` includes [resourcequotas] verbs; add them if absent." They're absent today. The KEP should commit: which file, namespaced `Role` vs `ClusterRole` (matters for Story 2's multi-tenant cluster), and the exact verbs.
> 
> **6. `status.used` semantics differ from the existing path, claimed equivalent.** `K8sWithGpuOperator.DiscoverUsage` sums requests from _scheduled_ pods only. `ResourceQuota.status.used` includes pending pods _and_ non-WVA workloads. Both defensible — but calling re-summing "redundant" understates the change. One-line acknowledgement is enough.
> 
> ### Minor
> * **Activation by env var with restart required** is inconsistent with the saturation ConfigMap hot-reload pattern. Defend or align.
> * **Hybrid mode unconsidered** in Alternatives: `min(nodes, quota)` would catch gap #3 cheaply. Worth a paragraph rejecting or accepting it.
> * **Multiple quotas with overlapping scopes** (e.g., a default quota + a `priorityClass`-scoped one) would double-count. The proposal disclaims scopes but doesn't address overlap among non-scoped quotas.
> * **`DiscoverNodes` returning empty** is safe today (no internal consumers outside `discovery/`); fine as forward-looking, but worth saying "no current caller."
> * **Naming axis** drifts: `K8sWithGpuOperator` names by label source, `NamespaceQuotaDiscovery` by mechanism. Not blocking; future `GKEDiscovery`/`EKSDiscovery` may want to slot in.
> 
> ### Bottom line
> The mechanism is well-grounded and the integration plan is minimal and correct. But the PR currently _claims_ to solve #205 and _appears_ to address Story 4 while delivering neither. Tighten scope to "quota-as-ceiling for quota-governed namespaces," fix the multi-namespace incoherence, commit to the RBAC change, and downgrade the over-quota risk to either a real mitigation or an explicit non-goal. After that, this is a reasonable Provisional KEP.
> 
> _Drafted with [Claude Code](https://claude.com/claude-code)._

@atantawi Thanks for the review. Addressed real gaps:
Re 1: Adding issue #205 was a mistake in the PR description, removed it.
Re 2: Thanks for catching, I made changes to the proposal re 3: In this proposal the assumption is quota’s are capacity.
Re 4: I have removed story 4 as it is correctly mentioned, and does not address this proposal 
Re 5: The multi-controller step is also stripped as it is not this proposal's concern
Re 6: removed redundant claim

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+---
+
+## Summary
+
+Add `NamespaceQuotaDiscovery`, a new implementation of the existing `CapacityDiscovery`
+interface that derives GPU capacity from Kubernetes `ResourceQuota` objects rather than
+from GPU Operator node labels. When enabled via `WVA_CAPACITY_NAMESPACES=ns-a,ns-b,ns-c`,
+the engine creates one quota-backed limiter per namespace and invokes the optimizer
+separately for each namespace's models — so each namespace's quota ceiling is enforced
+independently with no optimizer changes required.
```
ev-shindin: The per-GPU-type quota limiter for #1002 is already implemented in #1129 (`feat/quota-limiter`) per #1162's design, with chain composition tracked in #1003. This PR introduces a *second* quota path (a `CapacityDiscovery` backend) rather than feeding that limiter. Can we cross-reference #1129/#1162/#1003 and converge? The valuable kernel here — reading the budget from `ResourceQuota` — is exactly #1162's **option A2 (`ResourceQuotaReader`)**, designed as a composable `ConstraintProvider`. (Note #1129 uses a ConfigMap as its bootstrap quota surface, migrating to the cluster-default `ScalingPolicy` per #1162 §6.1 — so this isn't a ConfigMap-vs-`ResourceQuota` argument; the target surface is `ScalingPolicy` for both.)

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+> I want a single source of truth: the quota. I do not want to maintain GPU Operator
+> labels in parallel.
+
+### Notes and Constraints
+
+- **Quota equals capacity.** This design treats `ResourceQuota.spec.hard` as the total
+  GPU capacity available to workloads in the namespace. No node inventory, no GPU Operator
+  labels — the quota is the single source of truth for how many GPUs can be scheduled.
+  Keeping the quota aligned with actual node capacity is the operator's responsibility,
+  the same way correct node labels are the operator's responsibility with `K8sWithGpuOperator`.
```
ev-shindin: The doc states "the quota is the single source of truth for how many GPUs can be scheduled" and makes keeping it aligned with node capacity "the operator's responsibility." But a governance cap is not a capacity measurement: quota > physical silently over-schedules (`Pending` pods), and the risk table doesn't list this case. Please add it — and consider that composing with node discovery (`min(physical, quota)`) removes the failure mode entirely rather than documenting it away.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+**`DiscoverUsage(ctx) → map[resourceName]int`**
+
+1. List `ResourceQuotaList` in `d.Namespace`.
+2. Aggregate `Status.Used` values for GPU resource names across all quotas.
+3. Return the map. No pod listing. No inference.
+
+**`DiscoverNodes(ctx) → map[string]NodeInfo`**
+
+Returns an empty map. Quota-based discovery has no per-node information; callers that
+require node affinity must use `K8sWithGpuOperator`.
```
ev-shindin: As a `CapacityDiscovery` backend this switches off node discovery (`DiscoverNodes` returns empty). #1162's A2 deliberately uses a `ConstraintProvider` so quota composes with physical inventory (`min(physical, quota)`); the Alternative-1 rejection ("duplicates `TypeInventory` aggregation") trades the physical safety bound for interface tidiness. The aggregation concern is solvable; composability is the bigger prize.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+   ceiling enforced independently per namespace.
+4. Work on any Kubernetes distribution — GKE, EKS, AKS, OpenShift, kind — without
+   requiring GPU Operator or node-level RBAC.
+
+### Non-Goals
+
+- Replacing `K8sWithGpuOperator` — it remains the default and is required for
+  GPU-model-aware scheduling (e.g., differentiating A100 from H100).
+- GPU-model-aware scheduling: all GPUs in a namespace are treated as one pool keyed by
+  resource name (e.g. `nvidia.com/gpu`). Differentiating A100 from H100 within a
```
ev-shindin, line 93: Keying by `nvidia.com/gpu` (single counter) can't express `{H100:8, A100:4}`; the doc says use `K8sWithGpuOperator` for heterogeneous namespaces. That's the exact gap #1002/#1162 exist to close — and it means this only serves homogeneous, single-type namespaces, reintroducing the GPU-Operator dependency for mixed GPUs. Please state explicitly that this does not address #1002.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+
+- **Accelerator key is the resource name, not the model name.** `ResourceQuota` stores
+  `nvidia.com/gpu: 8`, not `NVIDIA-A100-PCIE-80GB: 8`. `TypeInventory` pools will be
+  keyed by `nvidia.com/gpu`.
+
+  **`VariantAutoscaling` is optional and will be deprecated.** When a VA is provided,
+  `acceleratorName` must be set explicitly so it matches the GPU resource name in the
+  `ResourceQuota` (e.g. `nvidia.com/gpu`).
+
+- **Usage from `status.used`, not pod listing.** `ResourceQuota.status.used` is
```
ev-shindin: A design-level question (independent of wiring): is a namespace's "used GPUs" `status.used` — which counts **all** admitted pods, including non-WVA workloads — or only WVA-managed replicas? The two give different available headroom, and it's a correctness-affecting *semantic* choice, not just an implementation detail. Please state which is authoritative and why. (The exact accounting in code can be settled on the implementation PR.)

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+  maintained by the quota admission controller and covers all admitted workloads in the
+  namespace — including pending pods and non-WVA workloads — whereas
+  `K8sWithGpuOperator.DiscoverUsage()` sums requests only from scheduled WVA-managed
+  pods. Both are defensible; `status.used` is the more conservative count and the natural
+  source of truth when quota is the capacity model.
+
+- **Each namespace's quota is enforced independently.** A `ResourceQuota` only governs
+  pods in its own namespace, so each namespace's ceiling must apply only to models running
+  in that namespace. The engine achieves this without any optimizer changes by invoking
```
ev-shindin: Enforcing each namespace's quota independently means the greedy optimizer fair-shares only *within* a namespace, not across all competing models on the shared physical cluster — a behavior change worth stating in the design. And there's no **cluster-aggregate** budget: N independent per-namespace ceilings, with no way for a cluster admin to cap total cluster GPUs (a goal for #1002/#1162). (Whether realizing this needs an engine change is an implementation question for later.)

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+[NVIDIA GPU Operator](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/overview.html)
+to label nodes with GPU model and memory information. This creates a hard dependency that
+rules out a class of deployment environments:
+
+- **Managed cloud services** (GKE, EKS, AKS) often do not expose GPU Operator labels on
+  their managed node pools.
+- **Multi-tenant clusters** commonly restrict node listing to cluster administrators,
+  making per-node inventory inaccessible to namespace-scoped controllers.
+- **Quota-governed namespaces** (common in OpenShift and enterprise on-prem clusters)
+  already encode GPU capacity in `ResourceQuota.spec.hard` — this is the authoritative
```
ev-shindin, line 62: The claim that quota-governed namespaces "already encode GPU capacity in `ResourceQuota.spec.hard` … the authoritative source" breaks under **Kueue** (budget lives in `ClusterQueue`/`ResourceFlavor`, and Kueue advises against `ResourceQuota` on managed namespaces) and OpenShift `ClusterResourceQuota` (aggregate cap not in a plain `ResourceQuota`). In those clusters `NamespaceQuotaDiscovery` reads empty/stale capacity. Note too: Kueue *can* express per-GPU-type via per-model `ResourceFlavor` — the surface this PR reads is the one that *can't* carry #1002's granularity.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+The optimizer receives only the requests for one namespace at a time, so its internal
+`available` map naturally reflects that namespace's quota ceiling. No changes to
+`GreedyByScoreOptimizer` or `mergeConstraints` are required.
+
+### Accelerator Name Convention
+
+| Discovery mode | `TypeInventory` pool key | VA `acceleratorName` |
+|---|---|---|
+| `K8sWithGpuOperator` | `A100`, `H100` (normalized from node label) | `A100` |
+| `NamespaceQuotaDiscovery` | `nvidia.com/gpu` (resource name, passed through) | `nvidia.com/gpu` |
```
ev-shindin: The `acceleratorName` a **tenant** writes must match the pool-key convention the **cluster admin** chose via `WVA_CAPACITY_NAMESPACES` (`nvidia.com/gpu` vs `A100`) — a tenant must know an install-time decision they can't see, and the VA breaks if the admin flips backends. Two things compound it through the VA-deprecation lens: (1) that field lives on the **`VariantAutoscaling` CRD, which is being deprecated** (#1092, merged; metadata moving to annotations on `ScaledObject`/`HPA`) — and the proposal itself notes VA "is optional and will be deprecated," so the convention is built on a surface scheduled for removal. (2) The deprecation + config-UX direction (#1194) is to *minimize and derive* per-workload metadata — #1194 even removes `model-id` as derivable from the `InferencePool`. A new, required, hand-authored `acceleratorName` runs against that.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+### API Changes
+
+No new CRDs, no new API types, no new flags in the saturation ConfigMap. The only user-
+facing surface is the environment variable and the `acceleratorName` convention on VA specs.
+
+### Activation
+
+`WVA_CAPACITY_NAMESPACES` accepts a comma-separated list of namespaces. When set, the
+engine creates one `NamespaceQuotaDiscovery` and one `DefaultLimiter` per namespace,
+keyed by namespace name. When unset, the existing `K8sWithGpuOperator` path is used.
```
ev-shindin: This adds a raw `ResourceQuota` to read + a `WVA_CAPACITY_NAMESPACES` env var + the `acceleratorName` convention — surfaces across all three #1194 personas, pulling quota toward `ResourceQuota`+env-var while #1194/#1162 consolidate onto a single `ScalingPolicy` (quota declared directly on the cluster-default `ScalingPolicy`, no ConfigMap/env-var). Competing directions; align before adding the env var.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,600 @@
...
+
+### Accelerator Name Convention
+
+| Discovery mode | `TypeInventory` pool key | VA `acceleratorName` |
+|---|---|---|
+| `K8sWithGpuOperator` | `A100`, `H100` (normalized from node label) | `A100` |
+| `NamespaceQuotaDiscovery` | `nvidia.com/gpu` (resource name, passed through) | `nvidia.com/gpu` |
+
+`VariantAutoscaling` is optional and will be deprecated. When a VA is provided,
+`acceleratorName` must be set so the limiter knows which pool to debit.
```
ev-shindin: The accelerator a variant uses is discoverable from the `InferencePool`'s pods (their GPU resource requests / node), the same "derive from the pool, don't make the tenant type it" principle that lets #1194 drop `model-id` and #1092 move metadata off the VA CRD. So the quota resource-name match could be derived by the limiter rather than imposed as a new required tenant field — keeping the tenant contract aligned with where discovery is heading (annotation-light, pool-derived) instead of adding a field that must be hand-set and then migrated off the deprecated VA spec.

ev-shindin (commented):
Thanks @asm582 — the operator-free-capacity motivation is real and valuable (managed cloud, locked-down multi-tenant clusters genuinely can't use GPU-Operator labels), the multi-namespace handling is clean, and the `status.used`-over-pod-summation reasoning is sound. Two framing items dominate:

**(a) This area is already implemented.** #1129  implements the per-GPU-type quota limiter for #1002 (per #1162's design): `QuotaInventory`, a composite limiter, factory, ~21 files with tests; chain composition (`min(physical, quota)`) is tracked in #1003. #1227 reads the GPU budget from `ResourceQuota` via a *parallel* discovery backend rather than feeding that limiter — I'd like to converge these rather than maintain two quota designs. (To be clear: #1129 wires quota through a **ConfigMap** today as its *bootstrap* surface; #1162 §6.1 commits to migrating it onto the cluster-default `ScalingPolicy`, no standalone ConfigMap. So the convergence target is the **`ScalingPolicy` surface + a composable limiter**, not "ConfigMap vs `ResourceQuota`" — both raw surfaces are transitional.)

**(b) "Quota = capacity" is a stated premise**  but the risk table doesn't cover the case where quota > node capacity → `Pending` pods. That's the core weakness of treating a governance cap as a capacity measurement, and it should be documented — or removed entirely by composing with node discovery (`min(physical, quota)`).

Specific items inline: replace-vs-compose, the per-type gap (= #1002), the usage-model question (what counts against the budget), per-namespace fair-share siloing, and the Kueue/`ClusterResourceQuota` case. Happy to fold the operator-free-capacity kernel into #1162/#1129 as the composable A2 (`ResourceQuotaReader`) provider.

ev-shindin:
> Kubernetes-based namespace quotas where kube quotas equal capacity.

@asm582 the operator-free-capacity insight is the part I most want to keep — many target clusters (managed cloud, locked-down multi-tenant) can't use GPU-Operator labels, and reading the budget from the namespace is right for them. You're also right that `ResourceQuota` gives cluster-admin-owned governance natively via RBAC.

Two things make me want to converge rather than run a parallel design:
- We already have the quota limiter implemented in **#1129** (per #1162's design), with chain composition tracked in **#1003**. It wires quota through a ConfigMap today as a *bootstrap* surface; #1162 §6.1 migrates that onto the cluster-default `ScalingPolicy` (no standalone ConfigMap) — so the end-state surface is `ScalingPolicy`, and this isn't a ConfigMap-vs-`ResourceQuota` debate.
- #1162 already models "read `ResourceQuota` as a GPU budget" as **option A2 (`ResourceQuotaReader`)** — but as a `ConstraintProvider` that **composes** with node discovery (`min(physical, quota)`) rather than replacing it.

Realizing #1227 as that A2 provider keeps your win and resolves the review concerns:
1. **Compose, don't replace** — quota becomes an additional budget while node discovery (when present) still gives the physical ceiling; `quota > nodes` can't over-schedule. No node labels (your case) → runs on quota alone.
2. **Quota is a budget, not "capacity"** — drops the false identity, and fixes the Kueue/`ClusterResourceQuota` case (the provider contributes nothing instead of a false zero).
3. **No tenant↔admin coupling** — the limiter resolves the accelerator key.
4. **One surface** — aligns with the single-`ScalingPolicy` direction (#1194) and #1162's ScalingPolicy-direct position; A2 reads native `ResourceQuota` as an *input* without making it the user surface.
5. **Per-type stays in #1129/#1162** — `ResourceQuota` is per-resource-name only; #1002's per-type lives in `spec.quota`. They compose.

Concretely: happy to land A2 in #1162/#1129 with you as co-author (largely the `status.used`/`DiscoverUsage` logic you've designed), and either close #1227 in favor of that or repoint it at the A2 provider shape. @lionelvillard — does converging these (and #1129/#1003) into one quota story match how you'd want this to land?

docs/proposals/namespace-quota-capacity-discovery.md
ev-shindin: Thanks — the `useDiscoveredUsage` flag + `NewDefaultLimiterWithDiscoveredUsage` defines the model cleanly, and ignoring `currentUsage` when the flag is set removes the double-count concern. One thing to land: the V2 path still uses WVA-only usage; your own recommendation to move V2 onto `status.used` via `RefreshAll()` would make V1/V2 consistent — good to track as the follow-up you noted.

docs/proposals/namespace-quota-capacity-discovery.md
ev-shindin: `resolveUnknownAccelerators()` auto-resolving empty/unknown `acceleratorName` → `nvidia.com/gpu` for single-pool namespaces fixes the "tenant must know the admin's resource-name convention" leak for the homogeneous case 👍. Three things still open:
1. **WVA already resolves accelerators by derivation.** `GetAcceleratorNameFromScaleTarget()` (`internal/utils/utils.go`) derives the accelerator from the scale target's `nodeSelector`/`nodeAffinity` GPU-product keys, falling back to the `AcceleratorNameLabel` on the VA — and it's already used across the collector, scale-from-zero, and saturation V1/V2. The new `resolveUnknownAccelerators()` is a *separate, weaker* path (single-pool fallback + a "clear stale `A100` names" footgun) that doesn't reuse it. The resource-name key you need (`nvidia.com/gpu`) is itself derivable from the pods' GPU resource requests via the same scale-target accessor — so this should extend the existing derivation rather than add a parallel heuristic.
2. The field lives on the `VariantAutoscaling` spec/label, which **#1092 deprecates** — metadata moving to annotations on `ScaledObject`/`HPA`. What's the plan post-VA, and does the empty/auto-resolve default carry to the annotation form?
3. The auto-resolve only triggers for a single pool key — fine, since quota mode is single-type by design, but worth stating it does *not* cover heterogeneous namespaces (consistent with the per-type non-goal).

Net: same "derive, don't author" trajectory as #1194 dropping `model-id` — and the derivation infrastructure already exists, so this is reuse, not new work.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Thanks, the proposal was changed based on feedback and implementation. Closing this outdated comment.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: the proposal was changed based on feedback and implementation. Closing this outdated comment.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: The proposal was changed based on feedback and implementation. Closing this outdated comment.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: True, this is a limitation of this proposal. As per the use case; this could be resolved in future proposals.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Kueue is a different quota management system; this proposal does not consider a specific quota management system but provides a path to read quotas.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Trying to understand this comment better, why would a user not see a quota set on the namespace? This should be a standard read-only operation. 

> that field lives on the VariantAutoscaling CRD, which is being deprecated (https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1092, merged; metadata moving to annotations on ScaledObject/HPA) — and the proposal itself notes VA "is optional and will be deprecated," so the convention is built on a surface scheduled for removal.

All this proposal says is that it requires a change to the VA, which should not be the case. I will generalize this statement as you correct the set acceleratorName, which is moving to annotations.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: This proposal only provides a way to read quotas without any new policy or configuration, a feature that is missing. 

The proposal was changed based on feedback and implementation.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: I am not proposing any CR changes to CRD; in fact, the intent is to avoid any. will work on this: https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1227#discussion_r3348183386

ev-shindin: But you are adding another configuration surface to existing ones. The overall configuration become very fragmented. I triying to avoid this.

docs/proposals/namespace-quota-capacity-discovery.md
ev-shindin: I mean Kueue, because its not compatible with your proposal. For me your proposal is acceptable as **additional**, probably pluggable way to read quotas. But we should agree on the main way first, which is provided in #1162 .

asm582:
> We already have the quota limiter implemented in https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1129 (per https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1162's design), with chain composition tracked in https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1003. It wires quota through a ConfigMap today as a bootstrap surface; https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1162 §6.1 migrates that onto the cluster-default ScalingPolicy (no standalone ConfigMap) — so the end-state surface is ScalingPolicy, and this isn't a ConfigMap-vs-ResourceQuota debate.

This proposal tries to start small by reading a quota and passing it to the limiter. It reuses existing objects already present in the Kubernetes cluster. Obviously, this could be expanded based on reference proposals.

> https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1162 already models "read ResourceQuota as a GPU budget" as option A2 (ResourceQuotaReader) — but as a ConstraintProvider that composes with node discovery (min(physical, quota)) rather than replacing it.

One of the requirements, as I understand it from the user, is to use fixed quotas and rebalancing (FYI, rebalancing is not part of this proposal ). Hence, we want to start small with a quota as capacity for now. As you pointed out, this has limitations that can be resolved in subsequent proposals/PRs.


> Compose, don't replace — quota becomes an additional budget while node discovery (when present) still gives the physical ceiling; quota > nodes can't over-schedule. No node labels (your case) → runs on quota alone.

Is this statement correct entirely? Kubernetes does not allow pods to spawn when hard quotas are set.

> Quota is a budget, not "capacity" — drops the false identity, and fixes the Kueue/ClusterResourceQuota case (the provider contributes nothing instead of a false zero).

As outlined, the quota for now is capacity; when evolving to budget, the constraints you mentioned should apply.

> No tenant↔admin coupling — the limiter resolves the accelerator key.

Not proposing to add a key, I was under the impression that this is required in VA CRD, and as you pointed out, it has been moved to annotation, hence I will adapt the language in the proposal.

> One surface — aligns with the single-ScalingPolicy direction (https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1194) and https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1162's ScalingPolicy-direct position; A2 reads native ResourceQuota as an input without making it the user surface.

>Per-type stays in https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1129/https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1162 — ResourceQuota is per-resource-name only; https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1002's per-type lives in spec.quota. They compose.

Sure, we can align when such a proposal and implementation get merged.

> Concretely: happy to land A2 in https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1162/https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1129 with you as co-author (largely the status.used/DiscoverUsage logic you've designed),

Sure, let's review the proposal so it can be merged, and then we can decide on implementation.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: ok, what is the fragmentation concern here when I am reading the existing quota object present on the Kubernetes cluster?

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+
+## Summary
+
+Add `NamespaceQuotaDiscovery`, a new implementation of the existing `CapacityDiscovery`
+interface that derives GPU capacity from Kubernetes `ResourceQuota` objects rather than
+from GPU Operator node labels. When enabled via `WVA_CAPACITY_NAMESPACES=ns-a,ns-b,ns-c`,
+the engine replaces `e.GPULimiter` with a `NamespaceLimiter` that holds one quota-backed
+`DefaultLimiter` per namespace. Both the V1 (`optimizeV1`) and V2 (`optimizeV2`) paths
+consume `e.GPULimiter` and therefore enforce per-namespace quota ceilings without any
+changes to the optimizers, `DefaultLimiter`, or `TypeInventory`.
```
mamy-CS: Says without any changes here, but I see DefaultLimiter getting a new bool field and a constructor below. Maybe adjust the claim.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+current allocation even while it is still saturated. That optimizer cannot be built without
+a reliable GPU ceiling to divide. This KEP provides that ceiling; the redistributive
+optimizer is addressed in a follow-up proposal.
+
+### Goals
+
+1. Add a `NamespaceQuotaDiscovery` backend that reads GPU capacity from
+   `ResourceQuota.spec.hard` and current usage from `ResourceQuota.status.used`.
+2. Plug it into the existing `CapacityDiscovery` / `UsageDiscovery` interfaces with zero
+   changes to `TypeInventory`, `DefaultLimiter`, or `GreedyByScoreOptimizer`.
```
mamy-CS: Same here, claim needs adjustment

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+| Mixed GPU types in one namespace share one pool | Low | Low | Documented as a known limitation. Use `K8sWithGpuOperator` for heterogeneous GPU namespaces. |
+
+---
+
+## Design Details
+
+### Overview
+
+Two new files, two small edits. No new Go interfaces, no new fields on `Engine`, no
+changes to `optimizeV1`, `DefaultLimiter`, `TypeInventory`, or any optimizer.
```
mamy-CS: Same here, claim needs adjustment

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+- **Multi-tenant clusters** commonly restrict node listing to cluster administrators,
+  making per-node inventory inaccessible to namespace-scoped controllers.
+- **Quota-governed namespaces** (common in OpenShift and enterprise on-prem clusters)
+  already encode GPU capacity in `ResourceQuota.spec.hard` — this is the authoritative
+  source and scanning nodes would duplicate it.
+
+In each of these cases a namespace-scoped quota is a practical fallback: the quota hard
+limit equals the GPU capacity visible to workloads in the namespace. Where GPU Operator
+labels are accessible, `K8sWithGpuOperator` remains the preferred backend — it provides
+GPU-model-aware scheduling and more precise per-node inventory that quota cannot express.
```
mamy-CS, line 68: how does an operator decide which backend to use, one sentence on detection heuristics would help hete.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+
+> As a platform team that manages GPU capacity via quota rather than node labels,
+> I want a single source of truth: the quota.
+
+### Notes and Constraints
+
+- **Quota equals capacity.** This design treats `ResourceQuota.spec.hard` as the total
+  GPU capacity available to workloads in the namespace. No node inventory, no GPU Operator
+  labels — the quota is the single source of truth for how many GPUs can be scheduled.
+  Keeping the quota aligned with actual node capacity is the operator's responsibility.
```
mamy-CS, line 123: If quota drifts from actual node capacity, the optimizer silently underallocates with no signal. I think this is a bigger operational risk right?

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+  are only evaluated against ns-a's `ResourceQuota`, and ns-b's against ns-b's. A saturated model in ns-a consuming its full quota does not reduce the
+  GPU headroom available to models in ns-b.
+
+
+
+### Risks and Mitigations
+
+| Risk | Likelihood | Impact | Mitigation |
+|---|---|---|---|
+| `status.used` transiently stale during rapid pod churn | Low | Low | The quota admission controller updates `status.used` synchronously on pod admission. Staleness is bounded by API server response time, well within the 30-second optimization interval. |
```
mamy-CS, line 175: When does status.used decrease? After pod deletion completes or when termination starts? during scale-down churn the limiter sees inflated usage and blocks valid scale-up correct? I think some info missing here for completeness.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+**`Limit()` — replace the first two steps:**
+
+```go
+if l.useDiscoveredUsage {
+    // reads spec.hard (capacity) + status.used (usage) in one call
+    if inv, ok := l.inventory.(*TypeInventory); ok {
+        if err := inv.RefreshAll(ctx); err != nil {
+            return fmt.Errorf("failed to refresh quota inventory: %w", err)
+        }
+    }
```
mamy-CS: Here, if the assertion fails the block is silently skipped

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+
+**`ComputeConstraints()` — same pattern:**
+
+```go
+if l.useDiscoveredUsage {
+    if inv, ok := l.inventory.(*TypeInventory); ok {
+        if err := inv.RefreshAll(ctx); err != nil {
+            return nil, fmt.Errorf("failed to refresh quota inventory: %w", err)
+        }
+    }
```
mamy-CS: Same silent-failure risk as above

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+**V2 already reads `spec.hard` from the quota.** `ComputeConstraints()` calls
+`inventory.Refresh(ctx)` → `NamespaceQuotaDiscovery.Discover()` — reading
+`ResourceQuota.spec.hard`. The V2 change below is needed only to dispatch per-namespace
+rather than treating all requests as a single global pool.
+
+> **Recommendation:** For consistency with V1, V2 should also adopt
+> `NewDefaultLimiterWithDiscoveredUsage()` so `ComputeConstraints()` reads
+> `ResourceQuota.status.used` via `RefreshAll()` instead of using
+> `computeCurrentGPUUsage(nsRequests)`. This ensures non-WVA GPU workloads are
+> accounted for in V2 constraints. Tracked as a follow-up.
```
mamy-CS, line 453: Where is this tracked? in mixed-workload namespaces as mentioned above V2 actively exceeds the quota ceiling. I think this should be called a known correctness limitation with a note on when it's unsafe to use V2 quota mode.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+### Accelerator Name Convention
+
+| Discovery mode | `TypeInventory` pool key | VA `acceleratorName` |
+|---|---|---|
+| `K8sWithGpuOperator` | `A100`, `H100` (normalized from node label) | `A100` |
+| `NamespaceQuotaDiscovery` | `nvidia.com/gpu` (canonical resource name) | `nvidia.com/gpu` or empty (auto-resolved) |
+
+`DefaultLimiter.resolveUnknownAccelerators()` auto-resolves an empty or unrecognised
+`acceleratorName` to `nvidia.com/gpu` when the namespace inventory has exactly one pool
+key. Operators do not need to set `acceleratorName` for homogeneous GPU namespaces.
```
mamy-CS, line 474: fails silently for multi-vendor quotas, the VA gets dropped with no scale decision and no log. Needs an explicit warning.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+Operators switching from `K8sWithGpuOperator` must clear any stale model-specific names
+(e.g. `A100`) that no longer match the quota pool key.
```
mamy-CS, line 476: Whats the migration path suggested here?

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+environments. This KEP provides it. The redistributive optimizer that acts on it is a
+follow-up KEP scoped to the optimizer: a fair-share mode that activates when available
+GPUs = 0 and multiple models are simultaneously saturated.
+
+---
+
+## Drawbacks
+
+- Introduces a second discovery backend that must be maintained alongside
+  `K8sWithGpuOperator`. The risk is low: both implement the same `FullDiscovery`
```
mamy-CS, line 798: Worth also noting the observability tradeoff. quota mode returns no per-node data correct? so debugging allocation decisions has less information than K8sWithGpuOperator.

kaushikmitr:
Looks like TPUs are an implicit non-goal here, `gpuResourceName()` only matches
`nvidia.com/gpu`/`amd.com/gpu`/`intel.com/gpu`, so `google.com/tpu` falls through and
a TPU-quota namespace would read as zero with no signal. It's not in the stated
Non-Goals either, and it's a different case from the "GPU-model-aware" one (different
resource name, and a topology/slice problem rather than a model-identity one).

How do we plan to cover TPUs? "quota integer = placeable capacity" premise is even weaker for TPUs, a chip count
doesn't say whether a contiguous slice of the required topology can actually land. Given
the motivation calls out GKE, should we make this an explicit non-goal for now, or is
TPU coverage something we want this backend to handle?

asm582:
> Looks like TPUs are an implicit non-goal here, `gpuResourceName()` only matches `nvidia.com/gpu`/`amd.com/gpu`/`intel.com/gpu`, so `google.com/tpu` falls through and a TPU-quota namespace would read as zero with no signal. It's not in the stated Non-Goals either, and it's a different case from the "GPU-model-aware" one (different resource name, and a topology/slice problem rather than a model-identity one).
> 
> How do we plan to cover TPUs? "quota integer = placeable capacity" premise is even weaker for TPUs, a chip count doesn't say whether a contiguous slice of the required topology can actually land. Given the motivation calls out GKE, should we make this an explicit non-goal for now, or is TPU coverage something we want this backend to handle?

> is TPU coverage something we want this backend to handle?

We can mention `google.com/tpu` as another provider string as part of the resource quota and expand the proposal to include a TPU-specific string type for workloads (that need a single TPU or non-contiguous?) that don't have topology, aka continuous requirement, would that work?

> How do we plan to cover TPUs? "quota integer = placeable capacity" premise is even weaker for TPUs, a chip count doesn't say whether a contiguous slice of the required topology can actually land. 

Hmm, this is a topology concern, and a fragmented cluster can always have more resources in sum. The expectation, then, would be that another system (like Kueue) should provide a guarantee around startup or place the workload back in the queue when resources are not available (contiguous). Would that help?

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,699 @@
...
+same namespace, each initially using 10% of the cluster's GPUs. Model A's request volume
+increases 10× and scales to 90%, leaving model B at 10%. When model B's request volume
+subsequently increases 10×, the correct allocation is 50/50 — but a naive independent
+autoscaler will not rebalance: model A still has a queue, so it sees no reason to release
+GPUs. Solving this requires a redistributive optimizer that can force model A below its
+current allocation even while it is still saturated. That optimizer cannot be built without
+a reliable GPU ceiling to divide. This KEP provides that ceiling; the redistributive
+optimizer is addressed in a follow-up proposal.
+
+### Goals
```
lionelvillard, line 81: In general goals describe the outcome, not the implementation details. 4 is a goal. 3 except `via WVA_CAPACITY_NAMESPACES ...`. 1 and 2 are not goals.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: I think this would be the operator's choice driven by the privilege cloud provider gives to an operator.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Trying to start small with quota equals capacity, meaning no node-based autoscaling for now, enabled in the cluster, could be a future improvement if you agree.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Good Catch, this pattern exists in the code base. The implementation of this proposal should fix it.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: This proposal does not change v2 behavior; it is a follow-up, as v2 is experimental. It does create plumbing that v2 can consume to improve quota calculation.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Thanks, I added a migration path.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Done.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Thanks, made changes.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: This is now removed.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Done.

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Hmm, why is per-node data needed?

docs/proposals/namespace-quota-capacity-discovery.md
asm582: Agree, this quota reading proposal is for homogeneous GPUs, as already mentioned in the proposal, and is not solving the mentioned issue.

kaushikmitr:
> > Looks like TPUs are an implicit non-goal here, `gpuResourceName()` only matches `nvidia.com/gpu`/`amd.com/gpu`/`intel.com/gpu`, so `google.com/tpu` falls through and a TPU-quota namespace would read as zero with no signal. It's not in the stated Non-Goals either, and it's a different case from the "GPU-model-aware" one (different resource name, and a topology/slice problem rather than a model-identity one).
> > How do we plan to cover TPUs? "quota integer = placeable capacity" premise is even weaker for TPUs, a chip count doesn't say whether a contiguous slice of the required topology can actually land. Given the motivation calls out GKE, should we make this an explicit non-goal for now, or is TPU coverage something we want this backend to handle?
> 
> > is TPU coverage something we want this backend to handle?
> 
> We can mention `google.com/tpu` as another provider string as part of the resource quota and expand the proposal to include a TPU-specific string type for workloads (that need a single TPU or non-contiguous?) that don't have topology, aka continuous requirement, would that work?
> 
> > How do we plan to cover TPUs? "quota integer = placeable capacity" premise is even weaker for TPUs, a chip count doesn't say whether a contiguous slice of the required topology can actually land.
> 
> Hmm, this is a topology concern, and a fragmented cluster can always have more resources in sum. The expectation, then, would be that another system (like Kueue) should provide a guarantee around startup or place the workload back in the queue when resources are not available (contiguous). Would that help?

Discussed offline with @asm582 since we have not tested this in TPUs keep it a non goal for now. TPUs have architechtural difference so we need to test if this can be easily extended to TPUs

asm582:
> > > Looks like TPUs are an implicit non-goal here, `gpuResourceName()` only matches `nvidia.com/gpu`/`amd.com/gpu`/`intel.com/gpu`, so `google.com/tpu` falls through and a TPU-quota namespace would read as zero with no signal. It's not in the stated Non-Goals either, and it's a different case from the "GPU-model-aware" one (different resource name, and a topology/slice problem rather than a model-identity one).
> > > How do we plan to cover TPUs? "quota integer = placeable capacity" premise is even weaker for TPUs, a chip count doesn't say whether a contiguous slice of the required topology can actually land. Given the motivation calls out GKE, should we make this an explicit non-goal for now, or is TPU coverage something we want this backend to handle?
> > 
> > 
> > > is TPU coverage something we want this backend to handle?
> > 
> > 
> > We can mention `google.com/tpu` as another provider string as part of the resource quota and expand the proposal to include a TPU-specific string type for workloads (that need a single TPU or non-contiguous?) that don't have topology, aka continuous requirement, would that work?
> > > How do we plan to cover TPUs? "quota integer = placeable capacity" premise is even weaker for TPUs, a chip count doesn't say whether a contiguous slice of the required topology can actually land.
> > 
> > 
> > Hmm, this is a topology concern, and a fragmented cluster can always have more resources in sum. The expectation, then, would be that another system (like Kueue) should provide a guarantee around startup or place the workload back in the queue when resources are not available (contiguous). Would that help?
> 
> Discussed offline with @asm582 since we have not tested this in TPUs keep it a non goal for now. TPUs have architechtural difference so we need to test if this can be easily extended to TPUs

@kaushikmitr added TPU support as a non-goal to the proposal.

docs/proposals/namespace-quota-capacity-discovery.md
ev-shindin: Just now in order to use WVA user needs:
1. Add label on pod template of Deployment/LWS
2. Create VA (or with VA deprecation put labels and annotations on HPA/ScaledObject)
3. Configure HPA
4. Create and configure several ConfigMaps to define thresholds and other things related to WVA behavior

Its too much, so I trying to reduce this configuration surface.

Your proposal adds another dimension - ResourceQuota.

I afraid that at the end of the day user will say: may be this WVA is good, but it is not usable, because it is very hard to configure it and to manage its configuration.
My question **why user should configure WVA behavior in several different places?** So I prefer to have single configuration object where user can configure all: thresholds, behaviors and quotas and do not put different configuration parts in the different places.

ev-shindin (changes requested):
Thanks for the substantial reworking — reusing `CapacityDiscovery`/`TypeInventory`/`DefaultLimiter` is the right instinct, the assumptions are now explicit, and scoping fair-share rebalancing out into a follow-up is exactly right.

I'm fully on board with "start simple and reuse existing concepts." My pushback is that the PR reuses existing **code** but introduces new **concepts** — specifically the ones we've already agreed are changing. So "simple now" here is really deferred work plus a second quota/capacity path to maintain until we undo it.

The better place to start, in my view, is the **configuration system** itself. #1162 (with #1194) moves WVA toward a model where an operator declares *all* WVA behavior — capacity and quota included — in **one place**, the `ScalingPolicy` surface. "Start simple" should mean starting from that single declarative surface and adding quota as one more configurable behavior within it — not starting from several disconnected places (an env var, a hand-authored `ResourceQuota`, an `acceleratorName` convention) that we then have to consolidate. Building the capability on top of #1162's configuration model is the simpler long-term starting point, not the harder one.

Concretely, the PR starts from the scattered end:

**1. It fragments the configuration surface.** `ResourceQuota` itself is native and the admin already manages it — fair. But the *new WVA-specific* config this adds is an env var (`WVA_CAPACITY_NAMESPACES`) plus conventions the operator has to hold in their head (quota = capacity, the pool key is the resource name, `acceleratorName` must match it). That's WVA behavior configured through env vars and conventions instead of the single `ScalingPolicy` surface (#1194). #1162 reads the *same* native `ResourceQuota` — the difference is that WVA learns about it through the config system, not an env var. The proposal references neither #1194 nor #1162.

**2. Per-GPU-type is an explicit non-goal — but that's #1002.** Keying by resource name (`nvidia.com/gpu`) and sending heterogeneous namespaces back to `K8sWithGpuOperator` means the "no GPU-Operator dependency" win only holds for homogeneous namespaces, and leaves the harder, already-specified per-type case (#1002, which #1162 is built around) to the dependency this set out to remove. Deferring per-type as a *feature* is fine — the issue is building a per-resource-name-only *structure* that precludes it, when #1162's model is per-type-ready and gives the homogeneous case for free.

**3. It replaces node discovery instead of composing — and Alternative 1 doesn't rebut that.** The `engine.go` sketch does `gpuLimiter = NewNamespaceLimiter(perNS)`, and `NamespaceQuotaDiscovery` implements no `NodeDiscovery` — so in quota mode there's no physical-capacity signal and no `min(physical, quota)` bound. Alternative 1 rejects a `ConstraintProvider` "that bypasses `TypeInventory`," but that isn't the ask: keep the `TypeInventory` reuse, just register quota as a constraint *alongside* node discovery so the optimizer takes the min, rather than swapping one backend for the other. #1162's A2 does exactly this.

Composing doesn't depend on always having a physical signal, either: where nodes aren't readable (e.g. Autopilot) the quota provider is simply the only constraint and you get today's behavior; where they are readable, physical GPU *count* comes from `node.status.allocatable` — core API, the GPU-Operator labels are only needed for the per-*model* key — so `min(physical, quota)` is available without GPU Operator. Composing is a superset of replacing: it never does worse, and it catches an over-set quota where the PR can't.

**4. "Quota = capacity" is now explicit (good) — but the failure mode isn't in the risk table.** If `spec.hard` exceeds real node capacity you get Pending pods, not scaling; that's delegated to "the operator's responsibility" rather than listed in Risks. Composing `min(physical, quota)` (point 3) removes the failure mode instead of documenting it away.

**5. No cluster aggregate; Kueue / `ClusterResourceQuota` unconsidered.** Per-namespace isolation is a stated goal, but it means no cluster-wide budget — a cluster admin can't cap *total* GPUs (part of #1002/#1162's scope). And where GPUs are governed by Kueue (`ClusterQueue`/`ResourceFlavor`) or OpenShift `ClusterResourceQuota`, the namespace `ResourceQuota` this reads is empty or stale.

**What I'd like:** keep the capability and the `CapacityDiscovery`/`TypeInventory` reuse, but surface quota as a **composable `ConstraintProvider`** (the A2 `ResourceQuotaReader` in #1162) registered *alongside* node discovery, on the `ScalingPolicy` surface (#1194). Same discovery kernel, different registration point — it keeps the native-RBAC governance strength you've built, gives `min(physical, quota)` for free, inherits per-GPU-type from #1162, and adds no surface we'll migrate later. And it isn't blocked on #1194 landing: the `ConstraintProvider` seam (and `mergeConstraints`) is already in `main`, so registering a quota provider alongside node discovery is a change against today's code — #1162/#1194 are just its eventual config home. Happy to co-author.

In short: I'm not arguing against starting simple — I'm arguing against introducing things we already know we'll have to undo.

cc @lionelvillard

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,845 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+
+---
+
+## Summary
+
+Add `NamespaceQuotaDiscovery`, a new implementation of the existing `CapacityDiscovery`
+interface that derives GPU capacity from Kubernetes `ResourceQuota` objects rather than
+from GPU Operator node labels. When enabled via `WVA_CAPACITY_NAMESPACES=ns-a,ns-b,ns-c`,
```
lionelvillard, line 39: Why WVA_CAPACITY_NAMESPACES needed?

asm582: To make it explicit that resource quotas are read from the configured namespace. A simplification here could be that WVA operates either in resource-quota mode or in GPU-operator mode. The switch is a single boolean: WVA_QUOTA_CAPACITY_ENABLED=true. No namespace list needed, as WVA already knows which namespaces it manages and reads quotas from all of them. 
Mixed-mode/Hybrid mode (some namespaces quota-based, others operator-based) is explicitly out of scope.

asm582: Actually thinking more, if mixed/hybrid mode is out of scope, we can remove this boolean flag too, agree?

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,845 @@
...
+
+Beyond fixing the discovery problem, knowing a trustworthy total GPU ceiling is a
+prerequisite for a broader class of fair-share problems. Consider two models served in the
+same namespace, each initially using 10% of the cluster's GPUs. Model A's request volume
+increases 10× and scales to 90%, leaving model B at 10%. When model B's request volume
+subsequently increases 10×, the correct allocation is 50/50 — but a naive independent
+autoscaler will not rebalance: model A still has a queue, so it sees no reason to release
+GPUs. Solving this requires a redistributive optimizer that can force model A below its
+current allocation even while it is still saturated. That optimizer cannot be built without
+a reliable GPU ceiling to divide. This KEP provides that ceiling; the redistributive
```
lionelvillard: nit: this is not a KEP. Replace by "proposal"

lionelvillard: > That optimizer cannot be built without a reliable GPU ceiling to divide

I would add "and also without a reliable Pod ceiling to divide". Something to keep in mind the future.

asm582: Can you explain more, `and also without a reliable Pod ceiling to divide?`

docs/proposals/namespace-quota-capacity-discovery.md
asm582: This proposal does not solve the configuration problem. It reads the objects Kubernetes already provides. When WVA has such a feature(s), reviewed, merged, and implemented, we can plan to use it for sure.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,845 @@
...
+  their managed node pools.
+- **Multi-tenant clusters** commonly restrict node listing to cluster administrators,
+  making per-node inventory inaccessible to namespace-scoped controllers.
+- **Quota-governed namespaces** (common in OpenShift and enterprise on-prem clusters)
+  already encode GPU capacity in `ResourceQuota.spec.hard` — this is the authoritative
+  source and scanning nodes would duplicate it.
+
+In each of these cases a namespace-scoped quota is a practical fallback: the quota hard
+limit equals the GPU capacity visible to workloads in the namespace. Where GPU Operator
+labels are accessible, `K8sWithGpuOperator` remains the preferred backend — it provides
```
lionelvillard, line 67: > `K8sWithGpuOperator` remains the preferred backend

I guess it depends on the requirements. AFAIK `K8sWithGpuOperator` does not support namespace-scoped inventory. Also there is a difference between quota and inventory.

asm582: If needed, we can change the word preferred. It all depends on how the term "quota" is defined in one's cluster configuration. The proposal starts with a quota as capacity (inventory).

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,845 @@
...
+  ```yaml
+  spec:
+    hard:
+      nvidia.com/gpu: 8   # ← this string becomes the TypeInventory pool key
+  ```
+
+  `TypeInventory` pools are therefore keyed by `nvidia.com/gpu` (or `amd.com/gpu`,
+  `intel.com/gpu`). All GPUs in a namespace are treated as one undifferentiated pool.
+  Differentiating by model (A100 vs H100) within a namespace is not supported — use
+  `K8sWithGpuOperator` for heterogeneous GPU namespaces.
```
lionelvillard, line 138: `K8sWithGpuOperator` and `ResourceQuota` are not mutually exclusive.  `ResourceQuota` can carry per-GPU-type quota (via annotations) enforced at the WVA-level. `K8sWithGpuOperator` is not a quota management system.

docs/proposals/namespace-quota-capacity-discovery.md
```diff
@@ -0,0 +1,845 @@
...
+
+- **`VariantAutoscaling` is optional and deprecated.** VA is the older mechanism for
```
lionelvillard, line 140: I would remove this section as it adds no values to this proposal. This is a distraction.

PR #1232, opened 2026-06-03
feat(chart): expose analyzerName and V2 thresholds in helm values

### LINKED ISSUE #1231: feat(chart): expose analyzerName and V2 thresholds in helm values
## Problem

The WVA helm chart's `templates/manager/wva-cp-configmap.yaml` only
templates the V1 saturation fields (`kvCacheThreshold`,
`queueLengthThreshold`, `kvSpareTrigger`, `queueSpareTrigger`).

The V2 (token-based, cost-aware) saturation analyzer is already
implemented and production-ready — `internal/config/saturation_scaling.go`
has full support for `analyzerName: saturation`, `scaleUpThreshold`, and
`scaleDownBoundary`. But there is no way to activate V2 via `helm
install/upgrade --set` or a values file. Users who want V2 today must
apply a manual `kubectl patch` or a standalone ConfigMap after standup,
which is fragile and easy to forget.

This affects **any** WVA deployment (single-variant or multi-variant) —
anyone who wants V2's superior KV-memory-based capacity accounting has
to reach outside the helm release to configure it.

## Proposed solution

Add `analyzerName`, `scaleUpThreshold`, and `scaleDownBoundary` to
`wva.capacityScaling.default` (and to the overrides schema) in
`values.yaml`, and **make V2 the default** (`analyzerName: saturation`).

V2 is the more capable analyzer: it reads real KV-memory capacity from
`vllm:cache_config_info`, aggregates demand across running requests,
waiting queue, and EPP queue, and steers scale-up toward the cheapest
variant first. There is no reason for new deployments to default to V1.

Users who need to stay on V1 can opt out explicitly:

```yaml
wva:
  capacityScaling:
    default:
      analyzerName: ""
```

## Breaking change

Changing the default from `""` (V1) to `"saturation"` (V2) is a
behavior-changing upgrade for existing deployments. See the PR for
details and the recommended CHANGELOG / version-bump guidance.

## Related

- `internal/config/saturation_scaling.go` — V2 config struct already
  supports all three fields.
- `llm-d/llm-d-benchmark#1451` — multi-variant WVA benchmark (immediate
  consumer, but this fix benefits every WVA standup).
- PR: #1232

biranofer:
Closes #1231

## What

Adds `analyzerName`, `scaleUpThreshold`, and `scaleDownBoundary` to
`wva.capacityScaling.default` (and to the per-model overrides schema)
in the WVA helm chart, and makes **V2 the default analyzer**.

### Default behavior after this PR

A plain `helm install workload-variant-autoscaler` now produces:

```yaml
data:
  default: |
    analyzerName: saturation
    kvCacheThreshold: 0.8
    queueLengthThreshold: 5
    kvSpareTrigger: 0.1
    queueSpareTrigger: 3
```

The controller's `ApplyDefaults()` fills in `scaleUpThreshold: 0.85`
and `scaleDownBoundary: 0.70` when those fields are absent.

### Opt out (revert to V1)

```yaml
wva:
  capacityScaling:
    default:
      analyzerName: ""
```

### Override thresholds

```yaml
wva:
  capacityScaling:
    default:
      analyzerName: saturation
      scaleUpThreshold: 0.80    # tighter than the 0.85 default
      scaleDownBoundary: 0.65
```

## ⚠️ Breaking change note

Changing `analyzerName` default from `""` (V1) to `"saturation"` (V2)
is a **behavior-changing default**. Anyone upgrading the chart without
explicitly pinning `analyzerName: ""` in their values file will silently
switch from V1 to V2.

V2 uses real KV-memory capacity from `vllm:cache_config_info` instead
of percentage thresholds, which requires the model server to emit that
metric (available in vLLM ≥ 0.9.x). Deployments on older vLLM versions
or those that suppressed `cache_config_info` will fall back gracefully
(V2 has a `computeReplicaCapacityFallback` path), but the scaling
behavior will differ from V1.

**Recommendation for maintainers:** call this out in the CHANGELOG and
consider bumping the chart's minor version when this is released.

## Files changed

| File | Change |
|------|--------|
| `charts/workload-variant-autoscaler/values.yaml` | `analyzerName: saturation` as default; add `scaleUpThreshold: 0`, `scaleDownBoundary: 0`; update overrides example comment |
| `charts/workload-variant-autoscaler/templates/manager/wva-cp-configmap.yaml` | Conditionally emit all three fields for `default` and per-model override entries |

## Testing

```bash
# Default — V2 is active
helm template r ./charts/workload-variant-autoscaler \
  | grep -A6 'default: |'
# analyzerName: saturation
# kvCacheThreshold: 0.8  ...

# Opt-out — V1 only
helm template r ./charts/workload-variant-autoscaler \
  --set wva.capacityScaling.default.analyzerName="" \
  | grep -A6 'default: |'
# kvCacheThreshold: 0.8  ...  (no analyzerName line)
```

## Related

- `llm-d/llm-d-benchmark#1451` — multi-variant WVA benchmark; once this
  PR is released, the standalone `two-variant-wva-v2-config.yaml`
  ConfigMap and the separate `kubectl apply` step in that guide can be
  removed.

### DIFF
--- a/charts/workload-variant-autoscaler/templates/manager/wva-cp-configmap.yaml
+++ b/charts/workload-variant-autoscaler/templates/manager/wva-cp-configmap.yaml
@@ -23,15 +23,27 @@ metadata:
 data:
   # Global defaults applied to all variants unless overridden
   default: |
+{{- if .Values.wva.capacityScaling.default.analyzerName }}
+    analyzerName: {{ .Values.wva.capacityScaling.default.analyzerName }}
+{{- end }}
     kvCacheThreshold: {{ .Values.wva.capacityScaling.default.kvCacheThreshold }}
     queueLengthThreshold: {{ .Values.wva.capacityScaling.default.queueLengthThreshold }}
     kvSpareTrigger: {{ .Values.wva.capacityScaling.default.kvSpareTrigger }}
     queueSpareTrigger: {{ .Values.wva.capacityScaling.default.queueSpareTrigger }}
+{{- if .Values.wva.capacityScaling.default.scaleUpThreshold }}
+    scaleUpThreshold: {{ .Values.wva.capacityScaling.default.scaleUpThreshold }}
+{{- end }}
+{{- if .Values.wva.capacityScaling.default.scaleDownBoundary }}
+    scaleDownBoundary: {{ .Values.wva.capacityScaling.default.scaleDownBoundary }}
+{{- end }}
 {{- if .Values.wva.capacityScaling.overrides }}
 {{- range $name, $override := .Values.wva.capacityScaling.overrides }}
   {{ $name }}: |
     model_id: {{ $override.modelID | quote }}
     namespace: {{ $override.namespace | quote }}
+{{- if $override.analyzerName }}
+    analyzerName: {{ $override.analyzerName }}
+{{- end }}
 {{- if $override.kvCacheThreshold }}
     kvCacheThreshold: {{ $override.kvCacheThreshold }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


ev-shindin (approved):
This PR simplifies benshmarking of v2 saturation engine.
I think we need test this engine more, because v1 is going to be deprecated. Since HELM is deprecated think also how this should be done in kustomize and prepare follow-up PR

PR #1237, opened 2026-06-04
fix(optimizer): role-aware scale-down for disaggregated models

ev-shindin:
## Summary

`costAwareScaleDown` sheds replicas using the **model-level** `SpareCapacity`,
which aggregates prefill and decode. Because prefill and decode capacity are not
fungible, this lets a role with slack drive removal of replicas from a **saturated**
role — and because removal is most-expensive-variant-first regardless of role, it
can even remove the wrong role's replicas entirely.

This PR makes scale-down **role-aware**: for disaggregated models it sheds each
role independently against that role's own spare and never trims a role with no
spare. Non-disaggregated models are unchanged.

## The bug

For a disaggregated model, model-level spare is
`totalSupply − totalDemand/boundary` summed over **all** variants
(`saturation_v2/analyzer.go`). So when prefill is saturated (`spare_prefill = 0`)
but decode has slack (`spare_decode > 0`), the aggregate is positive and
`costAwareScaleDown` runs — then removes the **most-expensive** variant first,
which can be prefill. The saturated role loses replicas while the role with actual
slack keeps them.

Concretely (1 GPU/replica):

| variant | role | cost | replicas | role spare |
|---|---|---:|---:|---:|
| prefill | prefill | 15 | 2 | 0 (saturated) |
| decode | decode | 5 | 3 | 20000 (2 replicas) |

Model-level spare = `20000` (decode's), so scale-down fires. **Before:** sorted by
cost desc → prefill removed first → `prefill 2 → 0`, decode untouched. The model
loses its entire prefill role. **After:** prefill (no spare) is skipped; decode
sheds its own spare → `decode 3 → 1`, prefill stays at 2.

Latent in the additive optimizer (scale-down only fires on small aggregate spare,
so the skew is small and self-corrects), but it becomes significant for any caller
that drives large reclaims through this primitive.

## The fix

- When `RoleCapacities` is present (disaggregation active), shed **each role**
  against its own `RoleCapacities[role].SpareCapacity`; skip roles with no spare.
- When it is absent, keep the existing model-level shed verbatim.
- Extract the per-variant removal loop into `scaleDownVariantSet` and add an
  exact-match `variantsForRole` helper so per-role sheds operate on disjoint
  variant sets.

Both optimizers (`GreedyByScoreOptimizer`, `CostAwareOptimizer`) call
`costAwareScaleDown`, so both benefit with no other changes.

## Behavior change to note

The cheapest-variant-at-1 protection is now scoped **per role**, so scale-down
keeps **≥ 1 replica in each active role** — a disaggregated model can no longer
have its prefill *or* decode role zeroed by normal scale-down (the old model-wide
protection could zero a whole role). This is more correct for P/D serving; full
model scale-to-zero is a separate path and is unaffected.

### DIFF
--- a/internal/engines/pipeline/cost_aware_optimizer.go
+++ b/internal/engines/pipeline/cost_aware_optimizer.go
@@ -144,6 +144,9 @@ func costAwareScaleDown(
 	if len(result.RoleCapacities) > 0 {
 		// Each role owns a disjoint set of variants and sheds against its own
 		// spare, so the map's iteration order does not affect the outcome.
+		// Assumes RoleCapacities keys partition VariantCapacities by role
+		// (one role per variant; no overlap); mixed-role configurations would
+		// break disjointness.
 		for role, rc := range result.RoleCapacities {
 			if rc.SpareCapacity <= 0 {
 				continue // saturated or under-supplied role — never trim it
--- a/internal/engines/pipeline/cost_aware_optimizer.go
+++ b/internal/engines/pipeline/cost_aware_optimizer.go
@@ -121,29 +121,58 @@ func costAwareScaleUp(
 	}
 }
 
-// costAwareScaleDown removes replicas from the most expensive variant.
-// Sorts by absolute cost descending, removes from most expensive first.
-// Respects minReplicas per variant — will not scale below the annotation floor.
-// The cheapest variant is protected at min 1 replica only when no other variant
-// has replicas — this prevents scale-down deadlocks where the expensive variant's
-// per-replica capacity exceeds spare but cheaper replicas could be removed.
+// costAwareScaleDown removes replicas to shed spare capacity, most-expensive
+// variant first, respecting minReplicas.
+//
+// For disaggregated (prefill/decode) models it sheds each role independently
+// against that role's own spare. Prefill and decode capacity are not fungible, so
+// removing by the model-level SpareCapacity — which aggregates all roles — would
+// let a role with slack drive removal of replicas from a saturated role (e.g.
+// trimming prefill because decode has spare). RoleCapacities is non-nil only when
+// disaggregation is active; non-disaggregated models keep the model-level shed.
 func costAwareScaleDown(
 	ctx context.Context,
 	result *interfaces.AnalyzerResult,
 	targets map[string]int,
 	stateMap ...map[string]interfaces.VariantReplicaState,
 ) {
-	logger := ctrl.LoggerFrom(ctx)
-
-	sorted := sortByCostDesc(result.VariantCapacities)
-	cheapest := findCheapestVariant(result.VariantCapacities)
-	remaining := result.SpareCapacity
-
-	// Build state lookup if provided
 	var states map[string]interfaces.VariantReplicaState
 	if len(stateMap) > 0 {
 		states = stateMap[0]
 	}
+
+	if len(result.RoleCapacities) > 0 {
+		// Each role owns a disjoint set of variants and sheds against its own
+		// spare, so the map's iteration order does not affect the outcome.
+		for role, rc := range result.RoleCapacities {
+			if rc.SpareCapacity <= 0 {
+				continue // saturated or under-supplied role — never trim it
+			}
+			scaleDownVariantSet(ctx, variantsForRole(result.VariantCapacities, role), rc.SpareCapacity, targets, states)
+		}
+		return
+	}
+
+	scaleDownVariantSet(ctx, result.VariantCapacities, result.SpareCapacity, targets, states)
+}
+
+// scaleDownVariantSet removes replicas from the given variant set, most-expensive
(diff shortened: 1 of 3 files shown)


asm582 (commented):
/lgtm

deanlorenz:
nit (cost_aware_optimizer.go, variantsForRole): use the `interfaces.RoleBoth` constant consistently. The new helper uses `interfaces.RoleBoth` for the empty-role canonicalization, but the sibling `filterVariantCapacitiesByRole` in `greedy_score_optimizer.go` still uses the string literal `"both"`. Aligning both on the constant is a trivial change.

deanlorenz:
suggestion (cost_aware_optimizer.go, costAwareScaleDown disag branch): the "iteration order does not affect outcome" comment is correct under a standing invariant — one role per variant; `RoleCapacities` keys partition `VariantCapacities` by role. That invariant is broader than scale-down and isn't asserted anywhere visible from this code path. A one-line note here locks it in code:

    // Assumes RoleCapacities keys partition VariantCapacities by role
    // (one role per variant; no overlap). Mixed-role configurations
    // would break disjointness.

deanlorenz:
test gap (cost_aware_optimizer_test.go, disag scale-down context): both new tests populate `RoleCapacities` with both prefill and decode entries. Adding a test where `RoleCapacities` has only one role (e.g., decode-only during a data-collection gap, or prefill-only registered) exercises the one-iteration path and pins the cheapest-at-1 protection within that single role.

deanlorenz:
test gap (cost_aware_optimizer_test.go, disag scale-down context): variant with `Role: ""` alongside non-empty `RoleCapacities` entries. The standing design assumption forbids this state, but `variantsForRole` silently drops the empty-role variant from per-role sheds (its canonicalized role is `RoleBoth`, which has no matching `RoleCapacities` key in a P/D model). One short test pinning the intended behavior — skipped silently, or treated as defensive — closes the loop on what happens when the assumption is violated.

deanlorenz:
behavior-change confirmation: PR body says full ScaleToZero across both roles is unchanged. Worth one explicit test asserting it: configure `scale-to-zero` policy active, drive zero demand on both roles, confirm both go to zero. The new per-role cheapest-at-1 protection layers on top of the ScaleToZero path; asserting they compose as described prevents accidental regression in the protection logic.

internal/engines/pipeline/cost_aware_optimizer.go
```diff
@@ -121,30 +121,59 @@ func costAwareScaleUp(
...
+	ctx context.Context,
+	variants []interfaces.VariantCapacity,
+	spare float64,
+	targets map[string]int,
+	states map[string]interfaces.VariantReplicaState,
+) {
+	logger := ctrl.LoggerFrom(ctx)
+
+	sorted := sortByCostDesc(variants)
+	cheapest := findCheapestVariant(variants)
```
deanlorenz: issue (cheapest-protection): collapsing the redundancy + correctness-under-ties:

1. `findCheapestVariant(variants)` is redundant after `sortByCostDesc`. Drop the call; use `i == len(sorted)-1` instead. Same semantics in the common case, and self-correcting if the sort ever changes.

2. Implicit assumption today: protection only works because cheapest is processed last (i.e., descending sort). The `i == len(sorted)-1` change makes that explicit.

3. The `for _, other := range variants` walk runs on every cheapest iteration, including when the unprotected shed would never drop current below 1. Move it inside the conditional `if i == len(sorted)-1 && current-n < 1` and walk `sorted[:i]` instead of the full set. (The annotation-min check the current code has — `minReplicas < 1` — is redundant once the lazy form clamps after computing n: if minReplicas ≥ 1, then removable ≤ current-1, so n ≤ current-1, so current-n ≥ 1 and the block never enters.)

4. Tie-breaking under equal cost. `sort.Slice` is not stable, and `findCheapestVariant` returns the first-by-input-order minimum, so the two can name different \"cheapest\" variants under ties. Suggest making the sort tie-break by PRC ascending so that the variant with the highest PRC at the cheapest cost-tier ends up at \`len-1\`:

\`\`\`go
sort.Slice(sorted, func(i, j int) bool {
    if sorted[i].Cost != sorted[j].Cost {
        return sorted[i].Cost > sorted[j].Cost
    }
    return sorted[i].PerReplicaCapacity < sorted[j].PerReplicaCapacity
})
\`\`\`

This matches the principle \"among equal-cost variants, prefer keeping the one with higher capacity per replica.\"

deanlorenz (approved):

internal/engines/pipeline/cost_aware_optimizer.go
ev-shindin: fixed in [c86495a](https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1237/commits/c86495a95e29455e0722cb4af96d14ef2abfea27)

PR #1240, opened 2026-06-05
docs(dev): modernize Kind setup docs and add simulator sample stack

mamy-CS:
The Kind developer setup docs were pointing at deploy-wva-emulated-on-kind which no longer exists after the llm-d infra refactor. This updates all docs to use CREATE_CLUSTER=true make deploy-e2e-infra and adds config/samples/simulator/ — a complete, ready-to-apply stack (Deployment + Service + ServiceMonitor + VA + HPA) using the simulator so developers can get a fully working environment with metrics flowing in two commands. Verified end-to-end on Kind.


shuynh2017:
@mamy-CS, can we quickly remove VA object and annotate HPA object in this PR, or should be done in another PR?

shuynh2017 (dismissed):

shuynh2017 (approved):

PR #1245, opened 2026-06-07
docs(proposals): minimal ScalingPolicy CRD core (split from #1194)

ev-shindin:
## What

A **bare-minimum** design for the WVA configuration CRD — `docs/proposals/design-scalingpolicy-crd.md` split out from #1194 so we can agree the CRD's *boundary* and *shape* before anything is layered on top.

**#1194 stays open as the reference** for the full analysis (migration, deprecation phases, status catalog, alternatives, comparison matrices); none of that is decided here.

## Why

Per review of #1194: *"The current proposal covers too much ground and should be trimmed to the bare minimum. Having a CRD is a good start. The questions to answer are what is the relationship between the Config CRD and the HPA object, and what is the generic shape (schema) of the CRD."*

This PR answers exactly those two questions and settles only the minimum they force (the resolution tiers and the status shape).

## The two questions

**1. CRD ↔ HPA relationship (§2).** They are config-decoupled. `ScalingPolicy` configures WVA's *decision*; the HPA/`ScaledObject` *actuates* it by reading the `wva_desired_replicas` metric. Neither references the other's config — so adding WVA config never means editing a tenant's HPA. `maxReplicas` stays as a complementary hard ceiling; scale-to-zero is a policy toggle the HPA layer actuates. (WVA *does* read the HPA/`ScaledObject` to identify the workloads it manages — but the discovery *mechanism* belongs to the separate VA-deprecation proposal, see below, not this CRD's shape.)

**2. Generic schema shape (§3).** A thin **typed envelope** (`priority`, `scaleToZero`, `inferencePoolRef`) around **pluggable `{type, name, parameters}` lists** (`analyzers`, `limiters`) whose `parameters` are `x-kubernetes-preserve-unknown-fields` — the EPP `EndpointPickerConfig` shape. Adding an analyzer or limiter never changes the CRD schema.

## On "even multiple levels is too much"

Kept **three tiers** (cluster-default → namespace → per-pool), but minimally and structurally, not by preference (§4):

- **Quota forces cluster + namespace.** Quota is cluster-admin-owned and cluster-scoped; thresholds/`priority`/`scaleToZero` are tenant-owned and namespace-scoped. Different owners, different RBAC — they can't share one object without letting tenants edit quota or admins own every tenant's thresholds.
- **Per-pool costs zero extra schema** — it's the *same* CRD with `inferencePoolRef` set, in the tenant's namespace; override granularity, not a new surface.

So "multiple levels" is three placements of **one** CRD resolved by a small deterministic merge — not three configuration systems. **`limiters`/`quota` are cluster-default-only** (CEL-rejected on namespace/per-pool, and tenants lack RBAC on the system namespace), so a tenant can't drop the quota limiter and escape their own cap.

## Status (§4) and what's deferred

`status` surfaces `effectivePolicy`/`sources` (the tier-merge result) and `managedVariants` — the workloads this policy manages, each with `workloadRef`, `role`, and per-replica `cost`. This proposal fixes only the **shape** of those fields.

**Two dependencies/open items (§5):**
- **Discovery and cost-sourcing depend on VA-deprecation proposal**. That proposal owns how WVA identifies managed workloads — currently the `llm-d.ai/managed` label on the HPA/`ScaledObject` (#1130) — and where per-variant `cost` comes from (`VariantAutoscaling` `cost` today; co-located there post-VA). This proposal is independent of the mechanism.
- **Open question:** where the effective policy is published for a pool governed only by defaults (no per-pool object). Candidates: `InferencePool` status, a status-only object, or a `wva-config explain` view.

**Out of scope, deferred to #1194 (§6):** migration tooling, deprecation phases, the full status-condition catalog, per-role config, alternatives/comparison matrices, and the `quota` `parameters` schema (owned by #1162).

---

Looking for agreement on **§2 (boundary)** and **§3 (shape)** first; the rest layers on top without changing either. #1194 remains open as the reference companion.

### DIFF
--- a/docs/proposals/design-scalingpolicy-crd.md
+++ b/docs/proposals/design-scalingpolicy-crd.md
@@ -1,144 +1,220 @@
-# Proposal: ScalingPolicy CRD — minimal core
+# Proposal: WVA scaling configuration — schema and resolution (ConfigMap first, CRD later)
 
-## Scope
+## Problem — what this addresses
 
-This is the **bare-minimum** core of the WVA configuration CRD, split out from the broader analysis in **#1194** (kept open for reference). Per review, it answers the two questions that gate the design — and settles only the minimum they force (the resolution tiers and the status surface):
+Today WVA is configured by **one cluster-wide ConfigMap** plus **per-workload labels/annotations** on the KEDA `ScaledObject` (`llm-d.ai/managed`, `llm-d.ai/model-id`, `inference.optimization/acceleratorName`). That is enough for a single global configuration, but it cannot express:
 
-1. **What is the relationship between the Config CRD and the HPA object?** (§2)
-2. **What is the generic shape (schema) of the CRD?** (§3)
+- **per-namespace or per-pool policy** — every workload shares one global `priority`, `scaleToZero`, and analyzer configuration;
+- **an ownership boundary** — cluster-admin-owned enforcement (GPU limiters / quota) and tenant-owned tuning (thresholds, priority) live in the same undifferentiated blob, with no way to give them different RBAC;
+- **validation** — a typo in the ConfigMap is silently ignored;
+- **a resolved-configuration surface** — an operator cannot see *what policy WVA actually applied* to a given pool.
 
-Everything else from #1194 — migration tooling, deprecation phases, the full status-condition catalog, per-role detail, alternatives, comparison matrices — is **deliberately deferred** (§6); the dependency on the VA-deprecation proposal (discovery, cost source) and one remaining open question are in §5. The goal here is to agree the CRD's *boundary* and *shape* first; the rest layers on top without changing either.
+This proposal fixes the configuration **schema** and **resolution** (the fields, the scope tiers, the pluggable-plugin envelope, and the merge order). It is **transport-neutral** and lands in two phases:
 
-## 1. The object
+- **Phase 1 — extend the existing ConfigMap** with this schema. No CRD, no migration, no new install step. Validation via `ValidatingAdmissionPolicy` (CEL) + the controller; the ownership boundary via RBAC. This is the minimal core.
+- **Phase 2 — promote the schema to a `ScalingPolicy` CRD** when field-level typed validation, `kubectl explain`, and a native `.status` are worth the operational cost of a CRD (§6). Same schema, same resolution — a delivery upgrade, not a redesign.
 
-One namespaced CRD, `ScalingPolicy` (`scaling.llm-d.ai/v1alpha1`), carries WVA's scaling configuration. It is matched to a workload through that workload's **`InferencePool`** (defined by the Gateway API Inference Extension; the workload→pool link is the pool's pod selector) — not through the workload, and not through its HPA:
+It does **not** change how WVA *actuates* (it still emits `wva_desired_replicas` for a `ScaledObject`, §3) or how it *identifies* the workloads it manages (§7).
+
+## Optional and backward-compatible
+
+The structured configuration is **optional and additive**. A workload with only the managed labels keeps scaling under WVA's defaults exactly as today — the schema adds *scoped overrides*, it does not become required. Upgrading an existing cluster needs **no migration**: phase 1 *is* the ConfigMap WVA already reads, with more structure in it. Fields that exist today (`priority`, `scaleToZero`, analyzer settings, limiter selection) keep their current meaning at the cluster tier; per-namespace and per-pool scoping is the new capability.
+
+**Before / after — the cluster tier is today's ConfigMap, restructured.** Today's flat cluster config gains a scope label and the typed shape; the values carry over:
 
 ```yaml
+# today (illustrative): one global ConfigMap
+data: { priority: "1.0", scaleToZero: "false", analyzer: "saturation", scaleUpThreshold: "0.9" }
+# phase 1: same values, structured, labelled as the cluster-default tier (§2)
+metadata: { labels: { scaling.llm-d.ai/config: "true", scaling.llm-d.ai/tier: cluster } }
+data:
+  config: |
+    priority: 1.0
+    scaleToZero: { enabled: false }
+    analyzers: [ { type: saturation, parameters: { scaleUpThreshold: 0.9 } } ]
+    limiters:  [ { type: gpu-inventory } ]      # admin-only, cluster tier
+```
+
+**Minimal usage.** A pool with no `ScalingPolicy`/ConfigMap of its own scales under the cluster default (unchanged behavior). To raise one pool's priority, a tenant adds *just* the per-pool object with only the delta:
+
+```yaml
+metadata: { namespace: production, labels: { scaling.llm-d.ai/config: "true", scaling.llm-d.ai/tier: pool, scaling.llm-d.ai/pool: granite-premium-pool } }
+data: { config: "inferencePoolRef: { name: granite-premium-pool }\npriority: 2.0" }
+```
+
+Everything else (`scaleToZero`, `behavior`, `analyzers`) inherits from the tiers above (§2).
+
+## 1. The configuration object and how it is matched
+
+The configuration is **one schema** carried by one object per scope:
+
+- **Phase 1:** a ConfigMap whose `data` holds the schema as a YAML document (one per tier, §2).
+- **Phase 2:** a namespaced `ScalingPolicy` custom resource (`scaling.llm-d.ai/v1alpha1`).
+
+It is matched to a workload through that workload's **`InferencePool`** (the Gateway API Inference Extension object; the workload→pool link is the pool's pod selector) via `inferencePoolRef` — **not** through the workload and **not** through its `ScaledObject`. The config points at the pool; the pool's workloads carry no reference back.
+
+```yaml
+# Phase 2 (CRD) form — the phase-1 ConfigMap carries the same document under data:
 apiVersion: scaling.llm-d.ai/v1alpha1
 kind: ScalingPolicy
-metadata:
-  name: granite-premium-priority
-  namespace: production
+metadata: { name: granite-premium-priority, namespace: production }
 spec:
-  inferencePoolRef:
-    name: granite-premium-pool      # the pool this policy configures
-  priority: 2.0
-  scaleToZero: { enabled: false }
+  inferencePoolRef: { name: granite-premium-pool }   # the pool this config governs (per-pool tier)
+  priority: 2.0                                       # per-model; see §2
+  scaleToZero: { enabled: false }                     # per-model
+  behavior:                                           # per-model; autoscaling/v2 HPA behavior (§4)
+    scaleDown: { stabilizationWindowSeconds: 300 }
   analyzers:
     - type: saturation
-      parameters: { scaleUpThreshold: 0.95 }
+      parameters: { scaleUpThreshold: 0.95 }          # plugin-owned; see §4
 ```
 
-The policy **points at the pool**; the pool's workloads carry no reference back to the policy. Why that direction matters is §2.
+The tunable field families and their meaning, defined here so nothing below is referenced before it is defined:
 
-## 2. Relationship to the HPA / ScaledObject (review question 1)
+- **`priority`** *(per-model)* — the model's weight when models contend for a shared GPU budget. It is only meaningful **relative to other models drawing on the same budget**: within a namespace quota if one exists, otherwise the shared cluster budget (so priority *does* compose across un-quota'd namespaces on the same accelerator type — how it does is engine/rescale territory, not this schema). It is not comparable across *independent* budgets. Default `1.0`.
+- **`scaleToZero`** *(per-model)* — whether WVA may emit a desired count of `0` for an idle model (§3).
+- **`behavior`** *(per-model, tenant-tunable)* — the HPA-style stabilization behavior WVA applies to its own recommendation (trailing windows, per-period rate policies, tolerance) before emitting it. This is a **typed** field reusing `k8s.io/api/autoscaling/v2` `HorizontalPodAutoscalerBehavior` verbatim (§4); its *shape* is fixed here, its *algorithm* is the stabilization proposal's.
+- **`analyzers`** *(per-model, tenant-tunable)* — the saturation/queueing analysis plugins and their thresholds (§4). Per-**role** (prefill/decode) tuning lives *inside* an analyzer's `parameters`, not as top-level fields (§5).
 
-**They own different things:** WVA owns the scaling decision, the HPA owns actuation. `ScalingPolicy` configures the decision; the HPA actuates it by reading WVA's `wva_desired_replicas` metric. Neither the policy nor the HPA references the other's *config*. WVA reads the HPA/`ScaledObject` to identify the workloads it manages, but the *mechanism* (the #1130 label) is the **VA-deprecation proposal's** — and this CRD's shape does not depend on it.**
+`limiters`/`quota` are **not** in that list — they are admin-only and cluster-scoped (§2).
 
-| | Owns | Reads | Writes |
+## 2. Scope tiers — one schema, up to three placements
+
+The surface is **one schema**, resolved at up to **three placements of it** — not three schemas, not three kinds — by a deterministic lookup in fixed order. In phase 1 each placement is a ConfigMap; in phase 2 each is a `ScalingPolicy` object.
+
+1. **Cluster-default** — in the system namespace, no `inferencePoolRef`. **Admin-owned.**
+2. **Namespace-default** — in the workload's namespace, no `inferencePoolRef`. **Tenant-owned.**
+3. **Per-pool** — in the workload's namespace, `inferencePoolRef` set. **Tenant-owned.** A workload-specific policy is *just this placement* — it needs only the fields it overrides.
+
+**How the controller finds a tier (phase 1).** A CRD has identity from its kind, namespace, and `spec.inferencePoolRef`; a ConfigMap does not, so phase 1 classifies ConfigMaps by a fixed label set the controller watches (`scaling.llm-d.ai/config: "true"`):
+
+| tier | namespace | labels |
+|---|---|---|
+| cluster-default | system (`workload-variant-autoscaler-system`) | `scaling.llm-d.ai/tier: cluster` |
+| namespace-default | workload's | `scaling.llm-d.ai/tier: namespace` |
+| per-pool | workload's | `scaling.llm-d.ai/tier: pool`, `scaling.llm-d.ai/pool: <inferencePool>` |
+
+The controller resolves a workload's policy from the labelled ConfigMaps in (system-ns, workload-ns); the `inferencePoolRef` inside the document must agree with the `pool` label (admission-checked). In phase 2 this labelling goes away — the CR's kind + namespace + `spec.inferencePoolRef` are the identity.
+
+**Resolution is inheritance, not replacement.** Fields merge **cluster → namespace → per-pool**, higher tier winning **per field** — a per-pool object that sets only `priority` inherits `scaleToZero`, `behavior`, and `analyzers` from the tiers above it. `analyzers` merge **by `name`** (§4): a lower tier can add or replace an analyzer entry without redeclaring the others. Note the granularity: because a plugin's `parameters` is **opaque** to the envelope (§4), it cannot be structurally deep-merged — a same-`name` entry **replaces the whole `parameters` block**, it does not patch one key. So a tenant writes the minimum set of *entries*, but retuning one threshold means restating that analyzer's `parameters`. This is the "narrower than intended" point from review: real per-field/per-entry inheritance across tiers, with parameters as the replace-granular leaf.
+
+**One kind, two purposes — deliberately.** The same schema serves two owners because it is placed differently and gated by different RBAC: at the **cluster-default** tier it is the admin's **enforcement** carrier (limiters/quota); at the **namespace/per-pool** tiers it is the tenant's **tuning** surface (priority, scaleToZero, analyzers). This dual purpose is intentional — one schema and one merge, partitioned by placement and RBAC, rather than two objects to keep in sync.
+
+**Which fields live at which tier:**
+
+| field family | tiers allowed | owner | rationale |
 |---|---|---|---|
-| **HPA / KEDA `ScaledObject`** (tenant-owned) | replica **actuation** | the `wva_desired_replicas` metric | the workload's `spec.replicas` |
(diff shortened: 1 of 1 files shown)


lionelvillard:
This proposal is about providing a CRD for configuring `wva_desired_replicas` metric behavior. Considering the scope, LGTM.

asm582:
In another review, it was recommended to explore [EPP pattern](https://github.com/llm-d/llm-d-router/blob/088c5edbde5235fff4e188e8d9323f61bdc7302d/docs/architecture.md?plain=1#L90) to reduce operation complexity. 

CRD-caused operational complexity is also agreed upon by a community user. 

Since this CRD is a detail for improving `wva_desired_replicas` metric, for this scope and area

/lgtm

ev-shindin:
cc @vivekk16

vivekk16 (commented):
@ev-shindin @lionelvillard  to confirm my understanding, with the current proposal, the P/D single inference pool case would be handled entirely within the analyzer's parameters (since priority and scaleToZero are per model concerns, not per role):

```
apiVersion: scaling.llm-d.ai/v1alpha1
kind: ScalingPolicy
metadata:
  name: granite-policy
  namespace: production
spec:
  inferencePoolRef:
    name: granite-pool
  priority: 2.0
  scaleToZero: { enabled: false }
  analyzers:
    - type: saturation
      parameters:
        scaleUpThreshold: 0.90
        scaleDownBoundary: 0.70
        roleOverrides:
          prefill:
            scaleUpThreshold: 0.80
            scaleDownBoundary: 0.60
          decode:
            scaleUpThreshold: 0.95
            scaleDownBoundary: 0.80
```
Is this how you envision it working? The engine today passes the same saturationConfig to every role group in analyzeRoleGroups, so beyond the CRD, we'd need engine changes to resolve and apply per role thresholds from the analyzer parameters. That's out of scope for this proposal but should be tracked separately.

cc @Gregory-Pereira @shuynh2017 @pierDipi

deanlorenz:
Agree with the scope and the §2/§3 answers. A few clarity items, mostly on §3–§5 where the text describes a model narrower than the design seems to intend.

**1. Inheritance is described as merge only (§4).** The text covers what happens when multiple CRs exist (merge, higher tier wins) but not what happens when a node has *no* CR — the common case. Suggest stating it: *a pool/namespace with no `ScalingPolicy` inherits its parent's resolved policy.* WVA could also generate an empty per-pool CR so the effective policy is visible through its status — which directly answers §5's open question (default-only pools): the question becomes "where is inherited policy surfaced," not "is there policy."

**2. Root CR is dual-purpose (§3).** `analyzers[]` has two meanings by tier: at the root tier it is **definitional** (the enabled-analyzer set + default thresholds/scores — effectively the registration list); at lower tiers it is an **override**. The text treats it uniformly. One line distinguishing the two would prevent reading the root list as just another override.

**3. Controller modes (§4).** §4 assumes a cluster-scoped controller with namespace tenants. WVA can also run namespace-scoped (root CR = the namespace CR, no cluster tier) — the "cluster-default + namespace" framing and the RBAC argument change shape in that mode. Worth one sentence stating which modes are in scope.

**4. `limiters`/`quota` ownership (§3 lines 77–90; §4 line 100).** As written, external quota (#1162) sits *inside* `ScalingPolicy` as a limiter, while §3 also says it isn't WVA-owned. "Excluded from the merge / cluster-default-only" is correct for an external constraint, but conflicts with the inheritance model for anything WVA *does* own. Suggest treating any WVA-owned cap as ordinary inheritable policy — validated at each level so the effective cap stays ≤ quota and ≤ physical limits, reflected in `effectivePolicy`. That removes the apparent contradiction between "limiters excluded from merge" and "lower tiers inherit resolved policy."

**5. Priority across role-pools (§3).** With one model per pool today, a single per-pool `priority` is unambiguous — no coordination needed. Recommend stating that explicitly, plus that if a future EPP splits one model into separate role-pools, cross-pool priority precedence is decided then. Converts a silent assumption into a bounded deferral.

**6. Schema vs. current engine (§3).** The schema permits per-pool analyzer enable/disable and per-pool thresholds; the engine applies a single global threshold set today. Suggest noting these as schema-supported / engine-future so the schema isn't read as current behavior.

(2) and (5) are one-line clarifications; (1), (3), (6) are short additions; (4) is the only one that may touch the schema.

shuynh2017:
May be the CRD can be more flexible by using more generic tags, not `inferencePoolRef`, `prefill`, `decode`, and may be these can be some kind of types.

shuynh2017:
@ev-shindin, did we address this comment https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1245#pullrequestreview-4454337956 from @vivekk16, and https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1245#issuecomment-4662740902 from @deanlorenz ?

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
+# Proposal: ScalingPolicy CRD — minimal core
+
```
shuynh2017, line 2: Can we list what are the current issues this proposal aims to address?

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+## Scope
+
+This is the **bare-minimum** core of the WVA configuration CRD, split out from the broader analysis in **#1194** (kept open for reference). Per review, it answers the two questions that gate the design — and settles only the minimum they force (the resolution tiers and the status surface):
```
shuynh2017: I think we should just keep this proposal clean by not referencing #1194. It's easier on reader/reviewer.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+spec:
+  inferencePoolRef:
+    name: granite-premium-pool      # the pool this policy configures
+  priority: 2.0
+  scaleToZero: { enabled: false }
+  analyzers:
+    - type: saturation
+      parameters: { scaleUpThreshold: 0.95 }
+```
+
```
shuynh2017, line 85: immediate questions:
- is this ScalingPolicy resource optional? (upgrade case). On second thought, it should be optional since WVA can scale a target with just labels/annotations.
- how are the fields such as `priority`, `scaleToZero`, `analyzers`, etc ... relate to those currently exist?

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+The policy **points at the pool**; the pool's workloads carry no reference back to the policy. Why that direction matters is §2.
+
+## 2. Relationship to the HPA / ScaledObject (review question 1)
```
shuynh2017: Let's remove HPA for the whole doc?

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+
+Everything else from #1194 — migration tooling, deprecation phases, the full status-condition catalog, per-role detail, alternatives, comparison matrices — is **deliberately deferred** (§6); the dependency on the VA-deprecation proposal (discovery, cost source) and one remaining open question are in §5. The goal here is to agree the CRD's *boundary* and *shape* first; the rest layers on top without changing either.
+
+## 1. The object
+
+One namespaced CRD, `ScalingPolicy` (`scaling.llm-d.ai/v1alpha1`), carries WVA's scaling configuration. It is matched to a workload through that workload's **`InferencePool`** (defined by the Gateway API Inference Extension; the workload→pool link is the pool's pod selector) — not through the workload, and not through its HPA:
+
+```yaml
+apiVersion: scaling.llm-d.ai/v1alpha1
+kind: ScalingPolicy
```
shuynh2017, line 73: If we were to introduce a CRD, then may be we should name it a bit more general as such it can grow, in this case more than just policy. May be name it `ScalingResource`, `WVAScalingObject` like Keda doing here https://github.com/kedacore/keda-docs/blob/main/content/docs/2.15/reference/scaledobject-spec.md. As such we won't have to introduce a new CRD when needed as I recall we've reduced by removing variant auto scaling CRD.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+  analyzers:
+    - type: saturation
+      parameters: { scaleUpThreshold: 0.95 }
+```
+
+The policy **points at the pool**; the pool's workloads carry no reference back to the policy. Why that direction matters is §2.
+
+## 2. Relationship to the HPA / ScaledObject (review question 1)
+
+**They own different things:** WVA owns the scaling decision, the HPA owns actuation. `ScalingPolicy` configures the decision; the HPA actuates it by reading WVA's `wva_desired_replicas` metric. Neither the policy nor the HPA references the other's *config*. WVA reads the HPA/`ScaledObject` to identify the workloads it manages, but the *mechanism* (the #1130 label) is the **VA-deprecation proposal's** — and this CRD's shape does not depend on it.**
```
shuynh2017: scaledObject (which is the only scaler WVA supports) can scale w/o WVA so this is confusing:
```
*They own different things:** WVA owns the scaling decision, the HPA owns actuation
```

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+
+The data flow is one-directional:
+
+```
+ScalingPolicy ──configures──▶ WVA ──emits──▶ wva_desired_replicas ──read by──▶ HPA / ScaledObject ──sets──▶ spec.replicas
+```
+
+The consequences answer the question directly:
+
+- **No config cross-reference.** The HPA / `ScaledObject` carries no reference to a WVA *config* object, and `ScalingPolicy` points at the `InferencePool`, **not** the HPA — so configuring WVA never means editing a tenant's HPA. The HPA's tie to WVA is that it reads the `wva_desired_replicas` metric to actuate. (An `HPA → ScalingPolicy` config reference would put a WVA-specific field on every workload — that's what this avoids.)
```
shuynh2017: This is confusing. Can we state what we are doing right now? and what's the change? Context: This is a help message in dashboard:  `Number of models processed in the last optimization cycle. For a variant to be found, its ScaledObject must be properly annotated with llm-d.ai/managed and llm-d.ai/model-id. The ScaledObject must also set the label inference.optimization/acceleratorName. Metric is wva_models_processed.`

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+- **WVA never *writes* tenant-owned objects.** It does not set `spec.replicas` or modify the HPA / `ScaledObject`; it publishes the `wva_desired_replicas` metric, which the HPA consumes. WVA *does* **read** the HPA/`ScaledObject` to identify its managed set — the mechanism (the `llm-d.ai/managed` label on those objects, #1130) belongs to the VA-deprecation proposal, not this one.
```
shuynh2017: This statement doesn't add any more clarity. I think if we address the above comment then we don't need this statement.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+- **HPA `maxReplicas` stays, as a complementary hard ceiling.** `ScalingPolicy` does not replace it: WVA's desired count is clamped by the HPA exactly as today. `ScalingPolicy` sets the *policy* WVA decides within; `maxReplicas` is the tenant's final per-workload safety cap.
+- **Scale-to-zero is a policy toggle, actuated by the HPA layer.** `spec.scaleToZero.enabled` lets WVA emit a desired of 0; KEDA/HPA actuate it. WVA decides 0; it does not force it.
```
shuynh2017: I think this phrase `WVA emit a desired of 0; KEDA/HPA actuate it` is sufficient, the additional phrase `WVA decides 0; it does not force it.` doesn't help to clarify, so suggest to remove.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+
+**Net:** a workload is *configured* via its pool (§1's "not through the HPA") and *actuated* via the metric; the CRD sits *upstream* of that metric, so it slots into the existing HPA/KEDA flow without modifying it. How a workload is *identified as WVA-managed* is the VA-deprecation proposal's, not here.
```
shuynh2017: let's clarify this here since I don't think we can remember all the details of the "VA-deprecation proposal". Let's clarify specially if there are changes from current implementation.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+
+## 3. The generic schema shape (review question 2)
+
+The CRD is a **thin typed envelope around pluggable, schemaless plugin lists** — so the schema itself is small and fixed, and new analyzers/limiters never change it. Two validation tiers, deliberately:
+
+```yaml
+spec:
+  # (a) typed, cross-cutting fields — OpenAPI/CEL-validated, `kubectl explain`-able
+  inferencePoolRef: { name: granite-premium-pool }   # per-pool only; mutually exclusive with limiters (§4)
+  priority: 2.0
```
shuynh2017, line 118: add a comment on priority when there are multiple priorities in same namespace, among namespaces?

... I can see it explains a bit later line 103. The reading is harder when things are referenced before defined.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+**Net:** a workload is *configured* via its pool (§1's "not through the HPA") and *actuated* via the metric; the CRD sits *upstream* of that metric, so it slots into the existing HPA/KEDA flow without modifying it. How a workload is *identified as WVA-managed* is the VA-deprecation proposal's, not here.
+
+## 3. The generic schema shape (review question 2)
+
+The CRD is a **thin typed envelope around pluggable, schemaless plugin lists** — so the schema itself is small and fixed, and new analyzers/limiters never change it. Two validation tiers, deliberately:
+
+```yaml
+spec:
+  # (a) typed, cross-cutting fields — OpenAPI/CEL-validated, `kubectl explain`-able
+  inferencePoolRef: { name: granite-premium-pool }   # per-pool only; mutually exclusive with limiters (§4)
```
shuynh2017: why `mutually exclusive with limiters` ?

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+  priority: 2.0
+  scaleToZero: { enabled: false }
+
+  # (b) pluggable lists — {type, name, parameters}; `parameters` is plugin-owned
+  analyzers:                          # one or more; TENANT-tunable, any tier
+    - type: saturation
```
shuynh2017, line 121: Is the analyzer type (V1, V2) global and can't be overriden? If so, why it's here?

... I can see it explains a bit later line 95. The reading is harder when things are referenced before defined.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+      name: sat                       # identity, used for cross-tier merge
+      parameters:                     # x-kubernetes-preserve-unknown-fields
+        scaleUpThreshold: 0.95
+  limiters:                           # zero or more; chain; CLUSTER-DEFAULT ONLY (admin)
+    - type: gpu-inventory
+    - type: quota                     # parameters schema owned by #1162
+```
+
+*Field catalog, not a single valid object: a real `ScalingPolicy` is one tier, and `inferencePoolRef` and `limiters`/`quota` never co-occur — limiters live only on the cluster-default, which has no `inferencePoolRef` (§4).*
```
shuynh2017: Same comment for `limiters` as for `analyzers`.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+- **Tenant-tunable vs admin-only — an isolation boundary, not a style choice.** `analyzers`, `priority`, and `scaleToZero` are tenant-tunable at any tier. **`limiters` (and `quota`) are cluster-default-only**: they are *instance-wide enforcement*, not tenant knobs. A **CEL rule rejects `limiters`/`quota` on any namespace or per-pool `ScalingPolicy`**, and tenants have no RBAC to write the cluster-default object (it lives in the system namespace). This is load-bearing for security: if a tenant could set `limiters` on a policy they own, they could drop the `quota` limiter — or set `limiters: []` — and **escape their own cap**. So `limiters`/`quota` are never authored, merged, or overridable below the cluster-default tier (§4).
```
shuynh2017: what does `Tenant-tunable vs admin-only` mean? How does it work?

... I can see it explains a bit later line 95. The reading is harder when things are referenced before defined.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+
+Two `ScalingPolicy` objects matching the **same** scope (two per-pool policies for one pool, or two namespace-defaults in one namespace) are a configuration error, **rejected at admission** — so resolution is always deterministic, never order- or name-dependent.
+
+**Two tiers are forced by ownership, not preference.** Quota is **cluster-admin-owned and cluster-scoped** (a tenant must not raise their own cap, and a cluster aggregate needs a cluster-wide view — per-namespace instances would each see full node inventory and collectively overcommit); thresholds/`priority`/`scaleToZero` are **tenant-owned and namespace-scoped**. Different owners, different RBAC — they can't share one object without letting tenants edit quota or admins own every tenant's thresholds. So **cluster-default + namespace are structural, not convenience.**
```
shuynh2017: I think this statement is already explained in line 99. Just wondering if there's anything new in this statement.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,144 @@
...
+
+The **third (per-pool) tier costs no extra schema** — it is the *same* CRD with `inferencePoolRef` set, in the tenant's namespace, adding override *granularity*, not a new surface. So "multiple levels" is three placements of one object resolved by a small deterministic merge — not three configuration systems.
+
```
shuynh2017, line 145: I think it may not be easy for operator to look at this resource and immediately remember which fields are cluster, which are namespaced, which  are per workload (inferencepool), may be we can organize? May be with a `cluster` or `global` section, and  `namespace` or `tenant` section, and `workload` section? As such, if all wanted is a policy for `workload` then that should just need to contain the `workload` section which should contain only the parameters that can be overridden.

shuynh2017:
I've reviewed and commented up to line 109. In general, I don't see the whole story for this proposal (impact on current implementation, upgrade current clusters, how to use, is it optional/not optional? etc ...) and this may be due to the longer proposal before this one, and this proposal has been out for a while and it lags the current state of WVA. Looking forward to another revision.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,236 @@
+# Proposal: WVA scaling configuration — schema and resolution (ConfigMap first, CRD later)
+
+## Problem — what this addresses
+
+Today WVA is configured by **one cluster-wide ConfigMap** plus **per-workload labels/annotations** on the KEDA `ScaledObject` (`llm-d.ai/managed`, `llm-d.ai/model-id`, `inference.optimization/acceleratorName`). That is enough for a single global configuration, but it cannot express:
+
+- **per-namespace or per-pool policy** — every workload shares one global `priority`, `scaleToZero`, and analyzer configuration;
+- **an ownership boundary** — cluster-admin-owned enforcement (GPU limiters / quota) and tenant-owned tuning (thresholds, priority) live in the same undifferentiated blob, with no way to give them different RBAC;
+- **validation** — a typo in the ConfigMap is silently ignored;
```
shuynh2017: ok, but validation is a general enhancement, as part of this proposal but not core to the proposal.

docs/proposals/design-scalingpolicy-crd.md
shuynh2017: ok, I get the issues.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,236 @@
...
+- **a resolved-configuration surface** — an operator cannot see *what policy WVA actually applied* to a given pool.
+
+This proposal fixes the configuration **schema** and **resolution** (the fields, the scope tiers, the pluggable-plugin envelope, and the merge order). It is **transport-neutral** and lands in two phases:
+
+- **Phase 1 — extend the existing ConfigMap** with this schema. No CRD, no migration, no new install step. Validation via `ValidatingAdmissionPolicy` (CEL) + the controller; the ownership boundary via RBAC. This is the minimal core.
```
shuynh2017: `schema`: need to define schema.

shuynh2017: not clear which of the issues above phase1 aims to address. The aim is no CRD, no migration, no new install step. Ok, let's see if the implementation reflects.

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,236 @@
...
+# cluster-default tier == the 'default' entry above
+priority: 1.0
+scaleToZero: { enabled: false }
+analyzers: [ { type: saturation, name: saturation, score: 1.0, parameters: { scaleUpThreshold: 0.85 } } ]
+limiters:  [ { type: gpu-inventory } ]              # admin-only, cluster tier (§2)
+
+# per-pool tier for granite-premium (== the override above), only the delta:
+#   inferencePoolRef: { name: granite-premium-pool }
+#   priority: 2.0
+```
```
shuynh2017, line 56: what is this yaml file?

docs/proposals/design-scalingpolicy-crd.md
```diff
@@ -0,0 +1,236 @@
...
+
+**Minimal usage.** A pool with no config of its own scales under the cluster default (unchanged). To raise one pool's priority, a tenant adds *just* the per-pool object carrying the delta (`inferencePoolRef` + `priority: 2.0`); everything else (`scaleToZero`, `behavior`, `analyzers`) inherits from the tiers above (§2).
+
```
shuynh2017, line 59: I think this is end of phase 1. I'd think it's cleaner to split the 2 phases into 2 .md files. For phase 1, I think I understand the intention and the goals, but can't visualize the changes. May be we can add examples of full before and after configmaps, in different namespaces, pool if applicable.

shuynh2017:
I've read to end of phase 1 (I think) and added comments. We need examples to clarify, to verify against claims such as no upgrade, additively-only, etc ...

PR #1228, opened 2026-06-02
engines: universal threshold post-step + analyzer aggregation helpers

deanlorenz:
## Summary

- Engine post-step applies a pure threshold formula to `RequiredCapacity` / `SpareCapacity` at model level and every `RoleCapacity` entry, replacing the saturation-only override-resolution loop. `TotalAnticipatedSupply` is read as-is — zero is a literal value, not a sentinel; no fallback walk over `VariantCapacities`.
- New `internal/engines/aggregation/` package with pure helpers (`SumTotalSupply`, `SumTotalAnticipatedSupply`, `SumTotalDemand`, `AggregateByRole`) over `[]VariantCapacity`. Analyzers use these to populate per-scope `Total*` fields and enforce the linearity invariant the optimizer's per-variant scaling math depends on.
- Saturation V2 simplified: drops in-analyzer RC/SC computation (engine post-step is now the sole writer); calls helpers for per-scope aggregation. Phase 4 redundancy removed.
- Per-analyzer threshold overrides resolved per analyzer (`AnalyzerScoreConfig.ScaleUpThreshold` / `ScaleDownBoundary` over the model-level globals) and applied uniformly at every scope. No per-role overrides.
- Dev-guide rewritten to document the architecture: per-variant `VariantCapacity` is canonical; analyzer publishes per-variant primitives + per-scope `Total*`; engine post-step computes RC/SC; `SchedulerQueue` is shared input across analyzers, but demand extraction is per-analyzer.

## Context

This PR is **stacked on #1225** (\`engines/saturation: multi-analyzer registry\`). The 4 threshold commits sit on top of #1225's 3 registration commits. When #1225 merges, this branch will be rebased onto main and the diff will reduce to the 4 threshold commits.

Implements Item 2 of the design discussion in \`planning/PR1113-review.md\`. Items 1 (combine deletion / per-analyzer slice → optimizers) and 3 (analyzer registration) are tracked separately.

## Test plan

- [x] \`make build\` — clean
- [x] \`golangci-lint run\` — 0 issues
- [x] \`make test\` — all packages pass
- [x] \`go test -race ./internal/engines/saturation/...\` — clean
- [x] \`internal/engines/aggregation/aggregation_test.go\` — full coverage of helper edge cases (empty input, single/multi variant, mixed roles, empty-role canonicalization, zero replicas, zero PRC)
[...]

PR #1246, opened 2026-06-08
engines/pipeline: per-analyzer slice optimizer — delete engine-side combine

deanlorenz:
## Summary

Completes the multi-analyzer engine rework: the engine no longer combines
analyzer results into a single scalar pair before optimizing. Instead it passes
a **per-analyzer slice** (`[]NamedAnalyzerResult`) straight to the optimizers,
which make scaling decisions over it via shared helpers. This is the third and
final piece of the split out of the original combine-based PR #1113, after
#1225 (race-safe analyzer registry) and #1228 (universal threshold post-step),
both merged.

## Why

The previous engine combined N analyzers into one `RequiredCapacity` /
`SpareCapacity` via dimensionless normalization, then handed that scalar to the
optimizer. But the optimizer fundamentally needs per-analyzer, per-variant data
(each analyzer's `PerReplicaCapacity` per variant) for its bottleneck math —
combining threw that away. This PR deletes the combine and gives the optimizer
the full per-analyzer slice.

## What changed

- **Slice contract** (`pipeline/optimizer_interfaces.go`): `ModelScalingRequest`
  carries `AnalyzerResults []NamedAnalyzerResult`. Each entry pairs an analyzer's
  result with mutable working counters the allocation loop decrements; the
  underlying `AnalyzerResult` is never mutated.
- **Shared helpers** (`pipeline/analyzer_helpers.go`): any-up / all-down gates,
  per-variant bottleneck sizing, and allocation/deallocation — implemented as
  free functions over the slice. The old combine's any-up/all-down semantics are
  preserved as slice predicates (scale up if *any* analyzer needs it; scale down
  only if *all* analyzers permit it).
- **Both optimizers migrated** to the slice: `CostAwareOptimizer` (cost-minimizing,
  unlimited) and `GreedyByScoreOptimizer` (fair-share, GPU-limited).
- **Unified (model, role) path**: a model is decomposed into per-role units —
  non-disaggregated is the single `"both"` role; disaggregated is `{prefill,
  decode}`. One scale-up path (role-generic joint allocation with a
  min-utilization joint commit so P and D advance together) and one
  role-iterated scale-down path serve both cases; the non-disaggregated case is
  the arity-1 reduction. No `if disaggregated` branching in dispatch.
- **`AnalyzerResult.Score` removed** from the interface; the fair-share priority
  metric is computed on demand from per-analyzer `Score` config weights.
- Saturation runs first and remains the keeper of per-variant metadata (Cost,
  AcceleratorName, Role) the optimizer reads — a transitional arrangement noted
  for future extraction.

## Behavior notes for reviewers

- For GPU-limited (`GreedyByScore`) disaggregated models, exhausting one role's
  accelerator budget stalls the joint scale-up for that model: the
  min-utilization joint commit advances both roles together, so a one-role cap
  pins the model at its current size.
- Fair-share `Score` affects model *ordering*, not per-model replica sizing.

## Testing

`make test` and `go test -race ./internal/engines/...` pass. New specs cover the
slice helpers, paired (P/D) scale-up and role-iterated scale-down, decode-only
scale-up, the min-utilization joint commit, and engine population of per-analyzer
config (scores, thresholds).

## Follow-ups (out of scope, tracked separately)

- Fold the queueing-model engine onto the same multi-analyzer slice path.
- Cost picker is efficiency-greedy; under integer rounding it can over-pick a
  high-capacity variant when required capacity is small — refine to rank by
  actual rounded allocation cost.
- Expand the multi-analyzer developer-guide (currently a stub) once the pipeline
  is settled.

### DIFF
--- a/docs/developer-guide/multi-analyzer-pipeline.md
+++ b/docs/developer-guide/multi-analyzer-pipeline.md
@@ -1,0 +1,50 @@
+# Multi-Analyzer Pipeline (developer reference)
+
+> **Status: STUB.** This doc summarises the multi-analyzer pipeline at a
+> high level and points to the design doc for detailed architecture,
+> alternatives considered, and future direction. Full developer-guide
+> content covering all three multi-analyzer PRs (registration, threshold,
+> optimizer) will be added post-review.
+
+The Workload Variant Autoscaler's scaling engine runs multiple **analyzers**
+in series each cycle. Each analyzer consumes the same per-replica metrics
+and produces an `*interfaces.AnalyzerResult` carrying per-variant capacity,
+model-level totals, and (for P/D disaggregated models) per-role capacity.
+The engine post-step calibrates `RequiredCapacity` / `SpareCapacity` at
+every scope using a uniform threshold formula. The optimizer reads a
+per-analyzer slice (`[]NamedAnalyzerResult`) and decides scaling actions
+over it via shared free functions in `internal/engines/pipeline/`.
+
+## Components
+
+- **Registration** — `internal/engines/saturation/engine.go`:
+  `RegisterAnalyzer(name, analyzer) error`. `cmd/main.go` registers external
+  analyzers (e.g., throughput) before `StartOptimizeLoop`. Saturation V2 is
+  pre-registered. Registry is snapshotted at Start; late registration
+  returns an error.
+- **Engine post-step** — `internal/engines/saturation/engine_v2.go`:
+  `applyUniversalThreshold(*AnalyzerResult, scaleUp, scaleDown)` applies a
+  pure formula `RC = max(0, TD/scaleUp − Anticipated)` /
+  `SC = max(0, TS − TD/scaleDown)` at model scope and every role.
+- **Aggregation helpers** — `internal/engines/aggregation/`:
+  `SumTotalSupply`, `SumTotalAnticipatedSupply`, `SumTotalDemand`,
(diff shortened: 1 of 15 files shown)


ev-shindin (approved):
This is the cleanest of the #1113 split. Verified while reading: mutation safety holds (working counters on `NamedAnalyzerResult` are separate from the shared, read-only `*AnalyzerResult`); the unified (model, role) path has no `if disaggregated` branching and is deterministic (sorted roles); any-up/all-down semantics are preserved as slice predicates; and `Score` is cleanly removed from `AnalyzerResult`. Approving — the items below are the only things I'd want addressed or acknowledged.

---

## Problems

### 1. Local variable shadows the builtin `max`
`analyzer_helpers.go` — `roleBottleneckReplicas` (~L132) and `roleAggRemaining` (~L151) declare a local `max`, shadowing the Go builtin `max` that `allocateForModelPaired` uses a few lines below. Legal but confusing, and `predeclared`-style linters flag it.

### 2. `prcForVariant` is an O(V) scan in the hot loop (non-blocking)
`analyzer_helpers.go:~51` — `prcForVariant` linear-scans `VariantCapacities`, and it's called per-analyzer × per-role × per-iteration inside the bottleneck/util sizing → O(N·V) per commit. Fine at realistic N/V; only a concern if either grows.

---
Can you squash 17 commits before merge?

Issue #1247, opened 2026-06-08
Data analysis of autoscaling experiments


asm582:
We have rich autoscaling data for all scenarios mentioned here: https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/main/docs/benchmark.md. We need to analyze the data to determine which component causes a run to be slow or fast, and develop a concrete hypothesis about what should be improved to inform the next set of autoscaling experiments.

asm582:
Data analysis has many knobs to evaluate; this work will happen over a few releases.

PR #1249, opened 2026-06-08
Data analysis of autoscaling experiments

kahilam:
## Summary

Root cause analysis of autoscaling benchmark data from [docs/benchmark.md](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/main/docs/benchmark.md), addressing #1247.

## Phase 1: Single-Replica Capacity Baseline 

I ran a 12-stage load sweep (0.5 → 18 RPS, 300s per stage) on **Qwen3-32B with 1 fixed replica** to characterize the model's capacity envelope before evaluating autoscaling. This gives us a scientific baseline for what a single replica can handle.

| Point | RPS | Observation |
|-------|-----|-------------|
| **Knee** | 2.0 | P99 TTFT jumps from 136ms to 18.8s (138x). Queueing begins. |
| **Peak** | 3.0 | Max throughput: 1901 output tok/s. 0% failures. Server fully loaded. |
| **Saturation** | 4.0 | Failures begin (5.9%). Throughput degrades. |

**Key takeaway:** A single Qwen3-32B replica can sustainably serve **~2-3 RPS**. Beyond 3 RPS, failures appear. This means **20 RPS requires ~7-10 replicas**.

**Phase 1 Charts: Single-Replica Load Sweep (Qwen3-32B)**
Chart 1: Throughput & Failure Rate vs. Requests Sent

Shows how achieved throughput (successful responses) diverges from offered load at the knee (2 RPS), peaks at 3 RPS, and saturates at 4 RPS where failures begin.

image

Throughput & Failure Rate
**Chart 2: Latency (TTFT & TPOT) vs. Requests Sent**

The P99 TTFT (blue, log scale) jumps 138× at the knee — from 136ms to 18.8s — while TPOT (orange) only doubles across the entire range. This proves the bottleneck is queueing, not per-token compute.

Latency (TTFT & TPOT)

image

These charts provide the scientific baseline for Phase 2: when we run autoscaling experiments at 20 RPS, we know each replica can only handle ~2-3 RPS, so ~7-10 replicas are needed.

## Phase 2: WVA Default(v1) Analysis Across Scenarios

Using the data from `benchmark.md`, all Qwen3-32B WVA experiments run at **20 RPS** — which requires ~7-10 replicas based on Phase 1.

### WVA Default(v1) — Qwen3-32B, 20 RPS, 600s

| Metric | Prefill Heavy | Decode Heavy | Symmetrical |
|--------|--------------|-------------|-------------|
| **Workload** | 4000/1000 tok | 1000/4000 tok | 1000/1000 tok |
| P99 TTFT (s) | 98.4 | 78.1 | 100.2 |
| P99 ITL (ms/tok) | 54.8 | 47.1 | 67.3 |
| Avg replicas | 1.73 | 1.84 | 1.70 |
| Max replicas | 3 | 3 | 3 |
| KV cache % | 66.3 | 79.2 | 70.2 |
| Queue depth | 236.5 | 108.8 | 166.8 |
| Errors | 4,184 | 3,563 | 3,729 |
| Pod startup (s) | 110 | 109 | 103 |
| Cost (GPU/hr) | 1.73 | 1.89 | 1.70 |

### WVA Default(v1) — Qwen3-32B, 20 RPS, 1800s

| Metric | Prefill Heavy | Decode Heavy | Symmetrical |
|--------|--------------|-------------|-------------|
| **Workload** | 4000/1000 tok | 1000/4000 tok | 1000/1000 tok |
| P99 TTFT (s) | 96.1 | 70.9 | 88.2 |
| P99 ITL (ms/tok) | 54.1 | 46.1 | 66.4 |
| Avg replicas | 2.55 | 2.40 | 3.05 |
| Max replicas | 4 | 4 | 5 |
| KV cache % | 59.9 | 72.2 | 59.0 |
| Queue depth | 182.0 | 88.4 | 114.2 |
| Errors | 12,822 | 10,762 | 10,272 |
| Pod startup (s) | 107 | 109 | 103 |
| Cost (GPU/hr) | 2.55 | 2.40 | 3.05 |

WVA 600s vs 1800s Comparison

### Key Observations

**600s runs:**
- WVA scales to max 3 replicas across all scenarios, with avg ~1.7-1.8. This is far short of the ~7-10 replicas needed for 20 RPS.
- Decode Heavy has the highest KV cache utilization (79.2%) — closest to the 80% saturation threshold, which may explain why it has the lowest TTFT (78.1s) — WVA reacts slightly faster when KV cache pressure is visible.
- Prefill Heavy has the deepest queue (236.5) because large prompts (4000 tokens) occupy each replica longer, causing more requests to pile up.
- All scenarios show 3,500–4,200 errors, meaning ~35% of requests are dropped regardless of workload type.

**1800s runs — what changes with more time:**
- WVA scales further (avg 2.55-3.05, max 4-5), showing the autoscaler *does* continue reacting given time.
- Queue depth drops 23-30% (e.g., Prefill: 236→182, Decode: 109→88) as extra replicas absorb some load.
- Despite more replicas, errors **triple** (e.g., Prefill: 4,184→12,822) because the longer duration means more total requests hit the overloaded system.
- TTFT improves only marginally (98.4→96.1s for Prefill) — more replicas help slightly, but the system still has far fewer replicas than the ~7-10 needed for 20 RPS.

**Summary**

At 20 RPS on Qwen3-32B, the system requires ~7-10 replicas but WVA Default(v1) never exceeds 3.05 avg / 5 max even with 1800s of runtime. The capacity gap is the dominant bottleneck: every scenario operates at 3-4x overload. WVA's apparent low TTFT (~78-100s) is an artifact of error shedding — ~35% of requests are dropped, and TTFT is only measured on surviving requests. Given more time (1800s), WVA scales further but errors triple proportionally, indicating the autoscaler reacts but cannot close the gap fast enough.


kahilam:
## Phase 3: Fair Comparison — First 600s of 1800s Runs vs Standalone 600s Runs

Addressing [#1258](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1258): *"Is there any correlation between the two runs for a fair comparison?"*

### Methodology

To determine whether 600s and 1800s runs are directly comparable, I extracted the **first 600s window** of raw Prometheus metrics from the 1800s runs and compared them against the standalone 600s runs. Both use the same parsing logic on the same metric types (`vllm:kv_cache_usage_perc`, `vllm:num_requests_running`, `vllm:num_requests_waiting`) from raw per-pod metrics files collected every ~15s.

**Run identification:** Runs were matched to `benchmark.md` by their error counts (e.g., Prefill Heavy 32B 1800s runs have 12,495 / 12,977 / 12,993 errors, averaging to the reported 12,822).

### Time-Series Overlay

The plot below overlays the 600s runs (blue solid) against the first-600s window of 1800s runs (red dashed) for KV Cache %, Requests Running, and Requests Waiting across all three scenarios. Data is averaged across 3 runs per configuration and binned into 30s intervals. Each subplot shows the mean difference between the two curves.

600s vs First 600s of 1800s

### Key Findings

1. **All three scenarios match** in the first 600s — Prefill Heavy, Decode Heavy, and Symmetrical all track the same patterns across KV cache, requests running, and requests waiting. The system behavior is reproducible regardless of total run duration.

2. **The improvement in full-1800s averages comes primarily from the later phase (600–1800s)** when WVA has scaled to more replicas. The "better" 1800s numbers are not because the system starts differently — it's because extra time dilutes the initial metrics with the improved steady-state.

3. **For fair comparison:** The first 600s of a 1800s run reproduces the 600s run. Comparing full-run averages across different durations is misleading when asking whether WVA behaves the same, because 1800s averages mix transient (cold-start) and steady-state phases.

---
*Made with [Cursor](https://cursor.com)*

kahilam:
## Phase 4: Impact on TTFT and ITL — Short (600s) vs Long (1800s)

Addressing [#1258](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1258) Q1: *"Impact on ITL and TTFT when we compare experiments that are run for short and long durations"*

### Methodology

Extracted per-request `time_to_first_token_ms` and `inter_token_latency_ms` from `results.json` across all successful requests. For the 1800s runs, I also windowed to only requests started within the first 600s for a fair comparison. All data is from Qwen3-32B, 20 RPS, WVA Default(v1).

### Part 1: Reproducibility — 600s Run vs First 600s of 1800s Run

TTFT and ITL: 600s vs first 600s of 1800s

The blue (600s) and orange (1800s first-600s) bars are nearly identical across all scenarios, confirming that the first 600s of a 1800s run reproduces the 600s run for both TTFT and ITL.

### Part 2: Duration Effect — 600s Run vs Full 1800s Run

TTFT and ITL: 600s vs full 1800s

| Scenario | TTFT Mean | | TTFT P99 | | ITL Mean | | ITL P99 | |
|----------|-----------|--------|----------|--------|----------|--------|---------|--------|
| | 600s | 1800s | 600s | 1800s | 600s | 1800s | 600s | 1800s |
| **Prefill Heavy** | 49.3s | 52.4s (+6%) | 98.6s | 96.6s (-2%) | 35.0ms | 34.4ms (-2%) | 55.0ms | 54.2ms (-2%) |
| **Decode Heavy** | 43.1s | 45.1s (+5%) | 75.5s | 71.0s (-6%) | 34.5ms | 34.2ms (-1%) | 47.1ms | 45.9ms (-2%) |
| **Symmetrical** | 45.2s | 43.4s (-4%) | 100.0s | 90.1s (-10%) | 42.2ms | 41.7ms (-1%) | 67.3ms | 66.4ms (-1%) |

### Key Findings

1. **ITL is unaffected by duration.** Across all scenarios, ITL stays within 1–2% whether the run is 600s or 1800s. ITL measures per-token GPU compute, which depends on concurrent requests per pod — not on total run duration.

2. **TTFT Mean is stable** (+/- 6%). Mean TTFT gets slightly worse for Prefill/Decode Heavy in 1800s runs because more requests queue up over time, but Symmetrical actually improves because WVA scales to more replicas (avg 3.05).

3. **TTFT P99 improves in longer runs** (-2% to -10%). WVA scales to more replicas over 1800s, which reduces the worst-case queue wait times. Symmetrical benefits most (P99 drops from 100s to 90s) because it reaches the highest replica count.

4. **Duration does not degrade per-request latency.** Longer runs produce more total requests and more total errors, but per-request TTFT and ITL stay in the same range.

---
*Made with [Cursor](https://cursor.com)*

kahilam:
## Phase 5: KV Cache Impact and EPP as the Driving Factor

Addressing [#1258](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1258) Q3: *"How is KVCache usage impacted when experiments are run for short and long durations, and can we pinpoint a factor (plugin) that is causing the difference in usage across the run?"*

### Prefill Heavy — Qwen3-32B, 20 RPS, WVA Default(v1)

#### Side-by-Side: 600s vs 1800s

Prefill Heavy 600s vs 1800s

- **600s run**: WVA scales to max 3 replicas but the run ends before steady state. KV avg 65%, EPP Queue avg 237.
- **1800s run**: Same initial pattern, but WVA has time to scale to 4 replicas. KV drops to avg 54%, Queue to avg 145 in the later phase.

#### 1800s Correlation: Replicas → KV Cache → Queue

Prefill Heavy Correlation

Time-aligned data shows the causal chain:

| Time | Event | Replicas | KV Cache | EPP Queue |
|------|-------|----------|----------|-----------|
| 0-210s | Cold start, 1 replica | 1 | 92-98% | 490-501 |
| 240s | WVA scales up | 1→2 | 98%→52% | 501→414 |
| 270-510s | Stable at 2 replicas | 2 | 43-71% | 185-245 |
| 540-570s | WVA scales up again | 2→3 | 47%→40% | 143-160 |
| 660-870s | Scale-down to 2 | 3→2 | 72-84% ↑ | 220-236 ↑ |
| 990-1800s | Scales back to 3-4 | 3-4 | 35-65% | 95-156 |

---

### Decode Heavy — Qwen3-32B, 20 RPS, WVA Default(v1)

#### Side-by-Side: 600s vs 1800s

Decode Heavy 600s vs 1800s

---

### Symmetrical — Qwen3-32B, 20 RPS, WVA Default(v1)

#### Side-by-Side: 600s vs 1800s

Symmetrical 600s vs 1800s

---

### Can We Pinpoint EPP?

**Yes.** EPP (Endpoint Picker Plugin) is the factor driving KV cache differences between short and long runs. The same pattern holds across all three scenarios:

1. **EPP controls how many requests reach each pod.** Its flow control holds requests in its own queue rather than forwarding them to saturated pods.

2. **More replicas → EPP distributes more → lower per-pod KV.** When WVA scales from 2 to 3 replicas, EPP has more targets to route to. Each pod gets fewer requests → KV cache drops.

3. **The Prefill Heavy scale-down at t=660s proves causality.** When replicas drop from 3 back to 2, KV cache immediately rises from 40% to 84% and queue jumps from 160 to 236 — because EPP now has fewer pods to distribute to.

4. **EPP masks saturation from WVA.** Even with queue depth at 237, pods show only 65% KV cache because EPP is absorbing the overload in its own queue. WVA sees "healthy" pods and doesn't scale aggressively enough.

---
*Made with [Cursor](https://cursor.com)*

asm582:
@kahilam, please share new updates on the data analysis part, thanks.

kahilam:
I did a deep dive into the prefill heavy benchmark results for EPP+KEDA, comparing runs with flow control enabled versus disabled. Here's the analysis.

**Prefill Heavy: Flow Control Impact Analysis
(Averaged across 3 runs each)**



image


### 1- KV cache utilization

image


avg KV = mean of all pods' vllm:kv_cache_usage_perc values
max KV = the highest vllm:kv_cache_usage_perc among all pods

FC max KV (red dashed) = the KV cache utilization of the single most loaded pod at each timestamp, with flow control enabled.

No FC max KV (blue dashed) = the KV cache utilization of the single most loaded pod at each timestamp, without flow control.

The reason I plotted both avg and max is to expose the problem:

FC: The avg KV is only ~11% (red solid), but the max is ~95% (red dashed). That huge gap means one pod is nearly full while the other 9 are empty — the average masks reality.
No FC: The avg KV is ~68% (blue solid) and the max is ~95% (blue dashed). The gap is much smaller because all 10 pods are sharing the load — traffic is distributed evenly.
The gap between the solid and dashed lines of the same color is the key insight: a large gap = uneven distribution, a small gap = healthy load balancing.

### 2- Active Pods

image

With flow control enabled, KEDA scales to 10 pods but only 1 pod actually receives traffic. The other 9 sit completely idle (0% KV cache, 0 requests running).

### 3- Request running & waiting

image


image

Requests Running: No FC reaches ~110 requests running across all pods — traffic is distributed. FC stays flat at ~10 — only one pod is processing.

Requests Waiting: Both peak at ~500, but No FC drops quickly as new pods absorb load. FC stays high because requests are stuck on the single active pod.


**4- Queue Depth** 

image

EPP Queue Depth: FC shows a sharp spike to ~500 then slowly declines but stays elevated (~50). No FC also spikes but drops faster and stabilizes lower — because all 10 pods are draining the queue, not just one.

asm582:
Thanks @kahilam. Can you point me to the Google Drive link for the above experiments?

kahilam:
> Thanks @kahilam. Can you point me to the Google Drive link for the above experiments?

https://drive.google.com/drive/u/0/folders/1ek48GnygtDU5xXPCVVsWocBHv4ZmWcwg

kahilam:
### Decode Heavy: Flow Control Impact Analysis
### (Averaged across 3 runs each)

image


Screenshot 2026-07-24 at 2 33 05 PM

Screenshot 2026-07-24 at 2 33 15 PM

Screenshot 2026-07-24 at 2 33 24 PM

Screenshot 2026-07-24 at 2 33 32 PM


Screenshot 2026-07-24 at 2 33 42 PM

FC actually outperforms No FC here — more completions, more requests running, higher KV utilization. This is the opposite of prefill. Why?

With decode heavy (1000 input / 4000 output), each request spends most of its time in autoregressive decoding. FC is concentrating requests on fewer pods initially (8 active at mid-run), which means more concurrent decodes per GPU → better batching efficiency → higher throughput. vLLM's continuous batching benefits from running more concurrent requests on the same GPU for decode-heavy workloads.

No FC spreads requests thinly across all 10 pods early (only 44 running total vs 113 for FC), meaning each pod is underutilized. The KV stays low (43%) because there aren't enough concurrent requests per pod to fill GPU memory.

kahilam:
### Symmetrical Heavy: Flow Control Impact Analysis
### (Averaged across 3 runs each)

image



Screenshot 2026-07-24 at 2 39 41 PM
Screenshot 2026-07-24 at 2 39 32 PM
Screenshot 2026-07-24 at 2 39 21 PM
Screenshot 2026-07-24 at 2 39 13 PM
Screenshot 2026-07-24 at 2 39 03 PM

With balanced workloads (1000/1000 tokens), both FC and No FC manage to activate all 10 pods — unlike prefill where FC starved 9 pods.

asm582:
@kahilam Below are the initial comments:
- The same plots are not plotted across the scenarios; please fix.
- How can there be queue depth for the no-flow-controller case?
- What is the definition of total vs active pods? Why are they important?
- Why is the KVCache plot at the inference pool level generated by llm-d- benchmark not reused?
- Total running and waiting requests are plotted from which source? why?

kahilam:
1- I am not sure what do you mean by same plots to be fixed.
2- Queue depth exists even without flow control because the metric inference_pool_average_queue_size measures the vLLM-level waiting queue across all pods, not an EPP-side holding queue.

```
Client → EPP → immediately forwards to pod → vLLM accepts request
                                                ↓
                                          vLLM internal scheduler:
                                          - Can it fit in GPU memory (KV cache)?
                                            Yes → running
                                            No  → waiting (queued inside vLLM)
```

3- Total pods = pods that exist and report metrics (they're running on the cluster).
Active pods = pods that are actually processing requests (KV cache > 5%).
Why it matters: In the prefill FC case, 10 pods exist (total) but only 1 is actually doing work (active). The other 9 are wasting GPU resources sitting idle. 

4- Here are the plots KVCache plot at the inference pool level generated by llm-d- benchmark:

image

image


5- Total running and waiting requests are plotted from vllm:num_requests_running and vllm:num_requests_waiting — raw per-pod Prometheus metrics scraped every ~16s, stored in each run's results/guidellm-*/metrics/raw/ logs

image

image

image

image

PR #1250, opened 2026-06-08
engines/analyzers/throughput: ThroughputAnalyzer — ITL model, scaling signal, and engine wiring

deanlorenz:
## Summary

Implements the ThroughputAnalyzer — a new scaling analyzer that estimates
decode-token supply (μ_dec) from an ITL model calibrated per variant, and
decode-token demand (λ_dec) from EPP arrival rates or vLLM request rates.
Registers the analyzer into the multi-analyzer engine alongside saturation V2.

This is a sibling of the multi-analyzer optimizer PR (#1246). On the current
main the analyzer runs and produces output; its signal becomes effective once
#1246 merges and the optimizer consumes the per-analyzer slice. The smoke e2e
(wiring health check) is already passing; full TA-driven scale-up/scale-down
e2e exercises are comment-triggered and validated once #1246 lands.

## What changed

- **`internal/engines/analyzers/throughput/`** (new package): per-variant
  workload-shape tracking (`ShapeTracker`, `ObservationWindow`), two-tier ITL
  model calibration (OLS tier-1 when window ready; constrained OLS tier-2 with
  `lastFittedB` persistence across shape resets), supply estimation (μ_dec_sat
  via KV-saturation operating point), three-priority demand estimation (EPP
  primary → vLLM fallback → k*-based local), GPS supply-model verification and
  SC suppression, queue-demand contribution, model-level RC/SC via the universal
  threshold post-step contract, and per-role aggregation for P/D disaggregated
  models.
- **`internal/collector/registration/throughput_analyzer.go`**: 9 Prometheus
  queries registered (generation token rate, KV utilization, KV total,
  avg ITL, avg output tokens, avg input tokens, prefix cache hit rate,
  scheduler dispatch rate, vLLM request rate).
- **`cmd/main.go`**: registers `ThroughputAnalyzer` into the engine via
  `RegisterAnalyzer` (error-return wiring per #1225 API).
- **`test/e2e/throughput_analyzer_test.go`**: wiring health-check smoke spec
  (passing) + scale-up and TA-only-mode full specs (comment-triggered).

## Analyzer contract

`ThroughputAnalyzer` implements `interfaces.Analyzer`. It publishes raw
`Total*` fields (`TotalSupply`, `TotalDemand`, `TotalAnticipatedSupply`,
`RoleCapacities[role].Total*`) and leaves `RequiredCapacity`/`SpareCapacity`
zero — the engine's universal threshold post-step writes those (per #1228).
The linearity invariant is preserved: `TotalSupply = aggregation.SumTotalSupply(VariantCapacities)`.

## Known limitation (tracked follow-up)

The EPP-presence and GPS-mismatch gates that previously suppressed
`SpareCapacity` are not yet restored under the new contract — the engine
post-step always computes SC. Will be restored via a `SuppressSpareCapacity`
opt-out or the deferred `ThresholdApplied` flag in a follow-up PR.

## Testing

`make test` and `go test -race ./internal/engines/analyzers/throughput/...` pass.
`make lint` 0 issues. E2E smoke (Step 2e): VA reconciles to steady state with
both analyzers enabled — PASSED.


internal/collector/registration/throughput_analyzer.go
```diff
@@ -100,26 +100,24 @@ func RegisterThroughputAnalyzerQueries(sourceRegistry *source.SourceRegistry) {
...
 	// Computed as the rate of the _sum histogram counter over 1m.
-	// Uses sum by (pod, llm_d_ai_variant) because request_generation_tokens_sum is an additive
+	// Uses sum by (pod) because request_generation_tokens_sum is an additive
 	// counter — multiple histogram buckets per pod must be summed.
-	// Preserves llm_d_ai_variant for direct pod-to-VA mapping.
 	registry.MustRegister(source.QueryTemplate{
 		Name:        QueryGenerationTokenRate,
 		Type:        source.QueryTypePromQL,
-		Template:    `sum by (pod, llm_d_ai_variant) (rate(vllm:request_generation_tokens_sum{namespace="{{.namespace}}",model_name="{{.modelID}}"}[1m]))`,
+		Template:    `sum by (pod) (rate(vllm:request_generation_tokens_sum{namespace="{{.namespace}}",model_name="{{.modelID}}"}[1m]))`,
```
ev-shindin: This PR's base maps pods via `podVAMapper.FindVAForPod` (owner traversal), so dropping `llm_d_ai_variant` here is harmless **on the old base**. But current `main` attributes pod→VA by reading the `llm_d_ai_variant` label **off the metric series** (`replica_metrics.go:319`, `buildInstanceKey`), and every query file preserves it in `by()` (saturation.go ×12, queueing_model.go ×4, throughput_analyzer.go ×8). After you rebase onto current main, `by (pod)` strips that label → these three throughput metrics arrive with empty `vaName` → the collector skips them → `groupByVariant` sees nothing → **the analyzer silently gets no decode-rate/k\* data and produces no signal.**

Two options: **(A)** restore the label now, or **(B)** hold this PR until #1260 (pod→VA derivation, implementing #1210) merges and rebase onto that main — then the collector derives pod→VA itself, `by (pod)` is correct as-is, and these suggestions are unnecessary. Either works; pick B if #1260 is close to avoid restore-then-re-remove churn. If landing now (A), restore the label in all three queries (and the `Preserves llm_d_ai_variant …` comments above them):

internal/collector/registration/throughput_analyzer.go
```diff
@@ -100,26 +100,24 @@ func RegisterThroughputAnalyzerQueries(sourceRegistry *source.SourceRegistry) {
...
-	// Uses max by (pod, llm_d_ai_variant) to consolidate any duplicate series to a single per-pod value.
+	// Uses max by (pod) to consolidate any duplicate series to a single per-pod value.
 	// Does NOT use max_over_time: the throughput analyzer needs the current
 	// operating point k*, not the worst-case peak used by the saturation analyzer.
-	// Preserves llm_d_ai_variant for direct pod-to-VA mapping.
 	registry.MustRegister(source.QueryTemplate{
 		Name:        QueryKvUsageInstant,
 		Type:        source.QueryTypePromQL,
-		Template:    `max by (pod, llm_d_ai_variant) (vllm:kv_cache_usage_perc{namespace="{{.namespace}}",model_name="{{.modelID}}"})`,
+		Template:    `max by (pod) (vllm:kv_cache_usage_perc{namespace="{{.namespace}}",model_name="{{.modelID}}"})`,
```
ev-shindin: same as line 108

internal/collector/registration/throughput_analyzer.go
```diff
@@ -129,11 +127,10 @@ func RegisterThroughputAnalyzerQueries(sourceRegistry *source.SourceRegistry) {
 	// completed request). Used as a fallback for λ_dec when EPP/scheduler metrics
 	// are unavailable; per variant V, the analyzer falls back to:
 	//   λ_dec_vllm = Σ_{r∈V}(VLLMRequestRate_r × AvgOutputTokens_r)
-	// Preserves llm_d_ai_variant for direct pod-to-VA mapping.
 	registry.MustRegister(source.QueryTemplate{
 		Name:        QueryVLLMRequestRate,
 		Type:        source.QueryTypePromQL,
-		Template:    `sum by (pod, llm_d_ai_variant) (rate(vllm:request_generation_tokens_count{namespace="{{.namespace}}",model_name="{{.modelID}}"}[1m]))`,
+		Template:    `sum by (pod) (rate(vllm:request_generation_tokens_count{namespace="{{.namespace}}",model_name="{{.modelID}}"}[1m]))`,
```
ev-shindin: same as line 108

internal/engines/analyzers/throughput/analyzer.go
```diff
@@ -123,22 +162,198 @@ func (a *ThroughputAnalyzer) Observe(
...
+		totalDemand += queueDemand
+		queueDemandByRole = distributeQueueDemandByRole(queueDemand, variantCapacities)
+	}
+
+	// TA publishes raw Total* fields; RequiredCapacity and SpareCapacity are left
+	// zero — the engine's universal threshold post-step writes them after Analyze returns.
+	// The GPS/EPP gate that previously suppressed SpareCapacity is dropped here
+	// (see docs/developer-guide/throughput-analyzer.md Known Regression).
+	// anyEPP and anyGPSMismatch are retained for potential future use.
+	_ = anyEPP
```
ev-shindin, line 371: The gate that previously suppressed spare capacity on GPS mismatch / no-EPP is gone (computed, then discarded). Documented as a Known Regression, but the consequence is a scale-**down** risk: a wrong ITL model or EPP-absent variant can now publish spare capacity and drive scale-down on uncertain data. Please link a follow-up issue and call out the scale-down risk explicitly in the PR/doc.

internal/engines/analyzers/throughput/analyzer.go
```diff
@@ -123,22 +162,198 @@ func (a *ThroughputAnalyzer) Observe(
...
+		if !ok {
+			continue
+		}
+
+		shape, hasShape := state.shapeTracker.Current()
+		if !hasShape || shape.KVreq <= 0 {
+			continue
+		}
+
+		// TODO: skip variant when state.lastSanityReport is not OK — stale/invalid
```
ev-shindin: As written, an `OK`-failing sanity report doesn't skip the variant, so stale/invalid metrics still flow into `computeDemand` → potential spurious scaling on stale data. Confirm this is acceptable, or gate the demand path on `lastSanityReport` before merge.

internal/engines/analyzers/throughput/analyzer.go
```diff
@@ -123,22 +162,198 @@ func (a *ThroughputAnalyzer) Observe(
...
 	now := time.Now()
+
+	// Build lookup tables from VariantStates before taking any locks.
+	pendingByVariant := make(map[string]int, len(input.VariantStates))
+	for _, vs := range input.VariantStates {
+		pendingByVariant[vs.VariantName] = vs.PendingReplicas
+	}
+
+	// Update variant roles so state.role is current when Observe() runs.
+	a.mu.Lock()
```
ev-shindin, line 214: `Analyze` does lock(role-update)/unlock → `Observe` (locks internally) → lock(main)/defer-unlock. Safe under the single-threaded optimize loop, but a concurrent `VariantState()` snapshot could observe partial state across the gaps. Fine to leave; worth a one-line note that `Analyze` is assumed single-flight.

ev-shindin (commented):
Really solid analyzer — divisions are guarded throughout (`safeDivide`, the `<=0 continue` guards, count/weight fallbacks in `averageShapeMetrics`), the doc comments carry the domain notation, and the test coverage (+1385: OLS tier-1/2, scaling-signal RC/SC, role-aware aggregation, linearity invariant) is excellent. Engine wiring in `cmd/main.go` is clean (`RegisterAnalyzer` error handled).

One blocker tied to a stale base, plus two documented regressions to acknowledge; inline. Note this PR's base (`badc48be`) is **behind** current `main` — please rebase, because the pod→VA attribution mechanism differs between them (see the `throughput_analyzer.go` comment). The rebase also resolves the #1246 dependency: **#1246 is already merged** (main's tip), so rebasing brings the companion optimizer in automatically — no separate merge-order decision.

deanlorenz:
Thanks for the thorough review.

**`by (pod)` attribution (lines 108/120/133):** Going with **Option B** — the queries are already in the right shape for the post-#1260 world (no `llm_d_ai_variant` in `by()`). Holding the rebase until #1260 lands on main; no code changes needed in this PR for that issue.

**SC-gate regression (line 343) and sanity-report gate (line 243):** Both are acknowledged and tracked in #1261 (analyzer interface: extended result type with accept-for-SC/RC/all flags, plus sanity helper mechanism). The correct fix is a richer return type from the analyzer interface rather than a point gate, and it needs to apply consistently across all analyzers. Will link from the PR description and the relevant TODO comments once the issue is open.

**Single-flight note (line 208):** Will add the one-line comment before the rebase.

deanlorenz:
**Update on `by (pod)` attribution — switching to Option A:**

After further analysis, the queries had a deeper issue: the three throughput processing loops in `replica_metrics.go` were keying `podData` by bare pod name (`value.Labels["pod"]`) while all other loops use `buildInstanceKey()` → composite `"pod:port"` key. The entries never merged; `GenerationTokenRate`, `KvUsageInstant`, and `VLLMRequestRate` were always zero regardless of rebase state. This bug predates this PR (introduced in #1051).

The fix requires `instance` in the `by()` clause so `buildInstanceKey` has what it needs, making Option B insufficient on its own. Going with **Option A**: updated all three queries to `by (instance, pod, llm_d_ai_variant)` (matching the saturation pattern) and fixed the processing loops to use `buildInstanceKey`. Also added `hasGenTokenRate`/`hasKvInstant`/`hasVLLMRate` boolean flags to `podMetricData` as the internal half of #1264's minimum fix.

`llm_d_ai_variant` in the `by()` clause is temporary — tracked for removal in #1263 after #1260 lands.

Rebased onto `main@0e977b3b` and pushed (`f5385168`, 28 commits). Bug B items (lines 208, 243, 343) also applied in the same push.

deanlorenz:
Rebased onto `main@04f95779` (#1267 locator, #1270, #1271) and pushed (`b0284253`, 28 commits). Also folded in a small collector improvement: a per-VA `Warning/UnattributedReadyPods` K8s event when a scale target has Ready pods but none are attributed in the current cycle (scrape-lag gated, deduped per cycle).

One merge-order suggestion: if #1266 (multi-analyzer-addendum — disabled-analyzer veto fix + pipeline doc rewrite) can land first, this PR's rebase onto the resulting `main` will be clean (no file overlap). Either order works, but #1266 first avoids a third rebase here.

deanlorenz:
**Round-2 update — all four review threads addressed; rebased onto current `main` (`f11f5120`).**

**`by (pod)` on the three throughput queries (108/120/133):** went with **Option A**. Restoring the label surfaced a pre-existing key-mismatch (since #1051): the three throughput loops keyed `podData` by bare pod name while every other loop uses `buildInstanceKey` → `pod:port`, so entries never merged and `GenerationTokenRate`/`KvUsageInstant`/`VLLMRequestRate` were always zero. Fixed — all three queries now `by (instance, pod, llm_d_ai_variant)`, loops use `buildInstanceKey`, plus `TestCollectReplicaMetrics_ThroughputKeyMerge`. After rebasing onto #1267's locator this composes (owner-walk resolves pod→VA, label is the fast path); the retained `llm_d_ai_variant` is tracked for the eventual groupby cleanup.

**GPS/EPP spare-capacity gate → scale-down risk (356):** acknowledged, linked #1261, scale-down risk called out in the doc. **Also** — the analyzer is now **opt-in / off by default** (below), so this can't affect default deployments until an operator enables throughput.

**Sanity-report gate:** deferred to the same #1261 (needs the engine to accept an AnalyzerStatus opt-out); `computeDemand`'s fallback cascade degrades gracefully meanwhile, and the opt-in default bounds the blast radius.

**Single-flight note (210):** added.

**New in round 2 — throughput is opt-in.** Registering the analyzer unconditionally let it join the optimizer's cross-analyzer scale-down aggregation (post-#1246); with no throughput data it contributed zero spare and vetoed a legitimate saturation scale-down (failed the V2 scale-down smoke test). Fix: `cmd/main.go` registers the ThroughputAnalyzer only when the saturation config enables it; the shipped default lists only `saturation`, so it's off by default and behaves as if absent. Startup-time stopgap — the per-cycle consumption gate (`effectiveEnabled` opt-in from #1266) is the long-term home. Also folded in: ITL fit filters stale replicas, `FitITLModel` guards ITL-at-saturation > 0, dead code removed, dev-guide enablement note.

ev-shindin (dismissed):
Approved conditionally - please provide follow-up PR:

## Should-fix

### 1. (correctness) Booting replicas inflate `TotalSupply` → SC inflation → spurious scale-down during scale-out
`analyzer.go` (the `variantCapacities` append, ~line 331):
```go
ReplicaCount:       len(variantMetrics),                       // includes booting KV=0 replicas
PerReplicaCapacity: perReplicaSupply,                          // mean over KV-capable replicas only
TotalCapacity:      float64(len(variantMetrics)) * perReplicaSupply,
```
`computeVariantSupply` sums only KV-capable replicas (`supply = n_kv × perReplicaSupply`), but `ReplicaCount = len(variantMetrics)` includes still-booting KV=0 replicas. The engine post-step computes SC from `TotalSupply`:
```go
// engine_v2.go applyUniversalThreshold
sc := r.TotalSupply - r.TotalDemand/scaleDown        // TotalSupply = Σ ReplicaCount × PerReplicaCapacity
```
So a variant with 2 ready + 1 booting replica publishes `TotalSupply = 3 × perReplicaSupply` while only `2 ×` is actually serving → **SC inflated → scale-down signal while a scale-out is still in progress.** That is the same direction as the V2 scale-down flake.

The comment claims this is *"consistent with saturation_v2"*, but it isn't: `saturation_v2.aggregateByVariant` uses `ReplicaCount: readyCount` and `TotalCapacity: readyCount × perReplicaCapacity`, putting booting/pending replicas only in `PendingReplicas` — so booting replicas **suppress RC** (via `TotalAnticipatedSupply = (ReplicaCount+Pending)×…`) but **do not inflate SC**. Note also `state.lastTotalSupply = supply` already stores the honest (ready-only) value, so the published `TotalCapacity` is internally inconsistent with the analyzer's own state snapshot.

**Fix (mirror saturation_v2):** set `ReplicaCount = n_kv` (the KV-capable count `computeVariantSupply` already counts — return it), fold the booting delta into pending (`PendingReplicas = pending + (len(variantMetrics) - n_kv)`), and set `TotalCapacity = supply`. RC suppression during scale-out is preserved (anticipated supply still includes them); SC becomes honest.

---

##  Nits / follow-ups

- **Warm-up gap not fully closed (doc accuracy).** The comment at the `demand == 0 → computeLocalDemand` site says this keeps a busy replica from being mis-scaled down, but `computeLocalDemand` has its own OL floor (`shape.AvgOutputTokens <= DefaultMinDecodeOLForLocalDemand`, 20 tokens) and can still return 0 → `demand=0` with `supply>0` for short-generation / prefill-only windows. This is tracked by #1261; worth a one-line caveat on the comment so it doesn't read as fully closed.
- **`FitITLModel` guards B but not A** (`itl_model.go:52`): `A <= 0` doesn't catch `+Inf`/`NaN`. Add the symmetric `if math.IsNaN(A) || math.IsInf(A, 0) { return ITLModel{}, false }` next to the B guard. Downstream `itlSat<=0` masks it (variant skipped, no crash), hence nit.
- **Logging under `a.mu` in `Observe`** (and `resolveITLModel`/`checkVariantGPSMismatch` called under the second lock): single-flight per model so no race, but logging while holding the mutex is fragile. Collect-and-emit-after, or tighten the lock scope.
- **`computeDemand` EPP loop lacks a symmetric `AvgOutputTokens > 0` guard** — harmless (adds `×0`) but the asymmetry vs the vLLM fallback reads as a missing guard; adding it makes the warm-up intent explicit.
- **`SanityIssueStaleMetrics` is unreachable** — the collector hardcodes `FreshnessStatus: "fresh"` on every emitted `ReplicaMetrics`, so the stale branch never fires. Pre-existing; either wire freshness into the field or document the case as a placeholder.

ev-shindin (dismissed):

ev-shindin (approved):
Fixed issues from previous review

PR #1252, opened 2026-06-09
fix(optimizer): prefer cheaper variant when it covers overflow at lower absolute cost

### LINKED ISSUE #1251: cost_aware_optimizer: greedy efficiency-first allocation overpays when cheaper variant covers overflow at lower absolute cost
## Problem

`costAwareScaleUp` sorts variants by cost efficiency (cost/perReplicaCapacity, ascending) and greedily allocates replicas to the most efficient variant first. This is correct in the steady state but misses a class of cases where a cheaper-by-absolute-cost variant can cover the required capacity at lower total spend.

**Example:**
- Required capacity: 100 tokens
- Variant A (most efficient): cost=10, perReplicaCapacity=200 → needs 1 replica, total cost=10
- Variant B (cheaper): cost=6, perReplicaCapacity=100 → needs 1 replica, total cost=6

The greedy approach picks A (better tokens/cost ratio). But B covers the same demand for 40% less. The optimizer should prefer B.

This occurs specifically when `ceil(remaining / vc.PerReplicaCapacity) = N` but the Nth replica's capacity is mostly wasted — a cheaper variant covers the overflow at lower absolute cost.

## Fix

Before committing to N replicas of the most efficient variant, compare the marginal cost of the Nth replica against the cost of covering the overflow (`remaining - (N-1) * vc.PerReplicaCapacity`) using the cheapest-by-absolute-cost variant:

```go
overflow := remaining - float64(replicasNeeded-1)*vc.PerReplicaCapacity
fallbackReplicas := int(math.Ceil(overflow / cheapestVC.PerReplicaCapacity))
if float64(fallbackReplicas)*cheapestVC.Cost < vc.Cost {
    replicasNeeded--
}
```

The N-1 terms cancel algebraically, so this reduces to: *is 1 replica of the efficient variant more expensive than k replicas of the cheapest variant to cover the overflow?* On equality the efficient variant is kept (strict `<`), which is correct since it provides more capacity per cost unit.

## Validation

Empirically confirmed on pokprod001 with a two-variant Llama-3.1-8B setup (primary TP=2 cost=10, v2 TP=1 cost=5):
- WVA consistently adds v2 replicas before touching primary when v2 covers the overflow more cheaply
- Primary only scales beyond minReplicas when v2 hits its maxReplicas cap (overflow correctly falls back to primary)
- 25-min run at rate=15: primary held at peak=2, v2 absorbed all remaining demand up to max=10, 98.6% SLO

biranofer:
## Summary

- `costAwareScaleUp` greedily picked the most cost-efficient variant (lowest cost/perReplicaCapacity) but overpaid when a cheaper-by-absolute-cost variant could cover the required overflow at lower total spend
- Add a per-iteration fallback check: if covering the overflow with the cheapest variant costs strictly less than one additional replica of the current variant, round down and let the loop assign the remainder to the cheaper variant
- Guard added: fallback only applies when the cheapest variant has headroom below its `maxReplicas` cap

## Details

The N-1 terms cancel algebraically, so the check reduces to:
```
fallbackReplicas * cheapest.Cost < vc.Cost
```
On equality the more efficient variant is kept (strict `<`) — at equal cost it provides more capacity per dollar.

Fixes #1251

## Test plan

- [ ] Existing unit tests pass (`make test` — all 113 specs green including `TestScaleToZero/should respect maxReplicas during scale-up`)
- [ ] Empirically validated on pokprod001: two-variant Llama-3.1-8B (primary TP=2 cost=10, v2 TP=1 cost=5), rate=15 req/s, 25-min run — primary held at peak=2 replicas, v2 absorbed all remaining demand up to max=10, 98.6% SLO

### DIFF
--- a/internal/engines/pipeline/cost_aware_optimizer.go
+++ b/internal/engines/pipeline/cost_aware_optimizer.go
@@ -88,6 +88,7 @@ func costAwareScaleUp(
 
 	sorted := sortByCostEfficiencyAsc(result.VariantCapacities)
 	remaining := result.RequiredCapacity
+	cheapestVC := cheapestVariantCapacity(result.VariantCapacities)
 
 	for _, vc := range sorted {
 		if remaining <= 0 {
@@ -109,6 +110,26 @@ func costAwareScaleUp(
 			if replicasNeeded > maxAdd {
 				replicasNeeded = maxAdd
 			}
+		}
+
+		// Round down if covering the overflow with cheapest replicas costs less than
+		// one additional replica of this variant. The remaining capacity will be
+		// assigned to the cheapest variant when the loop reaches it.
+		// Only applies when the cheapest variant has enough headroom to absorb the overflow.
+		if cheapestVC != nil && cheapestVC.VariantName != vc.VariantName && cheapestVC.PerReplicaCapacity > 0 {
+			overflow := remaining - float64(replicasNeeded-1)*vc.PerReplicaCapacity
+			fallbackReplicas := int(math.Ceil(overflow / cheapestVC.PerReplicaCapacity))
+			cheapestAvailable := math.MaxInt32
+			if cs := stateMap[cheapestVC.VariantName]; cs.MaxReplicas != nil && *cs.MaxReplicas > 0 {
+				cheapestAvailable = *cs.MaxReplicas - targets[cheapestVC.VariantName]
+			}
+			if fallbackReplicas <= cheapestAvailable && float64(fallbackReplicas)*cheapestVC.Cost < vc.Cost {
+				replicasNeeded--
+			}
+		}
+
+		if replicasNeeded <= 0 {
+			continue
 		}
 
 		targets[vc.VariantName] += replicasNeeded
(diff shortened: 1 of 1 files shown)


internal/engines/pipeline/cost_aware_optimizer.go
```diff
@@ -111,6 +112,26 @@ func costAwareScaleUp(
...
+		if cheapestVC != nil && cheapestVC.VariantName != vc.VariantName && cheapestVC.PerReplicaCapacity > 0 {
+			overflow := remaining - float64(replicasNeeded-1)*vc.PerReplicaCapacity
+			fallbackReplicas := int(math.Ceil(overflow / cheapestVC.PerReplicaCapacity))
+			cheapestAvailable := math.MaxInt32
+			if cs := stateMap[cheapestVC.VariantName]; cs.MaxReplicas != nil && *cs.MaxReplicas > 0 {
+				cheapestAvailable = *cs.MaxReplicas - targets[cheapestVC.VariantName]
+			}
+			if fallbackReplicas <= cheapestAvailable && float64(fallbackReplicas)*cheapestVC.Cost < vc.Cost {
+				replicasNeeded--
+			}
```
deanlorenz, line 128: could add up for multiple variants.
fallback would need to cover for all removed.
eg A, B, C (cheapest) ==> both A,B may meet fallback criterion. C will have to cover both. could be more than cheapestAvailable.

biranofer: The "removed" are not aggregated.  If C is enough to cover remaining after removing 1 of planned A,  then B can only contribute - at worst it'll contribute 0 and C is still OK.

biranofer:
I'm considering to add a CS (Cost Save) configuration parameter [0..1] that prefer the cheapest only if it saves CS of the cost, i,e.
`if fallbackReplicas <= cheapestAvailable && float64(fallbackReplicas)*cheapestVC.Cost <(1-CS)*vc.Cost {
				replicasNeeded--
			}`

A reasonable default can be 0.2

deanlorenz (commented):
- please add unit tests.
- please resolve conflicts after rebase to main.

NIT
- cheapestVariantCapacity does not break ties. order may change. may be harder to debug and test. 
- if current is capped then cheapest must cover all remaining -- unlikely to be cheaper than one current replica.

ev-shindin:
@biranofer please rebase and address @deanlorenz review

PR #1253, opened 2026-06-09
Fix usage of Intel extended k8s resource and label names

### LINKED ISSUE #1104: Bug: discovery uses incorrect Intel resource & label names
WVA device discovery and `kind` simulation use incorrect Intel extended resource & label names, and mismatching product names:
* `intel.com/gpu.product` + `intel.com/gpu.memory` GPU node labels
* `Intel-Gaudi-2-96GB` product name for a GPU resource
* `intel.com/gpu` GPU resource

Whereas the extended resource names provided by Intel device plugins are actually following:
* `habana.ai/gaudi` for Gaudi AI products
* `gpu.intel.com/i915` for legacy GPU products
* `gpu.intel.com/xe` for recent GPU products

And node labels they provides are of the form:
* `habana.ai/product.name`
* `gpu.intel.com/product`

There's currently no automatic node labeling for device memory amounts, so those would need to be set manually, using the same pattern as rest of their labels:
* `habana.ai/device.memory`
* `gpu.intel.com/memory`

References:
* https://intel.github.io/intel-device-plugins-for-kubernetes/cmd/gpu_plugin/README.html#introduction
* https://intel.github.io/intel-device-plugins-for-kubernetes/cmd/gpu_plugin/labels.html
* https://github.com/HabanaAI/gaudi-device-plugin/blob/main/examples/hl-smi-test.yml#L29
* https://github.com/HabanaAI/gaudi-feature-discovery/blob/main/collector/collector.go

(Faulty info seems to originate at least partially from https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/580?)

xuhui-lu:
Supersedes #1105 and fixes #1104, adopting @eero-t's groundwork. Adds the missing discovery fix so i915/xe GPU counts are no longer zeroed out when Intel variants share the `gpu.intel.com/product` label. Opened as draft for review.

### DIFF
--- a/deploy/kind-emulator/setup.sh
+++ b/deploy/kind-emulator/setup.sh
@@ -44,7 +44,7 @@ Options:
     -c CLUSTER_NAME    Cluster name (default: $DEFAULT_CLUSTER_NAME)
     -n NODES           Number of nodes (default: $DEFAULT_NODES)
     -g GPUS            GPUs per node (default: $DEFAULT_GPUS_PER_NODE)
-    -t TYPE            GPU type: nvidia, amd, intel, mix, nvidia-mix, amd-mix (default: $DEFAULT_GPU_TYPE)
+    -t TYPE            GPU type: nvidia, amd, intel-gaudi, intel-i915, intel-xe, mix, nvidia-mix, amd-mix (default: $DEFAULT_GPU_TYPE)
                        - nvidia-mix: H100, A100, MI300X heterogeneous (for limiter tests)
                        - amd-mix: MI300X, MI250, A100 heterogeneous (for limiter tests)
     -d MODEL           GPU model (default: $DEFAULT_GPU_MODEL)
@@ -59,9 +59,9 @@ EOF
 
 validate_gpu_type() {
     case "$1" in
-        nvidia|amd|intel|mix|nvidia-mix|amd-mix) return 0 ;;
+        nvidia|amd|intel-gaudi|intel-i915|intel-xe|mix|nvidia-mix|amd-mix) return 0 ;;
         *)
-            echo "Error: Invalid GPU type '$1'. Valid: nvidia, amd, intel, mix, nvidia-mix, amd-mix"
+            echo "Error: Invalid GPU type '$1'. Valid: nvidia, amd, intel-gaudi, intel-i915, intel-xe, mix, nvidia-mix, amd-mix"
             exit 1
             ;;
     esac
@@ -148,13 +148,29 @@ patch_node_gpu() {
     local gpu_count="$3"
     local gpu_product="$4"
     local gpu_memory="$5"
+    # NVIDIA defaults
+    local domain="${gpu_type}.com"
+    local count_label="gpu.count"
+    local memory_label="gpu.memory"
+    local product_label="gpu.product"
+
+    case "$gpu_type" in
+        "amd")
+            product_label="gpu.product-name" ;;
+        "intel-gaudi")
+            domain="habana.ai"; count_label="device.count"; memory_label="device.memory"; product_label="product.name" ;;
+        "intel-i915"|"intel-xe")
+            domain="gpu.intel.com"; count_label="device.count"; memory_label="memory"; product_label="product" ;;
+        "nvidia"|*)
+            ;;
+    esac
 
     kubectl patch node "${node_name}" --type merge --patch "
 metadata:
   labels:
-    ${gpu_type}.com/gpu.count: \"${gpu_count}\"
-    ${gpu_type}.com/gpu.product: \"${gpu_product}\"
-    ${gpu_type}.com/gpu.memory: \"${gpu_memory}\"
+    ${domain}/${count_label}: \"${gpu_count}\"
+    ${domain}/${memory_label}: \"${gpu_memory}\"
+    ${domain}/${product_label}: \"${gpu_product}\"
 "
 }
 
@@ -167,6 +183,8 @@ patch_node_custom_label() {
 }
 
 nodes_list=$(kubectl get nodes --no-headers -o custom-columns=":metadata.name")
+# K8s node names do not contain spaces
+# shellcheck disable=SC2206
 node_array=($nodes_list)
 
 for i in "${!node_array[@]}"; do
@@ -197,11 +215,13 @@ for i in "${!node_array[@]}"; do
             esac
             ;;
         "mix")
-            # Original mixed: nvidia, amd, intel
-            case $((i % 3)) in
-                0) current_type="nvidia"; current_model="NVIDIA-A100-PCIE-80GB"; current_memory=81920 ;;
-                1) current_type="amd";    current_model="AMD-MI300X-192G";       current_memory=196608 ;;
-                2) current_type="intel";  current_model="Intel-Gaudi-2-96GB";    current_memory=98304 ;;
+            # Mix of everything supported
+            case $((i % 5)) in
+                0) current_type="nvidia"; current_model="NVIDIA-A100-PCIE-80GB";   current_memory=81920 ;;
+                1) current_type="amd";    current_model="AMD-MI300X-192G";         current_memory=196608 ;;
+                2) current_type="intel-gaudi"; current_model="Intel-Gaudi-2-96GB"; current_memory=98304 ;;
+                3) current_type="intel-i915";  current_model="Max_1100";           current_memory=49152 ;;
+                4) current_type="intel-xe";    current_model="Pro-B60-Graphics";   current_memory=24576 ;;
             esac
             ;;
         *)
@@ -241,10 +261,12 @@ for i in "${!node_array[@]}"; do
             esac
             ;;
         "mix")
-            case $((i % 3)) in
+            case $((i % 5)) in
                 0) current_type="nvidia" ;;
                 1) current_type="amd" ;;
-                2) current_type="intel" ;;
+                2) current_type="intel-gaudi" ;;
+                3) current_type="intel-i915" ;;
+                4) current_type="intel-xe" ;;
             esac
             ;;
         *)
@@ -252,13 +274,22 @@ for i in "${!node_array[@]}"; do
             ;;
     esac
 
-    resource_name="${current_type}.com~1gpu"
+    case "$current_type" in
+        "intel-gaudi")
+            resource_name="habana.ai~1gaudi" ;;
+        "intel-i915")
+            resource_name="gpu.intel.com~1i915" ;;
+        "intel-xe")
+            resource_name="gpu.intel.com~1xe" ;;
+        *)
+            resource_name="${current_type}.com~1gpu" ;;
+    esac
 
     # Use kubectl patch with --subresource=status to directly update node status
     # This avoids the need for kubectl proxy and raw curl requests
     kubectl patch node "${node_name}" --subresource=status --type=json -p '[
-        {"op":"add","path":"/status/capacity/'${resource_name}'","value":"'${gpus_per_node}'"},
-        {"op":"add","path":"/status/allocatable/'${resource_name}'","value":"'${gpus_per_node}'"}
+        {"op":"add","path":"/status/capacity/'"${resource_name}"'","value":"'"${gpus_per_node}"'"},
+        {"op":"add","path":"/status/allocatable/'"${resource_name}"'","value":"'"${gpus_per_node}"'"}
     ]'
 done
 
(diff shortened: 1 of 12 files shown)


ev-shindin:
@xuhui-lu please rebase/resolve conflicts

test/utils/resources/llmdsim.go
```diff
@@ -13,7 +13,10 @@ import (
 	"k8s.io/utils/ptr"
 )
 
-const SimulatorImage = "ghcr.io/llm-d/llm-d-inference-sim:v0.9.0"
+const (
+	// gpuResourceDefault is the extended K8s resource type requested when none is specified.
+	gpuResourceDefault = "nvidia.com/gpu"
+)
```
eero-t, line 21: This part changed in the `main` (rebase adds another const), otherwise I don't think there are any conflicts.

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -98,6 +111,14 @@ func (d *K8sWithGpuOperator) listGPUNodes(ctx context.Context) (map[string]NodeI
...
 				}
 			}
+			// i915 and xe share the product label gpu.intel.com/product, so an
+			// i915 node is also selected during the xe iteration (and vice versa).
+			// In the non-matching iteration this resource has no allocatable, so
+			// count is 0. Don't let that zero overwrite a non-zero count already
+			// recorded for the same model in an earlier iteration.
+			if prev, seen := ni.Accelerators[model]; seen && count == 0 {
+				count = prev.Count
+			}
```
eero-t: Is there some reason why nodes having recognized product labels but without associated resources (`count = 0`) that could be allocated, would be relevant i.e. why not just continue loop when no resource matches?

(Does other code using mapping require those zero counts for product labels?)

PS. I'm asking because my quick fix in #1105 that just did continue, did not fix (all) the test issues.

eero-t (commented):
Thanks for continuing with this!  Below are couple of comments.

(I did minor updates for #1105 after @ev-shindin's comments, but it's better somebody who has time to really look into & address PR issues continues with it.)

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -98,6 +111,14 @@ func (d *K8sWithGpuOperator) listGPUNodes(ctx context.Context) (map[string]NodeI
...
 			ni.Accelerators[model] = AcceleratorModelInfo{
 				Count:  count,
 				Memory: node.Labels[memKey],
```
eero-t: Another issue is memory key:
```suggestion
			memory := node.Labels[memKey]
			if prev, seen := ni.Accelerators[model]; seen && count == 0 {
				count = prev.Count
				memory = prev.Memory
			}
			ni.Accelerators[model] = AcceleratorModelInfo{
				Count:  count,
				Memory: memory,
```
That's why just continuing on zero count seemed better alternative.

deploy/kind-emulator/setup.sh
```diff
@@ -241,24 +261,35 @@ for i in "${!node_array[@]}"; do
...
+    esac
 
     # Use kubectl patch with --subresource=status to directly update node status
     # This avoids the need for kubectl proxy and raw curl requests
     kubectl patch node "${node_name}" --subresource=status --type=json -p '[
-        {"op":"add","path":"/status/capacity/'${resource_name}'","value":"'${gpus_per_node}'"},
-        {"op":"add","path":"/status/allocatable/'${resource_name}'","value":"'${gpus_per_node}'"}
+        {"op":"add","path":"/status/capacity/'"${resource_name}"'","value":"'"${gpus_per_node}"'"},
+        {"op":"add","path":"/status/allocatable/'"${resource_name}"'","value":"'"${gpus_per_node}"'"}
     ]'
```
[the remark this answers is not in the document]
eero-t: Copilot does not seem to understand multi-line shell quoting.

(Even when being corrected on that issue multiple times.)

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -13,11 +13,23 @@ import (
...
-	"intel.com",
+var vendors = []gpuInfo{
+	{"NVIDIA", "nvidia.com/gpu", "nvidia.com/gpu.product", "nvidia.com/gpu.memory"},
+	{"AMD", "amd.com/gpu", "amd.com/gpu.product-name", "amd.com/gpu.memory"},
+	// NOTE: Node labeling rules installed for Node Feature Discovery (NFD) by Intel GPU operator,
+	// provide product labels only for Data Center products. Current Intel Gaudi / GPU operators
+	// do not label nodes with device memory information, that info needs to be labeled separately.
+	{"Intel", "habana.ai/gaudi", "habana.ai/product.name", "habana.ai/device.memory"},
+	{"Intel", "gpu.intel.com/i915", "gpu.intel.com/product", "gpu.intel.com/memory"},
+	{"Intel", "gpu.intel.com/xe", "gpu.intel.com/product", "gpu.intel.com/memory"},
```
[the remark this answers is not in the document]
eero-t: maybe `vendorResources`?

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -98,9 +111,20 @@ func (d *K8sWithGpuOperator) listGPUNodes(ctx context.Context) (map[string]NodeI
...
+				count = prev.Count
+				memory = prev.Memory
+			}
 			ni.Accelerators[model] = AcceleratorModelInfo{
 				Count:  count,
-				Memory: node.Labels[memKey],
+				Memory: memory,
 			}
 			nodes[node.Name] = ni
 			accelerators[model] += count
```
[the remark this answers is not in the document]
eero-t: It's keyed by product model name, and (unlike the label names) model names for accelerators differ between resources.

test/utils/resources/llmdsim.go
```diff
@@ -15,6 +15,11 @@ import (
 
 const SimulatorImage = "ghcr.io/llm-d/llm-d-inference-sim:v0.9.0"
 
+const (
+	// gpuResourceDefault is the extended K8s resource type requested when none is specified.
+	gpuResourceDefault = "nvidia.com/gpu"
```
eero-t: Better to put them in the same `const` block.

test/utils/resources/llmdsim.go
eero-t: Please resolve, I don't have an option for that.

ev-shindin:
@xuhui-lu @eero-t Should we plan this for the next release?

xuhui-lu:
Pushed a cleanup pass for the remaining review threads and resolved them. CI is running on the latest commit; local focused tests passed:

- `go test ./internal/discovery ./internal/resources ./internal/utils ./internal/utils/scaletarget ./test/utils/resources`
- `bash -n deploy/kind-emulator/setup.sh`

Once CI is green, this should be ready from my side.

eero-t:
> @xuhui-lu @eero-t Should we plan this for the next release?

@ev-shindin Yes, please.  Could you trigger the pending checks and review it otherwise?

eero-t (approved):
LGTM.

(if remaining CI checks indicate that further changes are needed, I have this & next week some time to look into them & propose / review fixes.)

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -57,9 +69,10 @@ func (d *K8sWithGpuOperator) listGPUNodes(ctx context.Context) (map[string]NodeI
 
 	// Query nodes for each GPU vendor separately.
 	// K8s LabelSelectors don't support OR logic across different keys (e.g. nvidia OR amd).
-	for _, vendor := range vendors {
-		prodKey := vendor + "/gpu.product"
-		memKey := vendor + "/gpu.memory"
+	for _, gpuInfo := range vendorResources {
```
ev-shindin: The slice rename to `vendorResources` is good, but the loop variable is still named `gpuInfo`, which shadows the `gpuInfo` struct type for the body of the loop (the type name is unreferenceable inside). Suggest renaming the loop var (e.g. `gr` or `res`):

```suggestion
	for _, gr := range vendorResources {
		vendor := gr.vendor
```

(and update `gr.memoryLabel` / `gr.productLabel` / `gr.resourceName` accordingly).

internal/constants/constants.go
```diff
@@ -42,14 +42,22 @@ var (
...
+	// GpuResources lists extended resource names provided by K8s GPU device vendor plugins
+	GpuResources = []string{
+		"nvidia.com/gpu",
+		"amd.com/gpu",
+		"habana.ai/gaudi",
+		"gpu.intel.com/i915",
+		"gpu.intel.com/xe",
+	}
 
 	// GpuProductKeys are the node selector/affinity keys used to identify GPU products
```
ev-shindin: `GpuProductKeys` here and the `productLabel` column of `vendorResources` (k8s_with_gpu_operator.go:23) encode the same per-vendor label set, so they can drift independently.

internal/discovery/k8s_with_gpu_operator_test.go
```diff
@@ -171,15 +171,43 @@ func TestDiscover_MixedVendors(t *testing.T) {
...
+					"habana.ai/gaudi": resource.MustParse("8"),
+				},
+			},
+		},
+		&corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "node-intel-i915",
+				Labels: map[string]string{
+					"gpu.intel.com/product": "Max_1100",
+					"gpu.intel.com/memory":  "49152", // manually set label
```
eero-t: May be better to remove these comments (that were in my original commit):
```suggestion
					"gpu.intel.com/memory":  "49152",
```
While such labels would currently need to be manually added, they could be automated later e.g. by providing them to NFD [1] from a XPUMD side-car [2], or just by using NFD feature rules (for a limited set of GPU models) [3].

[1]: https://kubernetes-sigs.github.io/node-feature-discovery/stable/usage/customization-guide.html#local-feature-source

[2]: https://github.com/intel/xpumanager/tree/main/xpumd/exporter/intelxpuinfo#readme

[3]: https://kubernetes-sigs.github.io/node-feature-discovery/stable/usage/customization-guide.html#nodefeature-custom-resource

internal/discovery/k8s_with_gpu_operator.go
ev-shindin, line 72: i915 and xe share `gpu.intel.com/product`, so the per-entry `d.Client.List(...)` (inside the `for _, … := range vendorResources` loop) issues two identical node-list calls every reconcile — same selector, same nodes. Correctness is fine (the zero-count guard dedups), but it's a redundant API round-trip.

ev-shindin, line 116: `RecordAvailableGPUsMetric` is called per vendor pass from the per-iteration `accelerators` map. A real **xe** node (e.g. Pro-B60) is first seen during the **i915** pass with `count==0`, so it records and emits a `0` for that model; the later xe pass then emits the correct count. Because i915 precedes xe in the table and it's all one synchronous `Discover()`, the 0 is overwritten before any scrape — so it's **not observable in practice** — but it's fragile (relies on table ordering).

ev-shindin (commented):
Thanks for the revisions - please see my comments below.

internal/discovery/k8s_with_gpu_operator.go
eero-t: Loop above has now a check to take care of the case when zero comes afterwards, instead of before:
```
if _, seen := ni.Accelerators[model]; seen && count == 0 {		
	nodes[node.Name] = ni
	continue
```
EDIT: isn't that enough?

internal/constants/constants.go
eero-t: Is this a request for a change, or just acknowledgement of a problem with the assumptions in the original code?

IMHO getting rid of that disconnect belongs to some other PR.  This one is targeted to fixing / adding support for correct Intel (+AMD) resource names & associated info.  Once that works, and tests guarantee it, some later PR can do a targeted refactor / clean up.

ev-shindin: Fine to defer, but a short cross-reference comment on each ("keep in sync with the other") would help a future edit update both.

internal/discovery/k8s_with_gpu_operator.go
eero-t: Ditto. ~Issue in original code and not really the point of this PR.~ IMHO improving that belongs to a separate refactor / cleanup PR.

ev-shindin: Agreed, non-blocking. Note for the record: this duplicate is *introduced by this PR* (the original code used three distinct product-label selectors and listed each vendor once); it's a new artifact of the i915/xe split, not pre-existing. Left for a follow-up by mutual agreement.

ev-shindin: That guard does handle the case where the zero is read *after* a real count (e.g. an i915 node recorded in the i915 pass, then seen as 0 in the xe pass: `seen` is true, so it's skipped). 

What I had in mind is the reverse ordering: a real **xe** node gets first encountered in the **i915 pass** (i915 comes before xe in the table) with `count == 0`. There it isn't `seen` yet, so the guard doesn't fire — it records 0 and emits a transient `0` metric for that model, which the xe pass then corrects to the real count.

But it's not observable in practice: both emits happen in the same synchronous `Discover()`, so the gauge is correct by the time anything scrapes it. Purely an ordering nit — fine to leave as-is or fold into the cleanup PR. No change needed here.

internal/constants/constants.go
eero-t: It's the same problem with the `GpuResources` list.

Both of them seem to be referred only from couple of places in:
* `resources/resource.go`
* `utils/utils.go`

=> May be best to move `GpuInfo` type + `vendorResources` array from `discovery/k8s_with_gpu_operator.go` to `constants/constants.go`, and replace `GpuResources` & `GpuProductKeys` with `VendorResources`.

internal/discovery/k8s_with_gpu_operator.go
eero-t: Ah, I see.  And the zero counts are also needed, that's why just doing continue from here did not work?
```
if cap, ok := node.Status.Allocatable[corev1.ResourceName(gpuInfo.resourceName)]; !ok {
    continue
```

internal/constants/constants.go
eero-t: FYI: I changed #1105 to be based on this PR, rebased it to latest `main`, and added commit (https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1105/changes/a96ff87be93fd300d4be107aa617f64459488cd7, the last one) showing the result of replacing `GpuResources` & `GpuProductKeys` in `constants` package with `vendorResources` (+ handling 2 trivial items commented above).

@xuhui-lu I haven't tested it, but at least Emacs Go-LSP did not give any warnings for the "Avoid duplicating vendor resource & product label info" commit, so you could cherry-pick it from my PR (+ further test / fix it if needed), if @ev-shindin thinks the approach looks OK.

(I assume you're on a quite different timezone from mine, so hopefully this will expedite things.)

xuhui-lu: cherrypicked, thanks!

eero-t: Dang, I should have ran Copilot immediately on it.  It noticed that original code had a separate GKE product label for Nvidia GPUs, and `utils` package test for it.  See my comment in #1105.

eero-t: => It may be better to drop that commit, and instead just add comment like this here:

```// Note: keep `GpuResources` & `GpuProductKeys` in sync with `vendorResources` struct array content in discovery```

And let some further PR handle the discrepancy with the GKE label (used only by `utils_test.go`), as it's an issue in the original code.

ev-shindin (approved):
Thanks @xuhui-lu and @eero-t

PR #1255, opened 2026-06-10
fix(saturation_v2): include gpuCount in k2 history key to prevent cross-variant contamination

### LINKED ISSUE #1254: saturation_v2: k2 history key missing gpuCount causes cross-variant capacity contamination
## Problem

The `computeK2` rolling-average history key was `"modelID|accelerator|outputBucket"`. In multi-variant deployments where two variants serve the **same model on the same accelerator type** but with different GPU counts (e.g. TP=1 vs TP=2), their k2 observations share the same bucket.

This causes cross-contamination:
- A TP=2 replica's high `tokensInUse` (larger batch capacity) inflates the rolling average for TP=1 replicas, making them appear more capable than they are.
- Conversely, a lightly-loaded TP=2 replica can depress the estimate for TP=1.

The result is incorrect `PerReplicaCapacity` estimates, leading to wrong scale-up decisions (too few or too many replicas).

## Fix

Add `gpuCount` to the history key:

```go
// Before
historyKey := fmt.Sprintf("%s|%s|%s", modelID, accelerator, outputBucket)

// After
historyKey := fmt.Sprintf("%s|%s|%d|%s", modelID, accelerator, gpuCount, outputBucket)
```

This ensures TP=1 and TP=2 replicas maintain independent k2 histories.

## Observed impact

On a two-variant Llama-3.1-8B setup (primary TP=2 cost=10, v2 TP=1 cost=5), without the fix the contaminated k2 history caused the cost-aware optimizer to route demand incorrectly. With the fix, per-replica capacity estimates are stable and consistent with hardware expectations.

biranofer:
## Summary

- The `computeK2` rolling-average history key was `"modelID|accelerator|outputBucket"`, causing TP=1 and TP=2 variants of the same model to share k2 observations
- Add `gpuCount` to the key so each tensor-parallel width maintains an independent history
- Update the test to use the new key format

Fixes #1254

## Test plan

- [ ] `make test` passes — existing k2 history test updated to use `"modelID|accelerator|gpuCount|outputBucket"` key format
- [ ] Empirically validated on pokprod001 with two-variant Llama-3.1-8B (primary TP=2, v2 TP=1): with fix, per-replica capacity estimates are stable and consistent with hardware expectations

### DIFF
--- a/internal/engines/analyzers/saturation_v2/analyzer.go
+++ b/internal/engines/analyzers/saturation_v2/analyzer.go
@@ -20,7 +20,8 @@ type SaturationAnalyzer struct {
 	// mu protects computeCapacityHistory from concurrent access.
 	mu sync.Mutex
 	// computeCapacityHistory stores rolling averages of observed k2 values,
-	// keyed by "modelID|accelerator|outputBucket".
+	// keyed by "modelID|accelerator|gpuCount|outputBucket".
+	// TODO: check if we need to use other model parameters as key in the future.
 	computeCapacityHistory map[string]*rollingAverage
 	capacityStore          *CapacityKnowledgeStore
 }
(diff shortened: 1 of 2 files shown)


ev-shindin (approved):
Thanjs @biranofer !

PR #1260, opened 2026-06-11
feat(collector): derive pod → VariantAutoscaling mapping per cycle

ev-shindin:
## What

Replaces the tenant-authored `llm-d.ai/variant` pod-template label as the **default** path for attributing a pod's metrics to its `VariantAutoscaling`. WVA now derives the mapping itself, once per optimization cycle, from each VA's `scaleTargetRef` selector. Pointing a `VariantAutoscaling` at a running `Deployment`/`LeaderWorkerSet` is sufficient — **no required pod-template label, no rolling restart to adopt, no ServiceMonitor relabel rule.**

Implements the design in [`docs/proposals/design-pod-va-mapping.md`](docs/proposals/design-pod-va-mapping.md) (#1210).

## Why

Today the collector reads the VA name from the `llm-d.ai/variant` label that the tenant must stamp on the workload's **pod template** (twice for LWS) and propagate via a `ServiceMonitor` relabel rule. That's wrong for an autoscaler:

- **It blocks in-place adoption.** Adding the label to a running workload's pod template is a rolling restart of the whole fleet — forced purely to *start* autoscaling.
- **It fails silently.** A wrong label key/value, a missing relabel rule, or a half-templated LWS yields zero matching metrics and a "working" but wrong scaling decision, with no signal.
- **The anchor is being deprecated.** The label's value must equal the `VariantAutoscaling` name — the CRD that's being deprecated.

HPA/KEDA/VPA attach via a target reference and read the target; none edit the pod template. This change brings WVA in line with that norm.

## How

- **New package `internal/collector/podvamap`** — builds a per-cycle `"<namespace>/<pod>" → VariantAutoscaling` map: for each managed VA it lists the pods behind its scale target's selector and attributes each to that VA. `Lookup(namespace, pod)` is O(1) on the metrics hot path.
- **Uncached reads — no new watch.** The per-cycle pod LIST uses `mgr.GetAPIReader()` (a direct, bounded, selector-scoped API LIST per scale target), so it starts **no cluster-wide Pod informer/cache**. The engine gains an `APIReader` field (set in `cmd/main.go`), falling back to the cached client when unset (tests).
- **Ownership / overlapping selectors.** A pod is attributed by selector + namespace match; pods matched by more than one workload selector are disambiguated by an in-memory owner-reference walk (`Pod → ReplicaSet → Deployment`, `Pod → StatefulSet → LeaderWorkerSet`), with the workload→VA match resolved against the in-memory VA set (no field-index cache). Unresolved collisions are left unmapped rather than misattributed.
- **`ScaleTargetAccessor.GetPodSelector()`** — added for Deployment (`spec.selector`) and LeaderWorkerSet (the `leaderworkerset.sigs.k8s.io/name` label).
- **Label stays as a higher-precedence override.** `llm-d.ai/variant` (and its relabel rule) remain supported for custom kinds / non-standard owner chains and for back-compat; when present on a pod's metrics it wins over the derived mapping. The override's value names the same autoscaling object the derived path resolves to (deprecation-proof, not a frozen VA name).

## Observability

- **`wva_pod_mapping_miss_total{reason}`** — counts pods that couldn't be attributed: `reason="unresolved"` (matched no managed workload selector and no override label) or `reason="ambiguous"` (overlapping selectors, owner-ref tiebreak failed). Surfaces the previously-silent wrong-capacity fallback.
- **No VA status condition** is added: the `VariantAutoscaling` CRD is being deprecated, so the durable miss signal lives in Prometheus rather than on the object.

## Back-compat & RBAC

- Existing installs that keep the `llm-d.ai/variant` label + ServiceMonitor relabel rule keep working unchanged (the label is a higher-precedence override).
- `docs/design/controller-behavior.md` updated: the label/relabel rule are now documented as an **optional override**, with the new default and the miss-metric semantics.
- No RBAC change needed — the manager ClusterRole already grants `pods` list and `replicasets`/`statefulsets` get.

## Out of scope / follow-ups

- Migration steps, implementation phasing, and the full discovery mechanism after VA-CRD deprecation are tracked separately in the proposal.
- Optimization noted in code (`TODO`): collapsing the per-scale-target LIST to one LIST per namespace with in-memory selector matching, if the per-cycle API-call count grows.
- This change also unblocks dropping `llm_d_ai_variant` from the per-pod PromQL queries (e.g. the throughput analyzer) once derivation is the sole path.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -454,6 +454,9 @@ func main() {
 			sourceRegistry,
 			cfg, // Pass unified Config to engine
 		)
+		// Use the uncached API reader for the per-cycle pod LIST in pod→VA derivation,
+		// so it does not start a cluster-wide Pod informer/cache.
+		engine.APIReader = mgr.GetAPIReader()
 		go engine.StartOptimizeLoop(ctx)
 		return nil
 	}))
--- a/docs/design/controller-behavior.md
+++ b/docs/design/controller-behavior.md
@@ -243,14 +243,16 @@ This periodic reconciliation is why many Update and Delete events can be safely 
 
 ## Prerequisites
 
-### `llm-d.ai/variant` Label on the Scale Target
+### `llm-d.ai/variant` Label on the Scale Target (optional override)
 
-WVA identifies which pods belong to a given `VariantAutoscaling` resource by reading the `llm-d.ai/variant` label from Prometheus metrics. Two things must be true for this to work:
+By default WVA derives which pods belong to a `VariantAutoscaling` itself, once per optimization cycle, by listing the pods behind the scale target's selector (`scaleTargetRef` → Deployment/LeaderWorkerSet) and attributing each to that VA. **For the standard kinds (Deployment, LeaderWorkerSet) no pod-template label or ServiceMonitor relabeling rule is required** — pointing a `VariantAutoscaling` at a running workload is sufficient, with no rolling restart to adopt.
+
+The `llm-d.ai/variant` label remains supported as a **higher-precedence override**, for workloads the selector can't resolve (custom kinds or non-standard owner chains) and for back-compat with existing installs. When the label is present on a pod it takes precedence over the derived mapping. To use it, two things must be true:
 
 1. The `llm-d.ai/variant` label must be present on the **pod template** of the scale target (Deployment or LeaderWorkerSet), with a value equal to the name of the corresponding `VariantAutoscaling` resource.
 2. The `ServiceMonitor` or `PodMonitor` that Prometheus uses to scrape those pods must include a target relabeling rule that propagates the pod label into the scraped metrics as `llm_d_ai_variant`.
 
-**Both are your responsibility.** WVA does not configure either automatically.
+WVA reads the label but never writes it; both steps below are your responsibility only if you opt into the override.
 
 #### 1. Set the pod template label
 
@@ -315,15 +317,15 @@ spec:
 
 > **Important**: This rule must live under `relabelings` (target relabeling), **not** `metricRelabelings` (metric relabeling). The `__meta_kubernetes_pod_label_*` labels are only available during target relabeling and are stripped before metric relabeling runs.
 
-WVA uses the `llm_d_ai_variant` metric label to associate per-pod metrics with the correct `VariantAutoscaling` resource.
+WVA uses the `llm_d_ai_variant` metric label, when present, to associate per-pod metrics with the correct `VariantAutoscaling` resource, taking precedence over the derived mapping.
 
-If the pod label or the relabeling rule is absent, WVA will not collect metrics for the affected pods and scaling decisions for that variant will not be made.
+If the override label or relabeling rule is absent, WVA falls back to per-cycle selector derivation (the default). A pod that resolves to no scale target — and carries no override label — is skipped and counted in the `wva_pod_mapping_miss_total{reason="unresolved"}` metric; a pod matched by multiple overlapping selectors that the owner-reference tiebreak cannot disambiguate is counted with `reason="ambiguous"`.
 
 ## Best Practices
 
 ### For Operators
 
-1. **Set `llm-d.ai/variant` before creating the VA**: Ensure the pod template label is present on the scale target before creating the `VariantAutoscaling` resource. WVA begins collecting metrics on the first reconciliation and will miss pods that don't yet carry the label.
+1. **Adopt without touching the workload**: For Deployment/LeaderWorkerSet scale targets, no pod-template label is needed — create the `VariantAutoscaling` pointing at the running workload and WVA derives the pod mapping from the selector. Only reach for the `llm-d.ai/variant` override (and its relabeling rule) for custom kinds or non-standard owner chains; setting it requires a rolling restart of the workload, so prefer derivation where it works.
 
 2. **Create VAs after deployments are ready**: While the controller handles the race condition, creating VAs after deployments are fully initialized avoids unnecessary early reconciliation cycles.
 
--- a/internal/collector/podvamap/podvamap.go
+++ b/internal/collector/podvamap/podvamap.go
@@ -1,0 +1,199 @@
+// Package podvamap derives, once per optimization cycle, a mapping from pod
+// identity to the VariantAutoscaling whose scale target owns the pod.
+//
+// It replaces the tenant-authored llm-d.ai/variant pod-template label as the
+// default attribution path: for each managed VariantAutoscaling, WVA lists the
+// pods behind its scale target's selector and attributes them to that VA. Pods
+// matched by more than one workload selector are disambiguated by walking the
+// owner chain (Pod -> ReplicaSet/StatefulSet -> Deployment/LeaderWorkerSet ->
+// VA); if still ambiguous they are left unmapped and surface as a mapping miss.
+//
+// The map is read-only on the metrics hot path: callers resolve a pod with a
+// single Lookup, with no Kubernetes calls.
+package podvamap
+
+import (
+	"context"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	llmdVariantAutoscalingV1alpha1 "github.com/llm-d/llm-d-workload-variant-autoscaler/api/v1alpha1"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/constants"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/metrics"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/utils/scaletarget"
+)
+
+// Map holds a per-cycle "<namespace>/<pod-name>" -> VariantAutoscaling-name mapping.
+// A nil *Map is safe to Lookup against (it resolves nothing).
+type Map struct {
+	byPod map[string]string
+}
+
+// Lookup returns the VariantAutoscaling name a pod was attributed to, and whether
+// the pod was resolved. Pods are keyed by namespace and name (pod names are not
+// unique across namespaces).
+func (m *Map) Lookup(namespace, podName string) (string, bool) {
+	if m == nil || m.byPod == nil {
+		return "", false
+	}
+	va, ok := m.byPod[namespace+"/"+podName]
+	return va, ok
+}
+
+// Len returns the number of resolved pods. Used mainly in tests.
+func (m *Map) Len() int {
+	if m == nil {
+		return 0
+	}
+	return len(m.byPod)
+}
+
+// Build derives the pod -> VariantAutoscaling map for a set of managed VAs.
+//
+//   - reader is an uncached read interface (mgr.GetAPIReader()): a direct, bounded
+//     API LIST/GET per cycle that does NOT start a cluster-wide Pod informer/cache.
+//   - scaleTargets is keyed by "<namespace>/<scaleTargetName>" and provides the
+//     workload (Deployment/LWS) backing each VA's scaleTargetRef.
+//   - variantAutoscalings is keyed by "<namespace>/<vaName>".
+//
+// For each VA, the pods behind its scale target's selector are listed and attributed
+// to that VA. Overlapping-selector collisions are resolved by owner reference;
+// unresolved collisions increment the pod-mapping miss counter and are omitted.
+func Build(
+	ctx context.Context,
+	reader client.Reader,
+	scaleTargets map[string]scaletarget.ScaleTargetAccessor,
+	variantAutoscalings map[string]*llmdVariantAutoscalingV1alpha1.VariantAutoscaling,
+) *Map {
+	logger := ctrl.LoggerFrom(ctx)
+
+	// candidates: "<namespace>/<podName>" -> set of VA names whose selector matched it.
+	candidates := make(map[string]map[string]struct{})
+	// podObjs: same key -> Pod, retained only for collisions needing owner-ref tiebreak.
+	podObjs := make(map[string]*corev1.Pod)
+
+	// TODO: this issues one selector-scoped List per VA (O(VAs) API calls per cycle).
+	// Acceptable at realistic VA counts; if it grows, switch to one List per namespace
+	// plus in-memory selector matching.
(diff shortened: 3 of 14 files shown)


lionelvillard (changes requested):
Remove all mentions of VariantAutoscaler as a resource.

@dumb0002 Can you confirm the in case of FMA integration the llm-d.ai/variant label is still required ? If so can we treat requester pod as special ScaleTarget with its own resolution algorithm? If then the controller-behavior doc needs to clearly acknowledge the use case.

cmd/main.go
```diff
@@ -454,6 +454,9 @@ func main() {
 			sourceRegistry,
 			cfg, // Pass unified Config to engine
 		)
+		// Use the uncached API reader for the per-cycle pod LIST in pod→VA derivation,
+		// so it does not start a cluster-wide Pod informer/cache.
+		engine.APIReader = mgr.GetAPIReader()
```
lionelvillard, line 459: Change NewEngine signature and pass mgr.GetAPIReader(), cleaner.

dumb0002:
> Remove all mentions of VariantAutoscaler as a resource.
> 
> @dumb0002 Can you confirm the in case of FMA integration the llm-d.ai/variant label is still required ? If so can we treat requester pod as special ScaleTarget with its own resolution algorithm? If then the controller-behavior doc needs to clearly acknowledge the use case.

@lionelvillard, yes, indeed! We still need the`llm-d.ai/variant` label for FMA dual-pod technique. The label helps identify the target `server requesting` pod to associate the prometheus metrics collect from a FMA `launcher pod`.

docs/design/controller-behavior.md
```diff
@@ -243,14 +243,16 @@ This periodic reconciliation is why many Update and Delete events can be safely
 
 ## Prerequisites
 
-### `llm-d.ai/variant` Label on the Scale Target
+### `llm-d.ai/variant` Label on the Scale Target (optional override)
```
lionelvillard, line 246: I would describe it this way:
1. the use case is about associating metrics to HPA objects
2. Explain how it works in sort of generic terms, metric pod_name/ns label -> pod -> .... -> scaleRefTarget -> HPA. 
3. Describe the 3 traversal algorithms currently supported: Deployment, LWS and Dual-Pods. For dual-pods, the current solution requires `llm-d.ai/variant` and metric relabeling, not needed for the 2 first algos.


Finally, can Deployment, LWS be generalized to ownership traversal?

internal/collector/podvamap/podvamap.go
```diff
@@ -0,0 +1,199 @@
+// Package podvamap derives, once per optimization cycle, a mapping from pod
```
lionelvillard, line 1: podvamap, please rename.

internal/collector/podvamap/podvamap.go
```diff
@@ -0,0 +1,199 @@
...
+// identity to the VariantAutoscaling whose scale target owns the pod.
+//
+// It replaces the tenant-authored llm-d.ai/variant pod-template label as the
+// default attribution path: for each managed VariantAutoscaling, WVA lists the
+// pods behind its scale target's selector and attributes them to that VA. Pods
+// matched by more than one workload selector are disambiguated by walking the
+// owner chain (Pod -> ReplicaSet/StatefulSet -> Deployment/LeaderWorkerSet ->
+// VA); if still ambiguous they are left unmapped and surface as a mapping miss.
+//
```
lionelvillard, line 10: please remove, good in a design doc, not as code doc.

internal/collector/podvamap/podvamap.go
```diff
@@ -0,0 +1,199 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	llmdVariantAutoscalingV1alpha1 "github.com/llm-d/llm-d-workload-variant-autoscaler/api/v1alpha1"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/constants"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/metrics"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/utils/scaletarget"
+)
+
+// Map holds a per-cycle "<namespace>/<pod-name>" -> VariantAutoscaling-name mapping.
```
lionelvillard, line 31: remove per-cycle.

internal/collector/podvamap/podvamap.go
```diff
@@ -0,0 +1,199 @@
...
+			return ""
+		}
+		controllee = rs
+	case constants.StatefulSetKind:
+		ss := &appsv1.StatefulSet{}
+		if err := reader.Get(ctx, client.ObjectKey{Namespace: ns, Name: owner.Name}, ss); err != nil {
+			logger.V(logging.DEBUG).Info("pod-VA derivation: failed to get StatefulSet", "statefulset", owner.Name, "error", err)
+			return ""
+		}
+		controllee = ss
```
lionelvillard, line 181: consider using a generic algorithm

internal/collector/podvamap/podvamap.go
```diff
@@ -0,0 +1,199 @@
...
+//   - variantAutoscalings is keyed by "<namespace>/<vaName>".
+//
+// For each VA, the pods behind its scale target's selector are listed and attributed
+// to that VA. Overlapping-selector collisions are resolved by owner reference;
+// unresolved collisions increment the pod-mapping miss counter and are omitted.
+func Build(
+	ctx context.Context,
+	reader client.Reader,
+	scaleTargets map[string]scaletarget.ScaleTargetAccessor,
+	variantAutoscalings map[string]*llmdVariantAutoscalingV1alpha1.VariantAutoscaling,
```
lionelvillard, line 71: variantAutoscalings should be HPA/KEDA

internal/constants/metrics.go
```diff
@@ -185,6 +185,21 @@ const (
 	// WVAKvCacheTokensCapacity is a gauge that tracks total KV cache token capacity per variant.
 	// Labels: variant_name, namespace, model_name
 	WVAKvCacheTokensCapacity = "wva_kv_cache_tokens_capacity"
+
+	// WVAPodMappingMissTotal is a counter that tracks pods that could not be attributed
+	// to a VariantAutoscaling during per-cycle derivation from the scaleTargetRef selector.
```
lionelvillard, line 190: VariantAutoscaling -> HPA/KEDA

deanlorenz (commented):
**Missing test: owner-ref tiebreak success path.** `podvamap_test.go` covers the failure case (overlapping selectors, no owner chain → unresolved). There is no test where `resolveByOwnerRef` succeeds — pod has a ReplicaSet owner pointing to one of the two Deployments. Worth adding to validate the happy path.

**#1263 — suggested as part of this PR: finish separating attribution from the query.** This PR already makes the podMap the attribution authority for label-less pods. The natural completion is to stop carrying `llm_d_ai_variant` in the PromQL `max by` clauses at all (saturation: 6 queries, queueing-model: 2) and resolve VA attribution once, after the query, from the podMap. `buildInstanceKey` then builds replica identity (`instance`, `pod`) only. The `llm-d.ai/variant` override stays supported by capturing it during the pod `List` that `podvamap.Build` already does — no extra K8s calls, label still wins over the selector-derived mapping. This makes every per-replica query analyzer-independent in form and closes the same-namespace/same-model scoping gap (two VAs of one model are disambiguated by the podMap, not the metric label). Filed as #1263 with the detailed steps; folding it into this PR keeps the query and the attribution layer consistent in one change.

**#1264 — `float64` zero-as-sentinel in `ReplicaMetrics`.** Separate follow-up; the zero-vs-unavailable ambiguity becomes more visible once label-free deployments are common. Not a blocker for this PR.

docs/design/controller-behavior.md
```diff
@@ -243,14 +243,16 @@ This periodic reconciliation is why many Update and Delete events can be safely
 
 ## Prerequisites
```
shuynh2017, line 244: This should no longer be "Prerequisites" ?

docs/design/controller-behavior.md
```diff
@@ -243,14 +243,16 @@ This periodic reconciliation is why many Update and Delete events can be safely
...
 
-### `llm-d.ai/variant` Label on the Scale Target
+### `llm-d.ai/variant` Label on the Scale Target (optional override)
 
-WVA identifies which pods belong to a given `VariantAutoscaling` resource by reading the `llm-d.ai/variant` label from Prometheus metrics. Two things must be true for this to work:
+By default WVA derives which pods belong to a `VariantAutoscaling` itself, once per optimization cycle, by listing the pods behind the scale target's selector (`scaleTargetRef` → Deployment/LeaderWorkerSet) and attributing each to that VA. **For the standard kinds (Deployment, LeaderWorkerSet) no pod-template label or ServiceMonitor relabeling rule is required** — pointing a `VariantAutoscaling` at a running workload is sufficient, with no rolling restart to adopt.
+
```
shuynh2017, line 249: Statement on line 248 is a bit late in this doc as it's after the 4 examples where this derivation is implied.

docs/design/controller-behavior.md
```diff
@@ -243,14 +243,16 @@ This periodic reconciliation is why many Update and Delete events can be safely
...
+The `llm-d.ai/variant` label remains supported as a **higher-precedence override**, for workloads the selector can't resolve (custom kinds or non-standard owner chains) and for back-compat with existing installs. When the label is present on a pod it takes precedence over the derived mapping. To use it, two things must be true:
 
 1. The `llm-d.ai/variant` label must be present on the **pod template** of the scale target (Deployment or LeaderWorkerSet), with a value equal to the name of the corresponding `VariantAutoscaling` resource.
```
shuynh2017, line 252: this is a bit confusing since line 248 states "For the standard kinds (Deployment, LeaderWorkerSet) no pod-template label or ServiceMonitor relabeling rule is required"?

docs/design/controller-behavior.md
```diff
@@ -315,15 +317,15 @@ spec:
 
 > **Important**: This rule must live under `relabelings` (target relabeling), **not** `metricRelabelings` (metric relabeling). The `__meta_kubernetes_pod_label_*` labels are only available during target relabeling and are stripped before metric relabeling runs.
 
-WVA uses the `llm_d_ai_variant` metric label to associate per-pod metrics with the correct `VariantAutoscaling` resource.
+WVA uses the `llm_d_ai_variant` metric label, when present, to associate per-pod metrics with the correct `VariantAutoscaling` resource, taking precedence over the derived mapping.
 
-If the pod label or the relabeling rule is absent, WVA will not collect metrics for the affected pods and scaling decisions for that variant will not be made.
+If the override label or relabeling rule is absent, WVA falls back to per-cycle selector derivation (the default). A pod that resolves to no scale target — and carries no override label — is skipped and counted in the `wva_pod_mapping_miss_total{reason="unresolved"}` metric; a pod matched by multiple overlapping selectors that the owner-reference tiebreak cannot disambiguate is counted with `reason="ambiguous"`.
```
shuynh2017, line 322: may be useful to add this to dashboard, if so, pls create an issue and I can take.

internal/constants/metrics.go
shuynh2017: may be refer HPA/KEDA/VariantAutoscaling to something generic as `autoscaler` ?

internal/collector/replica_metrics.go
```diff
@@ -315,8 +321,15 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
...
-		// Extract VA name from llm_d_ai_variant label (propagated by ServiceMonitor)
+		// Extract VA name from llm_d_ai_variant label (propagated by ServiceMonitor).
+		// The label is a higher-precedence override; when absent, fall back to the
+		// per-cycle pod→VA map derived from scale-target selectors.
 		vaName := labels[constants.VariantLabelPrometheusKey]
+		if vaName == "" && podMap != nil {
+			if derived, ok := podMap.Lookup(namespace, podName); ok {
+				vaName = derived
+			}
+		}
```
shuynh2017, line 332: can `buildInstanceKey ` return empty `vaName`? If so, should the calls to `buildInstanceKey ` handle empty `vaName` immediately?

internal/collector/replica_metrics.go
```diff
@@ -752,6 +765,9 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
 		}
 
 		if vaName == "" {
+			// Neither the override label nor per-cycle derivation attributed this pod to
```
shuynh2017, line 768: should this check be done right after `vaName := data.vaName`?

internal/collector/podvamap/podvamap.go
```diff
@@ -0,0 +1,199 @@
...
+	podObjs := make(map[string]*corev1.Pod)
+
+	// TODO: this issues one selector-scoped List per VA (O(VAs) API calls per cycle).
+	// Acceptable at realistic VA counts; if it grows, switch to one List per namespace
+	// plus in-memory selector matching.
+	for vaKey, va := range variantAutoscalings {
+		if va == nil {
+			continue
+		}
+		accessor, ok := scaleTargets[va.Namespace+"/"+va.GetScaleTargetName()]
```
shuynh2017, line 87: there's a utility function in `utils/variant.go`:
```
// GetNamespacedKey is a helper for building namespaced resource keys.
func GetNamespacedKey(namespace, name string) string {
	return namespace + "/" + name
}
```

internal/collector/podvamap/podvamap.go
```diff
@@ -0,0 +1,199 @@
...
+			client.InNamespace(va.Namespace),
+			client.MatchingLabelsSelector{Selector: accessor.GetPodSelector()},
+		); err != nil {
+			logger.V(logging.DEBUG).Info("pod-VA derivation: failed to list pods", "va", vaKey, "error", err)
+			continue
+		}
+
+		for i := range podList.Items {
+			pod := &podList.Items[i]
+			key := pod.Namespace + "/" + pod.Name
```
shuynh2017, line 104: use util function mentioned above?

internal/collector/podvamap/podvamap.go
```diff
@@ -0,0 +1,199 @@
...
+			}
+		}
+		// Still ambiguous: leave unmapped and record a miss. variant_name is empty
+		// because the pod could not be attributed to a single VA — that is the signal.
+		ns, name := "", key
+		if pod != nil {
+			ns, name = pod.Namespace, pod.Name
+		}
+		metrics.IncPodMappingMiss("", ns, constants.PodMappingMissAmbiguous)
+		logger.V(logging.DEBUG).Info("pod-VA derivation: pod matched multiple workloads, owner-ref tiebreak failed",
```
shuynh2017, line 137: general suggestion: derivation -> resolution?

ev-shindin:
Closeing this as #1267  provide better design

PR #1267, opened 2026-06-11
feat(locator): make llm-d.ai/variant optional for Deployment / LWS via ownerReference walk

lionelvillard:
## Summary

- Adds a `PodLocator` (`internal/collector/locator/`) that resolves a pod (by name) to its managed `HorizontalPodAutoscaler` or KEDA `ScaledObject` by walking `ownerReferences` and consulting two new field indexes (`HPAByScaleTargetKey`, `ScaledObjectByScaleTargetKey`).
- Wires the locator into `ReplicaMetricsCollector` as a fallback when the `llm_d_ai_variant` Prometheus label is absent — making the label and its ServiceMonitor relabeling rule **optional for ordinary Deployment / LWS layouts** while keeping it as the supported linkage for shadow-pod layouts.
- Reuses `mgr.GetAPIReader()` for the owner-walk so no Pod / ReplicaSet / Deployment / LWS informer is started by the locator; pod-side memoization sits in a small in-process LRU (size 4096), keyed by `(namespace, podName)`. The `pod → top-level scale-target` relation is immutable per Kubernetes' ownerReference rules, so cached entries are correct for the lifetime of the cached pod; the scaler-side step (managed annotation, scaleTargetRef edits) always reads through the field index.
- Splits `internal/controller/indexers/indexers.go` into one file per indexed resource type (`variantautoscaling.go`, `hpa.go`, `scaledobject.go`).

Spec: `docs/superpowers/specs/2026-06-11-pod-to-managed-scaler-locator-design.md`. Plan: `docs/superpowers/plans/2026-06-11-pod-to-managed-scaler-locator.md`.

## What changes for operators

- **Deployment / LeaderWorkerSet scale targets**: the `llm-d.ai/variant` pod-template label and the ServiceMonitor / PodMonitor relabeling rule become optional. The locator derives the variant association from `ownerReferences`. Existing deployments that already stamp the label keep working unchanged — the label-driven fast path simply short-circuits the locator.
- **Shadow-pod layouts** (vLLM pod with no `ownerReferences` path to the HPA-scaled target): the label and relabeling rule remain required. `docs/design/controller-behavior.md` was updated to call this out.

## Test plan

- [x] `go build ./...` — clean.
- [x] `go test $(go list ./... | grep -v test/e2e) -count=1` — 36 packages pass.
- [x] `make lint` — 0 issues.
- [x] `make manifests` — produces no diff (Pod/ReplicaSet RBAC was already covered by VA reconciler markers; locator's `+kubebuilder:rbac` markers are documentation).
- [ ] **Reviewer**: walk a managed Deployment + HPA scenario without the `llm-d.ai/variant` pod label and confirm `wva_desired_replicas` is emitted for that variant.
- [ ] **Reviewer**: walk a shadow-pod scenario with the label set and confirm fast-path resolution still works.
- [ ] **Reviewer**: confirm the `defaultCacheSize = 4096` choice is reasonable for our largest target clusters.
- [ ] e2e suite (`test/e2e`) requires a live cluster; not run in this PR.

## Notes for review

- 16 commits on top of the design + plan docs. Reviewable commit-by-commit; logical groupings:
  - Indexer refactor + new indexes (5 commits).
  - Locator package (5 commits): `walk.go`, `cache.go`, `locator.go`, plus an LWS test and a doc-comment fix.
  - Integration into the metrics collector + saturation engine (2 commits).
  - Docs / RBAC / lint (3 commits).
  - One dependency: `github.com/hashicorp/golang-lru/v2 v2.0.7`.
- The locator only relies on controller-runtime's cached client for HPA + ScaledObject (small fan-out, bounded by variant count). All Pod / ReplicaSet / Deployment / LWS reads go through `apiReader` — no informer is started for those kinds.


internal/collector/locator/walk.go
```diff
@@ -0,0 +1,126 @@
...
+		APIVersion: apiVersion,
+		Kind:       kind,
+		Name:       obj.GetName(),
+	}
+}
+
+// newTypedFor returns a freshly allocated typed object matching the
+// ownerReference's apiVersion/kind. Returns false for unknown kinds, which
+// signals walkOwnersUp to stop.
+func newTypedFor(owner *metav1.OwnerReference) (client.Object, bool) {
```
ev-shindin, line 112: **Blocker: LWS pods aren't reachable here.** LWS (sigs.k8s.io/lws v0.8.0) is built on StatefulSets — a pod's `Controller:true` owner is a per-group StatefulSet, and the StatefulSet is owned by the LeaderWorkerSet:
- leader pod: `Pod → StatefulSet(<lws>) → LeaderWorkerSet`
- worker pod: `Pod → StatefulSet(<lws>-<i>) → Pod(leader) → StatefulSet(<lws>) → LeaderWorkerSet`

`newTypedFor` only knows ReplicaSet / Deployment / LeaderWorkerSet, so a real LWS pod hits the StatefulSet, is treated as an unknown kind, and the walk stops at `[Pod]` → `Locate` returns nil → LWS still requires `llm-d.ai/variant`, the opposite of this PR's goal.

Please add `StatefulSet` (and `Pod`, for the worker chain) to `newTypedFor` + `constants`, so the walk continues `Pod → StatefulSet → LWS`. Quick confirm on a live LWS: `kubectl get pod <lws-pod> -o jsonpath='{.metadata.ownerReferences}'` — I expect `kind: StatefulSet`, not `LeaderWorkerSet`. If walking workers is intentionally out of scope, let's document that LWS workers still need the label.

internal/collector/locator/locator_test.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+}
+
+func TestLocate_LWSChain(t *testing.T) {
+	ns := testNamespace
+	lws := &lwsv1.LeaderWorkerSet{ObjectMeta: metav1.ObjectMeta{Name: "lws", Namespace: ns, UID: "uid-lws"}}
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: "p", Namespace: ns,
+			OwnerReferences: []metav1.OwnerReference{{
+				APIVersion: "leaderworkerset.x-k8s.io/v1", Kind: "LeaderWorkerSet",
```
ev-shindin, line 256: This test passes only because the pod's controller ownerReference is the `LeaderWorkerSet` directly — a shape the LWS controller never produces (LWS pods are owned by a per-group StatefulSet)

internal/collector/locator/locator.go
```diff
@@ -0,0 +1,214 @@
...
+	ref := autoscalingv2.CrossVersionObjectReference{
+		APIVersion: target.APIVersion,
+		Kind:       target.Kind,
+		Name:       target.Name,
+	}
+	hpa, err := indexers.FindHPAForScaleTarget(ctx, asClient(l.cached), ref, target.Namespace)
+	if err != nil {
+		return nil, err
+	}
+	so, err := indexers.FindSOForScaleTarget(ctx, asClient(l.cached), ref, target.Namespace)
```
ev-shindin: **Blocker: breaks HPA attribution when KEDA isn't installed.** `SetupIndexes(…, kedaEnabled)` skips registering the ScaledObject index when KEDA CRDs are absent (`indexers.go:54`), but this call runs unconditionally and `New` (line 57) is never told whether KEDA is on. A `MatchingFields` query on an unregistered index errors, and we return on it *after* already finding the HPA:

```go
hpa, _ := FindHPAForScaleTarget(...)  // found
so, err := FindSOForScaleTarget(...)  // errors: SO index not registered
if err != nil { return nil, err }     // throws away the HPA
```

So on a common HPA-only / no-KEDA cluster, every label-less pod resolution fails. Suggest: thread `kedaEnabled` into `New(cached, apiReader, kedaEnabled)` (engine.go:217 already has the flag available via main.go), and skip `FindSOForScaleTarget` when `!kedaEnabled`, returning the HPA result directly. A KEDA-absent unit test would lock it in.

ev-shindin (changes requested):
Thanks @lionelvillard ! I found some blockers - please look in.

internal/collector/locator/walk.go
lionelvillard: thanks! Fixed.

internal/collector/locator/locator_test.go
lionelvillard: test has been updated.

internal/collector/locator/locator.go
lionelvillard: Fixed in 2027e7b179f2cd572107750763f52f76b270a315

internal/collector/locator/locator.go
```diff
@@ -0,0 +1,241 @@
...
+// Exactly one of HPA / ScaledObject is non-nil on a successful Locate.
+type ManagedScaler struct {
+	HPA          *autoscalingv2.HorizontalPodAutoscaler
+	ScaledObject *kedav1alpha1.ScaledObject
+}
+
+// kedaEnabled records whether the KEDA ScaledObject CRD is installed. It is set
+// once at startup via SetKEDAEnabled before any locator is constructed. Defaults
+// to true so that ScaledObject lookups remain enabled unless explicitly disabled.
+var kedaEnabled atomic.Bool
```
ev-shindin, line 36: This is process-global mutable state: tests must `SetKEDAEnabled`/reset it (cross-test pollution risk), and `init(){ Store(true) }` means a forgotten `SetKEDAEnabled` silently re-enables SO lookups. Threading it through the constructor — `New(cached, apiReader, kedaEnabled)` — keeps the flag local to each locator and avoids the global + the default-true footgun. Non-blocking; flagging as a cleaner alternative.

internal/collector/locator/locator.go
```diff
@@ -0,0 +1,241 @@
...
+	if !annotations.IsManaged(so) {
+		return nil, nil
+	}
+	return so, nil
+}
+
+// asClient adapts a client.Reader into the client.Client expected by the
+// indexers package. The locator only ever performs reads; the index
+// helpers don't write.
+func asClient(r client.Reader) client.Client {
```
ev-shindin, line 234: `New(cached client.Reader, …)` actually requires a `client.Client` and this panics otherwise (line 240). The panic is unreachable in practice, but the signature advertises a contract it doesn't keep. Either take `cached client.Client` directly (what `mgr.GetClient()` returns) and drop `asClient`, or relax `FindHPAForScaleTarget`/`FindSOForScaleTarget` to accept `client.Reader` (they only `List`). Non-Blocking.

internal/collector/locator/walk.go
```diff
@@ -0,0 +1,136 @@
...
+	return chain, nil
+}
+
+// nodeOf converts a typed object plus its namespace into a chainNode.
+func nodeOf(obj client.Object, namespace string) chainNode {
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	apiVersion := gvk.GroupVersion().String()
+	kind := gvk.Kind
+	if kind == "" {
+		// fake clients sometimes leave TypeMeta empty; infer from concrete type.
```
ev-shindin, line 87: Minor: the comment says "fake clients sometimes leave TypeMeta empty," but controller-runtime's typed client leaves TypeMeta empty in production too — and now that `StatefulSet`/`Pod` are in the concrete-type switch, this fallback is load-bearing on the real LWS path, not just in tests. Worth rewording so nobody removes it thinking it's test-only.

ev-shindin (commented):
LGTM. There some small issue that worstwhile to fix. Non-blocking.

internal/collector/locator/locator.go
lionelvillard: I changed my mind on this one.. I'm fixing it.

ev-shindin (approved):

internal/collector/locator/locator.go
lionelvillard: will be addressed as a follow-up PR.

internal/collector/locator/locator.go
lionelvillard: cleanup to be addressed as a follow-up PR.

PR #1271, opened 2026-06-12
feat(config): default to V2 token-based saturation analyzer

ev-shindin:
> ⚠️ **Behavior change.** This flips the out-of-the-box saturation path from **V1 (percentage-based)** to **V2 (token-based)**. Applying the kustomize base or upgrading the Helm chart will switch existing installs from V1 to V2 scaling on next reconcile. This is intentional — V2 is the actively-developed path — but it's a behavioral change, not a no-op. Opting back to V1 is one edit (see *How V2 is selected*).

## What

Make the **V2 token-based saturation analyzer** the default by shipping the saturation ConfigMap with a non-empty `analyzers` list in both install paths:

```yaml
data:
  default: |
    analyzers:
      - name: saturation     # non-empty list → engine routes to V2 (GetAnalyzerName() == "saturation")
    kvCacheThreshold: 0.80
    queueLengthThreshold: 5
    kvSpareTrigger: 0.1
    queueSpareTrigger: 3
    enableLimiter: false
```

- **Kustomize:** `config/base/manager/saturation-scaling-configmap.yaml`
- **Helm:** `charts/workload-variant-autoscaler/templates/manager/wva-cp-configmap.yaml`
- **Docs:** `docs/developer-guide/saturation-scaling-config.md` — the *Default Configuration* section now documents V2 as the shipped default and how to revert to V1.

## Why

V2 is the actively-developed saturation path (token-based capacity vs V1's percentage-based). Previously the shipped default had no `analyzers`/`analyzerName`, so fresh installs ran V1. This makes V2 the out-of-the-box behavior while keeping V1 one edit away.

## How V2 is selected (and how to revert to V1)

The engine routes to V2 when `GetAnalyzerName()` returns `"saturation"`, which is true whenever the `analyzers` list is non-empty (or `analyzerName: saturation` is set). To fall back to **V1**, remove the `analyzers` list **and** leave `analyzerName` unset — both must be V1-clean, because `ApplyDefaults()` re-populates the list from `analyzerName: saturation` if it's present.

## e2e: keep existing specs on V1

The general e2e scaling specs were written and validated against V1. To avoid silently changing what they exercise, `BeforeSuite` (`test/e2e/suite_test.go`) reverts the deployed saturation ConfigMap to the V1 default right after the controller comes up:

- New `revertSaturationConfigToV1()` patches the live ConfigMap's `default` entry back to the V1 content; the controller hot-reloads the labeled ConfigMap (no restart).
- The **V2-specific** suites (`saturation_v2_test.go`, `saturation_analyzer_path_test.go`) set and restore their own config, so they are unaffected.

This keeps the kustomize/Helm default at V2 (the product behavior) while the existing suite continues to run the path it was designed for.

## Note on Helm

The Helm chart is deprecated. This change is applied to the chart **intentionally**, for parity with kustomize, because bencmarking still use Helm — and this is not as new chart feature work.

### DIFF
--- a/docs/developer-guide/saturation-scaling-config.md
+++ b/docs/developer-guide/saturation-scaling-config.md
@@ -44,7 +44,25 @@ These parameters apply when `analyzerName: "saturation"` is set or when the `ana
 
 ### Default Configuration
 
-The recommended values for the V1 (percentage-based) saturation analyzer are:
+As shipped (kustomize base and helm chart), the **default ConfigMap selects the V2 token-based analyzer**:
+
+```yaml
+analyzers:
+  - name: saturation
+kvCacheThreshold: 0.80
+queueLengthThreshold: 5
+kvSpareTrigger: 0.1
+queueSpareTrigger: 3
+enableLimiter: false
+```
+
+The non-empty `analyzers:` list routes the engine to V2 (`GetAnalyzerName() == "saturation"`).
+The V2 thresholds (`scaleUpThreshold` / `scaleDownBoundary`) are filled from their defaults
+(0.85 / 0.70) when omitted, and the `kvCacheThreshold` / `queueLengthThreshold` / spare-trigger
+values are retained because V2 still uses them for the per-replica saturation check.
+
+**To use the V1 (percentage-based) analyzer instead,** remove the `analyzers:` list and leave
+`analyzerName` unset:
 
 ```yaml
 kvCacheThreshold: 0.80
@@ -53,10 +71,11 @@ kvSpareTrigger: 0.1
 queueSpareTrigger: 3
 ```
 
-> **Important:** These V1 threshold values are **not hardcoded** in the analyzer code.
-> If the ConfigMap is missing or has no `default` entry, all V1 thresholds default to zero,
+> **Important:** The V1 threshold values are **not hardcoded** in the analyzer code.
+> If a V1 ConfigMap is missing or has no `default` entry, all V1 thresholds default to zero,
 > which will cause every replica to appear saturated and trigger continuous scale-up.
-> Always deploy the ConfigMap with a `default` entry containing valid thresholds.
+> Always deploy a V1 `default` entry containing valid thresholds. (V2 has safe hardcoded
+> threshold defaults, so it degrades gracefully.)
 
 ### How Scale-Up Triggers Work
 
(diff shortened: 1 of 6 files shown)


biranofer:
How will this synced with the helm deprecation ?

biranofer:
BeforeSuite patches the deployed ConfigMap to V1 maybe add in AfterSuite a restore to V2 ?

ev-shindin:
> How will this synced with the helm deprecation ?

Helm fix provided as part of this PR

docs/developer-guide/saturation-scaling-config.md
```diff
@@ -44,7 +44,25 @@ These parameters apply when `analyzerName: "saturation"` is set or when the `ana
...
+
+```yaml
+analyzers:
+  - name: saturation
+kvCacheThreshold: 0.80
+queueLengthThreshold: 5
+kvSpareTrigger: 0.1
+queueSpareTrigger: 3
+enableLimiter: false
+```
```
deanlorenz, line 63: suggest adding the default scaleupthreshold and scaledownboundary as comments:
```yaml
analyzers:
  - name: saturation
  # Per-analyzer threshold overrides (optional):
  # scaleUpThreshold: 0.85    # default
  # scaleDownBoundary: 0.70   # default
# Global threshold overrides (apply to all analyzers):
# scaleUpThreshold: 0.85
# scaleDownBoundary: 0.70
kvCacheThreshold: 0.80
...
```
(or could skip the global overrides for now).

deanlorenz (changes requested):
missing unit test for yaml threshold configuration, which now become active. 
(missed these on the threshold per analyzer PR)

> Two small additions, both following existing patterns in the codebase:
> 
> saturation_scaling_test.go — add a Context("YAML round-trip") block (~30-40 lines):
> 
> Parse V2 shipped default (analyzers:\n - name: saturation) → verify ApplyDefaults() fills 0.85/0.70
> Parse top-level scaleUpThreshold: 0.90 → verify it survives unmarshal
> Parse per-analyzer analyzers: [{name: saturation, scaleUpThreshold: 0.90}] → verify *float64 pointer is set
> configmap_reconciler_test.go — extend one existing test case with V2 YAML (analyzers: ... + scaleUpThreshold) to verify the end-to-end parse path through parseSaturationConfig (~15-20 lines).

deanlorenz (approved):

lionelvillard:
/lgtm

@ev-shindin please make sure to update the user-facing documentation. Also it would be good to explain, at the user-level, what are the expected behavioral changes. For instance, is v2 more cost-effective than v1? Is yes, for which workload? Users would want to know what they are getting into.

PR #1274, opened 2026-06-15
feat(saturation): log resolved V2 thresholds + flag non-positive values

ev-shindin:
Scale decisions on the V2 path are driven by the post-recalibration capacity from applyUniversalThreshold(scaleUp, scaleDown), but the only existing log ("V2 saturation analysis completed") is emitted BEFORE that step, so it shows the analyzer's raw output and never the resolved thresholds. When a scale-down silently does not happen, there is no way to tell from the logs whether the analyzer saw no spare or the resolved scaleDownBoundary was non-positive (which makes applyUniversalThreshold skip the scale-down branch).

Add an INFO line right after the saturation recalibration with the resolved scaleUpThreshold / scaleDownBoundary and the post-threshold requiredCapacity / spareCapacity that actually drive the decision, plus an explicit warning when a threshold is non-positive (a misconfig, or a config not recognized as V2 when defaults were applied, so the thresholds were never filled). This makes V2 scale-decision flakes diagnosable from a single log line.

### DIFF
--- a/internal/engines/saturation/engine_v2.go
+++ b/internal/engines/saturation/engine_v2.go
@@ -109,6 +109,31 @@ func (e *Engine) runAnalyzersAndScore(
 	satUp, satDown := resolveThresholds(interfaces.SaturationAnalyzerName, config)
 	applyUniversalThreshold(baseResult, satUp, satDown)
 
+	// Log the resolved thresholds together with the post-recalibration capacity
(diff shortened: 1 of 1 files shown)


ev-shindin:
Part of this PR is already closed by #1306 and threshold warnings are not required - they are validate during admission time

PR #1275, opened 2026-06-15
feat(collector): resolve VA attribution behind an Attributor seam

### LINKED ISSUE #1263: refactor(collector): remove llm_d_ai_variant from PromQL query groupbys; move VA attribution fully to collector layer
## Summary

Separate **VA attribution** from the **metric query**. The per-replica PromQL queries should carry only replica identity (`instance`, `pod`); which `VariantAutoscaling` owns a replica is a collector-layer decision resolved after the query — from the selector-derived podMap (PR #1260) and/or the `llm-d.ai/variant` pod label. The query form is then identical regardless of the attribution mechanism.

Best delivered as additions to PR #1260 (`feat/pod-mapping-derivation`), since it builds directly on that PR's podMap. A concrete code sketch is in the comment below.

## Problem

All saturation and queueing-model per-replica queries currently use `max by (instance, pod, llm_d_ai_variant)`. The `llm_d_ai_variant` term in the `by (...)` clause does not filter selection — it only carries the label into the result so Go code can read it for attribution. This tangles two orthogonal concerns:

- **Replica identity** (`instance`, `pod`) — which scrape target the series belongs to.
- **VA attribution** (`llm_d_ai_variant`) — which VA owns it; a collector concern, not a query concern.

It also creates a scoping gap: two VAs in the same namespace with the same `model_id` (e.g. TP2 vs TP4 of one model) are only distinguishable in the result by `llm_d_ai_variant`. The podMap is the correct disambiguation path. See also #1072.

## Proposed change

1. `internal/collector/registration/saturation.go` (6 queries) and `queueing_model.go` (2 queries) — drop `llm_d_ai_variant` from every per-replica `max by`. Model-level queries (scheduler-queue, scale-to-zero) are unaffected; throughput queries are handled separately in #1250.
2. `internal/collector/replica_metrics.go` — `buildInstanceKey` returns `(instanceKey, podName)` only (no label read); remove `vaName` from `podMetricData`; resolve `vaName` once per pod entry in the final assembly loop. #1260 already provides the podMap fallback that makes this work — the change consolidates it into a single attribution step.

No behaviour change for standard Deployment/LWS deployments: attribution moves from the Prometheus label to the podMap, which covers them by selector.

## `llm-d.ai/variant` override — read from the pod, not Prometheus

Once the label is out of the query result, the override (for custom kinds / non-standard owner chains, and back-compat) must be read from the **K8s pod object**. `podvamap.Build` only lists pods via each VA's `scaleTargetRef` selector, so a custom-kind pod that no selector matches is never listed — exactly the case the override exists for. So `Build` adds **one bounded labeled LIST per namespace** (`HasLabels{llm-d.ai/variant}`) and applies the result as an override that wins over the selector-derived mapping, matching #1260's current label-wins semantics. This also drops the ServiceMonitor relabel-rule requirement for the override path. See the code sketch below.

## References

- PR #1260 — podMap derivation; provides the attribution path.
- PR #1121 (merged) — established the `(instance, pod)` base pattern.
- Issue #1072 — InferencePool-selector scoping; this is a step toward it.
- Issue #1133 — LWS multi-pod over-counting; same collector layer.

deanlorenz:
Implements #1263. Moves VariantAutoscaling (VA) attribution out of the PromQL queries and behind an `Attributor` seam resolved once per pod, after the query.

## What changed

- **Per-replica queries carry replica identity only.** Drops `llm_d_ai_variant` from all 11 per-replica query templates (saturation ×6, queueing-model ×2, throughput-analyzer ×3). The 3 throughput queries additionally gain `instance` so they group by `(instance, pod)` like the rest — this also fixes a latent key-mismatch where a KV-cache sample and a generation-token-rate sample for the same replica produced two `ReplicaMetrics` entries (the throughput one dropped); they now merge into one.
- **New `internal/collector/attribution` package** — an `Attributor` interface (`VAForPod`) plus a default label attributor that performs one bounded, labeled pod `List` per namespace via the uncached API reader (`mgr.GetAPIReader()`, so no Pod informer/cache is started) and maps pod → VA from the `llm-d.ai/variant` **pod-object** label. The ServiceMonitor/PodMonitor relabel rule is **no longer required for attribution** (kept optional for back-compat; the metric label and its constant remain).
- **Degraded-attribution handling** — on a `List` failure the builder returns a partial attributor plus a joined error; the engine logs once per cycle and proceeds with whatever resolved. Unresolved pods are skipped at the existing gate (demoted to DEBUG).
- **`UnattributedReadyPods` warning event** — a per-VA K8s Warning event when a VA's scale target has Ready pods but none are attributed this cycle (fires only when the model produced at least one attributed replica, so scrape-lag windows stay quiet; deduped per cycle). Surfaces a missing/wrong pod label directly in `kubectl describe va`.

Future attribution mechanisms (e.g. the #1267 owner-walk locator) plug in behind the seam without touching the queries, the collector hot path, or any analyzer.

## Docs

- `docs/developer-guide/variant-attribution.md` (new) — the seam, the default label attributor, how to add a mechanism.
- `docs/design/controller-behavior.md` — attribution reads the pod-object label; relabel rule reframed as optional; `UnattributedReadyPods` added to troubleshooting.

## Tests

Attribution unit tests (resolve/miss/nil-safe/empty; builder error + partial-map); the KV+throughput merge regression guard; `UnattributedReadyPods` positive + two negatives (ready==0, model-wide empty).

Fixes #1263

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -455,6 +455,9 @@ func main() {
 			sourceRegistry,
 			cfg, // Pass unified Config to engine
 		)
+		// Uncached reader for the per-cycle variant-label pod LIST used by VA
+		// attribution, so no Pod informer/cache is started for it.
+		engine.APIReader = mgr.GetAPIReader()
 		go engine.StartOptimizeLoop(ctx)
 		return nil
 	}))
--- a/docs/design/controller-behavior.md
+++ b/docs/design/controller-behavior.md
@@ -245,12 +245,13 @@ This periodic reconciliation is why many Update and Delete events can be safely 
 
 ### `llm-d.ai/variant` Label on the Scale Target
 
-WVA identifies which pods belong to a given `VariantAutoscaling` resource by reading the `llm-d.ai/variant` label from Prometheus metrics. Two things must be true for this to work:
+WVA identifies which pods belong to a given `VariantAutoscaling` resource by reading the `llm-d.ai/variant` label **directly from the pod objects**. Once per optimization cycle WVA lists the variant-labeled pods in the namespace and maps each pod to its `VariantAutoscaling` by the label value. Per-replica metrics are collected keyed by replica identity (`instance`, `pod`) and attributed to a VA in this separate step — attribution is not carried in the metric labels.
+
+For this to work:
 
 1. The `llm-d.ai/variant` label must be present on the **pod template** of the scale target (Deployment or LeaderWorkerSet), with a value equal to the name of the corresponding `VariantAutoscaling` resource.
-2. The `ServiceMonitor` or `PodMonitor` that Prometheus uses to scrape those pods must include a target relabeling rule that propagates the pod label into the scraped metrics as `llm_d_ai_variant`.
 
-**Both are your responsibility.** WVA does not configure either automatically.
+**This is your responsibility.** WVA does not configure it automatically. The `ServiceMonitor`/`PodMonitor` relabeling rule that propagates the label into the metric series (described below) is **no longer required for attribution**; it is optional and only useful if you want to filter your own metric queries or dashboards by variant.
 
 #### 1. Set the pod template label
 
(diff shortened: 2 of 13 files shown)


deanlorenz:
Closing — superseded by #1267, which makes `llm-d.ai/variant` optional for Deployment/LWS via an ownerReference-walk PodLocator (retaining the label as the fast path and for shadow-pod resolution). The remaining additive pieces are handled elsewhere: the throughput key-mismatch normalization is carried by #1250, and the `UnattributedReadyPods` diagnostic event will be folded into the #1250 rebase.

PR #1277, opened 2026-06-15
engine: emit saturation cycle summary log per model per cycle

deanlorenz:
## Summary

- Adds one permanent `"saturation cycle summary"` INFO log line per model per reconcile cycle, emitted after the V2 optimizer produces decisions
- Records model-level supply/demand/utilization, per-analyzer RC/SC signals (shows whether TA's RC was positive when sat_v2's was zero), and per-variant k1/k2/k2Source/cost/prc/efficiency/replicas/action — all in a single grep-friendly line
- No behavioral change; no config flags; existing LEVEL(-4) lines untouched

**Motivation:** currently there is no way to read per-cycle analyzer/optimizer outputs without scraping `--v=4` debug lines. This line is always-on at INFO level and is the foundation for the benchmark correctness analysis tooling.

**Implementation:**
- `saturation_v2.ReplicaCapacity` gains `K2Priority int`; `computeK2` returns `(int64, int)` so the call site can record which priority path fired (observed/history/derived/fallback)
- `k1Slice`, `k2Slice`, `k2SourceLabel` helpers added; `aggregateByVariant` now populates `MedianK1`, `MedianK2`, `K2SourceLabel` per variant
- `interfaces.VariantCapacity` gains `MedianK1`, `MedianK2`, `K2SourceLabel` (zero for non-saturation analyzers)
- `logDecisionSummary` package-level helper in `engine_v2.go`; called from `optimizeV2` after `optimizer.Optimize` returns

## Test plan

- [ ] `engine_v2_log_test.go` — three new tests using `zaptest/observer`: required keys present, skips model with no saturation result, `analyzerSignals` populated with multiple analyzers
- [ ] `make test` passes
- [ ] `make lint` clean
- [ ] `go build ./...` clean
- [ ] Verify log line appears in `kubectl logs` on a running controller: `kubectl logs -l app.kubernetes.io/name=workload-variant-autoscaler | grep '"saturation cycle summary"'`

### DIFF
--- a/internal/engines/analyzers/saturation_v2/analyzer.go
+++ b/internal/engines/analyzers/saturation_v2/analyzer.go
@@ -168,7 +168,7 @@ func (a *SaturationAnalyzer) computeReplicaCapacity(
 	if rec := a.capacityStore.Get(namespace, modelID, rm.VariantName); rec != nil {
 		vllmParams = rec.VLLMParams
 	}
-	k2 := a.computeK2(
+	k2, k2Priority := a.computeK2(
 		modelID, rm.AcceleratorName,
 		gpuCount,
 		rm.QueueLength, rm.TokensInUse,
@@ -210,6 +210,7 @@ func (a *SaturationAnalyzer) computeReplicaCapacity(
 		TotalKvCapacityTokens: rm.TotalKvCapacityTokens,
 		MemoryBoundCapacity:   k1,
 		ComputeBoundCapacity:  k2,
+		K2Priority:            k2Priority,
 		EffectiveCapacity:     effectiveCapacity,
 		IsSaturated:           isSaturated,
 		ReplicaDemand:         replicaDemand,
@@ -270,6 +271,7 @@ func (a *SaturationAnalyzer) computeReplicaCapacityFallback(
 // 2. Historical → rolling average from previous observations
 // 3. Derived (from deployment args) → formula-based estimate
 // 4. Fallback → k1 (memory-bound only)
+// Returns the k2 value and the priority level (1–4) that produced it.
 func (a *SaturationAnalyzer) computeK2(
 	modelID, accelerator string,
 	gpuCount int,
@@ -278,7 +280,7 @@ func (a *SaturationAnalyzer) computeK2(
 	queueThreshold float64,
 	vllmParams *VLLMEngineParams,
 	k1 int64,
-) int64 {
+) (int64, int) {
 	outputBucket := classifyOutputLength(avgOutput)
 	historyKey := fmt.Sprintf("%s|%s|%d|%s", modelID, accelerator, gpuCount, outputBucket)
 
@@ -293,7 +295,7 @@ func (a *SaturationAnalyzer) computeK2(
 		}
 		ra.Add(float64(k2Observed))
 		a.mu.Unlock()
-		return k2Observed
+		return k2Observed, 1
 	}
 
 	// Priority 2: Historical — lock must cover Average() since Add() mutates
@@ -305,16 +307,16 @@ func (a *SaturationAnalyzer) computeK2(
 	}
 	a.mu.Unlock()
 	if histAvg > 0 {
-		return int64(histAvg)
+		return int64(histAvg), 2
 	}
 
 	// Priority 3: Derived from deployment args
 	if k2Derived := estimateCapacityFromParams(vllmParams, avgInput, avgOutput); k2Derived > 0 {
-		return k2Derived
+		return k2Derived, 3
 	}
 
 	// Priority 4: Fallback to k1
-	return k1
+	return k1, 4
 }
 
 // aggregateByVariant groups replica capacities by variant and computes
@@ -360,6 +362,9 @@ func (a *SaturationAnalyzer) aggregateByVariant(
 			readyCount = 0
 		}
 
+		var medianK1, medianK2 int64
+		var k2Source string
+
 		if len(replicas) > 0 {
 			// Use median effective capacity from ready pods
 			capacities := make([]int64, 0, len(replicas))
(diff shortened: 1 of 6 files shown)


internal/interfaces/analyzer.go
ev-shindin, line 1: Changed to the interface is specific to saturation analyzer. I think we need to find cleaner solution.

internal/engines/analyzers/saturation_v2/analyzer.go
```diff
@@ -636,6 +647,49 @@ func estimateSchedulerQueueDemand(
...
+	if len(replicas) == 0 {
+		return ""
+	}
+	effs := make([]int64, len(replicas))
+	for i, r := range replicas {
+		effs[i] = r.EffectiveCapacity
+	}
+	medEff := median(effs)
+	k2PriorityLabels := map[int]string{1: "P1-obs", 2: "P2-hist", 3: "P3-deriv", 4: "P4-k1"}
+	for _, r := range replicas {
```
ev-shindin: Bug for **even** replica counts. `median()`  averages the two middle elements for even length — `(sorted[n/2-1] + sorted[n/2]) / 2` — but here you equality-match a replica against that median:
With 2/4/6 ready replicas whose `EffectiveCapacity` differs (e.g. `[100, 200]` → `medEff = 150`), **no replica equals `150`**, so this returns `""` — the new `k2Source` diagnostic is silently blank exactly in the common heterogeneous-replica case. (The "same replica that determines PerReplicaCapacity" comment is also off for even counts — `PerReplicaCapacity` is itself an averaged median, not any single replica's value.)

ev-shindin (changes requested):
Please address review commenys.

internal/engines/analyzers/saturation_v2/analyzer.go
```diff
@@ -371,6 +376,12 @@ func (a *SaturationAnalyzer) aggregateByVariant(
 			if accelerator == "" {
 				accelerator = replicas[0].AcceleratorName
 			}
+			satVCs = append(satVCs, interfaces.SaturationVariantCapacity{
```
ev-shindin, line 379: These three fields are each computed over a different ordering, so together they don't characterize one coherent representative replica:

- `MedianK1 = median(k1Slice(...))` — sorted by **K1**; `median()` (line 667) **averages the two middle values** for even counts.
- `MedianK2 = median(k2Slice(...))` — sorted by **K2**.
- `K2SourceLabel = k2SourceLabel(...)` — sorted by **EffectiveCapacity** (lines 649/659), returns the lower-median *element's* `K2Priority`.

With an even/heterogeneous replica set you can get e.g. `MedianK2=150` (the average of two replicas' K2) next to `K2SourceLabel="P1-obs"` taken from a replica chosen by a third key whose K2 is neither 150 nor near it.

internal/engines/analyzers/saturation_v2/analyzer.go
```diff
@@ -293,7 +296,7 @@ func (a *SaturationAnalyzer) computeK2(
 		}
 		ra.Add(float64(k2Observed))
 		a.mu.Unlock()
-		return k2Observed
+		return k2Observed, 1
```
ev-shindin, line 299: The 1–4 codes appear as bare returns in `computeK2` (`return k2Observed, 1` … `return k1, 4`), as a literal `map[int]string{1:"P1-obs",…}` in `k2SourceLabel` (line 660), and as a comment on the `K2Priority` field. Three sources of truth that can drift. Suggest a small named type + consts.

internal/engines/analyzers/saturation_v2/analyzer.go
```diff
@@ -305,27 +308,28 @@ func (a *SaturationAnalyzer) computeK2(
 	}
 	a.mu.Unlock()
 	if histAvg > 0 {
-		return int64(histAvg)
+		return int64(histAvg), 2
```
ev-shindin, line 311: see comment to line 299

internal/engines/analyzers/saturation_v2/analyzer.go
```diff
@@ -305,27 +308,28 @@ func (a *SaturationAnalyzer) computeK2(
...
 	}
 
 	// Priority 3: Derived from deployment args
 	if k2Derived := estimateCapacityFromParams(vllmParams, avgInput, avgOutput); k2Derived > 0 {
-		return k2Derived
+		return k2Derived, 3
```
ev-shindin, line 316: see comment to line 299

internal/engines/analyzers/saturation_v2/analyzer.go
```diff
@@ -305,27 +308,28 @@ func (a *SaturationAnalyzer) computeK2(
...
 	}
 
 	// Priority 4: Fallback to k1
-	return k1
+	return k1, 4
```
ev-shindin, line 320: see comment to line 299

internal/engines/analyzers/saturation_v2/analyzer.go
```diff
@@ -278,7 +281,7 @@ func (a *SaturationAnalyzer) computeK2(
 	queueThreshold float64,
 	vllmParams *VLLMEngineParams,
 	k1 int64,
-) int64 {
+) (int64, int) {
```
ev-shindin, line 284: Signature now returns `(int64, int)`. Please confirm every caller consumes the second (priority) return into `ReplicaCapacity.K2Priority` — a caller that drops it leaves `K2Priority=0`, which maps to no label (blank) in `k2SourceLabel`.

ev-shindin (changes requested):
Thanks @deanlorenz ! Please address review comments

deanlorenz:
Superseded by #1318, which contains a complete redesign of the logging approach.

Issue #992, opened 2026-04-13
[LWS] Detect LWS CRD whenever it's available - not just when controller starts


shuynh2017:

asm582:
@shuynh2017 do we have any updates on this issue?

shuynh2017:
@asm582 thanks for reminder, I should be able to get to this soon.

PR #1278, opened 2026-06-15
Improvement: Dynamic detect LWS CRD

shuynh2017:
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/992
- Before the fix:
  - start manager -> install lws CRD -> create lws -> create VA with lws as scaleRef -> Error in WVA manager "Failed to get scale target".
- After the fix:
  - start manager -> install lws CRD -> create lws -> create VA with lws as scaleRef -> "Target found".
- The same pattern can be applied KEDA

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -32,6 +32,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	flag "github.com/spf13/pflag"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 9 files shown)


lionelvillard:
@shuynh2017 is there an issue corresponding to this PR? This is not a common k8s pattern to support adding CRD after informers have started.

shuynh2017:
@lionelvillard the only issue is in the ref ^^

shuynh2017:
@lionelvillard, can you point to the preferred way? I can take a look. Thanks.

lionelvillard:
The preferred way is to restart controllers. Adding/Removing CRDs are rare enough events to justify dynamic detection.

shuynh2017:
@lionelvillard ok, less code. Thanks.

PR #1280, opened 2026-06-16
refactor: replace fragile string matching in sanitizeReasonForMetrics

### LINKED ISSUE #1262: refactor: sanitizeReasonForMetrics function in engine.go
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1114#pullrequestreview-4477897382
```
// sanitizeReasonForMetrics converts detailed reason strings into bounded categorical
// values safe for use as Prometheus label values. This prevents cardinality explosion
// from dynamic numeric values embedded in reason strings.
// TODO: look fragile, refactor this function for better solution
func sanitizeReasonForMetrics(reason string, action interfaces.SaturationAction) string {
```

Goutham-Annem:
## What this PR does

`sanitizeReasonForMetrics` converts raw decision reason strings into bounded Prometheus label values to prevent cardinality explosion. It previously used `strings.Contains` on free-form reason strings set across three different packages. Renaming any reason string in a producer would silently break metric categorization — no compile error, no failing test.

This PR replaces that approach:

- **Saturation-only path**: uses `decision.SaturationOnly` (an existing struct field set at the decision source) — no string matching at all
- **Scale-from-zero and V2 paths**: defined named prefix constants (`scalingReasonPrefixScaleFromZero`, `scalingReasonPrefixV2`) as a single source of truth, and switched from `strings.Contains` to `strings.HasPrefix` for precision
- **Tests**: updated `sanitize_reason_test.go` to drive the function via `*VariantDecision` and added a precedence test confirming the `SaturationOnly` flag beats any reason prefix

## Why

Closes #1262

The existing `// TODO: look fragile, refactor this function for better solution` comment (added in PR #1114) identified this as a known debt. The classification now relies on structured data where it exists, and on stable prefixes defined as constants where it does not.

## How tested

- `go test ./internal/engines/saturation/... -run TestSanitizeReasonForMetrics -v` — all 16 cases pass including cardinality bound check
- `go build ./...` — clean
- `go vet ./internal/engines/saturation/...` — clean

### DIFF
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -1591,7 +1591,7 @@ func (e *Engine) applySaturationDecisions(
 		if hasDecision {
 			if decision.Action != interfaces.ActionNoChange {
 				// Sanitize reason to prevent Prometheus cardinality explosion from dynamic numeric values
-				sanitizedReason := sanitizeReasonForMetrics(reason, decision.Action)
+				sanitizedReason := sanitizeReasonForMetrics(&decision)
 				if err := e.metricsEmitter.EmitReplicaScalingMetrics(ctx, &updateVa, string(decision.Action), sanitizedReason); err != nil {
 					logger.Error(err, "Failed to emit replica scaling metrics")
 				}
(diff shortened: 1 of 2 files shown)


ev-shindin:
Thanks @Goutham-Annem — you correctly identified the fragility in
`sanitizeReasonForMetrics` (the `strings.Contains` matching on free-form reason
strings). #1262 ended up being fixed by #1282, which took the more thorough route:
it replaces the free-form reason strings with a typed `DecisionReason` enum and
removes `sanitizeReasonForMetrics` (and its test) altogether, so the metric `reason`
label is bounded by construction with no string matching at all.

Since #1282 is now merged I'm closing this as superseded by #1282 — but the underlying call was a good one.
Appreciate the contribution! 🙏

PR #1282, opened 2026-06-16
Refactor brittle metric emit code

shuynh2017:
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1262

The goal is to prevent cardinality explosion from dynamic numeric values embedded in `reason` strings function `EmitReplicaScalingMetrics` in `metrics.go` (same for other functions). To address this, the main idea is for `EmitReplicaScalingMetrics` to accept `type` defined as `string` instead of `string`, and control the way `reason` string is built. So the main idea is: 
```
func (m *MetricsEmitter) EmitReplicaScalingMetrics(ctx context.Context, va *llmdOptv1alpha1.VariantAutoscaling, 
direction interfaces.SaturationAction,   <=== type
reason interfaces.DecisionReason <=== type
)
```
```
const (
	// DecisionReasonV2 indicates a V2 pipeline decision.
	DecisionReasonV2 DecisionReason = "V2"
	// DecisionReasonKVCacheUtilAboveThreshold indicates KV cache saturation triggered scale-up.
	DecisionReasonKVCacheUtilAboveThreshold DecisionReason = "KV cache utilization above threshold"
	// DecisionReasonKVCacheUtilBelowThreshold indicates low KV cache usage triggered scale-down.
	DecisionReasonKVCacheUtilBelowThreshold DecisionReason = "KV cache utilization below threshold"
	// DecisionReasonNoRequestsInRetention indicates scale-to-zero due to no recent requests.
	DecisionReasonNoRequestsInRetention DecisionReason = "No requests in retention period"
	// DecisionReasonNoScalingNeeded indicates no action required.
	DecisionReasonNoScalingNeeded DecisionReason = "No scaling needed"
	// DecisionReasonSaturationOnly indicates decision from saturation-only mode.
	DecisionReasonSaturationOnly DecisionReason = "saturation-only mode"
	// DecisionReasonScaleFromZero indicates scale-up from zero replicas.
	DecisionReasonScaleFromZero DecisionReason = "scale-from-zero"
	// DecisionReasonTest is used for test scenarios.
	DecisionReasonTest DecisionReason = "test"
)
```

and everything else is derived from this. Same pattern can/should be used for other functions in `metrics.go`.


internal/interfaces/saturation_analyzer.go
```diff
@@ -318,6 +350,24 @@ func (d *VariantDecision) LastStep() *DecisionStep {
 	return &d.DecisionSteps[len(d.DecisionSteps)-1]
 }
 
+// SetDecisionReason sets both the typed reason category and detailed reason string.
+// The decisionReason should be one of the DecisionReason* constants.
+// The detailedReason provides human-readable context for logs, events, and status.
+func (d *VariantDecision) SetDecisionReason(action SaturationAction, decisionReason DecisionReason, detailedReason string) {
```
ev-shindin, line 349: `action` is accepted but never used — the body only sets `decisionReason` and `reason`.
The signature implies it sets `d.Action`, but callers set that separately. Please either
drop the parameter or wire it up (`d.Action = action`). Wiring it up would also fix
scalefromzero passing `ActionScaleUp` here while leaving `d.Action == ""`.

internal/metrics/metrics.go
```diff
@@ -432,12 +433,12 @@ func NewMetricsEmitter() *MetricsEmitter {
...
 // EmitReplicaScalingMetrics emits metrics related to replica scaling
-func (m *MetricsEmitter) EmitReplicaScalingMetrics(ctx context.Context, va *llmdOptv1alpha1.VariantAutoscaling, direction, reason string) error {
+func (m *MetricsEmitter) EmitReplicaScalingMetrics(ctx context.Context, va *llmdOptv1alpha1.VariantAutoscaling, direction interfaces.SaturationAction, reason interfaces.DecisionReason) error {
 	labels := prometheus.Labels{
 		constants.LabelVariantName: va.Name,
 		constants.LabelNamespace:   va.Namespace,
-		constants.LabelDirection:   direction,
-		constants.LabelReason:      reason,
+		constants.LabelDirection:   string(direction),
+		constants.LabelReason:      fmt.Sprintf("%s: %s", reason, direction),
```
ev-shindin: `direction` is already emitted as its own label (LabelDirection, line 440), so this
produces reason="V2: scale-up" alongside direction="scale-up". The `: direction` suffix
adds no information and inflates the reason label's value count — counter to this PR's goal.
Suggest: `constants.LabelReason: string(reason)`.

internal/interfaces/saturation_analyzer_test.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"testing"
+)
+
+func TestVariantDecision_SetDecisionReason(t *testing.T) {
+	tests := []struct {
+		name           string
+		action         SaturationAction
+		reasonType     DecisionReason
+		detailedReason string
```
ev-shindin: Stale reference: `VariantDecision.ReasonType` doesn't exist (it's the unexported
`decisionReason`, read via `DecisionReason()`). Please point the comment at
`DecisionReason()` / `Reason()`.

internal/metrics/metrics.go
```diff
@@ -432,12 +433,12 @@ func NewMetricsEmitter() *MetricsEmitter {
 }
 
 // EmitReplicaScalingMetrics emits metrics related to replica scaling
-func (m *MetricsEmitter) EmitReplicaScalingMetrics(ctx context.Context, va *llmdOptv1alpha1.VariantAutoscaling, direction, reason string) error {
+func (m *MetricsEmitter) EmitReplicaScalingMetrics(ctx context.Context, va *llmdOptv1alpha1.VariantAutoscaling, direction interfaces.SaturationAction, reason interfaces.DecisionReason) error {
```
ev-shindin, line 436: This is the contract that replaced the deleted sanitize_reason_test.go (~200 lines), but
nothing now asserts the emitted `reason` label. Please add a metrics test that calls this
with (ActionScaleUp, DecisionReasonV2), gathers the registry, and asserts the concrete
`reason` label string.

internal/interfaces/saturation_analyzer.go
```diff
@@ -8,6 +8,31 @@ import (
...
+// Used to set both VariantDecision.ReasonType (typed category) and
+// VariantDecision.Reason (human-readable detail) via SetDecisionReason.
+type DecisionReason string
+
+// Decision reason type constants for categorizing scaling decisions.
+const (
+	// DecisionReasonV2 indicates a V2 pipeline decision.
+	DecisionReasonV2 DecisionReason = "V2"
+	// DecisionReasonKVCacheUtilAboveThreshold indicates KV cache saturation triggered scale-up.
+	DecisionReasonKVCacheUtilAboveThreshold DecisionReason = "KV cache utilization above threshold"
```
ev-shindin: `DecisionReasonKVCacheUtilAboveThreshold`, `...BelowThreshold`,
`DecisionReasonNoRequestsInRetention`, and `DecisionReasonNoScalingNeeded` are only used
from tests — no production path sets them (the V1 path uses the generic
`DecisionReasonSaturationOnly`). Either wire them into the V1 KV-threshold paths or remove
them as dead.

internal/engines/scalefromzero/engine.go
```diff
@@ -357,11 +357,12 @@ func (e *Engine) processInactiveVariant(ctx context.Context, scaleTargets map[st
 			SafetyOverride:     false,
 			ModelBasedDecision: false,
 			AcceleratorName:    accelerator,
-			Reason:             reason, // Reason for scaling up
 			MetricsAvailable:   true,
 			MetricsReason:      MetricsReasonAvailable,
 			MetricsMessage:     MetricsMessageAvailable,
-		})
+		}
+		d.SetDecisionReason(interfaces.ActionScaleUp, interfaces.DecisionReasonScaleFromZero, string(interfaces.DecisionReasonScaleFromZero)+": pending request - scale-up")
```
ev-shindin: This detailed-reason string is duplicated here and at line 375, and the package-level
`reason` constant already holds the same value — reuse the constant instead of the literal.

internal/interfaces/saturation_analyzer.go
```diff
@@ -318,6 +350,24 @@ func (d *VariantDecision) LastStep() *DecisionStep {
...
+// SetDecisionReason sets both the typed reason category and detailed reason string.
+// The decisionReason should be one of the DecisionReason* constants.
+// The detailedReason provides human-readable context for logs, events, and status.
+func (d *VariantDecision) SetDecisionReason(action SaturationAction, decisionReason DecisionReason, detailedReason string) {
+	d.decisionReason = decisionReason
+	d.reason = detailedReason
+}
+
+// DecisionReason returns the categorized reason used for Prometheus metric labels.
+func (d *VariantDecision) DecisionReason() DecisionReason {
```
ev-shindin: Nit: the getter shares its name with the type `DecisionReason`, so `d.DecisionReason()`
returns an `interfaces.DecisionReason` — a bit confusing. Consider `ReasonCategory()`.

ev-shindin (changes requested):
Thanks @shuynh2017 — the typed `DecisionReason` approach is a real improvement over the string-sanitizing hack, and the cardinality goal is met. A few things I'd like addressed before merge.

internal/interfaces/saturation_analyzer.go
shuynh2017: yep. I will change to `GetDecisionReason` to be consistent with the `SetDecisionReason` function just above it.

internal/engines/scalefromzero/engine.go
shuynh2017: for this, I'd like to keep all `DecisionReason` in one place to make sure they are not from random strings (keep this a s a pattern string in metric). So I will form the string like this:
```
const (
reasonDetails           = ": pending request - scale-up"
)
string(interfaces.DecisionReasonScaleFromZero)+reasonDetails
```

internal/interfaces/saturation_analyzer.go
shuynh2017: I will wire it up. Thanks.

shuynh2017: fixed.

internal/metrics/metrics.go
shuynh2017: fixed.

internal/interfaces/saturation_analyzer_test.go
shuynh2017: fixed.

internal/metrics/metrics.go
shuynh2017: fixed.

internal/interfaces/saturation_analyzer.go
shuynh2017: fixed.

ev-shindin (approved):
Hi @shuynh2017 — thanks for the updates, this looks good. The cardinality goal is
solid: `reason` is now a bounded typed enum, `direction` is its own label, and every
decision path sets a non-empty reason. I reviewed the whole PR and only had small
follow-ups, so I pushed a single fixup commit to the branch to reduce the time/effort. 
(`2666797c`, fast-forward on top of your two commits — nothing rewritten). Please
review/squash as you like.

PR #1284, opened 2026-06-16
fix(annotations): propagate HPA/ScaledObject labels to synthetic VA

lionelvillard:
## Summary

- VariantAutoscalingFromHPA and VariantAutoscalingFromScaledObject were not copying metadata.labels from the source object to the synthetic VariantAutoscaling, causing GetAcceleratorNameFromScaleTarget to miss the inference.optimization/acceleratorName label fallback and emit AcceleratorNotResolved events even when the label was present on the HPA/ScaledObject.
- Fix: copy Labels from the source HPA/ScaledObject into the synthetic VA ObjectMeta.
- Reorganize config/samples/simulator/ into label/ and nodeSelector/ subdirectories to document both accelerator-resolution approaches side by side.

## Test plan

- [ ] Unit tests pass: `make test`
- [ ] Deploy HPA with `inference.optimization/acceleratorName` label and no nodeSelector — confirm no AcceleratorNotResolved events in controller logs
- [ ] Deploy HPA with nodeSelector (nodeSelector variant) — confirm accelerator still resolves via nodeSelector path


ev-shindin:
@lionelvillard do we need this fallback at all? What is the use-case? We already have discovery mechanism that should work fir any realistic use-case,

PR #1286, opened 2026-06-17
Revert "feat(config): default to V2 token-based saturation analyzer (…

lionelvillard:
…#1271)"

This reverts commit 04f95779e39bca3767a6c65a35aa529f2f4de284.

### DIFF
--- a/charts/workload-variant-autoscaler/templates/manager/wva-cp-configmap.yaml
+++ b/charts/workload-variant-autoscaler/templates/manager/wva-cp-configmap.yaml
@@ -23,12 +23,6 @@ metadata:
 data:
   # Global defaults applied to all variants unless overridden
   default: |
-    # Select the V2 token-based saturation analyzer. A non-empty analyzers list
-    # routes the engine to the V2 path (GetAnalyzerName() == "saturation").
-    # Remove this list (and leave analyzerName unset) to fall back to the V1
-    # percentage-based analyzer.
-    analyzers:
-      - name: saturation
     kvCacheThreshold: {{ .Values.wva.capacityScaling.default.kvCacheThreshold }}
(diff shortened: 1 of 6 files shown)


ev-shindin (approved):
Thanks @lionelvillard

asm582:
/lgtm
It would be good to create a test case to test the algorithm version.

hexfusion:
@lionelvillard for historical context could you outline why this PR was reverted?

ev-shindin:
@hexfusion this revert isn't a step back from V2 — it's about *how* the switch lands.

#1271 had made the V2 token-based analyzer the shipped default. Because V1 and V2 compute saturation/scaling differently, that would have silently changed the scaling behavior of existing deployments on upgrade, with no opt-in. We'd rather not change WVA's default behavior without advance notice, so we restored the V1 default for now.

V2 is still fully supported and is where we're headed — you can opt in today with a one-line change to the saturation-scaling ConfigMap's `default` entry:

```yaml
analyzers:
  - name: saturation   # selects the V2 token-based path
```

A short "enable V2" guide is coming, and we'll flip the default to V2 in a later release once that notice is out.

hexfusion:
so from a testing perspective we feel v2 is ready for default but we are holding off for formal documentation/guide. is that correct? @ev-shindin

lionelvillard:
@hexfusion we didn't enable v2 by default to avoid the blast radius effect while updating from 0.7.0 to 0.8.0, due to the fact that it is currently not possible to have both analyzers coexisting. 

The guide has been updated: https://github.com/llm-d/llm-d/blob/main/guides/workload-autoscaling/README.wva.md#enabling-saturation-engine-v2-recommended

PR #1287, opened 2026-06-17
Observability Dashboard: support selectable data sources

shuynh2017:
Why we need this fix:
- Without this fix, dashboard will work for OOB upstream clusters where there's only one Prometheus datasource called "prometheus". However, this will not work for downstream clusters where there may be multiple datasources which can be named differently. With this fix, Grafana will automatically show all its data sources, and users can select the appropriate datasource for WVA. The sreenshot below shows Grafana with 2 datasources:

image

Without this fix, Openshift users will have to manually change datasource in many places in the .json file defining the dashboard.

### DIFF
--- a/deploy/grafana/operational-dashboard.json
+++ b/deploy/grafana/operational-dashboard.json
@@ -1,4 +1,89 @@
 {
+  "templating": {
+    "list": [
+      {
+        "current": {
+          "selected": false,
+          "text": "prometheus",
+          "value": "prometheus"
+        },
+        "hide": 0,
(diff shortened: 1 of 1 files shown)


lionelvillard (approved):
we may want to consider moving this dashboard over to llm-d repo.

shuynh2017:
> we may want to consider moving this dashboard over to llm-d repo.

hmm, not sure. Metrics and dashboard should be close together, i.e., if someone creates/updates/adds/deletes metrics then searching wva should show all places need to be considered.

PR #1289, opened 2026-06-18
fix(benchmark): correct scenario path in benchmark-standup for v0.6.3

asm582:
## Summary

- In `llm-d-benchmark` v0.6.3, the WVA scenario file is named `workload-autoscaling.yaml`, not `inference-scheduling-wva.yaml`
- The `sed` patch and backup/restore in `benchmark-standup` were referencing the old (non-existent) filename, causing an immediate `No such file or directory` failure
- Also removes the stale TODO comment that noted this needed verification

## Test plan

- [ ] Run `make benchmark-standup BENCHMARK_NAMESPACE=<ns>` and confirm it no longer errors on the `sed` line
- [ ] Confirm `workload-autoscaling.yaml` is correctly patched with `PYTORCH_ALLOC_CONF` and restored after standup


kahilam (approved):
Tested with make benchmark-standup BENCHMARK_NAMESPACE=kahila-bench MODEL_ID=Qwen/Qwen3-32B.

Scenario path fix works — no sed errors.
workload-autoscaling.yaml correctly patched during standup and restored after.
Smoketests passed.

PR #1296, opened 2026-06-18
fix(collector): resolve VA CRD name from HPA scaleTargetRef in buildInstanceKey

### LINKED ISSUE #1290: bug(collector): PodLocator returns HPA name as variant name, causing name mismatch with CRD-sourced VariantAutoscalings
## Problem

When a `VariantAutoscaling` CRD and its associated HPA have **different names** (which is the case for KServe LLMInferenceService, where VA = `{name}-kserve-va` and HPA = `{name}-kserve-hpa`), the saturation engine collects zero usable metrics and reports `MetricsAvailable: False` / `MetricsMissing`.

This is a regression from v0.7.0 → v0.8.0-rc3. All WVA autoscaling e2e tests in kserve pass on v0.7.0 and fail on v0.8.0-rc3 with:

```
MetricsAvailable: False
reason: MetricsMissing
message: "No saturation metrics available - pods may not be ready or metrics not yet scraped"
```

## Root cause

In v0.7.0, `PodVAMapper.FindVAForPod()` walked ownerReferences (Pod → RS → Deployment) and then performed an **indexed lookup for the VariantAutoscaling CRD** whose `scaleTargetRef` pointed to that Deployment. It returned the **VA CRD name**, which matched the keys in the engine's `variantAutoscalings` map.

In v0.8.0-rc3, `buildInstanceKey()` (`replica_metrics.go`) replaced this with the `PodLocator`, which walks the same ownerReference chain but then looks up the **HPA** (or ScaledObject) whose `scaleTargetRef` matches the Deployment. It returns `vaName = ms.HPA.Name`, the **HPA name**, not the VA CRD name.

When the HPA name ≠ VA CRD name, the following chain of failures occurs:

1. `buildInstanceKey()` sets `vaName = HPA.Name` (e.g., `foo-kserve-hpa`)
2. Pod data is stored with `data.vaName = "foo-kserve-hpa"`
3. `vaName` is non-empty, so the pod is NOT skipped at the `if vaName == ""` guard
4. `ReplicaMetrics` is appended with `VariantName = "foo-kserve-hpa"`
5. `len(replicaMetrics) > 0`, so `prepareModelData` returns non-nil
6. `filterReplicaMetricsByVariants()` filters metrics against `VariantReplicaState` names derived from the VA CRDs (e.g., `foo-kserve-va`)
7. `"foo-kserve-hpa"` ∉ allowed set → **all metrics filtered out**
8. Analysis produces zero decisions → `desiredOptimizedAlloc` never set

## Who is affected

Any integration where the VA CRD and HPA have different `metadata.name` values. This includes KServe, which uses `{isvc}-kserve-va` / `{isvc}-kserve-hpa` naming (see kserve `pkg/controller/v1alpha2/llmisvc/scaling.go`).

KServe has not yet migrated to the annotation-based (`llm-d.ai/managed=true`) variant discovery path introduced in #1123. KServe still relies on the CRD-based dual-mode path where the controller creates both a `VariantAutoscaling` CR and an HPA as separate resources with distinct names. Moving KServe to annotation-based discovery is a non-trivial migration that cannot happen immediately, the CRD path must continue to work correctly for existing integrations.

Setups where VA name == HPA name are unaffected.

## Suggested fix

In `buildInstanceKey()`, the locator should resolve to the **VariantAutoscaling name** (or synthetic VA name), not the raw HPA/ScaledObject name. Options:

1. After `Locate()` returns an HPA, look up the VA CRD that created/targets the same `scaleTargetRef` as the HPA, and return that VA's name.
2. Restore the v0.7.0 behavior as a fallback: if `Locate()` finds an HPA but the HPA name doesn't match any key in `variantAutoscalings`, fall back to `FindVAForScaleTarget()` using the Deployment name.
3. In `filterReplicaMetricsByVariants`, match by `scaleTargetRef` rather than by exact variant name.

lionelvillard:
## Summary

- **Bug:** buildInstanceKey() used ms.HPA.Name as the variant name when the PodLocator resolved a pod to an HPA. For CRD-based setups like KServe where the VA CRD name ({isvc}-kserve-va) and HPA name ({isvc}-kserve-hpa) differ, all collected metrics were subsequently filtered out by filterReplicaMetricsByVariants because the HPA name was not in the VA-CRD-keyed allowed set.
- **Fix:** After Locate() returns a scaler, call indexers.FindVAForScaleTarget() with the scaler scaleTargetRef. If a VA CRD is found (CRD-based path), use its name. If not (annotation-based path, where synthetic VAs are keyed by scaler name), fall back to the HPA/ScaledObject name, preserving existing behavior.
- **Test:** Adds TestBuildInstanceKey_VACRDNameDiffersFromHPAName regression test exercising the exact KServe naming pattern (foo-kserve-va / foo-kserve-hpa), verified to fail before the fix and pass after.

Fixes #1290

## Test plan

- [x] TestBuildInstanceKey_VACRDNameDiffersFromHPAName fails on the buggy code with VariantName = foo-kserve-hpa; want foo-kserve-va, passes after the fix
- [x] go test ./internal/collector/... all collector tests pass
- [x] Pre-existing test failures in unrelated packages confirmed pre-existing, not introduced by this change

### DIFF
--- a/internal/collector/replica_metrics.go
+++ b/internal/collector/replica_metrics.go
@@ -58,6 +58,7 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/source"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/config"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/constants"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/controller/indexers"
 	saturation_v2 "github.com/llm-d/llm-d-workload-variant-autoscaler/internal/engines/analyzers/saturation_v2"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/interfaces"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
@@ -65,6 +66,7 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/saturation"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/utils"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/utils/scaletarget"
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/record"
 )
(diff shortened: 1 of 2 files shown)


internal/collector/replica_metrics.go
```diff
@@ -250,9 +252,26 @@ func (c *ReplicaMetricsCollector) buildInstanceKey(ctx context.Context, namespac
...
+					APIVersion: ms.ScaledObject.Spec.ScaleTargetRef.APIVersion,
+					Kind:       ms.ScaledObject.Spec.ScaleTargetRef.Kind,
+					Name:       ms.ScaledObject.Spec.ScaleTargetRef.Name,
+				}
+				if va, lookupErr := indexers.FindVAForScaleTarget(ctx, c.k8sClient, soRef, namespace); lookupErr == nil && va != nil {
+					vaName = va.Name
+				} else {
+					vaName = ms.ScaledObject.Name
+				}
 			}
```
shuynh2017, line 275: is it possible that after all this vaName is still nil? If this is the case then we should to log and record an error metric.

lionelvillard: yes it is, for various reasons. logging and recording is beyond the scope of this PR though.

PR #1306, opened 2026-06-22
fix(saturation): log V2 capacity after threshold post-step, not before

### LINKED ISSUE #1305: bug(obsevability): "V2 saturation analysis completed" log always reports requiredCapacity=0 / spareCapacity=0
## Summary

The `"V2 saturation analysis completed"` log line in `runV2AnalysisOnly`
(`internal/engines/saturation/engine_v2.go`) logs `requiredCapacity` and
`spareCapacity` from the raw analyzer result. The V2 analyzer deliberately
leaves those two fields at `0` — they are filled in later by the engine's
`applyUniversalThreshold` post-step. So this log line reports
`requiredCapacity=0, spareCapacity=0` on **every** run, scaling or not.

This is purely misleading observability (not a scaling bug), but it reads as if
the model has zero demand / zero spare and has cost real debugging time chasing
a phantom "no scale-up" anomaly.

## Root cause

- `internal/engines/analyzers/saturation_v2/analyzer.go` (Phase 5, ~line 121):
  the result is built without setting `RequiredCapacity`/`SpareCapacity`
  ("left zero; the engine post-step overwrites them").
- `internal/engines/saturation/engine_v2.go` `runV2AnalysisOnly` logs
  `result.RequiredCapacity` / `result.SpareCapacity` **immediately after**
  `Analyze()` returns — before `applyUniversalThreshold` runs.
- The real values are computed afterward in `runAnalyzersAndScore`
  (`resolveThresholds` + `applyUniversalThreshold`) and are what the optimizer
  actually scales on.

ev-shindin:
Fixes #1305

## Problem

The `"V2 saturation analysis completed"` log reported `requiredCapacity` and `spareCapacity` straight from the raw analyzer result. The V2 analyzer **deliberately leaves those two fields at 0** — they are filled in by the engine's `applyUniversalThreshold` post-step that runs afterward in `runAnalyzersAndScore`. So the log printed `requiredCapacity=0, spareCapacity=0` on every cycle regardless of actual demand, which reads as a phantom "no demand / no scale" signal and has cost real debugging time.

## Fix

Move the single Info line to `runAnalyzersAndScore`, **after** `applyUniversalThreshold`, so it reports the real `RequiredCapacity`/`SpareCapacity` (plus the resolved scale-up/down thresholds) alongside `totalSupply`/`totalDemand`/`utilization`. `runV2AnalysisOnly` no longer logs, so there's no second per-cycle Info line.

No behavior change beyond logging.

## Testing

`go build` + `go vet` clean; `golangci-lint` clean; `internal/engines/saturation` (envtest) passes. No test asserts on log output, and the underlying `applyUniversalThreshold` capacity computation is already covered by `engine_v2_threshold_test.go`.

### DIFF
--- a/internal/engines/saturation/engine_v2.go
+++ b/internal/engines/saturation/engine_v2.go
@@ -64,14 +64,10 @@ func (e *Engine) runV2AnalysisOnly(
 		return nil, fmt.Errorf("V2 saturation analysis failed: %w", err)
 	}
 
-	logger.Info("V2 saturation analysis completed",
-		"modelID", modelID,
-		"totalSupply", result.TotalSupply,
-		"totalDemand", result.TotalDemand,
-		"utilization", result.Utilization,
-		"requiredCapacity", result.RequiredCapacity,
-		"spareCapacity", result.SpareCapacity)
-
+	// Analysis results are logged by the caller (runAnalyzersAndScore) after the
+	// applyUniversalThreshold post-step, so the single Info line can include the
+	// real RequiredCapacity/SpareCapacity. They are left zero here, so logging
+	// them at this point would always report 0 and be misleading.
 	return result, nil
 }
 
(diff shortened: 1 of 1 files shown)


biranofer:
Seems OK, I think we'll need a PR for more ordered manner for the analyzers / optimizers logging.

biranofer (approved):
As commented, hopefully future PR will make more ordered logging of analyzers and optimizers.

PR #1315, opened 2026-06-22
feat(benchmark): add post-processing script for tabular benchmark reports

### LINKED ISSUE #1239: Post processing script of benchmarking
After a benchmark is completed, there should be a post-processing step, either integrated in make targets or a separate script that allows data to be reported in a tabular format from the benchmark output folder, as reported in https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/main/docs/benchmark.md

kahilam:
## Summary

- Add `hack/benchmark/postprocess.py` that extracts metrics from `llm-d-benchmark` output folders and generates a markdown table matching the format in `docs/benchmark.md`
- Extracted metrics: P99 TTFT, P99 ITL, avg/max replicas, avg KV cache utilization, avg EPP queue depth, error count, avg pod startup, and cost
- Integrate into Makefile with a new `benchmark-report` target and auto-call after `benchmark-run`

### Usage

```bash
# Standalone — single run:
python hack/benchmark/postprocess.py results/guidellm-*_1

# Three runs (generates Run 1 | Run 2 | Run 3 | Avg columns):
python hack/benchmark/postprocess.py results/guidellm-*_1 \
                                      results/guidellm-*_2 \
                                      results/guidellm-*_3

# With scenario header and GPU count:
python hack/benchmark/postprocess.py --scenario "Prefill Heavy — Qwen/Qwen3-32B (600s)" \
    --gpus-per-replica 2 results/guidellm-*_1

# Via Makefile (auto-detects latest results):
make benchmark-report
make benchmark-report BENCHMARK_GPUS_PER_REPLICA=2

# Auto-runs after benchmark-run completes:
make benchmark-run BENCHMARK_NAMESPACE=kahila-bench \
    BENCHMARK_MODEL_ID=Qwen/Qwen3-32B \
    BENCHMARK_GPUS_PER_REPLICA=2
```

### Example output

```
| Metric | Run 1 |
|--------|------|
| P99 TTFT (ms) | 531,703 |
| P99 ITL (ms/token) | 60.78 |
| Avg replicas | 1.54 |
| Max replicas | 2 |
| Avg KV cache utilization | 55.0% |
| Avg queue depth (EPP) | 107.1 |
| Error count | 0 |
| Avg pod startup (s) | 128 |
| Cost (avg replicas × GPU/hr) | 1.54 |
```

Closes #1239

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -398,7 +398,7 @@ benchmark-report: ## Generate a markdown table from the latest benchmark results
 	fi; \
 	echo "Results directory: $$LATEST_DIR"; \
 	echo ""; \
-	python3 $(CURDIR)/hack/benchmark/postprocess.py \
+	python3 $(CURDIR)/scripts/postprocess.py \
 		--gpus-per-replica $(BENCHMARK_GPUS_PER_REPLICA) \
 		$$LATEST_DIR
 
--- a/Makefile
+++ b/Makefile
@@ -396,7 +396,7 @@ benchmark-report: ## Generate a markdown table from the latest benchmark results
 	fi; \
 	echo "Results directory: $$LATEST_DIR"; \
 	echo ""; \
-	python3 $(CURDIR)/scripts/postprocess.py $$LATEST_DIR
+	python3 $(CURDIR)/hack/benchmark/postprocess.py $$LATEST_DIR
 
 BURSTY_WORKLOAD    ?= bursty.yaml
 BENCHMARK_WAIT_TIMEOUT ?= 7200
--- a/Makefile
+++ b/Makefile
@@ -381,6 +381,26 @@ benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<name
 		-w $(BENCHMARK_WORKLOAD) \
 		$(if $(BENCHMARK_MODEL_ID),-m $(BENCHMARK_MODEL_ID),) \
 		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,)
+	@echo ""
+	@echo "========================================="
+	@echo "  Generating benchmark report..."
+	@echo "========================================="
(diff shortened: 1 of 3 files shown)


asm582:
@kahilam, the script should not be in the hack folder; we should add it as a new make target.

Makefile
```diff
@@ -381,6 +381,26 @@ benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<name
 		-w $(BENCHMARK_WORKLOAD) \
 		$(if $(BENCHMARK_MODEL_ID),-m $(BENCHMARK_MODEL_ID),) \
 		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,)
+	@echo ""
+	@echo "========================================="
+	@echo "  Generating benchmark report..."
+	@echo "========================================="
+	@$(MAKE) benchmark-report
+
+BENCHMARK_GPUS_PER_REPLICA ?= 1
```
asm582: why is this needed?

asm582 (changes requested):

Makefile
kahilam: if you think it's unnecessary complexity, I could simplify by removing the BENCHMARK_GPUS_PER_REPLICA variable and the Cost row entirely. ince the cost is just a derived value from avg replicas anyway. Want me to remove it and push the update?

asm582: The default value of GPUs per replica, set to 1, is incorrect as per my understanding, we should remove it.

kahilam: Removed.

asm582:
@kahilam I don't think we should create a new top-level dir, please create a folder under `hack`

kahilam:
> @kahilam I don't think we should create a new top-level dir, please create a folder under `hack`

Moved back to hack/benchmark

Issue #919, opened 2026-03-23
Observability: PrometheusRule Alerting Rules


ev-shindin:
Add a `PrometheusRule` CRD with baseline alerting rules.

**Alerts:**

```yaml
groups:
- name: wva.rules
  rules:
  - alert: WVAHighErrorRate
    expr: rate(wva_errors_total[5m]) > 0.1
    for: 5m
    labels:
      severity: warning
    annotations:
      summary: "WVA error rate elevated"

  - alert: WVAOptimizationLoopStalled
    expr: rate(wva_models_processed_total[10m]) == 0
    for: 15m
    labels:
      severity: critical
    annotations:
      summary: "WVA optimization loop has stopped processing models"

  - alert: WVAMetricsCollectionFailing
    expr: rate(wva_metrics_collection_errors_total[5m]) > 0.5
    for: 5m
    labels:
      severity: warning
    annotations:
      summary: "WVA metrics collection failing frequently"

  - alert: WVAGPUResourceExhausted
    expr: wva_available_gpus == 0
    for: 5m
    labels:
      severity: warning
    annotations:
      summary: "No GPUs available for WVA scaling"

  - alert: WVAReplicaScalingThrashing
    expr: rate(wva_replica_scaling_total[10m]) > 2
    for: 10m
    labels:
      severity: warning
    annotations:
      summary: "WVA scaling decisions changing rapidly (possible thrashing)"
```

**Implementation:**
- Add `PrometheusRule` CRD in `config/prometheus/`
- Include in Helm chart as optional (enabled via `values.yaml` flag)
- Thresholds should be configurable in Helm values

**Acceptance Criteria:**
- [ ] PrometheusRule CRD deploys with `make deploy` and Helm chart
- [ ] Alerts fire correctly when conditions are met (manual verification)
- [ ] Helm chart allows enabling/disabling and threshold overrides

ev-shindin:
@hemanth9398 : this issue have pre-requistics. You would wait before preceding issues from this epic will be closed.

PR #1073, opened 2026-04-29
feat: Observability - pipeline stage metrics

shuynh2017:
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/915 . Close https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/915

## Acceptance Criteria:
- [x] Limiter constraint events tracked with variant + limiter name
- [x] Available GPU gauge reflects current inventory
  - [x] e2e test will discover GPUs and compare against the metrics. 
- [x] Enforcer modification counter tracks each policy type
  - [ ] is there a policy type `stabilization_window`?
- [x] Active optimizer gauge reflects current config
  - [x] At any point in time, only 1 optimizer is active, the others are inactive. 
- [x] Unit tests for each pipeline stage metric

## New Metrics
Here are some available ones after running smoke test with sample dashboard. Note `wva_available_gpus` is only available when `enableLimiter` is `true`:

```
 {
    "metric": {
      "__name__": "wva_available_gpus",
      "accelerator_model": "AMD-MI300X-192G",
      "accelerator_type": "MI300X",
      "accelerator_vendor": "amd.com",
      "container": "manager",
      "endpoint": "https",
      "instance": "10.244.0.73:8443",
      "job": "workload-variant-autoscaler-metrics",
      "namespace": "workload-variant-autoscaler-system",
      "pod": "workload-variant-autoscaler-controller-manager-75b45dd7c-89g5s",
      "service": "workload-variant-autoscaler-metrics"
    },
    "value": [
      1778600936.410,
      "4"
    ]
  },
  {
    "metric": {
      "__name__": "wva_available_gpus",
      "accelerator_model": "NVIDIA-A100-PCIE-80GB",
      "accelerator_type": "A100",
      "accelerator_vendor": "nvidia.com",
      "container": "manager",
      "endpoint": "https",
      "instance": "10.244.0.73:8443",
      "job": "workload-variant-autoscaler-metrics",
      "namespace": "workload-variant-autoscaler-system",
      "pod": "workload-variant-autoscaler-controller-manager-75b45dd7c-89g5s",
      "service": "workload-variant-autoscaler-metrics"
    },
    "value": [
      1778600936.410,
      "4"
    ]
  },
  {
    "metric": {
      "__name__": "wva_available_gpus",
      "accelerator_model": "NVIDIA-H100-SXM5-80GB",
      "accelerator_type": "H100",
      "accelerator_vendor": "nvidia.com",
      "container": "manager",
[...]

PR #1328, opened 2026-06-24
feat: Observability: PrometheusRule Alerting Rules

shuynh2017:
## Proposed Changes

- :gift: Add new feature https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/919

### Pre-review Checklist

- [ ] **Docs PR** for any user-facing impact. 
   - See Docs section below.
- [ ] **Proposal PR** for any new enhancement or change to existing behavior. 
    - N/A since this requirement was previously created.

**Release Note**

This change adds user-visible feature, but there's no impact in terms of breaking changes, etc ...

```release-note

```

**Docs**

Doc changes are in  `docs/developer-guide/prometheus.md`, but if we can find a home in llm-d then will glad to move there.


shuynh2017:
Ref: https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/919#issue-4118902148
Implementation:

- [ ] Add PrometheusRule CRD in config/prometheus/
Include in Helm chart as optional (enabled via values.yaml flag)
Thresholds should be configurable in Helm values
  - @ev-shindin, pls note that instead of `config/prometheus`, I have it in `config/components/prometheus-alerts` for kustomization. Also, there's no Helm support.

Acceptance Criteria:

- [ ] PrometheusRule CRD deploys with make deploy and Helm chart 
  - @ev-shindin, pls note that installation is via kustomization, not Helm.

- [x] Alerts fire correctly when conditions are met (manual verification). See screen capture in user doc.

- [ ] Helm chart allows enabling/disabling and threshold overrides. @ev-shindin, here users can manually update `prometheusrule.yaml` for threshold overrides, and enable/disable by setting environment var DEPLOY_ALERTING_RULES as specified in the user doc.
- [ ] rate(wva_models_processed_total[10m]) == 0
  - @ev-shindin, on this rule, we don't have `wva_models_processed_total`, we have `wva_models_processed`. Also, if we were to go with `wva_models_processed`, pls review the expression as this may not be what we want.

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+      expr: rate(wva_errors_total[5m]) > 0.1
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: "WVA error rate elevated (>0.1 errors/sec)"
+        description: "Error rate exceeds threshold (>30 errors over 5 minutes). Check controller logs for error patterns."
+
+    - alert: WVAOptimizationLoopStalled
+      expr: rate(wva_models_processed_total[10m]) == 0
```
ev-shindin: There is no `wva_models_processed_total`. The actual metric is `wva_models_processed`
(`internal/constants/metrics.go:117`) — it was deliberately renamed away from the `_total`
suffix in #981. `rate()` over an absent series returns an empty vector, and `empty == 0`
matches nothing, so this `critical` alert is silently dead.

It's also a **gauge** (`internal/metrics/metrics.go:211`, registered as `GaugeVec`), so
`rate()` is semantically wrong even after fixing the name. To detect a stalled optimization
loop, you can use the histogram's `_count` counter, which increments once per cycle:

```yaml
expr: rate(wva_optimization_duration_seconds_count[10m]) == 0
```

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+      expr: wva_available_gpus == 0
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: "No GPUs available for WVA scaling"
+        description: "No GPUs available for WVA scaling. Consider adding GPU nodes."
+
+    - alert: WVAReplicaScalingThrashing
+      expr: rate(wva_replica_scaling_total[10m]) > 2
```
ev-shindin: `rate(wva_replica_scaling_total[10m]) > 2` means **more than 2 scaling operations per second**
sustained for 10 minutes (your own description says "~1200 operations over 10 minutes"). The
optimization loop runs on an interval of tens of seconds, so real thrashing is a few flips per
*minute*, never 2/sec.

Consider a realistic threshold, and update the summary/description to match.

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+      expr: rate(wva_metrics_collection_errors_total[5m]) > 0.5
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: "WVA metrics collection failing frequently (>0.5 errors/sec)"
+        description: "Metrics collection error rate exceeds threshold (>150 errors over 5 minutes). Scaling decisions may be based on stale data. Verify Prometheus/metrics endpoint connectivity."
+
+    - alert: WVAGPUResourceExhausted
+      expr: wva_available_gpus == 0
```
ev-shindin, line 41: Minor: `wva_available_gpus` is only emitted when GPU discovery runs (limiter enabled /
cluster-scoped node read access — see #1073). On clusters that never emit it, this alert is
inert (the series is absent, so `== 0` never matches). That's acceptable, but please add a
one-line caveat in `prometheus.md` so operators don't assume "no alert = GPUs available."

test/e2e/prometheus_alerts_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+		if err == nil {
+			By("Deleting existing PrometheusRule from previous deployment")
+			err = crClient.Delete(ctx, existingRule)
+			Expect(err).NotTo(HaveOccurred(), "Should be able to delete existing PrometheusRule")
+			GinkgoWriter.Printf("✓ Deleted existing PrometheusRule '%s'\n", prometheusRuleName)
+
+			// Wait for deletion to complete
+			Eventually(func(g Gomega) {
+				checkRule := &promoperator.PrometheusRule{}
```
ev-shindin: This test is never checks that the alert **expressions reference metrics that actually exist**. 

Two gaps:

1. Add an assertion that each `rule.Expr` only references known metric names (e.g. cross-check
   against the 5 expected metrics, or against `internal/constants/metrics.go`). This turns a
   silent metric-rename regression into a test failure.
2. The test creates the `PrometheusRule` directly via `crClient`, bypassing the
   `DEPLOY_ALERTING_RULES` install path (`install.sh` / `infra_wva.sh`). The PR description
   claims tests "validate alert firing," but neither firing nor the install wiring is
   exercised — consider toning down that claim or covering the install path.

ev-shindin (changes requested):
Thanks @shuynh2017 ! Please see my comments below.

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: I am proposing to use `wva_models_processed` to show this alert. I think the alert is now accurate as well as handles absent series. Here's the alert with a more friendly description which I applied to the rest of the rules:
```
 - alert: WVAOptimizationLoopStalled
      expr: max_over_time(wva_models_processed[10m]) == 0 or absent_over_time(wva_models_processed[10m])
      for: 15m
      labels:
        severity: critical
      annotations:
        summary: "WVA optimization loop has stopped processing models"
        description: "No models have been processed over a 10-minute window for at least 15 minutes (either processing 0 models or metric is absent). The optimization loop may be stalled or crashed. Check controller pod status and logs."

```

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: I've updated this alert to apply to variants since the manager can handle many variants. I think this is more realistic. Included is how it looks if manually triggered:
```
- alert: WVAReplicaScalingThrashing
      expr: sum by (namespace, variant_name) (rate(wva_replica_scaling_total[10m])) > 0.033
      for: 10m
      labels:
        severity: warning
      annotations:
        summary: "WVA scaling decisions changed rapidly for a variant"
        description: "Variant has sustained a scaling rate exceeding 2/min over a 10-minute window for at least 10 minutes, indicating possible thrashing. Review scaling thresholds for the variant and consider adjusting them."
```
image

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: Added a new section in `prometheus.md` so we can add more notes later:
```
### Alerting Rules Notes
- For `WVAGPUResourceExhausted` rule with `wva_available_gpus` metric, as described in [wva_available_gpus](#wva_available_gpus) this metric is not always available in which case this rule will not trigger. In other words, if there's no alert for this rule, it does not mean GPUs are available.
```

test/e2e/prometheus_alerts_test.go
shuynh2017: - Fixed for 1
- For 2, update to change to test the full install path instead of create `PrometheusRule` directly. The flow:
  - Update e2e-smoke which calls `make  test-e2e-smoke-with-setup`. These tests are very quick so it should be fine to run in smoke tests:
      ```
    test-e2e-smoke-with-setup:
	    DEPLOY_ALERTING_RULES=true $(MAKE) deploy-e2e-infra
	    $(MAKE) test-e2e-smoke
    ```
- Test Output:
  ```
  PrometheusAlerts should have PrometheusRule deployed [smoke, prometheus-alerts]
  /home/shuynh/workload-variant-autoscaler-alert/workload-variant-autoscaler/test/e2e/prometheus_alerts_test.go:143
    STEP: Checking if PrometheusRule CRD is available @ 06/26/26 13:57:31.624
    ✓ PrometheusRule CRD is available
    ✓ DEPLOY_ALERTING_RULES is set - testing deployed PrometheusRule
    STEP: Verifying PrometheusRule exists @ 06/26/26 13:57:31.625
    ✓ PrometheusRule 'controller-manager-alerts' exists in namespace 'workload-variant-autoscaler-system'
    STEP: Verifying PrometheusRule has correct structure @ 06/26/26 13:57:31.63
    ✓ PrometheusRule structure is valid
  • [0.006 seconds]
  PrometheusAlerts should have all expected alert rules defined [smoke, prometheus-alerts]
  /home/shuynh/workload-variant-autoscaler-alert/workload-variant-autoscaler/test/e2e/prometheus_alerts_test.go:164
    STEP: Retrieving PrometheusRule @ 06/26/26 13:57:31.63
    STEP: Verifying all 5 alert rules are present @ 06/26/26 13:57:31.641
      ✓ Found alert: WVAHighErrorRate
      ✓ Found alert: WVAOptimizationLoopStalled
      ✓ Found alert: WVAMetricsCollectionFailing
      ✓ Found alert: WVAGPUResourceExhausted
      ✓ Found alert: WVAReplicaScalingThrashing
    ✓ All 5 expected alert rules are present
  • [0.011 seconds]
  PrometheusAlerts should have valid alert rule structure [smoke, prometheus-alerts]
  /home/shuynh/workload-variant-autoscaler-alert/workload-variant-autoscaler/test/e2e/prometheus_alerts_test.go:198
    STEP: Retrieving PrometheusRule @ 06/26/26 13:57:31.641
    STEP: Validating each alert rule has required fields @ 06/26/26 13:57:31.649
      ✓ Alert 'WVAHighErrorRate' has valid structure
      ✓ Alert 'WVAOptimizationLoopStalled' has valid structure
      ✓ Alert 'WVAMetricsCollectionFailing' has valid structure
      ✓ Alert 'WVAGPUResourceExhausted' has valid structure
      ✓ Alert 'WVAReplicaScalingThrashing' has valid structure
    ✓ All alert rules have valid structure
  • [0.008 seconds]
  PrometheusAlerts should only reference known WVA metrics in alert expressions [smoke, prometheus-alerts]
  /home/shuynh/workload-variant-autoscaler-alert/workload-variant-autoscaler/test/e2e/prometheus_alerts_test.go:232
    STEP: Retrieving PrometheusRule @ 06/26/26 13:57:31.65
    STEP: Building a map of valid WVA metric names @ 06/26/26 13:57:31.665
    STEP: Validating each alert expression references only known metrics @ 06/26/26 13:57:31.665
      Checking alert 'WVAHighErrorRate':
        Expression: rate(wva_errors_total[5m]) > 0.1
        Referenced metrics: [wva_errors_total]
      ✓ Alert 'WVAHighErrorRate' references only known metrics
      Checking alert 'WVAOptimizationLoopStalled':
        Expression: max_over_time(wva_models_processed[10m]) == 0 or absent_over_time(wva_models_processed[10m])
        Referenced metrics: [wva_models_processed]
      ✓ Alert 'WVAOptimizationLoopStalled' references only known metrics
      Checking alert 'WVAMetricsCollectionFailing':
        Expression: rate(wva_metrics_collection_errors_total[5m]) > 0.5
        Referenced metrics: [wva_metrics_collection_errors_total]
      ✓ Alert 'WVAMetricsCollectionFailing' references only known metrics
      Checking alert 'WVAGPUResourceExhausted':
        Expression: wva_available_gpus == 0
        Referenced metrics: [wva_available_gpus]
      ✓ Alert 'WVAGPUResourceExhausted' references only known metrics
      Checking alert 'WVAReplicaScalingThrashing':
        Expression: sum by (namespace, variant_name) (rate(wva_replica_scaling_total[10m])) > 0.033
        Referenced metrics: [wva_replica_scaling_total]
      ✓ Alert 'WVAReplicaScalingThrashing' references only known metrics
    ✓ All alert expressions reference known WVA metrics
  • [0.016 seconds]
  ```

Makefile
```diff
@@ -284,7 +284,10 @@ test-e2e-full: ## Run full e2e test suite
 # Convenience target that deploys infra + runs smoke tests.
 # Set DELETE_CLUSTER=true to delete Kind cluster after tests (default: keep cluster for debugging).
 .PHONY: test-e2e-smoke-with-setup
-test-e2e-smoke-with-setup: deploy-e2e-infra test-e2e-smoke
+test-e2e-smoke-with-setup:
+	export DEPLOY_ALERTING_RULES=true
```
ev-shindin: No `.ONESHELL` (`SHELL=…bash`, `.SHELLFLAGS=-ec`), so each recipe line is a separate shell. `export DEPLOY_ALERTING_RULES=true` exits before the `$(MAKE)` lines run, so `deploy-e2e-infra` uses the default `false` (the install script then *deletes* the PrometheusRule) and the suite hits `Skip(...)`. The target exercises nothing. Use inline env like `Makefile:323`: `DEPLOY_ALERTING_RULES=true $(MAKE) deploy-e2e-infra` on each line, or a target-specific `export`.

go.mod
```diff
@@ -120,5 +120,5 @@ require (
 	k8s.io/utils v0.0.0-20251002143259-bc988d571ff4
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.2 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
-	sigs.k8s.io/yaml v1.6.0 // indirect
+	sigs.k8s.io/yaml v1.6.0
```
ev-shindin: Do we need this change? Looks like nothing in the module imports `sigs.k8s.io/yaml`

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+      expr: rate(wva_errors_total[5m]) > 0.1
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: "WVA error rate elevated"
+        description: "WVA error rate has exceeded 6/min over a 5-minute window for at least 5 minutes. Check controller logs for error patterns."
+
+    - alert: WVAOptimizationLoopStalled
+      expr: max_over_time(wva_models_processed[10m]) == 0 or absent_over_time(wva_models_processed[10m])
```
ev-shindin, line 23: `max_over_time(wva_models_processed[10m]) == 0` flags a healthy controller with no `VariantAutoscaling`s (legitimately 0/cycle) as "stalled or crashed." Gate on a heartbeat, or document the ≥1-managed-model assumption.

ev-shindin (changes requested):
Thanks @shuynh2017! Please address review.

test/e2e/prometheus_alerts_test.go
ev-shindin: Tests are good, but looks like they ran with adifferent `Makefile`. Please see my comment below.

go.mod
shuynh2017: reverted. Thanks.

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: Good point. Updated severity and description to help with the rule:
```
-        severity: critical
+        severity: warning
       annotations:
         summary: "WVA optimization loop has stopped processing models"
-        description: "No models have been processed over a 10-minute window for at least 15 minutes (either processing 0 models or metric is absent). The optimization loop may be stalled or crashed. Check controller pod status and logs."
+        description: "No models have been processed over a 10-minute window for at least 15 minutes. This can indicate the controller is healthy but there's no models to process, or there are models to process but the optimization loop may be stalled or crashed, or there are issues with metric collection.  Check controller pod status and logs, or verify Prometheus endpoint connectivity."
```

Makefile
shuynh2017: Fixed. Thank you.
image

config/components/prometheus-alerts/prometheusrule.yaml
ev-shindin: `wva_models_processed` is an unlabeled global gauge, so a healthy controller managing zero VariantAutoscalings legitimately emits `0` every cycle → this `warning` fires when nothing is wrong, and the description itself says "This can indicate the controller is healthy but there's no models to process." That's an alert-fatigue anti-pattern. Gate it on work-existing (AND with a managed-VA-count signal, e.g. `... and on() (count(...) > 0)`), or use an `up`/heartbeat signal (the actual "stalled/crashed" condition), or downgrade to `severity: info` with a single clear meaning. The `absent_over_time` branch already covers "metric/target gone."

test/e2e/prometheus_alerts_test.go
```diff
@@ -0,0 +1,270 @@
...
+// PrometheusAlerts test suite validates the deployed PrometheusRule resource.
+// This test:
+// - Verifies PrometheusRule exists (deployed via install.sh)
+// - Validates all expected alert rules are present with correct structure
+// - Validates alert expressions reference only known WVA metrics
+//
+// This test does NOT:
+// - Create or delete PrometheusRule resources (expects them to be deployed)
+// - Validate that alerts actually fire when conditions are met (would require metric injection)
+var _ = Describe("PrometheusAlerts", Label("smoke"), Label("prometheus-alerts"), Ordered, func() {
```
ev-shindin, line 209: Every spec only calls `crClient.Get()` on the PrometheusRule. If the Prometheus `ruleSelector`/`ruleNamespaceSelector` doesn't match, all specs pass while Prometheus never loads the rules. Add a query to the in-cluster Prometheus `/api/v1/rules` (Eventually, ~60s for operator reconcile) asserting group `wva.rules` is present and each rule's `health == "ok"`. Keep the existing "does NOT validate alerts fire" disclaimer — this is the weaker "rules are visible" check that's missing.

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+spec:
+  groups:
+  - name: wva.rules
+    rules:
+    - alert: WVAHighErrorRate
+      expr: rate(wva_errors_total[5m]) > 0.1
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
```
ev-shindin, line 18: The exprs are per-series (`wva_errors_total{component,error_type}`, `wva_available_gpus{accelerator_*}`, thrashing by `{namespace,variant_name}`) but no `summary`/`description` uses Go templating, so a firing alert can't name what's affected. Add `{{ $labels.component }}` / `{{ $labels.accelerator_type }}` / `{{ $labels.namespace }}/{{ $labels.variant_name }}` and `{{ $value | printf "%.2f" }}`. Also decide intentionally whether `WVAHighErrorRate`/`WVAMetricsCollectionFailing` should fan out per `error_type`/`reason` or roll up via `sum by (component)`.

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+        description: "WVA error rate has exceeded 6/min over a 5-minute window for at least 5 minutes. Check controller logs for error patterns."
+
+    - alert: WVAOptimizationLoopStalled
+      expr: max_over_time(wva_models_processed[10m]) == 0 or absent_over_time(wva_models_processed[10m])
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        summary: "WVA optimization loop has stopped processing models"
+        description: "No models have been processed over a 10-minute window for at least 15 minutes. This can indicate the controller is healthy but there's no models to process, or there are models to process but the optimization loop may be stalled or crashed, or there are issues with metric collection.  Check controller pod status and logs, or verify Prometheus endpoint connectivity."
```
ev-shindin: same as line 20

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+    - alert: WVAMetricsCollectionFailing
+      expr: rate(wva_metrics_collection_errors_total[5m]) > 0.5
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: "WVA metrics collection failed frequently"
+        description: "Metrics collection error rate has exceeded 30/min over a 5-minute window for at least 5 minutes. Scaling decisions may be based on stale data. Verify Prometheus/metrics endpoint connectivity."
```
ev-shindin: same as line 20

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+    - alert: WVAGPUResourceExhausted
+      expr: wva_available_gpus == 0
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        summary: "No GPUs available for WVA scaling"
+        description: "No GPUs available for WVA scaling. Consider adding GPU nodes."
```
ev-shindin: same as line 20

config/components/prometheus-alerts/prometheusrule.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+    - alert: WVAReplicaScalingThrashing
+      expr: sum by (namespace, variant_name) (rate(wva_replica_scaling_total[10m])) > 0.033
+      for: 10m
+      labels:
+        severity: warning
+      annotations:
+        summary: "WVA scaling decisions changed rapidly for a variant"
+        description: "Variant has sustained a scaling rate exceeding 2/min over a 10-minute window for at least 10 minutes, indicating possible thrashing. Review scaling thresholds for the variant and consider adjusting them."
```
ev-shindin: same as line 20

test/e2e/prometheus_alerts_test.go
```diff
@@ -0,0 +1,270 @@
...
+	BeforeAll(func() {
+		// Skip if DEPLOY_ALERTING_RULES is not set to true
+		deployAlertingRules := os.Getenv("DEPLOY_ALERTING_RULES")
```
ev-shindin, line 212: `DEPLOY_ALERTING_RULES` must be set at both deploy and test time; the direct `make test-e2e-smoke` recipe doesn't forward it (CI uses the `-with-setup` wrapper, so it runs today — not dead in CI). More robust: gate the Skip on PrometheusRule CR existence (`IsNotFound` → Skip) so the infra state drives the decision; optionally add `DEPLOY_ALERTING_RULES=$(DEPLOY_ALERTING_RULES)` to the direct recipe's env list + document it.

ev-shindin (commented):
Please address comments.

test/e2e/prometheus_alerts_test.go
shuynh2017: Actually my first implementation is to skip if PrometheusRule CR  not found, but I changed and want to make sure the test is always run and not skipped - silently not tested. With the current implementation, I am now sure both rule installation and checking are done.

shuynh2017:
For this: _wva_models_processed is an unlabeled global gauge, so a healthy controller managing zero VariantAutoscalings legitimately emits 0 every cycle → this warning fires when nothing is wrong, and the description itself says "This can indicate the controller is healthy but there's no models to process." That's an alert-fatigue anti-pattern. Gate it on work-existing (AND with a managed-VA-count signal, e.g. ... and on() (count(...) > 0)), or use an up/heartbeat signal (the actual "stalled/crashed" condition), or downgrade to severity: info with a single clear meaning. The absent_over_time branch already covers "metric/target gone."_ I decide to downgrade to info, and think the message is sufficiently clear.

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: I've chose fan-out over rollup since the cardinality is low/manageable, different label values can use different responses, more actionalble.

test/e2e/prometheus_alerts_test.go
shuynh2017: fixed.

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: fixed.

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: fixed

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: fixed.

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: fixed.

config/components/prometheus-alerts/prometheusrule.yaml
shuynh2017: fixed.

shuynh2017:
@ev-shindin, thanks much for always improving the quality. I've addressed most and commented the rest.

ev-shindin:
@shuynh2017 pushed a maintainer edit (d4844f1c) that folds in the some items so we can move toward merge. Summary of what changed:

- **`Makefile`** — thread `DEPLOY_ALERTING_RULES` into `deploy-e2e-infra` and the `test-e2e-smoke` / `-full` / `-multi-controller` `go test` env blocks (with a `DEPLOY_ALERTING_RULES ?= false` default), so a split *deploy → test* pipeline actually runs the alert specs instead of silently skipping them. Also switched `-with-setup` to `$(MAKE) DEPLOY_ALERTING_RULES=true …` so a command-line override behaves consistently.
- **`deploy/lib/infra_wva.sh`** — collapsed the disabled-rules cleanup to a single guarded `delete`; it now only logs/acts when a `PrometheusRule` actually exists, so fresh installs on the default path stay quiet.
- **`docs/developer-guide/prometheus.md`** — added the missing `wva_replica_scaling_total` metric section (it backs `WVAReplicaScalingThrashing`).
- **`test/e2e/prometheus_alerts_test.go`** — expanded the PromQL-keyword blocklist in `extractMetricNames` (`increase`, `histogram_quantile`, `clamp_*`, the `*_over_time` family, math funcs, …) so a future alert using one of those won't false-fail as an "unknown metric"; and changed the rules-loaded check to **`Fail` instead of `Skip`** when `PROMETHEUS_URL` is set but unreachable — an explicitly-configured endpoint that can't be reached is a real failure, not a silent pass.
- **`config/components/prometheus-alerts/prometheusrule.yaml`** — reworded two alert descriptions to match what they actually detect:
  - `WVAOptimizationLoopStalled`: an idle controller (zero managed `VariantAutoscaling`s) legitimately reports `0`, so the text now says that's *expected* and only points at stall/crash/collection when managed workloads exist.
  - `WVAGPUResourceExhausted`: clarified this reflects node **allocatable** capacity (not free/schedulable GPUs) and is only emitted when the GPU limiter is enabled (`enableLimiter=true`); the likely real cause is a NotReady node or a missing device plugin.

**Two items I left for you**:

1. **`WVAMetricsCollectionFailing` threshold** (`> 0.5/s`). At ~60s/variant that needs ~30 variants failing at once, so a full Prometheus outage on a small deployment wouldn't trip it. Worth a lower threshold or a freshness/ratio signal.
2. **`WVAGPUResourceExhausted` firing behavior.** I only reworded the text — the underlying alert still can't fire in a default deploy (no `wva_available_gpus` series unless the limiter is on) and an "absent" series won't reach `== 0`. If you want it to actually fire, it needs gating on `enableLimiter` and/or a different expression; otherwise the reworded description at least tells operators what it means.

Happy to take another pass once you've decided on those two. Thanks for the clean, well-scoped PR.

PS please rebase this PR it has conflicts on `Makefile`

shuynh2017:
> **Two items I left for you**:
> 
> 1. **`WVAMetricsCollectionFailing` threshold** (`> 0.5/s`). At ~60s/variant that needs ~30 variants failing at once, so a full Prometheus outage on a small deployment wouldn't trip it. Worth a lower threshold or a freshness/ratio signal.

From my experience with using the dashboard thus far, I see this metric is mostly "No Data". Since it's not often that metric collection fails, I think we should lower the threshold to something really low. I chose 5/min.

> 2. **`WVAGPUResourceExhausted` firing behavior.** I only reworded the text — the underlying alert still can't fire in a default deploy (no `wva_available_gpus` series unless the limiter is on) and an "absent" series won't reach `== 0`. If you want it to actually fire, it needs gating on `enableLimiter` and/or a different expression; otherwise the reworded description at least tells operators what it means.

Thanks for updating the text.

ev-shindin (approved):
Thanks @shuynh2017 !

zdtsw:
@shuynh2017 do i understand the new alert rules wont be dpeloyed out of the box, rather it is done via make target or require manually apply?

PR #1334, opened 2026-06-24
fix(saturation): log actual analyzer mode instead of hardcoded value

### LINKED ISSUE #962: Remove hardcoding of analyzer mode
https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/511d3ae0450bb85b81197e63c530511e6aea040b/internal/engines/saturation/engine.go#L311

The saturation-only mode is printed even if the WVA controller is run in queueing-model mode.

Goutham-Annem:
Fixes #962

## Proposed Changes

- :bug: Fix bug

- The "Optimization completed successfully" log entry in `internal/engines/saturation/engine.go` always hardcoded `mode=saturation-only`, regardless of whether the queueing-model, V2 saturation, or V1 (legacy saturation-only) analyzer path actually ran.
- Extracted `modeLabelForAnalyzer(analyzerName string) string`, which maps the resolved `analyzerName` to the same label used by the `switch` that selects the optimize path, and used it for the completion log instead of the literal string.

### Pre-review Checklist

- [x] **E2E tests** for any new behavior — N/A, this is a logging-correctness fix with no behavioral change to scaling decisions
- [ ] **Docs PR** for any user-facing impact — N/A, internal log field only
- [ ] **Proposal PR** for any new enhancement or change to existing behavior — N/A, bug fix

**Release Note**

```release-note

```

**Docs**

N/A — internal log field, no user-facing documentation impact.

## How tested

- Added `TestModeLabelForAnalyzer` (table-driven, `internal/engines/saturation/mode_label_test.go`) covering all three modes plus the empty/unset and unrecognized fallback cases.
- `go test ./internal/engines/saturation/...` passes (61.4% coverage maintained).
- `make test` run locally: only pre-existing, unrelated failure in `internal/engines/analyzers/throughput` (a flaky degenerate-input case in `itl_model_test.go`), confirmed present on `main` without this change too.
- `gofmt`/`go vet` clean.

### DIFF
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -438,6 +438,7 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 	//   - Queueing model: QueueingModelAnalyzer → AnalyzerResult → Optimizer.Optimize → Enforcer bridge
 	// V1 will be deprecated once V2 is fully validated.
 	// Queueing model is activated by presence of wva-queueing-model-config ConfigMap.
+	mode := modeLabelForAnalyzer(analyzerName)
 	switch analyzerName {
 	case interfaces.QueueingModelAnalyzerName:
 		allDecisions = e.optimizeQueueingModel(ctx, modelGroups, currentAllocations)
@@ -460,11 +461,25 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 	e.applySaturationDecisions(ctx, allDecisions, vaMap, currentAllocations)
 
 	logger.Info("Optimization completed successfully",
-		"mode", "saturation-only",
+		"mode", mode,
 		"modelsProcessed", len(modelGroups),
 		"decisionsApplied", len(allDecisions))
 
 	return nil
+}
+
+// modeLabelForAnalyzer returns the human-readable mode label for the given
+// analyzer name, used in the "Optimization completed successfully" log entry.
+// It mirrors the analyzer selection in optimize's switch statement.
+func modeLabelForAnalyzer(analyzerName string) string {
+	switch analyzerName {
+	case interfaces.QueueingModelAnalyzerName:
+		return interfaces.QueueingModelAnalyzerName
+	case interfaces.SaturationAnalyzerName:
+		return interfaces.SaturationAnalyzerName
+	default:
+		return "saturation-only"
+	}
 }
 
 // recordEvent ensures only one event is recorded per VA in an optimization cycle.
--- a/internal/engines/saturation/mode_label_test.go
+++ b/internal/engines/saturation/mode_label_test.go
@@ -1,0 +1,31 @@
+package saturation
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/interfaces"
+)
+
+// TestModeLabelForAnalyzer ensures the "Optimization completed successfully" log
+// entry reports the analyzer mode that was actually selected, rather than always
+// logging "saturation-only" regardless of which path ran.
+func TestModeLabelForAnalyzer(t *testing.T) {
+	tests := []struct {
+		name         string
+		analyzerName string
+		wantMode     string
+	}{
+		{"queueing model analyzer", interfaces.QueueingModelAnalyzerName, interfaces.QueueingModelAnalyzerName},
+		{"V2 saturation analyzer", interfaces.SaturationAnalyzerName, interfaces.SaturationAnalyzerName},
+		{"unset analyzer name falls back to V1 saturation-only", "", "saturation-only"},
+		{"unrecognized analyzer name falls back to V1 saturation-only", "not-a-real-analyzer", "saturation-only"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.wantMode, modeLabelForAnalyzer(tt.analyzerName))
+		})
+	}
+}


internal/engines/saturation/engine.go
```diff
@@ -460,13 +461,27 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
...
 		"modelsProcessed", len(modelGroups),
 		"decisionsApplied", len(allDecisions))
 
 	return nil
 }
 
+// modeLabelForAnalyzer returns the human-readable mode label for the given
+// analyzer name, used in the "Optimization completed successfully" log entry.
+// It mirrors the analyzer selection in optimize's switch statement.
+func modeLabelForAnalyzer(analyzerName string) string {
```
ev-shindin: Optional (maintainability): this is a second `switch analyzerName` that duplicates the one in
`optimize`. They must stay in lockstep — add an analyzer case to one and forget the
other and the log silently lies again (the exact bug this fixes). Consider folding the label into
the existing switch instead

internal/engines/saturation/engine.go
```diff
@@ -460,13 +461,27 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 	e.applySaturationDecisions(ctx, allDecisions, vaMap, currentAllocations)
 
 	logger.Info("Optimization completed successfully",
-		"mode", "saturation-only",
+		"mode", mode,
```
ev-shindin, line 464: Operators now see `mode=saturation` (V2) vs `mode=saturation-only` (V1), which is a touch ambiguous. `saturation-v2` / `saturation-v1` would read clearer

ev-shindin (commented):
LGTM, there are sime comments to address.

internal/engines/saturation/mode_label_test.go
```diff
@@ -0,0 +1,43 @@
...
+		{"V2 saturation analyzer", interfaces.SaturationAnalyzerName, "saturation-v2"},
+		{"unset analyzer name falls back to V1", "", "saturation-v1"},
+		{"unrecognized analyzer name falls back to V1", "not-a-real-analyzer", "saturation-v1"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Reproduce the same switch logic as optimize() so the expected labels
+			// are verified without needing a full engine instance.
+			var mode string
```
ev-shindin: Non-vlocking: The test body re-implements the `optimize()` switch rather than invoking it, so it can't catch the regression its doc comment promises. Rename `"saturation-v2"` → `"saturation-v3"` in `engine.go` and forget the test, and it still passes (its private copy and `wantMode` both still say `"saturation-v2"`). Extract a helper in `engine.go` and call it from both sides — this also removes the duplicated `"saturation-vN"` literals and the style asymmetry (queueing uses a constant, V1/V2 use literals):

```go
func analyzerModeLabel(analyzerName string) string {
    switch analyzerName {
    case interfaces.QueueingModelAnalyzerName: return interfaces.QueueingModelAnalyzerName
    case interfaces.SaturationAnalyzerName:    return "saturation-v2"
    default:                                   return "saturation-v1"
    }
}
```

internal/engines/saturation/engine.go
```diff
@@ -438,12 +438,16 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 	//   - Queueing model: QueueingModelAnalyzer → AnalyzerResult → Optimizer.Optimize → Enforcer bridge
 	// V1 will be deprecated once V2 is fully validated.
 	// Queueing model is activated by presence of wva-queueing-model-config ConfigMap.
+	var mode string
```
ev-shindin: Nit: The mode label is assigned inline across the three switch arms, with one arm using the `interfaces.QueueingModelAnalyzerName` constant and the other two using raw `"saturation-vN"` literals. Replacing the inline assignments with `mode := analyzerModeLabel(analyzerName)` (above) keeps the path-selection switch and puts all three labels in one place, making the asymmetry intentional and the test meaningful in one move.

ev-shindin (commented):
LGTM. Please sign-off your last commit, and either open follow-up issue to improve test (see comment below), or fix test in this PR

Goutham-Annem:
Hi, just bumping this for visibility — CI is passing and the change is ready for review. Happy to address any feedback!

Goutham-Annem:
The DCO check is now clean (dropped an extra stray commit that was missing sign-off). PR is back to a single commit. Happy to address any feedback.

Goutham-Annem:
Friendly bump — CI is passing and branch is clean. Let me know if there's anything to address.

ev-shindin:
LGTM: Correct, well-scoped fix for #962. `modeLabelForAnalyzer` faithfully mirrors the `optimize()` selection switch, so the completion log now reports the analyzer path that actually ran (`queueing-model` / `saturation` / `saturation-only`) instead of the hardcoded `saturation-only`. Nice that production and the test now share the same helper (the test exercises real code, not a copy of the switch), and that the V1 label is left as `saturation-only` so the change is non-disruptive.

@Goutham-Annem:  everything below is optional - nothing blocking, however I would like to ask you to address first item (we have more analyzers ongoing).

**Maintainability**
- [ ] `modeLabelForAnalyzer` and `optimize()`'s path switch are two parallel switches over `analyzerName`. They agree today, but a future analyzer added to the path switch and not here would silently log `mode=saturation-only`. Fine for three modes; if you want to prevent drift, a tiny test that asserts the two switches cover the same cases (or deriving both from one map) would lock it in. — `internal/engines/saturation/engine.go`

**Simplification (nit)**
- [ ] The two named cases just return their input, so the body could collapse to `case interfaces.QueueingModelAnalyzerName, interfaces.SaturationAnalyzerName: return analyzerName`. The explicit form is arguably clearer since it mirrors the `optimize` switch — take it or leave it. — `internal/engines/saturation/engine.go`

**Style (nit)**
- [ ] `mode` is declared right before the switch but only used at the completion log ~20 lines later; inlining `"mode", modeLabelForAnalyzer(analyzerName)` at the log call keeps it next to its use. — `internal/engines/saturation/engine.go`

Goutham-Annem:
Friendly ping — CI is passing and this PR is ready for review. Would appreciate a look when you have a moment!

ev-shindin (approved):
LGTM.  Thanks @Goutham-Annem Please fix issue I marked in my previous comment in follow-up PR

PR #1335, opened 2026-06-24
feat(config): allow opt-in plain HTTP Prometheus endpoints

### LINKED ISSUE #1074: ValidateTLSConfig unconditionally rejects http:// Prometheus URLs — no way to opt out

Goutham-Annem:
Fixes #1074

## Proposed Changes

- :gift: Add new feature

- Add `PROMETHEUS_ALLOW_HTTP` (default `false`) to opt in to a plain `http://` `PROMETHEUS_BASE_URL`.
- `ValidateTLSConfig` now branches on URL scheme: `https://` keeps existing validation; `http://` requires the new flag and explicitly rejects being combined with bearer-token auth or any TLS-specific setting (CA cert, client cert, `insecureSkipVerify`, server name) to avoid a false sense of security; any other scheme is rejected outright.
- `CreatePrometheusTransport` only builds/attaches a `tls.Config` for `https://` endpoints — plain HTTP uses the default transport unmodified.
- Updated the "Initializing Prometheus client" startup log to report the actual `tlsEnabled`/`allowHTTP` state instead of a hardcoded `tlsEnabled: true`.
- Documented the new variable in `docs/developer-guide/prometheus.md` (env var reference table + a dev/testing note in Security Considerations).

### Pre-review Checklist

- [ ] **E2E tests** for any new behavior — this only affects local config validation/transport setup, no new e2e coverage added; let me know if you'd like one added for an HTTP-Prometheus scenario
- [x] **Docs PR** for any user-facing impact — updated `docs/developer-guide/prometheus.md` in this PR
- [ ] **Proposal PR** for any new enhancement or change to existing behavior — this is a small, additive opt-in flag; happy to write a proposal if maintainers want one

**Release Note**

```release-note
Add `PROMETHEUS_ALLOW_HTTP` to allow connecting to a plain `http://` Prometheus endpoint. Disabled by default; cannot be combined with TLS settings or bearer token auth.
```

**Docs**

Updated `docs/developer-guide/prometheus.md` (env var table + Security Considerations section) in this PR.

## How tested

- `go build ./...` and `go vet ./...` clean.
- `go test ./internal/config/... ./internal/utils/...` passes, including new cases:
  - `TestValidateTLSConfig`: HTTP without opt-in (fail), HTTP with opt-in (pass), HTTP + CA cert (fail), HTTP + bearer token (fail), unsupported scheme (fail)
  - `TestLoad_PrometheusAllowHTTPFromEnv`: confirms the env var loads correctly
- `make test` run locally: the only failure is a pre-existing, unrelated flaky case in `internal/engines/analyzers/throughput/itl_model_test.go`, confirmed present on `main` without this change.
- Could not run `make lint` locally — `golangci-lint` panics in this environment on a Go toolchain version mismatch (`file requires newer Go version go1.26 (application built with go1.25)`), unrelated to this diff; `gofmt`/`go vet` are clean.

## Notes

I'm aware a prior PR (#1075) attempted this same issue and went stale from inactivity after a maintainer review (not rejection) — addressed similarly here but reimplemented independently rather than reusing that diff.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -421,15 +422,16 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Always validate TLS configuration since HTTPS is required
+	// Validate Prometheus transport configuration before creating the client.
 	if err := utils.ValidateTLSConfig(cfg); err != nil {
-		setupLog.Error(err, "TLS configuration validation failed - HTTPS is required")
+		setupLog.Error(err, "Prometheus transport configuration validation failed")
 		os.Exit(1)
 	}
 
 	setupLog.Info("Initializing Prometheus client",
 		"address", cfg.PrometheusBaseURL(),
-		"tlsEnabled", true,
+		"tlsEnabled", strings.HasPrefix(cfg.PrometheusBaseURL(), "https://"),
+		"allowHTTP", cfg.PrometheusAllowHTTP(),
 	)
 
 	// Create Prometheus client with TLS support
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -25,7 +25,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -430,7 +429,7 @@ func main() {
 
 	setupLog.Info("Initializing Prometheus client",
 		"address", cfg.PrometheusBaseURL(),
-		"tlsEnabled", strings.HasPrefix(cfg.PrometheusBaseURL(), "https://"),
+		"tlsEnabled", utils.IsHTTPS(cfg.PrometheusBaseURL()),
 		"allowHTTP", cfg.PrometheusAllowHTTP(),
 	)
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	goflag "flag"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"time"
@@ -427,8 +428,9 @@ func main() {
 		os.Exit(1)
 	}
 
+	promURL, _ := url.Parse(cfg.PrometheusBaseURL()) // already validated above
 	setupLog.Info("Initializing Prometheus client",
-		"address", cfg.PrometheusBaseURL(),
+		"address", promURL.Redacted(),
 		"tlsEnabled", utils.IsHTTPS(cfg.PrometheusBaseURL()),
 		"allowHTTP", cfg.PrometheusAllowHTTP(),
 	)
--- a/internal/config/prometheus.go
+++ b/internal/config/prometheus.go
@@ -12,6 +12,7 @@ type prometheusConfig struct {
 	bearerToken        string
 	tokenPath          string
 	insecureSkipVerify bool
+	allowHTTP          bool
 	caCertPath         string
 	clientCertPath     string
 	clientKeyPath      string
@@ -94,6 +95,14 @@ func (c *Config) PrometheusInsecureSkipVerify() bool {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 	return c.prometheus.insecureSkipVerify
+}
+
+// PrometheusAllowHTTP returns whether plain HTTP connections to Prometheus are allowed.
+// Thread-safe.
+func (c *Config) PrometheusAllowHTTP() bool {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.prometheus.allowHTTP
 }
 
 // PrometheusCACertPath returns the Prometheus CA certificate path.
--- a/internal/utils/prometheus_transport.go
+++ b/internal/utils/prometheus_transport.go
@@ -21,7 +21,7 @@ func CreatePrometheusTransport(cfg *config.Config) (http.RoundTripper, error) {
 	// Clone the default transport to get all the good defaults
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 
-	if strings.HasPrefix(cfg.PrometheusBaseURL(), "https://") {
+	if IsHTTPS(cfg.PrometheusBaseURL()) {
 		tlsConfig, err := CreateTLSConfig(cfg)
 		if err != nil {
 			return nil, err
@@ -72,7 +72,7 @@ func CreatePrometheusClientConfig(cfg *config.Config) (*api.Config, error) {
 	return clientConfig, nil
 }
 
-// bearerTokenRoundTripper adds bearer token authentication to HTTPS requests
+// bearerTokenRoundTripper adds bearer token authentication to HTTP and HTTPS requests
 type bearerTokenRoundTripper struct {
 	base  http.RoundTripper
 	token string
--- a/internal/utils/prometheus_transport.go
+++ b/internal/utils/prometheus_transport.go
@@ -72,14 +72,18 @@ func CreatePrometheusClientConfig(cfg *config.Config) (*api.Config, error) {
 	return clientConfig, nil
 }
 
-// bearerTokenRoundTripper adds bearer token authentication to HTTP and HTTPS requests
+// bearerTokenRoundTripper adds bearer token authentication to HTTPS requests.
+// The header is only set for https:// requests to prevent credential leakage
+// if a redirect leads to a plain http:// endpoint.
 type bearerTokenRoundTripper struct {
 	base  http.RoundTripper
 	token string
 }
 
-// RoundTrip adds the Authorization header with bearer token
+// RoundTrip adds the Authorization header for HTTPS requests only.
 func (b *bearerTokenRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
-	req.Header.Set("Authorization", "Bearer "+b.token)
+	if req.URL.Scheme == "https" {
+		req.Header.Set("Authorization", "Bearer "+b.token)
+	}
 	return b.base.RoundTrip(req)
 }
--- a/internal/utils/prometheus_transport.go
+++ b/internal/utils/prometheus_transport.go
@@ -13,24 +13,27 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
 )
 
-// CreatePrometheusTransport creates a custom HTTPS transport for Prometheus client with TLS support.
-// TLS is always enabled for HTTPS-only support with configurable certificate validation.
+// CreatePrometheusTransport creates a custom transport for the Prometheus client.
+// TLS configuration is only applied for https:// endpoints; plain http://
+// endpoints (allowed only when PROMETHEUS_ALLOW_HTTP is set, see ValidateTLSConfig)
+// use the default transport unmodified.
 func CreatePrometheusTransport(cfg *config.Config) (http.RoundTripper, error) {
 	// Clone the default transport to get all the good defaults
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 
-	// Configure TLS (always required for HTTPS-only support)
-	tlsConfig, err := CreateTLSConfig(cfg)
-	if err != nil {
-		return nil, err
+	if strings.HasPrefix(cfg.PrometheusBaseURL(), "https://") {
+		tlsConfig, err := CreateTLSConfig(cfg)
+		if err != nil {
+			return nil, err
+		}
+		transport.TLSClientConfig = tlsConfig
+		ctrl.Log.V(logging.VERBOSE).Info("TLS configuration applied to Prometheus HTTPS transport")
 	}
-	transport.TLSClientConfig = tlsConfig
-	ctrl.Log.V(logging.VERBOSE).Info("TLS configuration applied to Prometheus HTTPS transport")
 
 	return transport, nil
 }
 
-// CreatePrometheusClientConfig creates a complete Prometheus client configuration with HTTPS support.
+// CreatePrometheusClientConfig creates a complete Prometheus client configuration.
 // Supports both direct bearer tokens and token files for flexible authentication.
 func CreatePrometheusClientConfig(cfg *config.Config) (*api.Config, error) {
 	clientConfig := &api.Config{
--- a/internal/utils/prometheus_transport.go
+++ b/internal/utils/prometheus_transport.go
@@ -82,7 +82,7 @@ type bearerTokenRoundTripper struct {
 
 // RoundTrip adds the Authorization header for HTTPS requests only.
 func (b *bearerTokenRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
-	if req.URL.Scheme == "https" {
+	if req.URL.Scheme == schemeHTTPS {
 		req.Header.Set("Authorization", "Bearer "+b.token)
 	}
 	return b.base.RoundTrip(req)
--- a/internal/utils/prometheus_transport.go
+++ b/internal/utils/prometheus_transport.go
@@ -40,7 +40,7 @@ func CreatePrometheusClientConfig(cfg *config.Config) (*api.Config, error) {
 		Address: cfg.PrometheusBaseURL(),
 	}
 
-	// Create custom HTTPS transport with TLS support
+	// Create the custom transport; TLS is applied only for https:// endpoints.
 	transport, err := CreatePrometheusTransport(cfg)
 	if err != nil {
 		return nil, err
--- a/internal/utils/tls.go
+++ b/internal/utils/tls.go
@@ -15,7 +15,7 @@ import (
 )
 
 // IsHTTPS reports whether rawURL uses the https scheme.
-// It uses url.Parse so the comparison is case-insensitive per RFC 3986.
+// url.Parse normalizes scheme to lowercase, so the comparison is case-insensitive.
 func IsHTTPS(rawURL string) bool {
 	u, err := url.Parse(rawURL)
 	if err != nil {
@@ -105,6 +105,9 @@ func ValidateTLSConfig(cfg *config.Config) error {
 	case "https":
 		// Continue with the HTTPS-specific validation below.
 	case "http":
+		if u.User != nil {
+			return fmt.Errorf("refusing to use Prometheus URL with embedded credentials %q; use PROMETHEUS_BEARER_TOKEN or PROMETHEUS_TOKEN_PATH instead", u.Redacted())
+		}
 		if !allowHTTP {
 			return fmt.Errorf("plain HTTP Prometheus URL %q is not allowed; set PROMETHEUS_ALLOW_HTTP=true to permit http:// endpoints", baseURL)
 		}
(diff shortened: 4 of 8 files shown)


internal/config/prometheus.go
```diff
@@ -96,6 +97,14 @@ func (c *Config) PrometheusInsecureSkipVerify() bool {
 	return c.prometheus.insecureSkipVerify
 }
 
+// PrometheusAllowHTTP returns whether plain HTTP connections to Prometheus are allowed.
+// Thread-safe.
+func (c *Config) PrometheusAllowHTTP() bool {
```
ev-shindin, line 102: Scheme detection is inconsistent across the codebase and should be centralized here.
`ValidateTLSConfig` derives the scheme via `url.Parse(...).Scheme` (RFC 3986 → lowercased), but
`cmd/main.go:433` and `internal/utils/prometheus_transport.go:24` use case-sensitive
`strings.HasPrefix(url, "https://")`. For a URL like `HTTPS://prom:9090` these disagree (see the
two comments on those lines). Please add a single shared helper and use it at both call sites.

internal/utils/prometheus_transport.go
```diff
@@ -13,24 +13,27 @@ import (
...
+// use the default transport unmodified.
 func CreatePrometheusTransport(cfg *config.Config) (http.RoundTripper, error) {
 	// Clone the default transport to get all the good defaults
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 
-	// Configure TLS (always required for HTTPS-only support)
-	tlsConfig, err := CreateTLSConfig(cfg)
-	if err != nil {
-		return nil, err
+	if strings.HasPrefix(cfg.PrometheusBaseURL(), "https://") {
```
ev-shindin: Case-sensitive `HasPrefix("https://")` disagrees with `ValidateTLSConfig` (which lowercases via
`url.Parse`). For `HTTPS://...` it skips `CreateTLSConfig`, silently dropping custom CA / client
cert / `serverName` → runtime handshake failure after validation passed. Use shared helper.

cmd/main.go
```diff
@@ -421,15 +422,16 @@ func main() {
...
 	if err := utils.ValidateTLSConfig(cfg); err != nil {
-		setupLog.Error(err, "TLS configuration validation failed - HTTPS is required")
+		setupLog.Error(err, "Prometheus transport configuration validation failed")
 		os.Exit(1)
 	}
 
 	setupLog.Info("Initializing Prometheus client",
 		"address", cfg.PrometheusBaseURL(),
-		"tlsEnabled", true,
+		"tlsEnabled", strings.HasPrefix(cfg.PrometheusBaseURL(), "https://"),
```
ev-shindin: Same mismatch: `HTTPS://` logs `tlsEnabled=false`.

internal/utils/prometheus_transport.go
ev-shindin: Stale doc: `bearerTokenRoundTripper` now wraps HTTP too, not just HTTPS.

internal/utils/tls.go
ev-shindin: Stale doc: "TLS is always enabled for HTTPS-only support" — no longer true; only called for `https://` now.

internal/utils/tls_test.go
```diff
@@ -146,12 +147,45 @@ func TestValidateTLSConfig(t *testing.T) {
 			expectError: true,
 		},
 		{
-			name: "HTTP URL - should fail",
+			name: "HTTP URL without opt-in - should fail",
```
ev-shindin, line 174: Only `caCertPath` is tested of the 5-field TLS guard. Add `expectError: true` cases (HTTP + opt-in)
for: `INSECURE_SKIP_VERIFY`, `CLIENT_CERT_PATH`, `SERVER_NAME`, and `TOKEN_PATH` (only bearer is tested).

internal/utils/prometheus_transport.go
```diff
@@ -13,24 +13,27 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
 )
 
-// CreatePrometheusTransport creates a custom HTTPS transport for Prometheus client with TLS support.
-// TLS is always enabled for HTTPS-only support with configurable certificate validation.
+// CreatePrometheusTransport creates a custom transport for the Prometheus client.
+// TLS configuration is only applied for https:// endpoints; plain http://
+// endpoints (allowed only when PROMETHEUS_ALLOW_HTTP is set, see ValidateTLSConfig)
+// use the default transport unmodified.
 func CreatePrometheusTransport(cfg *config.Config) (http.RoundTripper, error) {
```
ev-shindin, line 20: Central change is untested. Add cases: `http://`+opt-in → `TLSClientConfig == nil`; `https://` → `!= nil`.

ev-shindin (changes requested):
Thanks @Goutham-Annem! Please address my review.

Goutham-Annem:
Hi @ev-shindin, I've addressed all the review points in the latest push — `IsHTTPS()` helper added and used at both call sites, stale docs updated, and test coverage added for the TLS guard and transport. Please take another look when you have a chance!

ev-shindin:
@Goutham-Annem please sign-off last commit and fix linter issues.

Goutham-Annem:
Fixed — amended the last commit with DCO sign-off and resolved both linter issues:
- `gofmt`: realigned map literal spacing after introducing the `envPrometheusAllowHTTP` constant
- `gocritic`: collapsed `else { if }` to `else if` in the TLS transport test
- `goconst`: extracted the 9-occurrence `"PROMETHEUS_ALLOW_HTTP"` string to a package-level constant

All tests pass and `golangci-lint run --new-from-rev` reports 0 issues on the changed files. Please take another look when you get a chance!

internal/utils/tls_test.go
```diff
@@ -146,12 +151,81 @@ func TestValidateTLSConfig(t *testing.T) {
...
+			name: "HTTP URL with opt-in and client cert path - should fail",
+			promConfig: testConfigFromEnv(t, map[string]string{
+				"PROMETHEUS_BASE_URL":         "http://prometheus:9090",
+				envPrometheusAllowHTTP:        "true",
+				"PROMETHEUS_CLIENT_CERT_PATH": "/tmp/client.crt",
+			}),
+			expectError: true,
+		},
+		{
+			name: "HTTP URL with opt-in and server name - should fail",
```
ev-shindin, line 251: `ValidateTLSConfig` rejects `insecureSkipVerify || caCertPath || clientCertPath || clientKeyPath || serverName` for `http://`. The new table covers all of those **except `clientKeyPath`** — so removing it from the `||` chain would be a silent regression that every existing case still passes. Add a case next to the server-name one.

internal/utils/tls.go
```diff
@@ -71,23 +81,43 @@ func CreateTLSConfig(cfg *config.Config) (*tls.Config, error) {
...
+		// Continue with the HTTPS-specific validation below.
+	case "http":
+		if !allowHTTP {
+			return fmt.Errorf("plain HTTP Prometheus URL %q is not allowed; set PROMETHEUS_ALLOW_HTTP=true to permit http:// endpoints", baseURL)
+		}
+		if cfg.PrometheusBearerToken() != "" || cfg.PrometheusTokenPath() != "" {
+			return fmt.Errorf("refusing to send bearer token authentication over plain HTTP Prometheus URL %q", baseURL)
+		}
+		if insecureSkipVerify || caCertPath != "" || clientCertPath != "" || clientKeyPath != "" || cfg.PrometheusServerName() != "" {
+			return fmt.Errorf("TLS-related settings are not supported with plain HTTP Prometheus URL %q; remove them or use an https:// URL", baseURL)
```
ev-shindin, line 128: The HTTP branch carefully refuses bearer tokens/token-paths, but `http://user:pass@host` passes validation and the client then sends HTTP Basic credentials in cleartext. To stay consistent with the PR's own "no credentials over plain HTTP" rule, reject userinfo in the `http` case (alongside the existing checks).

internal/utils/tls.go
```diff
@@ -71,23 +81,43 @@ func CreateTLSConfig(cfg *config.Config) (*tls.Config, error) {
...
+		}
+		ctrl.Log.Info("Plain HTTP Prometheus endpoint allowed by configuration", "address", baseURL)
```
ev-shindin: If the URL embeds userinfo, the password is logged in cleartext. Log `u.Redacted()` (or a redacted copy) instead of the raw `baseURL`.

internal/utils/tls.go
```diff
@@ -14,8 +14,18 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
 )
 
+// IsHTTPS reports whether rawURL uses the https scheme.
+// It uses url.Parse so the comparison is case-insensitive per RFC 3986.
+func IsHTTPS(rawURL string) bool {
```
ev-shindin, line 24: The new exported helper is exercised only indirectly via `TestCreatePrometheusTransport`. Add a `TestIsHTTPS` table: `http://…`→false, `https://…`→true, `""`→false, `HTTPS://…`→true (documents the case-insensitive claim), malformed→false.

internal/utils/tls.go
```diff
@@ -14,8 +14,18 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
 )
 
+// IsHTTPS reports whether rawURL uses the https scheme.
+// It uses url.Parse so the comparison is case-insensitive per RFC 3986.
```
ev-shindin: Nit: `IsHTTPS` doc says "case-insensitive per RFC 3986" — technically `url.Parse` normalizes the scheme to lowercase before the comparison; phrasing is slightly loose.

internal/utils/tls.go
```diff
@@ -71,23 +81,43 @@ func CreateTLSConfig(cfg *config.Config) (*tls.Config, error) {
...
+		if cfg.PrometheusBearerToken() != "" || cfg.PrometheusTokenPath() != "" {
+			return fmt.Errorf("refusing to send bearer token authentication over plain HTTP Prometheus URL %q", baseURL)
+		}
+		if insecureSkipVerify || caCertPath != "" || clientCertPath != "" || clientKeyPath != "" || cfg.PrometheusServerName() != "" {
+			return fmt.Errorf("TLS-related settings are not supported with plain HTTP Prometheus URL %q; remove them or use an https:// URL", baseURL)
+		}
+		ctrl.Log.Info("Plain HTTP Prometheus endpoint allowed by configuration", "address", baseURL)
+		return nil
+	default:
+		return fmt.Errorf("unsupported Prometheus URL scheme %q in %q; expected http or https", u.Scheme, baseURL)
```
ev-shindin, line 133: Empty `PROMETHEUS_BASE_URL` produces a misleading `unsupported scheme ""` error (pre-existing class). Optional: a clearer "PROMETHEUS_BASE_URL is required" message

cmd/main.go
```diff
@@ -421,15 +421,16 @@ func main() {
...
-	// Always validate TLS configuration since HTTPS is required
+	// Validate Prometheus transport configuration before creating the client.
 	if err := utils.ValidateTLSConfig(cfg); err != nil {
-		setupLog.Error(err, "TLS configuration validation failed - HTTPS is required")
+		setupLog.Error(err, "Prometheus transport configuration validation failed")
 		os.Exit(1)
 	}
 
 	setupLog.Info("Initializing Prometheus client",
 		"address", cfg.PrometheusBaseURL(),
```
ev-shindin: If the URL embeds userinfo, the password is logged in cleartext. Log `u.Redacted()` (or a redacted copy) instead of the raw `baseURL`.

internal/utils/prometheus_transport.go
```diff
@@ -69,7 +72,7 @@ func CreatePrometheusClientConfig(cfg *config.Config) (*api.Config, error) {
 	return clientConfig, nil
 }
 
-// bearerTokenRoundTripper adds bearer token authentication to HTTPS requests
+// bearerTokenRoundTripper adds bearer token authentication to HTTP and HTTPS requests
```
ev-shindin: `bearerTokenRoundTripper.RoundTrip` sets `Authorization` on every request with no per-request scheme check; a same-host `https→http` redirect would re-emit the token over cleartext (Go strips `Authorization` only across *different* hosts). Worth a follow-up `if req.URL.Scheme == "https"` guard.

ev-shindin (commented):
Thanks @Goutham-Annem . This now in a better shape and in a one step before approval. Please see my comments below.

internal/utils/tls_test.go
ev-shindin: This one resolved partially - see follow-up comments.

ev-shindin:
Hi @Goutham-Annem ! Do you plan this PR for the nearby release? If yes please try to fix it ASAP. I am going to vacation for a two weeks and wil be busy on return.

Goutham-Annem:
Thanks for the detailed review! I've addressed all the comments in the latest commit:

- **`TestIsHTTPS`**: added table-driven test covering `https://`, `HTTPS://` (uppercase), `http://`, empty string, invalid URL, and no-scheme cases
- **`clientKeyPath` test**: added `HTTP URL with opt-in and client key path - should fail` case in `TestValidateTLSConfig`
- **Embedded credentials in HTTP URL**: `ValidateTLSConfig` now rejects URLs with userinfo (e.g. `http://user:pass@host`) in the http branch, returning an error pointing to `PROMETHEUS_BEARER_TOKEN`/`PROMETHEUS_TOKEN_PATH`
- **Redacted URL logging**: both `tls.go` (plain HTTP allowed log) and `cmd/main.go` (Prometheus client init log) now use `u.Redacted()` instead of the raw URL string
- **`bearerTokenRoundTripper.RoundTrip`**: now only sets the `Authorization` header when `req.URL.Scheme == "https"` to prevent credential leakage on redirects
- **`IsHTTPS` doc nit**: updated to "url.Parse normalizes scheme to lowercase" (not RFC 3986)

Goutham-Annem:
Hi — just following up. I've addressed all the review comments from the last round and CI is green. Could you take another look when you have a moment? Happy to adjust anything further.

Goutham-Annem:
Hi @ev-shindin — thanks for the thorough review. Walking through each comment:

**Scheme detection inconsistency** (`strings.HasPrefix` vs `url.Parse`): Consolidated into the `IsHTTPS()` helper in `internal/utils/tls.go` which uses `url.Parse` throughout. Both `cmd/main.go` and `prometheus_transport.go` now call `utils.IsHTTPS()`.

**`HTTPS://` logs `tlsEnabled=false`**: Fixed — `cmd/main.go:434` now uses `utils.IsHTTPS()` which handles uppercase schemes correctly.

**Stale doc on `bearerTokenRoundTripper`**: The `RoundTrip` method already has `if req.URL.Scheme == schemeHTTPS` — it only injects the token for https requests, which also guards against same-host https→http redirect leakage.

**Stale doc "TLS is always enabled for HTTPS-only support"**: Updated — the transport comment now accurately reflects that HTTP is allowed when `PROMETHEUS_ALLOW_HTTP` is set.

**`http://user:pass@host` passes validation**: Rejected — `ValidateTLSConfig` checks `u.User != nil` for the `http` branch and returns an error with `u.Redacted()`.

**Password logged in cleartext**: The log statement uses `u.Redacted()` in `ValidateTLSConfig`. `CreatePrometheusClientConfig` doesn't log the URL.

**Missing `clientKeyPath` test case**: Added — `TestValidateTLSConfig` now has "HTTP URL with opt-in and client key path - should fail".

**Missing `INSECURE_SKIP_VERIFY`, `CLIENT_CERT_PATH`, `SERVER_NAME`, `TOKEN_PATH` cases**: All added to `TestValidateTLSConfig`.

**`http://`+opt-in → `TLSClientConfig == nil`; `https://` → `!= nil`**: Covered by `TestCreatePrometheusTransport`.

**`TestIsHTTPS` table**: Added — covers `https://`, `HTTPS://`, `http://`, empty string, invalid URL, no scheme.

**Doc nit on `IsHTTPS`**: Fixed in the latest commit — changed "case-insensitive per RFC 3986" to clarify it's `url.Parse` that normalizes the scheme to lowercase.

All tests pass locally. Would you be able to take another look?

Goutham-Annem:
Hi @ev-shindin — all the feedback from your last review (2026-06-29) has been addressed:

- Centralized scheme detection via `IsHTTPS` helper ✓
- Added `TestIsHTTPS` table-driven test ✓
- Reject URLs with embedded userinfo credentials ✓
- Redact logged URLs ✓
- Guard bearer token against plain HTTP ✓
- Fixed doc comment on `IsHTTPS` (url.Parse normalizes scheme) ✓

Could you take another look when you get a chance? Happy to make any further adjustments!

ev-shindin:
Really nice change — secure-by-default, the `http` branch exhaustively rejects unsafe combos, TLS scheme-gating is clean (no `tls.Config` on `http`, `https` always gets TLS 1.2+), the bearer round-tripper is `https`-only, validation is fail-fast at startup, and the `http` rejection matrix is thoroughly unit-tested. A few things:

## Should-fix

- **`https` doesn't reject embedded URL credentials (`tls.go`).** The `http` branch rejects `u.User`, but the `https` case is a pass-through — `https://user:pass@prom:9090` is accepted and sent as Basic auth automatically. Credentials in a URL leak even over TLS (config dumps, process listings, redirect `Referer`, un-redacted logs), which is exactly why you reject it for `http`. Move the `u.User != nil` check above the `switch` so both schemes reject it.

## Nice-to-have

- The `allowHTTP` gate is enforced only in `ValidateTLSConfig`, not in transport creation. Safe today (`main.go` validates + `os.Exit(1)` before the client is built), but a future caller that skips validation would silently get cleartext — consider gating in the transport layer so the invariant is local.
- `docs/developer-guide/prometheus.md`: the incompatibility list omits `PROMETHEUS_SERVER_NAME`, which `ValidateTLSConfig` also rejects over `http` (`tls.go`).
- `internal/utils/prometheus_transport.go`: the "Create custom **HTTPS** transport" comment now handles both schemes.
- Add a `schemeHTTP` const to match `schemeHTTPS` (bare `"http"` literal in the `switch`).
- Pre-existing but in scope: the three cert-file-missing branches on the `https` path (`tls.go`) have no test — a regression there would pass the suite. Three table rows would close it.

## Nit

- The embedded-credentials error suggests `PROMETHEUS_BEARER_TOKEN`, but bearer-over-`http` is rejected two lines later — the advice only applies once the user also switches to `https`.

Only the first item is worth blocking on; the rest are polish. Happy to push these as maintainer edits.

ev-shindin (approved):
approved after the fix, Thanks @Goutham-Annem

Goutham-Annem:
Hi @lionelvillard — ev-shindin approved this morning. Could you take a look and merge when you get a chance? Happy to address anything else if needed!

Goutham-Annem:
The `e2e-tests-full-keda` failure (LWS scale-from-zero timeout) is a pre-existing infrastructure issue unrelated to this PR — the same failure is occurring on the active `chore/retire-prom-adapter` branch as well (runs 29068077147 and 29064502442), indicating the KEDA e2e environment is currently unstable independently of this change. All other checks (smoke, smoke-keda, full non-KEDA, lint, unit tests) are green.

Could a maintainer merge this given the pre-existing infra issue and ev-shindin's approval? Happy to rebase or make any other adjustments needed.

ev-shindin:
@Goutham-Annem  thanks.  I merged this.

PR #1341, opened 2026-06-25
fix: exclude KEDA-generated HPAs from the collector's managed-HPA index

### LINKED ISSUE #1333: KEDA-generated HPA inherits llm-d.ai/managed, causing ambiguous scale target in the collector
## Summary

In annotation-discovery mode, a managed KEDA ScaledObject and the HPA KEDA generates from it both carry llm-d.ai/managed: "true". The metrics collector's resolveScaler then matches two managed scalers for the same scaleTargetRef and fails with ambiguous scale target. 

## Version

- WVA v0.8.0 (controller image v0.8.0).
- Annotation mode (KEDA ScaledObject with llm-d.ai/managed, no VariantAutoscaling CR).
- Deployment-only layout (no LeaderWorkerSet).

## Symptom

The controller is healthy, leader-elected, and querying Prometheus successfully, but logs every interval:

```
collector/replica_metrics.go:858  Skipping pod that doesn't match any scale target  {"pod":"<model>-...", "scale targets":["<model>"]}
saturation/engine.go:594          No saturation metrics available for model, skipping analysis  {"modelID":"<model-id>"}
saturation/engine.go:1736         Safety net activated: emitted fallback metrics  {"fallbackSource":"current-replicas"}
```

`wva_desired_replicas` is emitted but only ever equals the fallback current replica count. Real saturation optimization never runs even though the vLLM metrics are present and queryable.

## Root cause

This is what I believe to be the evaluation path.

1. KEDA propagates the ScaledObject's annotations onto its generated HPA at creation time (kedacore/keda#5468). So a ScaledObject annotated llm-d.ai/managed: "true" yields a keda-hpa-* that also carries llm-d.ai/managed: "true".

2. The collector's managed-HPA index filters only on the annotation, with no ownerRef exclusion e.g. https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/e0ad326a0c030779d644d93af5c190adf04499cb/internal/controller/indexers/hpa.go#L27

```go
if !annotations.IsManaged(hpa) {
    return nil
}
```

3. So the KEDA-generated child HPA is indexed as a managed scaler, and FindHPAForScaleTarget (same file, :41) returns it for the Deployment.

4. resolveScaler then sees both a managed HPA and the managed ScaledObject for the same target and [errors](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/e0ad326a0c030779d644d93af5c190adf04499cb/internal/collector/locator/locator.go#L230).

```go
case hpa != nil && so != nil:
    return nil, fmt.Errorf("ambiguous scale target %s/%s/%s: matched HPA %q and ScaledObject %q", ...)
```

5. In [buildInstanceKey](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/e0ad326a0c030779d644d93af5c190adf04499cb/internal/collector/locator/locator.go#L241) the vLLM series has no llm_d_ai_variant label, so it falls back to locator.Locate(...), which returns the ambiguity error; the error is logged at V(DEBUG) ("locator.Locate failed; treating pod as unmanaged") and vaName is left empty → the pod is skipped (:858) → the saturation set is empty → safety net.

This class of concern is already handled by the Coordinator [elsewhere](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/e0ad326a0c030779d644d93af5c190adf04499cb/internal/coordinator/selection.go#L35).

```go
func IsHPAUnderControl(hpa *autoscalingv2.HorizontalPodAutoscaler) bool {
    ...
    if isOwnedByKEDAScaledObject(hpa) {   // ownerRef Kind=ScaledObject, group=keda.sh
        return false
    }
    return true
}
```

## Reproduction

1. Install WVA v0.8.0 in annotation mode (no prometheus-operator; e.g. GMP, or any setup where the collector path is exercised).
2. Create a KEDA ScaledObject for a Deployment with metadata.annotations: { llm-d.ai/managed: "true", llm-d.ai/model-id: <id> } and a wva_desired_replicas trigger.
3. Observe that KEDA's generated keda-hpa-<name> inherits llm-d.ai/managed: "true":
kubectl get scaledobject <name> -o jsonpath='{.metadata.annotations.llm-d\.ai/managed}'   # true
kubectl get hpa keda-hpa-<name> -o jsonpath='{.metadata.annotations.llm-d\.ai/managed}'    # true

ieaves:
Fixes #1333

## Proposed Changes

- Exclude KEDA-generated child HPAs from the collector's managed-HPA index (`HPAByScaleTargetIndexFunc`), fixing the `ambiguous scale target` failure that left annotation-mode variants stuck on the current replicas safety net fallback instead of real saturation/optimization.
- Factor the KEDA-ownership check into a shared `utils.IsOwnedByKEDAScaledObject` (with `ScaledObjectKind` /`ScaledObjectAPIGroup` constants); the Coordinator's `IsHPAUnderControl` now uses the same predicate, so there is one source of truth. 
- Add a regression test: a KEDA-owned managed HPA is not indexed; a managed non-KEDA HPA still is.

### Pre-review Checklist

- [ ] **Proposal PR** for any new enhancement

This is a bug fix so no documentation / proposal pr was included and this does not introduce any new behavior.

**Release Note**

```release-note
Fixed annotation-mode discovery so the metrics collector no longer fails with "ambiguous scale target" when a managed KEDA ScaledObject's generated HPA inherits the llm-d.ai/uns real saturation/optimization for such variantsinstead of emitting only the current-replicas fallback.
```

**Docs**

No documentation change. This is a bug fix restoring intended behavior with no user facing config or usage change.

### DIFF
--- a/internal/controller/indexers/hpa_index_test.go
+++ b/internal/controller/indexers/hpa_index_test.go
@@ -1,19 +1,3 @@
-/*
-Copyright 2025.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
 package indexers
 
 import (
--- a/internal/controller/indexers/hpa_index_test.go
+++ b/internal/controller/indexers/hpa_index_test.go
@@ -1,0 +1,64 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package indexers
+
+import (
+	"testing"
+
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/annotations"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/constants"
+)
+
+// A KEDA-generated HPA inherits the ScaledObject's llm-d.ai/managed annotation
+// (kedacore/keda#5468). It must NOT be indexed as a managed scaler, or the
+// collector's locator sees two managed scalers for one target and fails with
+// "ambiguous scale target". Regression for #1333.
+func TestHPAByScaleTargetIndexFunc_SkipsKEDAOwnedHPA(t *testing.T) {
+	hpa := &autoscalingv2.HorizontalPodAutoscaler{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "keda-hpa-my-model",
+			Namespace:   "inference",
+			Annotations: map[string]string{annotations.Managed: "true"},
+			OwnerReferences: []metav1.OwnerReference{{
+				APIVersion: constants.ScaledObjectAPIGroup + "/v1alpha1",
+				Kind:       constants.ScaledObjectKind,
+				Name:       "my-model",
+				Controller: ptr.To(true),
+			}},
+		},
+		Spec: autoscalingv2.HorizontalPodAutoscalerSpec{
+			ScaleTargetRef: autoscalingv2.CrossVersionObjectReference{
+				Kind: constants.DeploymentKind,
+				Name: "my-model",
+			},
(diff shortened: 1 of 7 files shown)


internal/controller/indexers/hpa_index_test.go
```diff
@@ -0,0 +1,64 @@
+/*
```
lionelvillard: the license header should be removed.

internal/utils/utils.go
```diff
@@ -455,3 +455,26 @@ func extractGPUFromNodeSelectorTerm(term corev1.NodeSelectorTerm, gpuKeys []stri
...
+
+// IsOwnedByKEDAScaledObject reports whether obj has a controller OwnerReference
+// whose APIVersion is in the keda.sh API group and whose Kind is "ScaledObject".
+//
+// KEDA generates an HPA per ScaledObject and propagates the ScaledObject's
+// annotations (including llm-d.ai/managed) onto that child HPA. Callers that scan
+// for WVA-managed scalers must exclude such KEDA-generated HPAs: the ScaledObject
+// is the managed scaler, and treating its child HPA as a second managed scaler for
+// the same target is ambiguous.
+func IsOwnedByKEDAScaledObject(obj metav1.Object) bool {
```
lionelvillard: can you move this utility function to a new file, called `keda.go`?

lionelvillard:
thanks @ieaves for your contribution! I added 2 minor comments to address and then we can merge it.

ieaves:
@lionelvillard Absolute pleasure! I just pushed up the requested changes.

vivekk16 (approved):
/lgtm

lionelvillard:
@ieaves can you please rebase?

ieaves:
Of course!

lionelvillard (approved):

Issue #1106, opened 2026-05-07
Improved multi-vendor support


eero-t:
v0.6 release added partial support for other vendors' GPUs than NVIDIA, and setups mixing them, so that e.g. `kind` simulator can now be configured to mix devices from NVIDIA, AMD and Intel (real devices won't work yet though, due to #1104).

However, there are still places in Go code which refer only to Nvidia and AMD:
* internal/utils/scaletarget/lws_test.go::TestLWSAccessor_GetTotalGPUsPerReplica()
* internal/utils/utils_test.go::TestGetAcceleratorNameFromScaleTarget()
* test/e2e/limiter_test.go  (multiple places)

And which refer only to NVIDIA:
* .github/workflows/ci-e2e-openshift.yaml::gpu-check
* api/v1alpha1/variantautoscaling_types_test.go::makeValidVA()
* deploy/kubernetes/README.md#Troubleshooting
* hack/benchmark/dump_epp_fc_metrics/dump_all_metrics.py::get_node_gpus()
* hack/vllm-benchmark-deployment.yaml (resources/tolerations)
* test/benchmark/prefill_heavy_benchmark_test.go::monitorLoop  (Uses Nvidia product & "accelerator" labels)
* test/e2e/fixtures/model_service_builder.go::buildModelServiceResources()

Scripts supporting only NVIDIA:
* deploy/kubernetes/create-kind-cluster-with-nvidia.sh (AMD & Intel scripts missing)
* deploy/lib/discovery.sh (If no Nvidia resources, uses nvidia-smi)

And docs referring only to NVIDIA:
* deploy/openshift/README.md#Prerequisites
* docs/developer-guide/benchmark-guide.md#Prerequisites
* docs/benchmark.md#Environment

xuhui-lu:
hey @eero-t , does someone work on this issue already? If not, I could take a try here.

eero-t:
@xuhui-lu AFAIK nobody's working currently on these.

But note that current code supporting multiple GPUs is also a bit broken (see bug #1104) and needs fixes.  I have draft fix PR #1105 for those, which may need to be taken into account when looking into above files.

xuhui-lu:
> [@xuhui-lu](https://github.com/xuhui-lu) AFAIK nobody's working currently on these.
> 
[...]

PR #1350, opened 2026-06-27
test: cover LWS multi-vendor GPU resources

xuhui-lu:
## Summary
- add LWS GPU counting coverage for supported resource names: NVIDIA, AMD, Intel Gaudi, Intel i915, and Intel Xe
- add a mixed-vendor LWS case across leader and worker containers
- keep this as a unit-test-only follow-up for the multi-vendor GPU work

Part of #1106.

## Testing
- `go test ./internal/utils/scaletarget`
- `go test ./internal/utils/...`

Also tried `go test ./...`; it is not clean locally for unrelated reasons:
- `internal/engines/analyzers/throughput`: `FitITLModel` flat-line test currently fails
- `test/chart`: local `helm` binary is not installed
- `test/e2e`: entered cluster preflight / auth flow and was interrupted

### DIFF
--- a/internal/utils/scaletarget/lws_test.go
+++ b/internal/utils/scaletarget/lws_test.go
@@ -484,6 +484,135 @@ func TestLWSAccessor_GetTotalGPUsPerReplica(t *testing.T) {
 	}
 }
 
+func TestLWSAccessor_GetTotalGPUsPerReplica_SupportedGPUResources(t *testing.T) {
+	tests := []struct {
+		name         string
+		resourceName corev1.ResourceName
+		size         int32
+		leaderGPUs   string
+		workerGPUs   string
+		expected     int
+	}{
+		{
+			name:         "NVIDIA GPUs",
+			resourceName: "nvidia.com/gpu",
+			size:         4,
+			leaderGPUs:   "2",
+			workerGPUs:   "1",
+			expected:     5,
+		},
+		{
+			name:         "AMD GPUs",
+			resourceName: "amd.com/gpu",
+			size:         3,
+			leaderGPUs:   "2",
+			workerGPUs:   "3",
+			expected:     8,
+		},
+		{
+			name:         "Intel Gaudi GPUs",
+			resourceName: "habana.ai/gaudi",
+			size:         4,
+			leaderGPUs:   "1",
+			workerGPUs:   "2",
+			expected:     7,
+		},
+		{
+			name:         "Intel i915 GPUs",
+			resourceName: "gpu.intel.com/i915",
+			size:         2,
+			leaderGPUs:   "1",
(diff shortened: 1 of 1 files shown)


internal/utils/scaletarget/lws_test.go
```diff
@@ -484,6 +484,135 @@ func TestLWSAccessor_GetTotalGPUsPerReplica(t *testing.T) {
...
+			name:         "Intel Xe GPUs",
+			resourceName: "gpu.intel.com/xe",
+			size:         5,
+			leaderGPUs:   "3",
+			workerGPUs:   "1",
+			expected:     7,
+		},
+	}
+
+	for _, tt := range tests {
```
ev-shindin, line 541: The function name claims to cover all supported GPU resources, but it's 5 hardcoded rows with nothing tying it to `constants.VendorResources` (the authoritative 5-entry list). If a 6th vendor is added there, this test silently misses it while its name still implies full coverage. Add a one-line guard before the loop:

```go
assert.Len(t, tests, len(constants.VendorResources),
    "add a row when constants.VendorResources changes")
```

internal/utils/scaletarget/lws_test.go
```diff
@@ -484,6 +484,135 @@ func TestLWSAccessor_GetTotalGPUsPerReplica(t *testing.T) {
...
+
+			accessor := NewLWSAccessor(lws)
+			result := accessor.GetTotalGPUsPerReplica()
+
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestLWSAccessor_GetTotalGPUsPerReplica_MixedSupportedGPUResources(t *testing.T) {
```
ev-shindin, line 553: Nit: This puts two physical vendors (nvidia + gaudi) in one pod, which is an unusual real topology. Add a one-line comment that it's a synthetic check of intra-pod summation across all recognized GPU resource names, not a modeled deployment, so future readers don't mistake it for an expected production config.

ev-shindin (dismissed):
LGTM. Look to some minor (non-blocking) comments

internal/utils/scaletarget/lws_test.go
xuhui-lu: updated.

internal/utils/scaletarget/lws_test.go
xuhui-lu: updated.

ev-shindin (approved):
Thanks @xuhui-lu ! To follow-up: every existing and new test sets a non-zero GPU request, so the `if total == 0 { return 1 }` default is never exercised — change it to `return 0` and the whole suite still passes. Add a CPU-only (no GPU requests) LWS case asserting `1`

PR #1351, opened 2026-06-27
feat: add SGLang inference-engine backend support

MohanKumar21:
WVA is currently hard-wired to **vLLM**: every signal it collects assumes vLLM's
Prometheus metric names (`vllm:*`) and vLLM's deployment CLI flags. Operators
running **SGLang** model servers therefore can't use WVA at all — the collector
queries `vllm:*` series that SGLang pods never emit, so WVA sees no load and never
scales. This PR adds SGLang as a first-class inference-engine backend alongside
vLLM, with **zero behavior change for existing vLLM deployments**.

## Proposed Changes

- :gift: **Add SGLang as a second supported inference engine.** New
  `internal/inferenceengine` package detects the engine per variant from the pod
  spec (container image / launch command), defaulting to vLLM when no SGLang
  signal is present. Adds SGLang Prometheus metric constants and an SGLang
  deployment-arg parser (`ParseSGLangArgs` / `ParseEngineArgs`).
- :gift: **Engine-scoped metric queries.** Each engine-specific query now has an
  SGLang PromQL variant registered alongside the vLLM one (`EngineQuery` resolver;
  vLLM keeps its bare names). Includes the two *structural* differences, not just
  name swaps: KV-cache capacity via `sglang:max_total_num_tokens`, and prefix-cache
  hit rate via `sglang:cached_tokens_total / sglang:prompt_tokens_total`.
- :gift: **Engine-aware collection.** `CollectReplicaMetrics` detects the engines
  present for a model, refreshes the right query variants per engine, and merges
  per-engine results back under their logical names so per-pod processing stays
  engine-agnostic.
- :broom: **Refactor for reuse.** The vLLM arg-parse loop is extracted into a
  shared `parseArgsWith` helper used by both engines; the existing engine-params
  struct is reused (new `TotalKvTokensOverride` field for SGLang `--max-total-tokens`).
- :book: **Proposal doc** `docs/proposals/sglang-backend.md` with the full
  vLLM↔SGLang metric and CLI-flag mapping (sourced from SGLang's
  `observability/metrics_collector.py` and `server_args.py`).

vLLM remains the default and **all vLLM query/registration names are unchanged**,
so a deployment with no SGLang signal behaves exactly as before.

### Pre-review Checklist

- [x] **E2E tests** for any new behavior — `test/e2e/sglang_backend_test.go` plus a
  reusable fixture `test/e2e/fixtures/sglang_emulator_builder.go` (a CPU-only
  SGLang metrics emitter, launched with a faithful `sglang.launch_server` command):
  deploys it, wires a `ServiceMonitor`, creates a `VariantAutoscaling`, and asserts
  WVA detects SGLang and reconciles the variant from `sglang:*` metrics. Runs under
  `make test-e2e-*` (kind-emulator), like the rest of the e2e suite. Also covered by
  unit tests (detection, arg parsing, query registration, result merging) and an
  integration test that executes the registered SGLang template through the
  Prometheus source.
- [x] **Docs** for user-facing impact — `docs/user-guide/sglang-backend.md`
  (detection, metric/flag mapping, caveats). A guide in `llm-d/llm-d` is an optional
  follow-up (see Docs below).
- [x] **Proposal PR** for any new enhancement — `docs/proposals/sglang-backend.md`
  is included in this PR.

**Release Note**

```release-note
WVA now supports SGLang model servers in addition to vLLM. The inference engine is
auto-detected per variant from the deployment's pod spec (container image and launch
command), and SGLang's Prometheus metrics and serving flags are mapped onto WVA's
autoscaling signals (KV-cache utilization, queue depth, token throughput, latency,
and capacity).
```

**Docs**

TODO: link a follow-up docs/guide PR in llm-d/llm-d for SGLang backend usage.

### DIFF
--- a/internal/collector/engine_queries.go
+++ b/internal/collector/engine_queries.go
@@ -1,0 +1,105 @@
+package collector
+
+import (
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/registration"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/source"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/inferenceengine"
+)
+
+// engineSpecificReplicaQueries are the logical query names collected per replica
+// whose metric source is the inference engine (vLLM/SGLang). Each is refreshed
+// once per present engine and merged back under its logical name.
+var engineSpecificReplicaQueries = []string{
+	registration.QueryKvCacheUsage,
+	registration.QueryQueueLength,
+	registration.QueryCacheConfigInfo,
+	registration.QueryAvgOutputTokens,
+	registration.QueryAvgInputTokens,
+	registration.QueryPrefixCacheHitRate,
+	registration.QueryAvgTTFT,
+	registration.QueryAvgITL,
+	registration.QueryGenerationTokenRate,
+	registration.QueryKvUsageInstant,
+	registration.QueryVLLMRequestRate,
+}
+
+// agnosticReplicaQueries are the logical query names collected per replica whose
+// metric source is engine-independent (the EPP inference scheduler). They are
+// refreshed once and shared across engines.
+var agnosticReplicaQueries = []string{
+	registration.QuerySchedulerDispatchRate,
+}
+
+// buildEngineQueryList returns the physical query names to refresh for the given
+// set of present engines: each agnostic query once, plus each engine-specific
+// query for every present engine. For a single-engine vLLM model this is exactly
+// the agnostic queries plus the bare engine-specific names (unchanged behavior).
+func buildEngineQueryList(engines []inferenceengine.Engine, engineSpecific, agnostic []string) []string {
+	queries := make([]string, 0, len(agnostic)+len(engineSpecific)*len(engines))
+	queries = append(queries, agnostic...)
+	for _, logical := range engineSpecific {
+		for _, eng := range engines {
+			queries = append(queries, registration.EngineQuery(eng, logical))
+		}
+	}
+	return queries
+}
+
+// mergeEngineResults re-keys engine-specific query results under their logical
+// names so downstream per-pod processing is engine-agnostic.
+//
+//   - Single engine: the physical result is aliased to the logical name (a no-op
+//     for vLLM, where physical == logical; a rename for SGLang). The physical key
+//     is left in place so engine-specific consumers (e.g. the SGLang cache-config
+//     pass) can still read it.
+//   - Multiple engines: the per-engine series are concatenated into a new result
+//     under the logical name. Series are disjoint by pod, so concatenation is safe.
+func mergeEngineResults(results map[string]*source.MetricResult, engines []inferenceengine.Engine, logicalNames []string) {
+	if len(engines) == 1 {
+		eng := engines[0]
+		for _, logical := range logicalNames {
+			phys := registration.EngineQuery(eng, logical)
+			if phys == logical {
+				continue // vLLM: already keyed by logical name.
+			}
+			if r := results[phys]; r != nil {
+				r.QueryName = logical
+				results[logical] = r
+			}
+		}
+		return
+	}
+
+	for _, logical := range logicalNames {
+		var merged *source.MetricResult
+		for _, eng := range engines {
+			r := results[registration.EngineQuery(eng, logical)]
+			if r == nil {
+				continue
+			}
+			if merged == nil {
+				merged = &source.MetricResult{QueryName: logical, CollectedAt: r.CollectedAt}
+			}
+			merged.Values = append(merged.Values, r.Values...)
+			if merged.Error == nil && r.Error != nil {
+				merged.Error = r.Error
+			}
+			if !r.CollectedAt.IsZero() && (merged.CollectedAt.IsZero() || r.CollectedAt.Before(merged.CollectedAt)) {
+				merged.CollectedAt = r.CollectedAt
+			}
+		}
+		if merged != nil {
+			results[logical] = merged
+		}
+	}
+}
+
+// containsEngine reports whether the engine set includes the given engine.
+func containsEngine(engines []inferenceengine.Engine, target inferenceengine.Engine) bool {
+	for _, e := range engines {
+		if e == target {
+			return true
+		}
+	}
+	return false
+}
--- a/internal/collector/engine_queries.go
+++ b/internal/collector/engine_queries.go
@@ -1,6 +1,8 @@
 package collector
 
 import (
+	"slices"
+
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/registration"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/source"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/inferenceengine"
@@ -9,6 +11,13 @@ import (
 // engineSpecificReplicaQueries are the logical query names collected per replica
 // whose metric source is the inference engine (vLLM/SGLang). Each is refreshed
 // once per present engine and merged back under its logical name.
+//
+// This is the per-replica subset of registration.EngineSpecificQueries (the
+// authoritative list of engine-specific logical queries). Every entry here MUST
+// also appear there — TestEngineSpecificReplicaQueriesSubset enforces it so the
+// two lists cannot drift. It intentionally omits registration.QueryModelRequestCount,
+// which is engine-specific but a scale-to-zero query (collected on demand by the
+// enforcer, not per replica in this collector).
 var engineSpecificReplicaQueries = []string{
 	registration.QueryKvCacheUsage,
 	registration.QueryQueueLength,
@@ -20,7 +29,7 @@ var engineSpecificReplicaQueries = []string{
 	registration.QueryAvgITL,
 	registration.QueryGenerationTokenRate,
 	registration.QueryKvUsageInstant,
-	registration.QueryVLLMRequestRate,
+	registration.QueryRequestRate,
 }
 
 // agnosticReplicaQueries are the logical query names collected per replica whose
@@ -72,6 +81,7 @@ func mergeEngineResults(results map[string]*source.MetricResult, engines []infer
 
 	for _, logical := range logicalNames {
 		var merged *source.MetricResult
+		var firstErr error
 		for _, eng := range engines {
 			r := results[registration.EngineQuery(eng, logical)]
 			if r == nil {
@@ -81,14 +91,22 @@ func mergeEngineResults(results map[string]*source.MetricResult, engines []infer
 				merged = &source.MetricResult{QueryName: logical, CollectedAt: r.CollectedAt}
 			}
 			merged.Values = append(merged.Values, r.Values...)
-			if merged.Error == nil && r.Error != nil {
-				merged.Error = r.Error
+			if firstErr == nil && r.Error != nil {
+				firstErr = r.Error
 			}
 			if !r.CollectedAt.IsZero() && (merged.CollectedAt.IsZero() || r.CollectedAt.Before(merged.CollectedAt)) {
 				merged.CollectedAt = r.CollectedAt
 			}
 		}
 		if merged != nil {
+			// Only surface an error when NO engine produced any series. A partial
+			// success — one engine errored (e.g. transient timeout) while another
+			// returned values — must not mark the merged result errored, or the
+			// downstream HasError() checks would discard the healthy engine's pods
+			// and blackhole scaling for the whole mixed-engine model.
+			if len(merged.Values) == 0 {
+				merged.Error = firstErr
+			}
 			results[logical] = merged
 		}
 	}
@@ -96,10 +114,5 @@ func mergeEngineResults(results map[string]*source.MetricResult, engines []infer
 
 // containsEngine reports whether the engine set includes the given engine.
 func containsEngine(engines []inferenceengine.Engine, target inferenceengine.Engine) bool {
-	for _, e := range engines {
-		if e == target {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(engines, target)
 }
--- a/internal/collector/registration/saturation.go
+++ b/internal/collector/registration/saturation.go
@@ -2,6 +2,7 @@ package registration
 
 import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/source"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/inferenceengine"
 )
 
 // Query name constants for type-safe query references.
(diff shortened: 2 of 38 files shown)


internal/collector/registration/scale_to_zero.go
```diff
@@ -74,6 +84,23 @@ func CollectModelRequestCount(
...
+	// Defaults to vLLM for backward compatibility. Engine-aware scale-to-zero is
+	// wired via CollectModelRequestCountForEngine; see docs/proposals/sglang-backend.md
+	// (Phase 2) for threading the detected engine through the enforcer.
+	return CollectModelRequestCountForEngine(ctx, metricsSource, inferenceengine.EngineVLLM, modelID, namespace, retentionPeriod)
+}
+
+// CollectModelRequestCountForEngine is the engine-aware variant of
+// CollectModelRequestCount. It selects the request-count metric appropriate for
+// the given inference engine (vllm:request_success_total vs sglang:num_requests_total).
+func CollectModelRequestCountForEngine(
```
ev-shindin, line 102: `CollectModelRequestCount` hardcodes `EngineVLLM`, so an SGLang variant with scale-to-zero queries `vllm:request_success_total` (which SGLang never emits) → returns 0 → the enforcer treats the model as idle and sets every decision's `TargetReplicas` to 0. The engine-aware `CollectModelRequestCountForEngine` and the `sglang:num_requests_total` query both exist but aren't wired into the enforcer (`engine.go` `requestCountFunc`). This is "Phase 2 deferred", but there's no guard and the user-guide table lists the SGLang metric as if it works. Please gate scale-to-zero off for detected-SGLang variants, or add a loud caveat that it must not be enabled for SGLang yet.

internal/collector/registration/saturation.go
```diff
@@ -134,4 +135,77 @@ func RegisterSaturationQueries(sourceRegistry *source.SourceRegistry) {
...
+	//
+	// Structural difference from vLLM: derived from SGLang's token counters
+	// (cached prompt tokens / total prompt tokens) rather than hit/query counters.
+	// This is unit-safe (0.0-1.0) and parallels the vLLM hits/queries formula.
+	// SGLang also exposes sglang:cache_hit_rate directly, but its units are
+	// version-dependent (0-1 vs 0-100), so the counter ratio is preferred.
+	registerForEngine(registry, inferenceengine.EngineSGLang, source.QueryTemplate{
+		Name:        QueryPrefixCacheHitRate,
+		Type:        source.QueryTypePromQL,
+		Template:    `max by (instance, pod, llm_d_ai_variant) (rate(sglang:cached_tokens_total{namespace="{{.namespace}}",model_name="{{.modelID}}"}[5m]) / rate(sglang:prompt_tokens_total{namespace="{{.namespace}}",model_name="{{.modelID}}"}[5m]))`,
```
ev-shindin: The template divides `rate(sglang:cached_tokens_total{ns,model}) / rate(sglang:prompt_tokens_total{ns,model})` before the `max by(...)`. The two counters don't share an identical label set (per SGLang source, `cached_tokens_total` carries an extra `cache_source` label), so one-to-one vector matching finds no pairs and returns an empty vector — the hit rate is always 0/absent, the cache-hit demand reduction is disabled, and WVA over-provisions SGLang. The e2e fixture emits both counters with matching labels, so e2e can't catch this. Fix: aggregate the differing labels away first, e.g. `sum by(instance,pod,llm_d_ai_variant)(rate(cached_tokens_total[5m])) / sum by(...)(rate(prompt_tokens_total[5m]))` (or `... ignoring(cache_source) ...`).

test/e2e/sglang_backend_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	It("detects SGLang and reconciles the variant from sglang:* metrics", func() {
+		Eventually(func(g Gomega) {
+			va := &variantautoscalingv1alpha1.VariantAutoscaling{}
+			g.Expect(crClient.Get(ctx, client.ObjectKey{Namespace: cfg.LLMDNamespace, Name: vaName}, va)).To(Succeed())
+			// WVA sets status conditions once it has reconciled the variant...
+			g.Expect(va.Status.Conditions).NotTo(BeEmpty(), "WVA should set status conditions for the SGLang variant")
+			// ...and computes a desired allocation, which is only possible if it
+			// collected the SGLang metrics (vLLM queries would return nothing here).
+			g.Expect(va.Status.DesiredOptimizedAlloc.NumReplicas).NotTo(BeNil(),
```
ev-shindin, line 69: `NumReplicas != nil` passes via static arg-based capacity estimation even when engine routing is broken (it cannot catch the two HIGH bugs above). With the fixture's `token_usage=0.85` / `num_queue_reqs=3`, assert a scale-up: `g.Expect(*va.Status.DesiredOptimizedAlloc.NumReplicas).To(BeNumerically(">", 1))`.

internal/collector/replica_metrics.go
```diff
@@ -340,25 +341,19 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
 		source.ParamNamespace: namespace,
 	}
 
+	// Determine which inference engines this model's variants run. Engine-specific
+	// queries are refreshed once per present engine (vLLM pods emit vllm:* series,
+	// SGLang pods emit sglang:* series — the per-pod results are disjoint and are
+	// merged back under the logical query name below). For vLLM-only models this
+	// is identical to the previous fixed query list.
+	engines := inferenceengine.Present(scaleTargets)
```
ev-shindin, line 349: `engines := inferenceengine.Present(scaleTargets)` here determines the engines, but `Detect` defaults to vLLM on a nil/unresolvable leader template or an unmatched SGLang image/command; the variant then gets `vllm:*` queries, emits nothing, and yields zero replica metrics — indistinguishable from "no metrics." Add a debug log of the detected engine(s) per scale target here so operators can tell "wrong engine" from "no data."

internal/collector/registration/throughput_analyzer.go
```diff
@@ -139,4 +140,38 @@ func RegisterThroughputAnalyzerQueries(sourceRegistry *source.SourceRegistry) {
...
+		Name:        QueryKvUsageInstant,
+		Type:        source.QueryTypePromQL,
+		Template:    `max by (instance, pod, llm_d_ai_variant) (sglang:token_usage{namespace="{{.namespace}}",model_name="{{.modelID}}"})`,
+		Params:      []string{source.ParamNamespace, source.ParamModelID},
+		Description: "Instantaneous KV cache utilization per pod (0.0-1.0), used as k* in the ITL model (SGLang)",
+	})
+
+	// Per-pod request completion rate (req/s) from the generation histogram count.
+	registerForEngine(registry, inferenceengine.EngineSGLang, source.QueryTemplate{
+		Name:        QueryVLLMRequestRate,
```
ev-shindin: The metric is engine-agnostic (request completion rate per pod), so the SGLang physical key `sglang/vllm_request_rate` is self-contradictory. Rename the constant/logical name to `request_rate` (update the few references in `replica_metrics.go` and the throughput tests).

internal/inferenceengine/engine.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	for _, tok := range c.Args {
+		b.WriteString(tok)
+		b.WriteByte(' ')
+	}
+	cmd := strings.ToLower(b.String())
+
+	return strings.Contains(cmd, "sglang.launch_server") ||
+		strings.Contains(cmd, "sglang serve") ||
+		strings.Contains(cmd, "-m sglang")
```
ev-shindin: `-m sglang` substring is broader than the documented detection contract (matches `-m sglang_anything`). The first two checks (`sglang.launch_server` / `sglang serve`) already cover known launches; drop it or tighten, and add a false-positive guard test.

internal/engines/analyzers/saturation_v2/deployment_parser.go
```diff
@@ -7,22 +7,31 @@ import (
...
 
-// VLLMEngineParams holds vLLM configuration parameters parsed from a
+// VLLMEngineParams holds inference-engine configuration parameters parsed from a
 // Deployment/LWS's container args and environment variables. These are used
 // to derive compute-bound capacity (k2) when no live metrics are available.
+//
+// Despite the name, this struct is the shared engine-params type for all
+// supported engines: ParseVLLMArgs populates it from vLLM flags and
+// ParseSGLangArgs populates it from SGLang flags (mapped onto the same fields).
 type VLLMEngineParams struct {
```
ev-shindin: The `VLLMEngineParams`→`EngineParams` rename never happened despite the "despite the name…" comment and the proposal calling it `EngineParams`. Do the rename (`replace_all` across saturation_v2) or fix the docs.

internal/engines/analyzers/saturation_v2/deployment_parser.go
ev-shindin: This doesn't compare the new `TotalKvTokensOverride`; two same-engine SGLang variants differing only in `--max-total-tokens` are deemed compatible (wrong capacity on the no-metrics fallback).

internal/collector/engine_queries.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/registration"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/source"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/inferenceengine"
+)
+
+// engineSpecificReplicaQueries are the logical query names collected per replica
+// whose metric source is the inference engine (vLLM/SGLang). Each is refreshed
+// once per present engine and merged back under its logical name.
+var engineSpecificReplicaQueries = []string{
```
ev-shindin, line 21: We have this one as well as `internal/collector/registration/engine.go:16` (`EngineSpecificQueries`) - two hand-maintained engine-specific query lists; cross-reference or derive one from the other to prevent drift.

internal/constants/metrics.go
```diff
@@ -68,6 +68,62 @@ const (
...
+	SGLangTokenUsage = "sglang:token_usage"
+
+	// SGLangMaxTotalNumTokens is the total KV-cache token capacity (gauge).
+	// SGLang exposes capacity directly, unlike vLLM which derives it from
+	// vllm:cache_config_info (num_gpu_blocks x block_size).
+	SGLangMaxTotalNumTokens = "sglang:max_total_num_tokens"
+
+	// SGLangTimeToFirstTokenSecondsSum/Count are the TTFT histogram parts.
+	// SGLang equivalent of vllm:time_to_first_token_seconds_{sum,count}.
+	SGLangTimeToFirstTokenSecondsSum   = "sglang:time_to_first_token_seconds_sum"
```
ev-shindin: Nit: The SGLang `*Count` constants have no individual doc comment and the `*Sum` comments don't start with the identifier. Split into per-constant comments like the vLLM block above.

ev-shindin (changes requested):
Thanks @MohanKumar21 this PR is really important!There are some comments to address. Also it would be great if you will provide following tests:

- Detection: `sglang serve` branch and a `-m sglang` false-positive guard (`-m sglang_bench` → expect vLLM).
- Positive metric-name assertions for the 5 untested SGLang templates (output/input tokens, prefix-cache, gen-token-rate, request-rate).
- Unit test for the SGLang cache-config block (`sglang:max_total_num_tokens` → `TotalKvCapacityTokens`).
- `CollectModelRequestCountForEngine` with `EngineSGLang` (asserts `sglang/model_request_count` routing).
- `--max-prefill-tokens` flag in `sglang_parser_test.go`.

internal/collector/registration/scale_to_zero.go
MohanKumar21: Gated. Scale-to-zero enforcement is now skipped (with a loud Info log) for any model whose scale targets run a non-vLLM engine, in both the V1 and V2 optimize paths.

internal/collector/registration/saturation.go
MohanKumar21: Rewrote the SGLang prefix-cache-hit-rate template to aggregate each counter with sum by (instance, pod, llm_d_ai_variant) before the division, so the ratio is computed on matching label sets regardless of cache_source.

test/e2e/sglang_backend_test.go
MohanKumar21: Done.

internal/collector/replica_metrics.go
MohanKumar21: Added a DEBUG log listing the per-scale-target detected engine.

internal/collector/registration/throughput_analyzer.go
MohanKumar21: Renamed the logical query to request_rate (constant QueryRequestRate); the SGLang key is now sglang/request_rate. I also made the in-memory field generic: ReplicaMetrics.VLLMRequestRate → RequestRate.

internal/inferenceengine/engine.go
MohanKumar21: Dropped the -m sglang substring check.

internal/engines/analyzers/saturation_v2/deployment_parser.go
MohanKumar21: Done — renamed VLLMEngineParams → EngineParams

internal/engines/analyzers/saturation_v2/deployment_parser.go
MohanKumar21: Fixed - `IsCapacityCompatible` now also compares TotalKvTokensOverride, so two SGLang variants differing only in --max-total-tokens

internal/collector/engine_queries.go
MohanKumar21: Cross-referenced both lists with doc comments explaining the subset relationship (the collector list excludes the scale-to-zero model_request_count), and added a test asserting the collector list is a subset of registration.EngineSpecificQueries to catch drift.

internal/constants/metrics.go
MohanKumar21: Done.

MohanKumar21:
Thanks for the feedback @ev-shindin . Addressed your comments. I have also made sure the changes are such that we can easily adopt to other inference engines in the future. I refactored the vLLM-specific naming to be engine-neutral and added an explicit Engine discriminator so the collection/parsing layers are engine-agnostic and a new backend mostly just needs its query templates + arg parser. Tests cover detection, the SGLang templates, request-count routing, --max-prefill-tokens, the scale-to-zero gate, and cross-engine validation (mixed vLLM+SGLang collection with no cross-engine capacity corruption).

test/e2e/sglang_backend_test.go
```diff
@@ -0,0 +1,82 @@
...
+// SGLang backend e2e: deploys a synthetic SGLang model server (CPU-only emitter of
+// sglang:* metrics, launched with a faithful `sglang.launch_server` command),
+// wires Prometheus scraping, and creates a VariantAutoscaling for it. It asserts
+// that WVA — with no engine configuration — detects SGLang, collects the sglang:*
+// metrics, and reconciles the variant to a desired allocation.
+//
+// This exercises the same code path as vLLM, only with SGLang metric names and
+// flags. It runs in the kind-emulator environment (cfg.UseSimulator); a real
+// SGLang server requires a GPU.
+var _ = Describe("SGLang backend", Ordered, func() {
```
ev-shindin: This tests has no `Label`; CI filters are `smoke` and `full && !flaky`, so an unlabeled spec matches neither and the whole scenario (incl. the `>1` guard for both fixes) is skipped. Add `Label("full")` (or `Label("smoke","full")`).

internal/engines/saturation/engine.go
ev-shindin, line 1: The gate `scaleToZeroSupportedForEngines` (`engine.go:1177`) was wired into V1 and V2 but `optimizeQueueingModel` (`internal/engines/saturation/engine_queueing_model.go`) calls `EnforcePolicyOnDecisions` ungated. V1/V2/queueing-model are mutually exclusive, so a queueing-model deployment (the `wva-queueing-model-config` ConfigMap is present) still zeroes a busy SGLang model via the vLLM-only counter — the original bug.

internal/engines/saturation/engine.go
```diff
@@ -610,7 +611,14 @@ func (e *Engine) optimizeV1(
 		// keeping decode (or vice versa) makes the model non-functional — both
 		// stages must scale together.
 		if len(modelDecisions) > 0 {
-			if !hasMinReplicasAboveZero(data.variantStates) {
+			switch {
+			case !scaleToZeroSupportedForEngines(data.scaleTargets):
```
ev-shindin: `engine_scale_to_zero_gate_test.go` tests the `scaleToZeroSupportedForEngines` predicate in isolation; nothing drives `optimizeV1`/`optimizeV2` with an SGLang target and asserts the enforcer is skipped. Inverting the `!` in either caller would pass all existing tests while zeroing SGLang models. Add a wired test per path.

internal/engines/saturation/engine.go
```diff
@@ -854,6 +864,9 @@ func (e *Engine) optimizeV2(
 
 		requests = append(requests, *req)
 		modelReplicaMetrics[modelID] = data.replicaMetrics
+		if !scaleToZeroSupportedForEngines(data.scaleTargets) {
```
ev-shindin: same as line `615` above

internal/engines/saturation/engine.go
```diff
@@ -610,7 +611,14 @@ func (e *Engine) optimizeV1(
 		// keeping decode (or vice versa) makes the model non-functional — both
 		// stages must scale together.
 		if len(modelDecisions) > 0 {
-			if !hasMinReplicasAboveZero(data.variantStates) {
+			switch {
+			case !scaleToZeroSupportedForEngines(data.scaleTargets):
+				logger.Info("Skipping scale-to-zero enforcement: model runs a non-vLLM engine; engine-aware request counting is not yet wired through the enforcer (see docs/proposals/sglang-backend.md Phase 2)",
```
ev-shindin: Nit: per-cycle `logger.Info(...)` for the static engine-skip condition; use `logger.V(logging.DEBUG).Info(...)` to match the adjacent `hasMinReplicasAboveZero` checks — otherwise an Info line every cycle for every SGLang model.

internal/engines/saturation/engine.go
```diff
@@ -872,6 +885,14 @@ func (e *Engine) optimizeV2(
 
 	// Stage 3: Apply enforcer per-model (directly on decisions)
 	for _, req := range requests {
+		// Skip scale-to-zero enforcement for models running a non-vLLM engine:
+		// the request-count query is hardcoded to vLLM, so it would falsely read
+		// as idle and zero the model (see scaleToZeroSupportedForEngines).
+		if skipScaleToZero[utils.GetNamespacedKey(req.Namespace, req.ModelID)] {
+			logger.Info("Skipping scale-to-zero enforcement (V2): model runs a non-vLLM engine; engine-aware request counting is not yet wired through the enforcer (see docs/proposals/sglang-backend.md Phase 2)",
```
ev-shindin: Nit: see comment to line `616` above.

ev-shindin (changes requested):
Thanks @MohanKumar21 ! This PR converges, however one issue is overlooked in my previous review, and some tests should be wired to PR. Please address my review.

test/e2e/sglang_backend_test.go
mohankluisz: Done.

internal/engines/saturation/engine.go
MohanKumar21: Fixed.

internal/engines/saturation/engine.go
MohanKumar21: Fixed. The engine-skip log is now logger.V(logging.DEBUG).Info(...) (it lives once in the helper), matching the adjacent minReplicas skip.

internal/engines/saturation/engine.go
MohanKumar21: Fixed. The scale-to-zero gate was wired into V1 and V2 but optimizeQueueingModel called EnforcePolicyOnDecisions ungated, so a queueing-model deployment (wva-q…) running SGLang would have been zeroed by the vLLM-hardcoded request counter.

Rather than copy the gate a third time, I extracted a single (*Engine).applyScaleToZeroEnforcement(...) helper that owns the gating (non-vLLM engine and minReplicas > 0) plus the enforce call, and routed all three optimize paths (V1, V2, queueing-model) through it. The queueing path is now gated identically to the others, and there is no longer any per-caller !scaleToZeroSupportedForEngines(...) to get out of sync.

internal/engines/saturation/engine.go
MohanKumar21: Fixed. Added engine_scale_to_zero_enforce_test.go, which drives the shared enforcement seam with a real, idle (requestCount == 0) Enforcer and scale-to-zero enabled in config, then asserts end-to-end outcomes:

- idle all-vLLM model → scaled to 0 (canary: proves the enforcer actually runs),
- idle SGLang model → left at current replicas,
- mixed vLLM+SGLang → left at current replicas,
- vLLM with minReplicas > 0 → left at current replicas.

Because every optimize path now funnels through applyScaleToZeroEnforcement, this one spec locks the gate down for V1, V2, and the queueing model. Verified by mutation: inverting the gate flips the vLLM canary and zeroes the SGLang/mixed cases → 3 specs fail; restored → green.

internal/engines/saturation/engine.go
MohanKumar21: Fixed. Same as above

MohanKumar21:
Addressed the comments @ev-shindin . Thanks

ev-shindin (approved):
LGTM. Thanks @MohanKumar21

MohanKumar21:
@ev-shindin This branch's deploy script fetched the old llm-d scheduler/ guide path while main's workflow injects LLM_D_RELEASE=v0.8.1, where it's now router/ causing 404) and the e2e workflow failure. I've rebased onto main to pick up the fix. Kindly request that you retrigger the tests. Thanks.

MohanKumar21:
E2e test failed due to insufficient gpus

MohanKumar21:
Rebased on main, as there were some more workdlow changes

PR #1353, opened 2026-06-28
feat(engine): add HPA-style stabilization of scaling recommendations

### LINKED ISSUE #1352: Add HPA-style stabilization of scaling recommendations in WVA
## Background

WVA's optimizer emits a fresh per-variant replica target every cycle. Acting on
those raw targets flaps under noisy load, so today WVA relies on a downstream
HPA's stabilization window (notably the 300s scale-down delay) to damp the
signal. That coupling blocks two things:

1. **Rescale.** The priority-weighted rescale proposal [#1238](https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1238) reallocates GPUs
   by reclaiming them from over-target / saturated models to serve higher-priority
   ones — i.e. it deliberately drives scale-**downs**. The HPA's own scale-down
   stabilization window delays and second-guesses those decisions. For rescale to
   actuate cleanly, WVA needs to own the scale-up/down stabilization itself and
   run the HPA with `behavior` delays set to **0** (near-passthrough), so WVA is
   the single place that decides *when* to rescale and the HPA does not
   double-stabilize.

2. **Direct actuation.** ([#1316](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1316)) Stabilization-over-time is the prerequisite for WVA
   patching `/scale` directly (the 1→N range). Patching without a stabilization
   window would lose exactly the damping the HPA provides today and reintroduce
   flapping. A reusable in-process stabilizer is the foundation that later enables
   direct actuation.

## Proposed scope

- A self-contained stabilization component: a clean-room port of the Kubernetes
  HPA *configurable scaling behavior* — trailing stabilization window (min over
  the up-window / max over the down-window), per-period `Pods`/`Percent` rate
  policies with `selectPolicy`, tolerance deadband, and min/max clamp. Reuse the
  `autoscaling/v2` behavior types so it is configured the familiar HPA way.
- Apply it in the V2 engine right after the optimizer and before the enforcer,
  per `namespace/model/variant[/role]`, with a structured `stabilization-decision`
  cycle log.
- Gate it behind a config flag, default off, so there is no behavior change
  unless enabled. When enabled, apply HPA-style defaults; the intended deployment
  contract is **WVA owns stabilization, HPA `behavior` delays set to 0**.

## Out of scope / follow-ups

- Exposing the full per-policy behavior through the ConfigMap (start with the
  on/off flag and sensible defaults).
- Wiring direct `/scale` actuation for the 1→N range with a single-writer guard.
- The V1 saturation path (V2 first).

ev-shindin:
Fixes #1352

## Proposed Changes

- :gift: Add `internal/stabilization`: a clean-room port of the Kubernetes HPA *configurable scaling behavior* (trailing window — min over the up-window / max over the down-window — plus per-period `Pods`/`Percent` rate policies with `selectPolicy`, tolerance deadband, and min/max clamp). Reuses the `autoscaling/v2` behavior types so it is configured the familiar HPA way.
- :gift: Wire stabilization into the V2 engine right after the optimizer and before the enforcer, keyed per `namespace/model/variant[/role]`, with a structured `stabilization-decision` cycle log. Gated by a new `enableStabilization` saturation-config flag, **default off**. Intended deployment contract: WVA owns stabilization, HPA `behavior` delays set to `0`.
- :broom: Extract `VariantDecision.ActionForTarget` and reuse it from both the enforcer and the stabilizer (removes a duplicated action-recompute switch).

This is the foundation for owning scale-down so the priority-weighted rescale proposal (#1238) can actuate cleanly, and a prerequisite for direct `/scale` actuation in the 1→N range. Per-policy config via the ConfigMap, direct actuation, and the V1 path are follow-ups (see #1352).

### Pre-review Checklist

- [ ] **E2E tests** for any new behavior — N/A while default-off (no behavior change on upgrade). Covered by `internal/stabilization` unit specs and the saturation engine envtest suite; E2E will accompany enabling it by default / direct actuation.
- [ ] **Docs PR** for any user-facing impact — developer design note included (`docs/plans/engine/stabilization.md`); no user-facing impact while the flag is off. User-facing docs will accompany the per-policy config follow-up.
- [ ] **Proposal PR** for any new enhancement or change to existing behavior — rationale tracked in #1352 and related to the rescale proposal #1238; a dedicated proposal for the operator-facing config surface is a follow-up.

**Release Note**

```release-note
Added opt-in HPA-style stabilization of WVA scaling recommendations (a trailing
scale-down window plus per-period rate policies), enabled via
`enableStabilization` in the saturation scaling config. Disabled by default.
```

**Docs**

No user-visible impact while default-off; user-facing documentation will accompany the per-policy configuration follow-up.

### DIFF
--- a/docs/plans/engine/stabilization.md
+++ b/docs/plans/engine/stabilization.md
@@ -47,9 +47,13 @@ pipeline, in order, matches the HPA:
    since the optimizer already decided to act.
 3. **Per-period rate policies.** `Pods` and `Percent` policies cap the change
    per `periodSeconds`, budgeting against changes already actuated within the
-   period (tracked as per-key scale events). `selectPolicy` `Max`/`Min`/
-   `Disabled` combines competing policies. Defaults: scale-up the higher of
-   `+4 pods` or `+100%` per `60s`; scale-down `-100%` per `60s`.
+   period (the period start is reconstructed as `current − added + removed`
+   using both scale-up and scale-down events, as the HPA does). `selectPolicy`
+   `Max`/`Min`/`Disabled` combines competing policies. Defaults: scale-up the
+   higher of `+4 pods` or `+100%` per `60s`; scale-down `-100%` per `60s`. The
+   magnitudes and the `300s` down window match the HPA; the `60s` period is
+   chosen for WVA's ~30s optimize cadence rather than the controller's `15s`
+   default (which would make the budget a near no-op at that cadence).
 4. **Min/max clamp** to the variant's bounds.
 
 The `Stabilizer` is long-lived and concurrency-safe; it keeps the trailing
@@ -65,8 +69,8 @@ matches the HPA: stabilize the recommendation, then let the hard bounds have the
 last word. The existing emit path (Prometheus desired-replica gauge, internal
 decision cache, CRD status patch) carries the stabilized value unchanged.
 
-Each variant is keyed `namespace/variant[/role]`, so disaggregated prefill/decode
-scale targets are damped independently.
+Each variant is keyed `namespace/model/variant[/role]`, so disaggregated
+prefill/decode scale targets are damped independently.
 
 A `stabilization-decision` structured log line is emitted per model per cycle,
 grouped like the sibling `scaling-decision` line, with `curr`, `raw`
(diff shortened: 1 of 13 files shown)


lionelvillard:
Is that really necessary? User will have to configure stabilization in two different places, is that intended?

PR #1354, opened 2026-06-29
Create a dedicated component for Openshift patches and Openshift's cluster-scoped deployment

aleskandro:
For Openshift, an overlay is defined to deploy the namespace-scoped configuration. This commit moves the patches in that overlay to a dedicated component, migrate the current overlay to use the component and adds a new component for the cluster-scoped deployment in Openshift.

With this PR, downstream deployments (e.g., in ODH) can consume the cluster-scoped overlay instead of adding further patches that might need to be carried across rebases.


lionelvillard:
lgtm. I'm trying to see if we can run the llm-d nightlies before merging this PR.

lionelvillard:
not possible. Let's keep an eye on it

lionelvillard (approved):

lionelvillard:
~https://github.com/llm-d/llm-d/actions/runs/28471873869~ The WVA guide is still pinned to the 0.8 release

PR #1355, opened 2026-06-29
fix: allow annotation mode without VA CRD

### LINKED ISSUE #1345: WVA crashes on startup when VariantAutoscaling CRD is not installed
WVA v0.8.0 (and current `main`) crashes on startup if the `VariantAutoscaling` CRD is not present in the cluster, even in annotation-only mode which doesn't require it.

```
ERROR  setup  cmd/main.go:388  unable to setup indexes
{"error": "failed to set up index by scale target for VariantAutoscaling: no matches for kind \"VariantAutoscaling\" in version \"llmd.ai/v1alpha1\""}
```

`SetupIndexes` in `internal/controller/indexers/indexers.go` unconditionally indexes the VA type and `main.go` calls `os.Exit(1)` on failure.

xuhui-lu:
## Summary

Fixes #1345.

This makes the deprecated `VariantAutoscaling` CRD optional for annotation-mode deployments. When the VA CRD is absent, WVA now skips VA-specific startup wiring and continues with HPA / KEDA annotation discovery.

## Changes

- Detect whether the `VariantAutoscaling` CRD is installed at startup.
- Register the VA scale-target field index only when the VA CRD exists.
- Register the `VariantAutoscalingReconciler` only when the VA CRD exists.
- Treat `NoMatch` while listing `VariantAutoscalingList` as an empty CRD-sourced variant list, then continue to annotation-sourced HPA / ScaledObject discovery.
- Add regression coverage for:
  - `SetupIndexes(false, false)` skipping optional VA / ScaledObject indexes while keeping HPA indexing available.
  - `readyVariantAutoscalings` continuing to return annotation-sourced synthetic VAs when the VA CRD list path returns `NoMatch`.

## Testing

- `go test ./internal/controller/indexers`
- `go test ./internal/utils`
- `go test ./cmd/...`
- `go test ./internal/collector`
- `go test ./internal/utils/crd`
- `go test ./internal/controller/...`
- `gofmt -l changed-files`
- `git diff --check`

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -201,6 +201,17 @@ func main() {
 		setupLog.Info("LeaderWorkerSet CRD not found - support disabled (Deployment-only mode)")
 	}
 
+	vaCRDEnabled, err := crd.CheckVariantAutoscalingCRD(restConfig, setupLog)
+	if err != nil {
+		setupLog.Error(err, "failed to determine VariantAutoscaling CRD availability")
+		os.Exit(1)
+	}
+	if vaCRDEnabled {
+		setupLog.Info("VariantAutoscaling CRD detected - support enabled")
+	} else {
+		setupLog.Info("VariantAutoscaling CRD not found - VA reconciler disabled; annotation-based discovery remains enabled")
+	}
+
 	// Detect KEDA for annotation-based ScaledObject discovery (dual-mode, Phase 1)
 	kedaEnabled := crd.CheckKEDACRD(restConfig, setupLog)
 	if kedaEnabled {
@@ -384,7 +395,7 @@ func main() {
 
 	// Setup custom indexes for lookups on VariantAutoscalings
 	setupLog.Info("Setting up indexes")
-	if err := indexers.SetupIndexes(context.Background(), mgr, kedaEnabled); err != nil {
+	if err := indexers.SetupIndexes(context.Background(), mgr, vaCRDEnabled, kedaEnabled); err != nil {
 		setupLog.Error(err, "unable to setup indexes")
 		os.Exit(1)
 	}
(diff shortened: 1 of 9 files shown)


cmd/main.go
```diff
@@ -201,6 +201,13 @@ func main() {
 		setupLog.Info("LeaderWorkerSet CRD not found - support disabled (Deployment-only mode)")
 	}
 
+	vaCRDEnabled := crd.CheckVariantAutoscalingCRD(restConfig, setupLog)
```
ev-shindin: `CheckCRDInstalled` returns `false` on **any** discovery failure (client-create error, or `ServerGroupsAndResources()` error with nil lists), not only genuine CRD-absence — it can't distinguish "absent" from "couldn't determine." This PR newly gates the VA **reconciler** (`cmd/main.go:522`) and **field index** on that boolean. So a transient discovery hiccup at startup on a cluster that *has* the VA CRD (apiserver 503, discovery throttling on a large cluster, RBAC denial) → `vaCRDEnabled=false` for the pod's lifetime → the reconciler never runs and VA `.status`/actuation is dead until a restart.

It's a *degraded* state, not a crash (`readyVariantAutoscalings` still lists the real VAs and the engines optimize them; only status writeback and index attribution are lost), and it matches the existing fail-open LWS/KEDA pattern — but those are optional add-ons, whereas this gates the project's core resource. Suggest distinguishing "absent" from "couldn't determine": retry-with-backoff on a hard discovery error, or fail-closed (crash-loop) for the VA CRD, rather than silently degrading.

cmd/main.go
```diff
@@ -201,6 +201,13 @@ func main() {
...
+	if vaCRDEnabled {
+		setupLog.Info("VariantAutoscaling CRD detected - support enabled")
+	} else {
+		setupLog.Info("VariantAutoscaling CRD not found - support disabled")
```
ev-shindin: On a transient detection failure the operator sees `INFO  VariantAutoscaling CRD not found - support disabled`, which reads as an intentional annotation-only configuration. The hard-failure `Error` log inside `CheckCRDInstalled` is easy to miss in startup noise. Make the decision line distinguish "not found" from "detection failed" (and ideally emit a Warning when detection errored).

internal/utils/variant.go
```diff
@@ -187,7 +187,11 @@ func readyVariantAutoscalings(ctx context.Context, k8sClient client.Client) ([]w
 	// List VAs using the informer cache with optional label selector
 	var vaList wvav1alpha1.VariantAutoscalingList
 	if err := k8sClient.List(ctx, &vaList, listOpts...); err != nil {
-		return nil, err
+		if !apimeta.IsNoMatchError(err) {
```
ev-shindin, line 190: `if !apimeta.IsNoMatchError(err) { return nil, err }` — the swallow path is covered by the new "VA CRD NoMatch" test, but the **fatal** branch isn't: invert or delete the guard and no test fails. Add a test that a non-`NoMatch` VA `List` error (e.g. "api server unavailable") is still returned to the caller.

internal/controller/indexers/indexers_test.go
```diff
@@ -97,6 +98,63 @@ var _ = Describe("Indexers", Ordered, func() {
...
+			disabledClient := disabledMgr.GetClient()
+			Eventually(func() string {
+				got, err := FindHPAForScaleTarget(ctx, disabledClient, hpa.Spec.ScaleTargetRef, ns)
+				if err != nil || got == nil {
+					return ""
+				}
+				return got.Name
+			}).Should(Equal("managed-hpa-only"))
+
+			_, err = FindVAForDeployment(ctx, disabledClient, "target-deploy", ns)
```
ev-shindin, line 156: nit: The assertion that `FindVAForDeployment` errors containing `VAScaleTargetKey` runs in envtest where the VA CRD *is* installed (only the index is skipped), so it exercises index-gating, not the true CRD-absent path. Harmless (production callers guard on `err == nil`, not the message), but worth a comment noting it.

ev-shindin (commented):
Thanks @xuhui-lu ! Please address review comments.

ev-shindin (dismissed):
LGTM. Thanks @xuhui-lu !

lionelvillard:
@xuhui-lu can you please rebase?

lionelvillard (approved):
thanks @xuhui-lu !

lionelvillard:
@vivekk16 Do we need to backport this PR?

vivekk16:
Yes please @lionelvillard

Issue #1133, opened 2026-05-13
LWS replicas with multiple metric-emitting pods are over-counted in saturation analysis


ev-shindin:
### Problem

When a `LeaderWorkerSet` (LWS) workload runs with **all pods of a single LWS replica** exporting `vllm:*` Prometheus metrics (leader + workers), WVA's saturation analyzer treats each pod as an independent replica of the underlying `VariantAutoscaling` (VA). The result is:

1. **`ReplicaCount` is inflated** by the number of pods per LWS replica. An LWS with `K` replicas of `1 leader + N workers` where all pods emit yields `(1+N)·K` replica-metrics entries reported to the analyzer, but the operator's view of the LWS is `K` replicas.
2. **Per-replica averages are diluted.** KV-cache utilization and queue length are averaged over `(1+N)·K` samples that are not independent — worker-pod metrics correlate with their leader because they shard tensor state. Saturation thresholds (`KvCacheThreshold`, `QueueLengthThreshold`) are evaluated against the diluted average and can miss legitimate saturation.
3. **Scaling decisions drift.** The cost-aware and greedy-by-score optimizers receive `ReplicaCount = (1+N)·K`; downstream replica deltas are computed against this inflated baseline, producing scale-up/down actions that don't match the operator's mental model of the LWS.

This is independent of #1121's keying refactor. #1121 correctly resolves "multiple vLLM instances on a single pod" (one pod, multiple ports). The case here is the *opposite*: multiple pods of a single LWS replica each emitting once.

### Affected code paths

Trace from scrape → analyzer:

- `internal/collector/replica_metrics.go` — `buildInstanceKey` produces one `podMetricData` entry per `(pod_name, port)` tuple. With all-pods-emit, this is one entry per pod.
- `internal/collector/replica_metrics.go:617` — final loop appends one `interfaces.ReplicaMetrics` to `replicaMetrics` per `podMetricData` entry. All worker pods AND the leader of the same LWS replica resolve to the same VA via `FindVAForPod` (`internal/collector/source/pod_va_mapper.go:57`, `case "LeaderWorkerSet"`), so they share a `VariantName`.
- `internal/saturation/analyzer.go:138-156` — analyzer keys by `VariantName` and computes `ReplicaCount = len(metrics)`. No deduplication or aggregation by LWS replica index.

### Reproduction

[...]

PR #1358, opened 2026-06-30
Bug-fix: LWS replicas with multiple metric-emitting pods are over-cou…

shuynh2017:
Fixes # https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1133

## Proposed Changes

-  :bug: Fix bug

### Pre-review Checklist

- [N/A ] **E2E tests** for any new behavior

**Release Note**

```release-note
N/A
```

**Docs**

N/A

## Implementation
- The implementation follows strategy #1 in https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1133. 
```
Filter at the collector layer. Identify the metric-emitting pods that "represent" an LWS replica (e.g., the leader pod via the leaderworkerset.sigs.k8s.io/worker-index=0 label) and drop the others before building ReplicaMetrics. Simplest if the deployment convention is leader-emits.
```
- The flow:
    - In `saturation/analyzer.go`, total replicas is still `analysis.TotalReplicas = len(replicaMetrics)`. A search on places that set `TotalReplicas`, this is the only place where it's set (other than set to zero).
    - In `replicas_metrics.go`, `ReplicasMetrics` is filtered to not include LWS worker pods by calling `isLWSWorker` function.
    - `isLWSWorker` function calls `locator` to get pod labels. 
    - In `locator.go`, the function `resolveTarget` has already called `l.apiReader.Get` to get pod info. The update is to leverage this call (instead of making new K8S API call in `isLWSWorker`) and store pod labels in the cache. This does the job but as side-effect, it extends the function `resolveTarget` to more than just resolve target.
 
- Also added artifacts to create LWS for testing

## Unit Tests
- Added as in "Proposed acceptance criteria".

## Test
- Tested by created LWS with 1 and 2 replicas, 2 and 4 pods respectively, and with debugger, checked the  ``analysis.TotalReplicas ` is set correctly to 1 and 2, instead 2 and 4.


ev-shindin:
@shuynh2017 - please rebase

internal/collector/locator/locator.go
```diff
@@ -191,6 +197,30 @@ func (l *podLocator) LocateByVariant(ctx context.Context, namespace, variantName
...
+	}
+
+	// Not in cache, fetch the pod
+	pod := &corev1.Pod{}
+	if err := l.apiReader.Get(ctx, types.NamespacedName{Namespace: namespace, Name: podName}, pod); err != nil {
+		return nil
+	}
+
+	// Resolve the target to populate the cache (so subsequent Locate calls don't refetch)
+	target, _ := l.resolveScaleTarget(ctx, pod, namespace)
```
ev-shindin: This ignores the `resolveScaleTarget` error and then caches the zero-value `target`. On a transient owner-walk error (`walkOwnersUp` API hiccup / ownerRef cycle), a pod that *does* have a scaler ancestor gets a permanent negative cache entry, and later `Locate`/`ResolveScaleTarget` calls return "unmanaged" without retrying until eviction. Skip the `cache.add` (or propagate the error) when `resolveScaleTarget` fails.

internal/collector/locator/locator.go
```diff
@@ -191,6 +197,30 @@ func (l *podLocator) LocateByVariant(ctx context.Context, namespace, variantName
...
+
+	// Check cache first
+	key := podKey{Namespace: namespace, Name: podName}
+	if labels, hit := l.cache.getLabels(key); hit {
+		return labels
+	}
+
+	// Not in cache, fetch the pod
+	pod := &corev1.Pod{}
+	if err := l.apiReader.Get(ctx, types.NamespacedName{Namespace: namespace, Name: podName}, pod); err != nil {
```
ev-shindin, line 195: This swallows *every* `apiReader.Get` error, so a transient failure is indistinguishable from "pod not found," with no diagnostic trail.

internal/collector/locator/locator.go
```diff
@@ -74,6 +74,12 @@ type PodLocator interface {
 	// fallback in the collector (buildInstanceKey). Remove it when the
 	// VariantAutoscaling CRD is removed.
 	ResolveScaleTarget(ctx context.Context, namespace, podName string) (ref autoscalingv2.CrossVersionObjectReference, ok bool, err error)
+
+	// GetPodLabels returns the labels for the specified pod. This reuses the
```
ev-shindin, line 79: "results in only one API read" isn't true for the common path: when a pod carries the `llm_d_ai_variant` label, `buildInstanceKey` never calls `Locate`, so the cache is cold and `GetPodLabels` runs a full `resolveScaleTarget`/`walkOwnersUp` chain (Pod → ReplicaSet → Deployment/LWS GETs) just to read labels. Either soften the doc, or add a labels-only path that does a single Pod GET without the owner walk.

internal/collector/replica_metrics.go
```diff
@@ -324,6 +324,32 @@ func (c *ReplicaMetricsCollector) buildInstanceKey(ctx context.Context, namespac
...
+	if podName == "" || c.locator == nil {
+		return false
+	}
+
+	labels := c.locator.GetPodLabels(ctx, namespace, podName)
+	if labels == nil {
+		return false
+	}
+
+	workerIndex, hasLabel := labels[constants.LWSWorkerIndexLabel]
```
ev-shindin: The label key and the "worker-index == 0 -> leader" semantic are already in the imported LWS module: use `leaderworkersetv1.WorkerIndexLabelKey` , and consider `sigs.k8s.io/lws/pkg/utils/pod.LeaderPod` for the predicate. Keep your extra "has the label at all" check - that distinguishes non-LWS pods, which `LeaderPod` doesn't.

internal/collector/replica_metrics.go
```diff
@@ -324,6 +324,32 @@ func (c *ReplicaMetricsCollector) buildInstanceKey(ctx context.Context, namespac
...
+// with a value other than "0" (leader pods have worker-index="0").
+// Returns false for non-LWS pods or LWS leader pods.
+// Uses the locator's GetPodLabels which reuses the same pod fetch that Locate performs.
+func (c *ReplicaMetricsCollector) isLWSWorker(ctx context.Context, namespace, podName string) bool {
+	if podName == "" || c.locator == nil {
+		return false
+	}
+
+	labels := c.locator.GetPodLabels(ctx, namespace, podName)
+	if labels == nil {
```
ev-shindin, line 300: Fail-open (nil labels -> not a worker -> pod included) is a reasonable default, but it's silent. Combined with the swallowed errors in `GetPodLabels`, a transient API blip here re-introduces the double-count invisibly. Log at DEBUG when labels are nil due to an error (vs. a genuine non-LWS pod).

internal/collector/replica_metrics.go
```diff
@@ -324,6 +324,32 @@ func (c *ReplicaMetricsCollector) buildInstanceKey(ctx context.Context, namespac
 	return instanceKey, podName, vaName
 }
 
+// isLWSWorker checks if a pod is part of an LWS and is a worker (non-leader).
+// Returns true if the pod has the leaderworkerset.sigs.k8s.io/worker-index label
+// with a value other than "0" (leader pods have worker-index="0").
+// Returns false for non-LWS pods or LWS leader pods.
+// Uses the locator's GetPodLabels which reuses the same pod fetch that Locate performs.
```
ev-shindin, line 293: See the note on `locator.go` L78.

internal/constants/constants.go
```diff
@@ -102,6 +102,8 @@ const (
 	LeaderWorkerSetKind       = "LeaderWorkerSet"
 	LeaderWorkerSetAPIVersion = "leaderworkerset.x-k8s.io/v1"
 
+	// LeaderWorkerSet labels
+	LWSWorkerIndexLabel = "leaderworkerset.sigs.k8s.io/worker-index"
```
ev-shindin: This re-hardcodes a string the LWS module already exports: `sigs.k8s.io/lws/api/leaderworkerset/v1.WorkerIndexLabelKey`. Please reference or alias.

internal/interfaces/saturation_analyzer.go
```diff
@@ -39,6 +39,8 @@ const RolePrefill = "prefill"
 // and the queueing model analyzer. Saturation analysis uses KV cache, queue, and
 // token-capacity fields, while the queueing model analyzer uses
 // ArrivalRate and MaxBatchSize to model queue dynamics and estimate optimal capacity.
+// For LWS, ReplicasMetrics only holds leader pods (leaderworkerset.sigs.k8s.io/worker-index=0 label)
```
ev-shindin: Nit: typo `ReplicasMetrics` -> `ReplicaMetrics`

internal/collector/locator/locator_test.go
```diff
@@ -428,3 +428,148 @@ func TestResolveScaleTarget_ReusesLocateCache(t *testing.T) {
 		t.Errorf("cache miss: ok=%v ref=%+v, want ok=true Deployment d", ok, ref)
 	}
 }
+
+func TestGetPodLabels(t *testing.T) {
```
ev-shindin, line 434: Missing a negative-path test: force `walkOwnersUp` to error (e.g. an ownerReference cycle) during `GetPodLabels`, then assert a later `Locate` for the same pod still resolves correctly - i.e. the cache wasn't poisoned (see `locator.go` L218)

internal/collector/locator/locator_test.go
```diff
@@ -428,3 +428,148 @@ func TestResolveScaleTarget_ReusesLocateCache(t *testing.T) {
...
+			for k, v := range tt.want {
+				if gotV, ok := got[k]; !ok || gotV != v {
+					t.Errorf("GetPodLabels()[%q] = %v, want %v", k, gotV, v)
+				}
+			}
+		})
+	}
+}
+
+func TestGetPodLabels_CacheReuse(t *testing.T) {
```
ev-shindin, line 531: `TestGetPodLabels_CacheReuse` calls `GetPodLabels` twice and only asserts the returns are equal - that passes whether or not caching happens. Use the delete-then-verify pattern from `TestResolveScaleTarget_ReusesLocateCache`: delete the pod from the fake client after the first call, then assert the second call still returns the labels.

internal/collector/replica_metrics_test.go
```diff
@@ -798,3 +799,665 @@ func TestCollectReplicaMetrics_ThroughputOrphanSkipped(t *testing.T) {
 	assert.Equal(t, "pod-known", results[0].PodName)
 	assert.Equal(t, float64(0), results[0].GenerationTokenRate, "orphan entry must not contaminate pod-known")
 }
+
+func TestIsLWSWorker(t *testing.T) {
```
ev-shindin, line 1038: `TestIsLWSWorker` doesn't cover a non-nil locator whose `GetPodLabels` returns `nil` (fetch failed) - a distinct branch from the empty-labels-map case. Add a row (`podLabels: nil`, non-nil mock locator) asserting `false`, to lock in the fail-open choice.

ev-shindin (changes requested):
Thanks @shuynh2017! Nice fix - the strategy (drop LWS worker pods so only the leader counts) is right, and `TestCollectReplicaMetrics_LWSWorkerPodsFiltered` validates it end-to-end. Two things I'd address before merge: the `GetPodLabels` cache-poisoning on a transient `resolveScaleTarget` error (locator.go L218), and the inaccurate "only one API read" claim / extra owner-walk on the common label fast-path (locator.go L78). The reuse and test-gap notes are optional polish. Also - no automated e2e regression for the bug itself; `fixtures.EnsureModelServiceLWS` already supports `groupSize>1`, so an e2e asserting replica-count == replicas (not pods) would guard it.
Alsd please rebase the branch to resolve merge conflicts.

internal/collector/locator/locator.go
shuynh2017: fixed.

internal/collector/locator/locator.go
shuynh2017: fixed.

internal/collector/locator/locator.go
shuynh2017: fixed

internal/collector/replica_metrics.go
shuynh2017: fixed

internal/collector/replica_metrics.go
shuynh2017: fixed

internal/constants/constants.go
shuynh2017: fixed

internal/interfaces/saturation_analyzer.go
shuynh2017: fixed

internal/collector/locator/locator_test.go
shuynh2017: fixed

internal/collector/locator/locator_test.go
shuynh2017: fixed

internal/collector/replica_metrics_test.go
shuynh2017: fixed

shuynh2017:
@ev-shindin, thank you for taking time to review. Pls review again.

ev-shindin (approved):
Thanks @shuynh2017 !

PR #1366, opened 2026-07-01
fix: suppress repeated VariantAutoscaling deprecation warnings in tests

### LINKED ISSUE #1365: Reduce log pollution during e2e tests from deprecation warnings
## Description

When running `make test-e2e-full`, the test output is polluted with repeated deprecation warnings about VariantAutoscaling being deprecated every 5 seconds.

### Current Behavior

The deprecation message is logged repeatedly:
```
VariantAutoscaling is deprecated and will be removed in a future release. Migrate to the annotation-based path (add llm-d.ai/managed=true to your HPA or ScaledObject). See docs/developer-guide/migrating-from-va-crd.md for migration steps.
```

This makes it difficult to parse actual test output and adds unnecessary noise to e2e test logs.

### Expected Behavior

Deprecation warnings should either:
- Be logged only once per test run instead of repeatedly
- Be logged at a lower level (DEBUG instead of INFO)
- Be suppressed during e2e tests while still being shown in normal operation

### Suggested Solution

Consider adding a mechanism to detect when running in test mode and either suppress the deprecation warning or log it only once.

sats-23:
Fixes #1365 

## Proposed Changes

- 🧹Pre-set "llm-d.ai/deprecation-warned": "true" in the single private factory buildVariantAutoscaling that all e2e VA helpers delegate to. The annotation signals to the controller that the deprecation notice has already been acknowledged, so the block is skipped from the very first reconcile.

### Pre-review Checklist

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/e2e/fixtures/va_builder.go
+++ b/test/e2e/fixtures/va_builder.go
@@ -134,6 +134,12 @@ func buildVariantAutoscaling(namespace, name, deploymentName, modelID, accelerat
 			Name:      name,
 			Namespace: namespace,
 			Labels:    labels,
+			// Pre-acknowledge the deprecation warning so the controller does not
+			// emit it repeatedly during e2e tests (it fires once per VA object
+			// absent this annotation, which would spam logs on every test cycle).
+			Annotations: map[string]string{
+				"llm-d.ai/deprecation-warned": "true",
+			},
 		},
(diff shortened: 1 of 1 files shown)


lionelvillard:
small PR and failure is unrelated. Merging.

lionelvillard (approved):

PR #1368, opened 2026-07-01
fix(engine): populate V2 saturation decision metrics

ev-shindin:
## Fixes

`wva_saturation_utilization`, `wva_required_capacity`, and `wva_spare_capacity` always read **0** on the V2 saturation path.

## Proposed Changes

- On the V2 path, `buildDecisionsWithOptimizer` (shared by the cost-aware and greedy-by-score optimizers) never copied `Utilization` / `RequiredCapacity` / `SpareCapacity` onto the `VariantDecision` the actuator records — the values were computed on the `AnalyzerResult` but dropped, so the three gauges emitted 0 (despite the Stage-4 comment claiming they were "already set").
- Copy them onto each decision:
  - `Utilization` ← per-variant `VariantCapacity` (demand/capacity ratio).
  - `RequiredCapacity` / `SpareCapacity` ← saturation `AnalyzerResult` — **per-role** for P/D-disaggregated models (`RoleCapacities`), model-level otherwise. `SpareCapacity` is now absolute KV-cache **tokens** (the scale-down companion to `RequiredCapacity`), not the `1 − utilization` ratio previously documented (redundant with the utilization gauge, and never emitted).
- Give `wva_spare_capacity` the same `unit` label as `wva_required_capacity` (shared analyzer-path unit: `continuous` for V2 tokens; fraction for V1) and switch it to the model/role-level label set — the two matched gauges are now symmetric and self-describing.
- Sourced from the saturation analyzer entry (always present on the V2 path); the throughput analyzer also leaves these zero and is not reflected in the `saturation_*` gauges by design.

Updated `VariantDecision` + `WVASpareCapacity` doc comments, the spare/required metric help, and `docs/developer-guide/prometheus.md` (labels, V2 token semantics, and the now-invalid `wva_spare_capacity < 0.2` alert). Added optimizer propagation tests for **both** the cost-aware and greedy-by-score paths (model-level + per-role); updated metrics tests for the spare `unit` label.

## Pre-review Checklist

- [x] **E2E tests** — unit tests added (`cost_aware_optimizer_test.go`, `metrics_test.go`); envtest/E2E via CI
- [ ] **Docs PR** — companion docs PR (`docs/v2-saturation-engine`) documents the now-working metrics

## Release Note

```release-note
Fix wva_saturation_utilization, wva_required_capacity, and wva_spare_capacity reading 0 on the V2 saturation path; wva_spare_capacity is now KV-cache tokens with a `unit` label matching wva_required_capacity.
```

## Notes for reviewers

- **Semantics change:** `wva_spare_capacity` on V2 is now KV-cache tokens (was documented as `1 − utilization`, but always emitted 0). Not plotted in the shipped Grafana dashboard, so no panel breaks.
- **Label change:** `wva_spare_capacity` drops `accelerator_type` and gains `unit` (now matches `wva_required_capacity`, a model/role-level signal). Safe because the metric was always 0.

### DIFF
--- a/docs/developer-guide/prometheus.md
+++ b/docs/developer-guide/prometheus.md
@@ -534,12 +534,12 @@ With WVA metrics, the value for the label `namespace` is the WVA controller name
 
 ### `wva_spare_capacity`
 - **Type**: Gauge
-- **Description**: Spare capacity from saturation analysis; >0 indicates scale-down headroom. Use the `unit` label to interpret the value: `continuous` (V2) = absolute spare in KV-cache **tokens**, `max(0, TotalSupply - TotalDemand/scaleDownBoundary)` — the companion to `wva_required_capacity`; otherwise (V1) = a 0.0-1.0 threshold-relative fraction (`kvCacheThreshold - avg KV usage`).
+- **Description**: Spare capacity; >0 indicates scale-down headroom (per-role for P/D-disaggregated models, model-level otherwise). Use the `unit` label to interpret the value: `continuous` → **token surplus** from the **Token-based analyzer** (`max(0, TotalSupply - TotalDemand/scaleDownBoundary)`); empty → a 0.0-1.0 threshold-relative fraction from the **Percentage-based analyzer** (`kvCacheThreshold - avg KV usage`).
 - **Labels**:
   - `variant_name`: Name of the variant
   - `namespace`: Kubernetes namespace
   - `model_name`: Model name served by the variant
-  - `unit`: `continuous` (V2, token value) or empty (V1, 0.0-1.0 fraction)
+  - `unit`: `continuous` (Token-based analyzer — a token magnitude) or empty (Percentage-based analyzer — 0.0-1.0 fraction)
 - **Use Case**: Track available capacity headroom to prevent saturation and optimize resource allocation
 - **Example**:
   ```
@@ -567,12 +567,12 @@ With WVA metrics, the value for the label `namespace` is the WVA controller name
 
 ### `wva_required_capacity`
 - **Type**: Gauge
-- **Description**: Required capacity; >0 indicates scale-up needed (V2: per-role for P/D-disaggregated models, model-level otherwise). Use the `unit` label to interpret the value: `binary` → 0/1 scale-up signal (V1), `continuous` → token demand (V2).
+- **Description**: Required capacity; >0 indicates scale-up needed (per-role for P/D-disaggregated models, model-level otherwise). Use the `unit` label to interpret the value: `continuous` → **token demand** from the **Token-based analyzer**; `binary` → 0/1 scale-up signal from the **Percentage-based analyzer**.
 - **Labels**:
   - `variant_name`: Name of the variant
   - `namespace`: Kubernetes namespace
   - `model_name`: Model name served by the variant
-  - `unit`: Interpretation of the value (`binary` or `continuous`)
+  - `unit`: `continuous` (Token-based analyzer — a token magnitude) or `binary` (Percentage-based analyzer — 0/1 signal)
 - **Use Case**: Identify when additional capacity is needed and understand the magnitude of demand
 - **Example**:
   ```
--- a/docs/developer-guide/prometheus.md
+++ b/docs/developer-guide/prometheus.md
@@ -534,19 +534,19 @@ With WVA metrics, the value for the label `namespace` is the WVA controller name
 
 ### `wva_spare_capacity`
 - **Type**: Gauge
-- **Description**: Per-variant spare KV-cache capacity (0.0-1.0) from saturation analysis. V1 path: threshold-relative spare (kvCacheThreshold - avg KV usage). V2 path: 1.0 - utilization.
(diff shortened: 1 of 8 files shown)


docs/developer-guide/prometheus.md
```diff
@@ -534,19 +534,19 @@ With WVA metrics, the value for the label `namespace` is the WVA controller name
 
 ### `wva_spare_capacity`
 - **Type**: Gauge
-- **Description**: Per-variant spare KV-cache capacity (0.0-1.0) from saturation analysis. V1 path: threshold-relative spare (kvCacheThreshold - avg KV usage). V2 path: 1.0 - utilization.
+- **Description**: Spare capacity from saturation analysis; >0 indicates scale-down headroom. Use the `unit` label to interpret the value: `continuous` (V2) = absolute spare in KV-cache **tokens**, `max(0, TotalSupply - TotalDemand/scaleDownBoundary)` — the companion to `wva_required_capacity`; otherwise (V1) = a 0.0-1.0 threshold-relative fraction (`kvCacheThreshold - avg KV usage`).
```
shuynh2017: Can we use "Utilization-based analyzer" for V1 and "Token-based analyzer" for V2, instead of 'V1' and 'V2'?

shuynh2017: otherwise (V1) = a ?

docs/developer-guide/prometheus.md
```diff
@@ -534,19 +534,19 @@ With WVA metrics, the value for the label `namespace` is the WVA controller name
...
 - **Labels**:
   - `variant_name`: Name of the variant
   - `namespace`: Kubernetes namespace
   - `model_name`: Model name served by the variant
-  - `accelerator_type`: Type of accelerator being used
+  - `unit`: `continuous` (V2, token value) or empty (V1, 0.0-1.0 fraction)
```
shuynh2017: why `continuous` and not `token` to be precise?

docs/developer-guide/prometheus.md
```diff
@@ -560,14 +560,14 @@ With WVA metrics, the value for the label `namespace` is the WVA controller name
...
-      "0.8"
+      "12000"
     ]
   }
   ```
 
 ### `wva_required_capacity`
 - **Type**: Gauge
-- **Description**: Model-level required capacity; >0 indicates scale-up needed. Use the `unit` label to interpret the value: `binary` → 0/1 scale-up signal (V1), `continuous` → token demand (V2).
+- **Description**: Required capacity; >0 indicates scale-up needed (V2: per-role for P/D-disaggregated models, model-level otherwise). Use the `unit` label to interpret the value: `binary` → 0/1 scale-up signal (V1), `continuous` → token demand (V2).
```
shuynh2017: to be consistent, swap to describe V2 first: e.g. Use the `unit` ltable to interpret the value: `continous` -> ...(V2), `binary` ... (V1)

shuynh2017: To be consistent with `wva_spare_capacity`, either expand `token demand` here, or shorten the `absolute spare` in `wva_spare_capacity` description. I'd suggest to shorten to reduce possibility of drifting.

internal/constants/metrics.go
```diff
@@ -166,12 +166,16 @@ const (
...
+	// Value semantics differ by analyzer (use the "unit" label to distinguish):
+	//   - unit="continuous" (V2): absolute spare in KV-cache tokens
+	//   - otherwise         (V1): 0.0-1.0 threshold-relative KV-cache fraction
+	// Labels: variant_name, namespace, model_name, unit
 	WVASpareCapacity = "wva_spare_capacity"
 
-	// WVARequiredCapacity is a gauge that tracks model-level required capacity.
-	// >0 means scale-up needed.
+	// WVARequiredCapacity is a gauge that tracks required capacity; >0 means scale-up
+	// needed (V2: per-role for P/D-disaggregated models, model-level otherwise).
```
shuynh2017: "(V2: per-role for P/D-disaggregated models, model-level otherwise)", this doesn't apply to WVASpareCapacity?

internal/engines/pipeline/cost_aware_optimizer_test.go
```diff
@@ -73,6 +73,67 @@ var _ = Describe("CostAwareOptimizer", func() {
...
+				}),
+			}
+
+			dm := decisionMap(optimizer.Optimize(ctx, requests, nil))
+			Expect(dm["p"].RequiredCapacity).To(Equal(100.0))
+			Expect(dm["p"].SpareCapacity).To(Equal(10.0))
+			Expect(dm["d"].RequiredCapacity).To(Equal(200.0))
+			Expect(dm["d"].SpareCapacity).To(Equal(20.0))
+		})
+
```
shuynh2017, line 136: Missing test case: A variant with an empty role (Role: "") when RoleCapacities contains a "both" entry. This would exercise the role == "" →
  RoleBoth normalization path (lines 306-308) followed by the lookup at line 309 in cost_aware_optimizer.go

internal/constants/metrics.go
shuynh2017: Do we need to update descriptions in grafana dashboard?

shuynh2017:
@ev-shindin pls address comments.

shuynh2017 (approved):

PR #1371, opened 2026-07-01
test(e2e): add KEDA smoke test suite and non-blocking CI job

mamy-CS:
Fixes # (part of prometheus-adapter deprecation migration)

## Proposed Changes

- :gift: Add `smoke_keda_test.go` - dedicated KEDA smoke test suite covering infrastructure readiness, basic autoscaling (ScaledObject Ready condition, KEDA-created HPA), and error handling; labeled `smoke && keda` so it runs independently from the HPA smoke suite
- :gift: Add `test-e2e-smoke-keda` and `test-e2e-smoke-keda-with-setup` make targets; update `test-e2e-smoke` filter to `smoke && !keda`
- :gift: Add non-blocking `e2e-tests-smoke-keda` CI job that spins its own Kind cluster with `SCALER_BACKEND=keda`, running in parallel with the existing HPA smoke job

### Pre-review Checklist

- [x] **Proposal PR** for any new enhancement or change to existing behavior

**Release Note**

```release-note
NONE

### DIFF
--- a/test/e2e/fixtures/scaled_object_builder.go
+++ b/test/e2e/fixtures/scaled_object_builder.go
@@ -124,7 +124,7 @@ func buildScaledObject(namespace, name, scaleTargetName, vaName string, minRepli
 	prometheusURL := "https://kube-prometheus-stack-prometheus." + monitoringNamespace + ".svc.cluster.local:9090"
 	// Use "namespace" not "exported_namespace": WVA controller emits the metric with label namespace;
 	// exported_namespace is only used by Prometheus Adapter for the external metrics API.
-	query := fmt.Sprintf("max(wva_desired_replicas{variant_name=%q,namespace=%q})", vaName, namespace)
+	query := fmt.Sprintf("wva_desired_replicas{variant_name=%q,namespace=%q}", vaName, namespace)
 
 	spec := kedav1alpha1.ScaledObjectSpec{
 		ScaleTargetRef: &kedav1alpha1.ScaleTarget{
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: test/e2e/fixtures/scaled_object_builder.go [hunk 1]
```diff
@@ -124,7 +124,7 @@ func buildScaledObject(namespace, name, scaleTargetName, vaName string, minRepli
 	prometheusURL := "https://kube-prometheus-stack-prometheus." + monitoringNamespace + ".svc.cluster.local:9090"
 	// Use "namespace" not "exported_namespace": WVA controller emits the metric with label namespace;
 	// exported_namespace is only used by Prometheus Adapter for the external metrics API.
-	query := fmt.Sprintf("wva_desired_replicas{variant_name=%q,namespace=%q}", vaName, namespace)
+	query := fmt.Sprintf("max(wva_desired_replicas{variant_name=%q,namespace=%q})", vaName, namespace)
```
lionelvillard: why `max`?

test/e2e/smoke_keda_test.go
```diff
@@ -0,0 +1,428 @@
...
+		modelServiceName = "smoke-test-ms"
+		deploymentName   = modelServiceName + "-decode"
+		vaName           = "smoke-test-va"
+		scalerName       = "smoke-test-hpa" // base name; ScaledObject will be scalerName+"-so"
+		minReplicas      = int32(1)
+	)
+
+	BeforeAll(func() {
+		By("Cleaning up any existing smoke test resources")
+		cleanupSmokeTestResources()
```
lionelvillard, line 103: Eventually, each test should run in isolation (ie different namespace) because 1. there is no need to cleanup resources 2. tests can run in parallel (faster). Can we done as a follow-up.

test/e2e/fixtures/scaled_object_builder.go
mamy-CS: To take care of the case where two pods briefly coexist, each emitting the metric with the same labels. In this case prom ends up with two time series. And KEDA's prom trigger requires a scalar, if it receives a vector it errors with "multiple time series returned" and stops scaling. max collapses the vector to a single value regardless of how many controller pods are running. Itsa a safe guard for this case.

test/e2e/smoke_keda_test.go
mamy-CS: Yes, that will be a good improvement. Tracked as a follow-up.

lionelvillard (approved):

Issue #1018, opened 2026-04-16
[Autoscaling] RFC: EPP-predictor-based autoscaling


kaushikmitr:
# RFC: EPP-predictor-based autoscaling

## Summary

Adding a new `epp-saturation` analyzer that consumes the pool-level saturation
signal emitted by the gateway-api-inference-extension (EPP) latency detector,
instead of computing saturation from per-pod vLLM metrics.

**Code** (on my fork, not yet upstreamed):
- Analyzer: [`internal/engines/analyzers/epp_saturation/`](https://github.com/kaushikmitr/llm-d-workload-variant-autoscaler/tree/main/internal/engines/analyzers/epp_saturation)
- Engine wiring: [`internal/engines/saturation/engine_epp_saturation.go`](https://github.com/kaushikmitr/llm-d-workload-variant-autoscaler/blob/main/internal/engines/saturation/engine_epp_saturation.go)
- Query registration: [`internal/collector/registration/epp_saturation.go`](https://github.com/kaushikmitr/llm-d-workload-variant-autoscaler/blob/main/internal/collector/registration/epp_saturation.go)
- Deploy: [`deploy/deploy-epp-saturation.sh`](https://github.com/kaushikmitr/llm-d-workload-variant-autoscaler/blob/main/deploy/deploy-epp-saturation.sh) + [`deploy/README-epp-saturation.md`](https://github.com/kaushikmitr/llm-d-workload-variant-autoscaler/blob/main/deploy/README-epp-saturation.md)

This issue presents validation results from a GKE experiment and proposes a
few refinements based on what we observed.

## Key discussion points

Two questions I'd like community input on up front. They're the framing I
most want feedback on before going deeper into the proposal below.

### 1. How do we show this (or any analyzer) is actually working?

There's no agreed benchmark in this repo for evaluating an analyzer's quality.
The data I collected (Pearson correlation between signal and actual TTFT,
SLO violation rate, replica-count stability) is a reasonable first cut, but
it's ad-hoc, I made it up. Before investing in more analyzers, it would
be valuable to agree on a standard evaluation harness.

Some specific questions:

- **What defines "correct" behavior?** Is it (a) signal tracks actual
  latency, (b) replica recommendations keep SLO violations under X%,
  (c) total GPU cost minimized subject to SLO, (d) some composite? Each
  implies a different quality metric.
- **What workload patterns should we evaluate against?** Single-level
[...]

PR #1376, opened 2026-07-03
Epp saturation observability

kaushikmitr:
# EPP-saturation analyzer: SLO-driven autoscaling on the EPP's predicted latency (RFC #1018)

Implements the proposals from RFC #1018: an `epp-saturation` analyzer that
derives its scaling signal from the EPP's **P90 predicted latency** against
operator-configured SLOs, plus the observability, damping, and validation work
around it. Benchmarked end-to-end with active scaling on GKE (H100 × 24,
Qwen3-32B TP=2): the default configuration reaches **95.8 % combined SLO
attainment at ~26 % lower cost than a peak-sized static pool**.

Full results, figures, and the experiment frontier:
`docs/developer-guide/epp-saturation-benchmark.md`. Deployable objects:
`config/samples/epp-saturation-benchmark/`.

## What's in this PR

**The analyzer & signal**
- `saturation = max(P90 predTTFT / ttftSLOMs, P90 predTPOT / tpotSLOMs)`
  (defaults 3000/100 ms) — SLO policy lives in WVA config, not the EPP.
  Predicted-preferred with a NaN-proof fallback to actual latency; idle pools
  read 0 and scale toward minReplicas.
- **Why P90, not the mean:** the SLO objective is per-request tail attainment,
  and the mean is flat below the saturation knee (a trailing indicator); the
  P90 rises as soon as queueing variance appears. Measured as part of the
  default configuration: 95.8 % vs 84.7 % attainment at equal episode cost.
- **Defaults are calibrated to the P90 signal's measured healthy band**:
  `scaleUpThreshold` 0.55, `scaleDownBoundary` 0.40, `smoothingAlpha` 0.6.
  Workloads whose SLO sits close to base latency should raise the band
  (documented on the fields and in the benchmark doc).

**Overshoot & warmup damping**
- `saturationCap` (default 2.0): clamps the raw signal ahead of the EMA — knee
  signals reach 10–90× SLO and otherwise poison the smoothed signal (measured:
  smoothed peak ~32 → ~2, overshoot recovery halved).
- Credit for in-flight replicas: scale-up asks only for the deficit beyond
  pods already warming; scale-down deliberately uses the uncredited signal.
- `holdScaleUpWhileWarming` (default **off**): optional hold that cuts
  step-load peaks (11→8 measured) but falls behind ramps (79 % scored) —
  opt-in for steady, cost-sensitive workloads.

**Observability (RFC #1018 proposals 2–4)**
- `wva_epp_saturation_raw` / `wva_epp_saturation_smoothed` gauges for tuning
  `smoothingAlpha` and observing the clamp.
- `ScalingCapped` condition + `wva_scale_capped` gauge — "fine at the cap" vs
  "wanted more but blocked", uncapped recommendation in the message. The
  condition appears on every VA (False on non-EPP paths) by design.
- `MetricsAvailable=False` with an EPP-specific reason on signal loss; last
  good target and cap state preserved in the decision cache.

**Hardening**
- 13 adversarially-verified review findings fixed, including: the in-flight
  credit could flip an overloaded pool into scale-down; a stalled predictor
  produced present-but-NaN samples that defeated the PromQL fallback;
  per-model SLO overrides were silently dropped by config merge; analyzer
  config validation was never wired in; namespace collisions in the warmup
  hold; stale `ScalingCapped` on the scale-from-zero and signal-loss paths.

**Benchmark & deployables**
- `docs/developer-guide/epp-saturation-benchmark.md`: headline scorecard vs
  static-at-peak, the violation-bill decomposition
  (trigger lag + ask lag + grant lag + warmup), predictor-calibration
  sensitivity (cold vs calibrated predictor = 85.6 % vs 95.8 % on identical
  config — burn-in protocol documented), reproduction protocol, and the full
  experiment frontier including negative results (burst grants alone, HPA
  decay windows, low thresholds on the mean signal).
- `config/samples/epp-saturation-benchmark/`: the exact VA, HPA (benchmarked
  scale policies), EPP plugin config, scrape-job fragments + RBAC, vLLM warmup
  patch (compile-cache hostPath + fast startupProbe → pod-ready ~2 m 15 s →
  ~100 s), and the inference-perf scored profile, with a deploy-order README.

## Benchmark summary

Prefill-Heavy 4000/1000, staged ramp 2→10→1 rps, SLO ttft ≤ 3 s / tpot ≤ 100 ms,
bounds 3–8 replicas:

| | WVA (defaults) | static-at-peak (8) |
|---|---|---|
| Combined SLO attainment | **95.8 %** | 100 % |
| Avg replicas (full episode) | **5.93** | 8 |

Residual violations sit in one shallow ~2-minute transient at the rate-6
capacity crossing; the rate-10 peak stage runs violation-free (capacity is in
place before it arrives, p90 TTFT 0.59 s).

## Testing

- `make test` green (unit + envtest); new tests for the metrics emitters,
  latency derivation (max-of-terms, NaN→0), clamp, credit direction guards,
  warmup-hold helper, `ScalingCapped` condition, and the signal-unavailable
  cache path.
- Full active-scaling loop validated live on two GKE clusters
  (WVA → prometheus-adapter → external-metric HPA → decode), bidirectional,
  cap firing, all gauges.

## Known limitations / follow-ups

- A broken scrape reads as idle (empty result = 0 saturation, intended for
  idle pools); an `up{}` health guard is deferred.
- Multi-variant cap detection reports False rather than Unknown.
- Leading-indicator signals (queue depth / running requests) and sleep-mode
  warm pools are the measured path beyond ~96 % on step loads.
- Predicted-TPOT falls back to actual TPOT on builds that don't emit it.


atantawi:
## Solution-approach assessment + clarifications (not a code review)

Ahead of a line-level review, I wanted to step back and assess this at the **architecture / methodology** level — is the approach valid, sensible, and practical, and how does it contrast with the analyzers already in `main`? Code-quality review will come separately. Nothing below is blocking a *review*; the handful of items I'd flag as needing an answer on approach grounds are called out explicitly at the end.

First — this is a strong, unusually honest contribution, and the benchmark/validation work is the standout piece (more on that below).

### Where it sits in the analyzer landscape

The engine selects **one** top-level path per reconcile (`switch analyzerName`). This adds a fourth, mutually-exclusive path:

| | Signal source | Where the capacity model lives | Stance |
|---|---|---|---|
| **V2 pipeline** (`saturation_v2` + `throughput`) | vLLM per-pod (KV tokens, ITL) | **inside WVA** — white-box; composed via max-required / min-spare | reactive, measured |
| **queueing-model** | EPP arrival rate + vLLM latencies | **inside WVA** — M/M/c-style, learns α/β/γ online | model-based, needs traffic to converge |
| **epp-saturation** (this PR) | EPP predicted TTFT/TPOT (probe-driven) | **outside WVA** — delegated to the EPP latency detector | probe-based, black-box |

The defining contrast: every existing analyzer builds its own capacity model from ground-truth serving metrics; this one consumes a pre-digested latency prediction from an external component and turns it into a scaling signal by dividing by the SLO. Both its strengths and its risks flow from that one decision.

One refinement worth surfacing: the RFC and the `analyzer.go` package comment describe consuming a pre-computed `inference_extension_latency_detector_pool_saturation` gauge, but the code actually computes `max(predTTFT/SLO, predTPOT/SLO)` itself. That's a genuine improvement over the RFC — **the SLO policy stays in WVA**; only the latency *prediction* is delegated. The docs just lag the implementation here.

### Is the approach sensible? — Yes, with a clear thesis

The RFC motivation holds up: WVA and the EPP predictor would otherwise be "two parallel models of the same hardware," so consuming the EPP signal is a defensible **buy-vs-build** on the plant model; it also decouples WVA from vLLM metric-name churn (the reason `sglang_parser.go` exists in `saturation_v2`); and SLO-denominated saturation is the most direct proxy for the thing operators actually care about (attainment), where the token/throughput analyzers scale on *correlates* of latency. As a concept it's non-redundant — "delegate the model, keep the policy" is a distinct point the other three don't occupy.

### Technical validity — the core tension, and why it mostly survives

The open-loop model is inaccurate exactly where it matters, and the PR is candid about it: `perReplicaCapacity = 1/N` assumes saturation is inversely linear in replica count, but continuous batching makes latency flat until ρ≈1 then vertical (the doc's own finding #1 — a step, not a line). A linear `1/N` model is least accurate at the knee.

It nonetheless works because the design **doesn't trust the model's magnitude** — it (1) triggers *below* the knee (`scaleUpThreshold=0.55`, on the P90 pre-rise), where being wrong about the slope is cheap; (2) clamps + EMA-smooths so a knee-spike can't poison the signal; and (3) relies on closed-loop re-convergence via the HPA + next reconcile, with the in-flight credit and the credited-up / uncredited-down asymmetry preventing double-provisioning and mid-warmup flip-to-scale-down. This is sound control engineering: a crude plant model wrapped in a conservative, early-triggering feedback loop. Correctness rests on the loop, not the formula — which is a valid design, provided the feedback cadence stays fast relative to load dynamics (honestly disclosed in finding #7: discontinuous steps onto a drained pool remain the failure mode).

### The concerns I'd rank highest

1. **External-component coupling** the white-box analyzers don't have: a measured 10-point attainment swing (85.6% → 95.8%) from EPP *predictor calibration state alone*, and degraded scaling after any EPP/predictor restart. Correctness is conditional on an external online-trained model's convergence.
2. **Silent under-provisioning on scrape loss**: empty/NaN → `saturation=0` → drain toward `minReplicas`, *with `MetricsAvailable=True`*. A broken scrape under load looks identical to idle, and the single-signal black-box design has nothing to cross-check it against.
3. **Single-workload calibration shipped as code defaults**: 0.55/0.40 were derived from one workload/model/GPU; the doc says they must be re-derived for decode-heavy / symmetrical / SLO-near-base-latency. This partially undercuts the "simpler operator mental model" pitch.
4. **"Predictive" is overstated** — the P90 signal leads the *mean* (variance precedes the mean shift), i.e. *less lagging*, not leading. The benchmark doc walks this back correctly and names true leading indicators (queue depth / running-request count) as the path to ≥98%.

### The standout: validation methodology

RFC #1018's own open question was "we have no way to prove an analyzer works — the metrics were ad-hoc." This PR answers exactly that: an A/B against static-at-peak, cost quantified in GPU-replica-hours, a violation-bill decomposition (`(trigger + ask + grant + warmup lag) × arrival rate`), a full ablation frontier isolating each knob, a formalized control-law model, and `analyzer-checklists.md` as a reusable graduation gate. Against the cardinal sin of performance modeling — validating nothing — this is exemplary and honest about its own divergences. The warmup decomposition, HPA burst-policy tuning, and checklist are valuable **independent** of this analyzer.

---

## Clarifying questions & suggestions

**Safety / correctness**
1. **Scrape-loss aliases as idle** (concern #2). Can the `up{}` guard land in this PR rather than being deferred — or at minimum flip `MetricsAvailable=False` when the series is absent but replicas are actively serving?
2. **Predictor-calibration coupling** (concern #1). Is there a signal WVA can read to *detect* a cold/miscalibrated predictor and fall back to the safety-net (hold last replicas) instead of acting on a bad prediction?

**Defaults / calibration**
3. **Single-workload defaults.** What's the expected behavior on the other two reference workloads (decode-heavy, symmetrical)? Would it be safer to ship 0.55/0.40 as documented starting points rather than compiled-in defaults, or to gate them behind a stated workload assumption?
4. **Threshold derivation.** Is there a repeatable procedure (or helper) for an operator to find the healthy P90 band for *their* workload, so re-calibration isn't guesswork? The benchmark recipe seems distillable into a short "how to set your thresholds" note.

**Architecture / fit**
5. **Standalone path vs. V2 pipeline participant.** The analyzer already emits `VariantCapacity`/`RequiredCapacity`/`SpareCapacity` — the shape the V2 pipeline composes. Given the "leading-indicator blend" is the headline future work, why build the blend inside this analyzer rather than registering epp-saturation as a pipeline analyzer and reusing the existing max-required / min-spare aggregation? What specifically doesn't compose — the bespoke credited-up / uncredited-down scale-down logic?
6. **Cross-check.** As a standalone single-signal path it inherits none of the multi-analyzer cross-validation. Any appetite for a cheap sanity gate (e.g. a KV-util or queue-depth guard) even while it stays a separate path — which would directly mitigate #1?

**Model framing**
7. **`1/N` linearity.** Is the intent explicitly "crude open-loop model, correctness comes from early trigger + feedback re-convergence"? If so, stating that as the design principle would reframe several of the concerns above and make the clamp/EMA/credit choices read as deliberate.
8. **Stale language.** Can the `analyzer.go` header comment and the RFC be reconciled with what shipped (WVA computes saturation from predicted latency ÷ SLO; the signal is less-lagging, not "predictive")?

**Non-blocking suggestions**
9. Call out the warmup decomposition, HPA burst-policy tuning, and `analyzer-checklists.md` as reusable across *all* analyzers, so they don't get buried as epp-saturation-specific.
10. Consider filing the true leading indicators (queue depth / running-request count) as a tracked follow-up now, since the doc already identifies them as the measured path to ≥98%.

Items I'd flag as genuinely needing an answer on approach grounds before merge: **#1 (scrape-loss safety), #3 (defaults), and #5 (standalone vs. pipeline)**. Everything else is discussion.

kaushikmitr:
Focussing on these below we can tackle the first two them by adding guardrails in the code. 
1. (scrape-loss safety)
2. (defaults), 
For this:
4. (standalone vs. pipeline). I dont think the scale up, scale down, hysterisis, EMA together compose well with the pipeline apprroach. 

But a bigger point is the entire logic of this autoscaling can be expressed in these steps below. Its entirely analytical with hyper parameters and should it be done in WVA or just use KEDA + CEL directly?

https://github.com/kaushikmitr/llm-d-workload-variant-autoscaler/blob/epp-saturation-observability/docs/developer-guide/epp-saturation-benchmark.md#appendix-control-law-model

kaushikmitr:
as discussed with @asm582 this can be implemented directly in KEDA. Created this PR:
http://github.com/llm-d/llm-d/pull/2011

asm582:
@kaushikmitr can we close this PR?

kaushikmitr:
closing this as we moved the implementation to KEDA/EPP without using any wva component

PR #1378, opened 2026-07-04
feat: Add DRA-aware GPU accounting

MohanKumar21:
Fixes #

## Proposed Changes

- :gift: Add DRA-aware GPU accounting: WVA now counts accelerators requested through Kubernetes Dynamic Resource Allocation (DRA) in addition to standard extended resources (e.g. `nvidia.com/gpu`).
  - New `internal/resources` helpers resolve device counts from `pod.spec.resourceClaims` — direct `ResourceClaim` references, `ResourceClaimTemplate` references (including generated claims from `pod.status.resourceClaimStatuses`), `exactCount` allocation mode (with the Kubernetes default of 1 when count is omitted), and `firstAvailable` subrequests (max across alternatives). Allocated claims use `status.allocation.devices.results`, which also covers `allocationMode: All`.
  - Container-level DRA extended resources (`deviceclass.resource.kubernetes.io/<device-class>`) are summed alongside vendor GPU resources.
- :gift: Cluster GPU usage discovery (`K8sWithGpuOperator.DiscoverUsage`) now includes DRA device requests from running pods when computing per-accelerator usage.
- :gift: The saturation engine (both `BuildVariantStates` and the v2 analysis path) resolves `gpusPerReplica` via a new DRA-aware helper (`utils.GetDRAAwareGPUsPerReplica`) that handles leader/worker (LWS) pod templates and falls back to the legacy pod-template count if DRA objects cannot be resolved.
- :broom: Register `resource.k8s.io/v1` types in the scheme and extend the manager ClusterRole with read access (`get`/`list`/`watch`) to `deviceclasses`, `resourceclaims`, `resourceclaimtemplates`, and `resourceslices`. Lookups degrade gracefully (count 0) on clusters without the DRA API (`NoMatchError`) or when claims are not found.

### Pre-review Checklist

- [ ] **E2E tests** for any new behavior — unit tests cover the DRA counting logic (`internal/resources/resource_test.go`, `internal/discovery/k8s_with_gpu_operator_test.go`); no e2e added since the kind-emulator environment does not provision a DRA driver.

**Release Note**

```release-note
The workload-variant-autoscaler now accounts for GPUs requested via Kubernetes Dynamic Resource Allocation (DRA) — ResourceClaims, ResourceClaimTemplates, and deviceclass.resource.kubernetes.io/* extended resources — when computing per-replica GPU counts and cluster GPU usage. The manager ClusterRole gains read-only access to resource.k8s.io resources. Clusters without the DRA API are unaffected.
```

**Docs**


internal/utils/utils.go
```diff
@@ -374,6 +374,54 @@ func GetProductKeys() []string {
...
+	workerGPUs, err := resources.GetPodTemplateGPUs(ctx, c, scaleTarget.GetNamespace(), workerTemplate)
+	if err != nil {
+		ctrl.LoggerFrom(ctx).V(logging.DEBUG).Info("Failed to resolve DRA resources for worker pod template; falling back to legacy GPU count",
+			"namespace", scaleTarget.GetNamespace(),
+			"scaleTarget", scaleTarget.GetName(),
+			"error", err)
+		return scaleTarget.GetTotalGPUsPerReplica()
+	}
+
+	if leaderTemplate == workerTemplate {
```
ev-shindin: This guard undercounts LWS groups that use the single-template convention. Per LWS v0.8.0, `Size` is the **total** pods per group *including the leader*, and `LeaderTemplate` "defaults to the worker template if not specified" — so a leaderless group of `Size=N` runs N identical pods and needs `N × workerGPUs`. Because `GetLeaderPodTemplateSpec()` already falls back to the worker template (returning the same pointer), `leaderGPUs` here already equals `workerGPUs`; zeroing it makes `total = (N-1) × workerGPUs` — a one-pod undercount (e.g. `size:4`, worker `nvidia.com/gpu:1` → 3 instead of 4). It also makes two functionally identical specs disagree: nil `leaderTemplate` undercounts, but explicitly duplicating the template into both fields (distinct pointers) computes the correct total. Just delete the `if leaderTemplate == workerTemplate { leaderGPUs = 0 }` block — `total := leaderGPUs + (groupSize-1)*workerGPUs` is then correct in both the nil-leader and distinct-leader cases. This value feeds `BuildVariantStates` and the v2 capacity store, so the undercount skews capacity/headroom and quota math.

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -201,8 +202,13 @@ func (d *K8sWithGpuOperator) DiscoverUsage(ctx context.Context) (map[string]int,
 			continue
 		}
 
-		// Sum GPU requests from all containers
+		// Sum GPU requests from all containers and DRA claims.
 		gpuCount := getPodGPURequests(&pod)
+		draCount, err := resources.GetPodDRADeviceCount(ctx, d.Client, &pod)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve DRA device requests for pod %s/%s: %w", pod.Namespace, pod.Name, err)
```
ev-shindin: A single pod's DRA lookup error aborts *all* cluster GPU-usage discovery (`return nil, err`), dropping usage for every other pod that resolved fine. This is inconsistent with the sibling path added in this same PR - `GetDRAAwareGPUsPerReplica` treats the identical error class as non-fatal (logs at DEBUG, falls back). A transient API blip, or `Forbidden` before the new `resource.k8s.io` RBAC finishes propagating on a rolling upgrade, would then zero out usage for the whole cluster that cycle. Suggest degrading per-pod (log at DEBUG, skip the DRA contribution) to match the engine path.

internal/resources/resource.go
```diff
@@ -44,6 +50,172 @@ func GetContainersGPUs(containers []corev1.Container) int {
...
+	for _, status := range statuses {
+		if status.ResourceClaimName != nil {
+			statusByName[status.Name] = *status.ResourceClaimName
+		}
+	}
+
+	total := 0
+	for _, podClaim := range spec.ResourceClaims {
+		switch {
+		case podClaim.ResourceClaimName != nil:
```
ev-shindin, line 137: `DiscoverUsage` calls this per pod, so a `ResourceClaimName` (name-referenced, non-generated) claim shared by multiple pods - a legitimate DRA pattern, and directly relevant to the LWS gang-sharing motivation - is counted once per referencing pod: N pods sharing a k-device claim sum to `N×k` instead of `k`. The per-pod-generated `ResourceClaimTemplateName` path isn't affected. Either dedupe allocated claims by name across the pod loop in `DiscoverUsage`, or add a comment confirming shared named claims aren't a supported pattern here.

internal/resources/resource.go
```diff
@@ -30,6 +34,8 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
 )
 
+const draDeviceClassResourcePrefix = "deviceclass.resource.kubernetes.io/"
```
ev-shindin: `draDeviceClassResourcePrefix` duplicates `resourcev1.ResourceDeviceClassPrefix` (`= "deviceclass.resource.kubernetes.io/"`) from `k8s.io/api/resource/v1`, which is already imported in this file as `resourcev1`. Drop the local const and use `resourcev1.ResourceDeviceClassPrefix`

config/base/rbac/manager-clusterrole.yaml
```diff
@@ -153,3 +153,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - resource.k8s.io
+  resources:
+  - deviceclasses
```
ev-shindin: The code only `Get`s `resourceclaims` and `resourceclaimtemplates`; it never reads `deviceclasses` or `resourceslices`. Consider dropping those two resources from the grant (and the matching kubebuilder marker at `variantautoscaling_controller.go#L103`) until they're actually used, to keep the ClusterRole minimal.

internal/utils/utils.go
```diff
@@ -374,6 +374,54 @@ func GetProductKeys() []string {
 	return slices.Sorted(maps.Keys(labels))
 }
 
+// GetDRAAwareGPUsPerReplica resolves the accelerator count for a scale target,
+// including exact-count DRA ResourceClaimTemplates and ResourceClaims when the
+// Kubernetes API can be queried. It falls back to the legacy accessor behavior
+// when DRA objects cannot be resolved.
+func GetDRAAwareGPUsPerReplica(ctx context.Context, c client.Reader, scaleTarget scaletarget.ScaleTargetAccessor) int {
```
ev-shindin, line 179: GetDRAAwareGPUsPerReplica` has **zero** test coverage - including the `leaderTemplate == workerTemplate` branch, which is exactly the undercount above. Add cases (fake client with `resourcev1` registered): LWS `size>1` distinct leader/worker, LWS `size>1` nil leaderTemplate (assert `N × workerGPUs`), the error-fallback to `GetTotalGPUsPerReplica`, nil target, and all-zero floor.

internal/discovery/k8s_with_gpu_operator_test.go
ev-shindin, line 1: The DRA path in `DiscoverUsage` is untested — the fake clients register only `corev1`, and no pod fixture sets `Spec.ResourceClaims`. Add a case (register `resourcev1`) asserting a claim's device count is added to `usageByType`, plus one asserting the error-propagation behavior

internal/resources/resource_test.go
```diff
@@ -0,0 +1,208 @@
...
+		name     string
+		requests []resourcev1.DeviceRequest
+		want     int
+	}{
+		{
+			name: "exact count defaults to one",
+			requests: []resourcev1.DeviceRequest{{
+				Name: "gpu",
+				Exactly: &resourcev1.ExactDeviceRequest{
+					DeviceClassName: "gpu.example.com",
```
ev-shindin, line 251: The error branches (`meta.IsNoMatchError` -> 0, `apierrors.IsNotFound` -> 0, other error -> wrapped/propagated) and the direct `podClaim.ResourceClaimName` branch are untested. Add cases via `interceptor.Funcs` (force a non-NotFound Get error) and a `ResourceClaimName`-referenced fixture.

ev-shindin (changes requested):
Thanks @MohanKumar21! Nice feature please rebase PR and address review comments.

internal/utils/utils.go
```diff
@@ -370,6 +370,54 @@ func GetProductKeys() []string {
...
+			"scaleTarget", scaleTarget.GetName(),
+			"error", err)
+		return scaleTarget.GetTotalGPUsPerReplica()
+	}
+
+	// A leaderless group (nil LeaderTemplate) has leaderGPUs == workerGPUs because
+	// GetLeaderPodTemplateSpec falls back to the worker template. Size counts the
+	// leader, so total = leaderGPUs + (Size-1)*workerGPUs is correct in both the
+	// leaderless (N*workerGPUs) and distinct-leader cases.
+	total := leaderGPUs + (int(groupSize)-1)*workerGPUs
```
ev-shindin: The shared-claim dedup added to `DiscoverUsage` (the `seenClaims` map) is not applied on this capacity path. `GetPodTemplateGPUs` (called at L383 and L401) takes no `seen` map, so if a scale-target pod template references a literal `ResourceClaimName` (a single shared claim) rather than a `ResourceClaimTemplateName`, that one physical claim is counted in the leader term **and** again in `(groupSize-1)*workerGPUs`. Thread a shared `seen` map through the two `GetPodTemplateGPUs` calls, or explicitly treat `ResourceClaimName` in a scale-target template as unsupported/ignored and document why.

internal/resources/resource.go
```diff
@@ -44,6 +48,197 @@ func GetContainersGPUs(containers []corev1.Container) int {
...
+			return 0, nil
+		}
+		return 0, fmt.Errorf("get ResourceClaimTemplate %s/%s: %w", namespace, name, err)
+	}
+	return CountDRADeviceRequests(template.Spec.Spec.Devices.Requests), nil
+}
+
+// CountDRADeviceRequests returns the count for DRA requests that declare an
+// exact count. The Kubernetes default for ExactCount with omitted count is one.
+func CountDRADeviceRequests(requests []resourcev1.DeviceRequest) int {
```
ev-shindin, line 205: `CountDRADeviceRequests` (and the allocation-results path in `getResourceClaimDeviceCount` L169 / `getResourceClaimTemplateDeviceCount` L189) sum **every** device regardless of `DeviceClassName`. The standard path (`GetContainersGPUs`) filters against `constants.VendorResources`; here a claim mixing a GPU with a NIC/FPGA — or a non-GPU claim entirely — inflates the GPU total that drives scaling. Filter by a known GPU device-class allow-list, or document the "every DRA claim on a WVA-managed pod is GPU-only" assumption and why it holds.

internal/discovery/k8s_with_gpu_operator.go
```diff
@@ -272,6 +289,7 @@ func getPodGPURequests(pod *corev1.Pod) int {
 				containerGPUs += int(qty.Value())
 			}
 		}
+		containerGPUs += resources.GetContainersDRAExtendedResources([]corev1.Container{container})
```
ev-shindin, line 292: The regular-container equivalent (L279) is covered by `TestGetPodGPURequests_DRAExtendedResource`, but no test exercises an **init** container requesting a `deviceclass.resource.kubernetes.io/*` resource, leaving this line and its `max(initMax, regularTotal)` interaction uncovered.

internal/resources/resource.go
```diff
@@ -44,6 +48,197 @@ func GetContainersGPUs(containers []corev1.Container) int {
...
+		key := namespace + "/" + name
+		if seen[key] {
+			return true
+		}
+		seen[key] = true
+		return false
+	}
+
+	total := 0
+	for _, podClaim := range spec.ResourceClaims {
```
ev-shindin, line 135: The loop accumulates `total` and returns the partial value with the error (L143/L153/L161); `DiscoverUsage` adds `draCount` even when `err != nil`. No test has a pod with >1 `spec.resourceClaims` where the first resolves and a later one fails. Add a two-claim pod (first → 2, second → Get fails) asserting `2, err != nil`.

internal/discovery/dra_usage_test.go
```diff
@@ -0,0 +1,120 @@
...
+		draGPUClaim("shared-claim", 2),
+	).Build()
+
+	result, err := NewK8sWithGpuOperator(c).DiscoverUsage(context.Background())
+
+	require.NoError(t, err)
+	assert.Equal(t, 2, result["NVIDIA-H100-SXM5-80GB"])
+}
+
+func TestDiscoverUsage_DRAErrorDegradesPerPod(t *testing.T) {
```
ev-shindin, line 101: The pod has no non-DRA GPU request, so the result is 0 regardless of whether the DRA-error-degrades path runs. Give the pod a real `nvidia.com/gpu` request so the expected result is non-zero - the test then fails if a DRA error aborts the whole pod's accounting instead of just the DRA piece.

internal/utils/dra_gpus_test.go
```diff
@@ -0,0 +1,142 @@
...
+					Requests: corev1.ResourceList{
+						corev1.ResourceName("nvidia.com/gpu"): resource.MustParse(gpus),
+					},
+				},
+			}},
+		},
+	}
+}
+
+func TestGetDRAAwareGPUsPerReplica(t *testing.T) {
```
ev-shindin, line 66: All table cases use plain `nvidia.com/gpu` templates; only the fallback branch (`...FallsBackOnError`, L111) touches real DRA objects. Add a case where leader/worker templates reference a `ResourceClaimTemplateName` resolving in the fake client, e.g. leader 1 GPU + worker template 2 devices + size 3 -> expect `1 + 2*2 = 5`.

internal/resources/resource.go
```diff
@@ -44,6 +48,197 @@ func GetContainersGPUs(containers []corev1.Container) int {
...
+			}
+		}
+	}
+	return total
+}
+
+// GetPodTemplateGPUs returns the accelerator count declared by a pod template,
+// including standard extended GPU resources and exact-count DRA claims.
+func GetPodTemplateGPUs(ctx context.Context, c client.Reader, namespace string, template *corev1.PodTemplateSpec) (int, error) {
+	if template == nil {
```
ev-shindin, line 68: The nil guards in `GetPodTemplateGPUs` (L68), `GetPodDRADeviceCount` (L93), and `GetPodSpecDRADeviceCount` (L106) sit on a per-reconcile / per-pod hot path; a refactor that dereferences before the check would panic uncaught. Add three one-line tests asserting each returns `0, nil` for nil input.

internal/resources/resource_test.go
```diff
@@ -0,0 +1,313 @@
...
+	}).Build()
+	pod := podWithNamedClaim("pod", "some-claim")
+
+	_, err := GetPodDRADeviceCount(context.Background(), c, pod, nil)
+
+	require.Error(t, err)
+	require.True(t, apierrors.IsForbidden(err))
+}
+
+func TestCountDRADeviceRequests(t *testing.T) {
```
ev-shindin, line 240: Every table case passes a single-element `requests` slice, so the `total +=` accumulation across multiple `DeviceRequest`s in one claim is unverified. Add a case with two requests (Exactly count=2 + Exactly count=1) asserting the sum (3).

ev-shindin (commented):
Thanks @MohanKumar21 ! Please address two correctness gaps ans several provide missing test-cases.

ev-shindin:
@MohanKumar21 do you have a time to finilize this. We are several days before code freeze.

PR #1379, opened 2026-07-06
fix(pod): cache per-metric entries so Get(metricName) returns results

### LINKED ISSUE #619: Fix PodScrapingSource to support fetching individual metrics
## Problem
The `PodScrapingSource` currently returns all metrics when calling `Get("all_metrics")` instead of providing a way to return individual metrics (e.g., `Get("vllm:queue_size")`).

## Desired Behavior
Update the implementation so that `Get(<metric_name>)` returns the specific metric requested, such as `Get("vllm:queue_size")`, rather than always returning all metrics.  This will allow consumers to fetch only the metrics they need and improve API usability.

## Suggested Solution
- Refactor the `PodScrapingSource. Get()` method to support fetching a single metric by name. 
- Update all existing usages of `Get("all_metrics")` to return the queried metric.
- Add tests to validate individual metrics fetch behavior.

## Relevant Code
- [`PodScrapingSource.Get(queryName, params)`](https://github.com/llm-d-incubation/workload-variant-autoscaler/blob/main/internal/collector/source/pod/pod_scraping_source.go) currently returns cached `"all_metrics"`.
- To support fetching specific metrics (e.g. `"vllm:queue_size"`), update the cache and querying logic so that calling `Get("<metric_name>")` returns only that metric, not all.

---
**Labels:** bug, enhancement
**Type:** Bug

Goutham-Annem:
## Summary

Fixes #619.

`PodScrapingSource.Refresh()` aggregated all scraped pod metrics into a single `MetricResult` stored under the cache key `"all_metrics"`. As a result, `Get("vllm:queue_size", nil)` always returned `nil` — the lookup key never matched anything in the cache.

**Root cause**: `cache.Set` was only called once with `BuildCacheKey("all_metrics", nil)`, so any query name other than `"all_metrics"` produced a cache miss.

**Fix**: After building the aggregated result, split the values by their `__name__` label and store one `CachedValue` per distinct metric name. `Refresh()` also returns these per-metric keys alongside `"all_metrics"` in its result map, so callers that iterate over results see individual metrics immediately.

A new `splitByMetricName` helper (unexported) does the grouping; it skips values that lack a `__name__` label so no regression occurs for metrics that don't carry the label.

## Changes

- `pod_scraping_source.go`: add `splitByMetricName`, call it from `Refresh()`, store per-metric cache entries
- `pod_scraping_source_test.go`: add `Describe("splitByMetricName")` with 3 cases; extend `Get` tests with per-metric lookup cases; extend the Refresh integration test to assert per-metric keys appear in results and `Get()` works for individual metric names

## Test plan

- [x] All 37 existing + new unit tests pass (`go test ./internal/collector/source/pod/...`)
- [x] Full test suite passes (`go test ./...`)
- [x] `go vet` clean

### DIFF
--- a/internal/collector/source/pod/pod_scraping_source.go
+++ b/internal/collector/source/pod/pod_scraping_source.go
@@ -140,8 +140,12 @@ func (p *PodScrapingSource) Refresh(ctx context.Context, spec source.RefreshSpec
 
 	// Also cache per-metric slices keyed by __name__ so that
 	// Get("vllm:queue_size", nil) returns only the relevant values.
-	perMetric := p.splitByMetricName(aggregated)
+	perMetric := splitByMetricName(aggregated)
+
+	out := make(map[string]*source.MetricResult, len(perMetric)+1)
+	out["all_metrics"] = aggregated
 	for name, mr := range perMetric {
+		out[name] = mr
 		p.cache.Set(source.BuildCacheKey(name, nil), *mr, p.config.DefaultTTL)
 	}
 
@@ -151,11 +155,6 @@ func (p *PodScrapingSource) Refresh(ctx context.Context, spec source.RefreshSpec
 		"metricCount", len(aggregated.Values),
 		"distinctMetrics", len(perMetric))
 
-	out := make(map[string]*source.MetricResult, len(perMetric)+1)
-	out["all_metrics"] = aggregated
-	for name, mr := range perMetric {
-		out[name] = mr
-	}
 	return out, nil
 }
 
@@ -410,11 +409,12 @@ func (p *PodScrapingSource) parsePrometheusMetrics(reader io.Reader, podName str
 
 // splitByMetricName groups aggregated values by their "__name__" label,
 // returning one MetricResult per distinct metric name.
-func (p *PodScrapingSource) splitByMetricName(aggregated *source.MetricResult) map[string]*source.MetricResult {
+// Values without a "__name__" label and the reserved name "all_metrics" are skipped.
+func splitByMetricName(aggregated *source.MetricResult) map[string]*source.MetricResult {
 	grouped := make(map[string][]source.MetricValue)
 	for _, v := range aggregated.Values {
 		name := v.Labels["__name__"]
-		if name == "" {
+		if name == "" || name == "all_metrics" {
 			continue
 		}
 		grouped[name] = append(grouped[name], v)
--- a/internal/collector/source/pod/pod_scraping_source.go
+++ b/internal/collector/source/pod/pod_scraping_source.go
@@ -134,18 +134,29 @@ func (p *PodScrapingSource) Refresh(ctx context.Context, spec source.RefreshSpec
 	// Aggregate results
 	aggregated := p.aggregateResults(results)
 
-	// Cache the result
+	// Cache the aggregated result under the canonical "all_metrics" key.
 	cacheKey := source.BuildCacheKey("all_metrics", nil)
 	p.cache.Set(cacheKey, *aggregated, p.config.DefaultTTL)
+
+	// Also cache per-metric slices keyed by __name__ so that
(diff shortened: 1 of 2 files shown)


internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -397,6 +408,29 @@ func (p *PodScrapingSource) parsePrometheusMetrics(reader io.Reader, podName str
 	}, nil
 }
 
+// splitByMetricName groups aggregated values by their "__name__" label,
+// returning one MetricResult per distinct metric name.
+func (p *PodScrapingSource) splitByMetricName(aggregated *source.MetricResult) map[string]*source.MetricResult {
```
mamy-CS: p is never actually used in this method. You can make it a plain package level function instead. Methods on a struct should need the receiver. If they don't, they're just functions in disguise. Something like `func splitByMetricName(aggregated *source.MetricResult) map[string]*source.MetricResult` would be cleaner and makes it clearer the logic doesn't depend on any source state.

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -134,18 +134,29 @@ func (p *PodScrapingSource) Refresh(ctx context.Context, spec source.RefreshSpec
...
-		"metricCount", len(aggregated.Values))
+		"metricCount", len(aggregated.Values),
+		"distinctMetrics", len(perMetric))
 
-	return map[string]*source.MetricResult{
-		"all_metrics": aggregated,
-	}, nil
+	out := make(map[string]*source.MetricResult, len(perMetric)+1)
+	out["all_metrics"] = aggregated
+	for name, mr := range perMetric {
```
mamy-CS: You're ranging over perMetric twice, once to fill the cache and once to fill the return map. You could collapse these into a single pass and build out first, then loop over out to write to the cache (or vice versa). It's a minor thing but two nearly identical loops back to back can make a reader wonder if they're subtly different. The other one is on line 144 above.

internal/collector/source/pod/pod_scraping_source.go
```diff
@@ -397,6 +408,29 @@ func (p *PodScrapingSource) parsePrometheusMetrics(reader io.Reader, podName str
 	}, nil
 }
 
+// splitByMetricName groups aggregated values by their "__name__" label,
+// returning one MetricResult per distinct metric name.
+func (p *PodScrapingSource) splitByMetricName(aggregated *source.MetricResult) map[string]*source.MetricResult {
+	grouped := make(map[string][]source.MetricValue)
+	for _, v := range aggregated.Values {
+		name := v.Labels["__name__"]
+		if name == "" {
```
mamy-CS: Worth adding a quick guard for the `"all_metrics"` name here too, e.g. `if name == "" || name == "all_metrics" { continue }`. It's a theoretical edge case, but if a scraped metric ever had `__name__="all_metrics"` it would silently overwrite the canonical aggregated entry in both the return map and the cache. Probably never happens with vllm's naming conventions, but a one-liner protects against it and makes the invariant explicit.

mamy-CS (changes requested):
Thank you for the fix. Some nits suggested, thank you for adding the test coverage too. The OpenShift E2E failure is a GPU-availability infrastructure flake, we can rerun again when cluster is up again.

Goutham-Annem:
Hi @mamy-CS — thanks for the review! All three nits addressed in the latest push:

1. **Receiver removed** — `splitByMetricName` is now a plain package-level function
2. **Single loop** — the two `for name, mr := range perMetric` passes are collapsed into one that fills the return map and writes to cache simultaneously  
3. **`all_metrics` guard** — added `|| name == "all_metrics"` to the skip condition in `splitByMetricName`, plus a test case covering it

The OpenShift E2E GPU flake is noted — happy to `/retest` once the cluster has capacity.

mamy-CS (approved):
/lgtm

PR #1380, opened 2026-07-06
fix(variants): scope HPA/ScaledObject list calls to tracked namespaces

### LINKED ISSUE #1134: perf: scope annotationSourcedVariants List calls to tracked namespaces

Goutham-Annem:
## Summary

Fixes #1134.

`annotationSourcedVariants()` issued two unscoped cluster-wide `List` calls (HPAs and ScaledObjects) on every engine tick, reading the entire cluster cache even when only a small number of namespaces were actually tracked.

**Root cause**: `k8sClient.List(ctx, &hpaList)` and `k8sClient.List(ctx, &soList)` had no `client.ListOption` to restrict the query, as noted by the `TODO(#1134)` comments in the code.

**Fix**: Add a `trackedNamespaces []string` parameter through the call chain. When non-empty, perform one `client.InNamespace(ns)` scoped call per tracked namespace and union the results. When empty/nil, fall back to the original cluster-wide behaviour.

The parameter is threaded through:
- `annotationSourcedVariants(ctx, k8sClient, trackedNamespaces)`
- `readyVariantAutoscalings(ctx, k8sClient, trackedNamespaces)`
- `filterVariantsByScaleTargetAccessor(ctx, client, filter, filterName, trackedNamespaces)`
- `ActiveVariantAutoscaling(ctx, client, trackedNamespaces)`
- `InactiveVariantAutoscaling(ctx, client, trackedNamespaces)`
- `ActiveVariantAutoscalingByModel(ctx, client, trackedNamespaces)`
- `InactiveVariantAutoscalingByModel(ctx, client, trackedNamespaces)`

The scalefromzero engine already holds `Datastore` and now passes `e.Datastore.ListTrackedNamespaces()`. The saturation engine gains an exported `Datastore datastore.Datastore` field (zero-value safe — `nil` means cluster-wide) that is wired from `cmd/main.go` via `engine.Datastore = ds`.

## Changes

- `internal/utils/variant.go`: add `namespacedListCalls` helper; thread `trackedNamespaces` through the full call chain; remove TODO(#1134) comments
- `internal/engines/saturation/engine.go`: add `Datastore datastore.Datastore` field; pass `ListTrackedNamespaces()` to `ActiveVariantAutoscaling`
- `internal/engines/scalefromzero/engine.go`: pass `ListTrackedNamespaces()` to `InactiveVariantAutoscaling`
- `cmd/main.go`: wire `ds` into saturation engine via `engine.Datastore = ds`
- `internal/utils/variant_test.go`: add `TestAnnotationSourcedVariantsNamespaceScoping` (3 cases: single namespace, multi-namespace union, nil fallback); update all call sites to pass `nil`
- `internal/engines/scalefromzero/engine_test.go`: update call site to pass `nil`

## Test plan

- [x] New `TestAnnotationSourcedVariantsNamespaceScoping` tests pass
- [x] All existing unit tests pass (`go test ./internal/utils/... ./internal/engines/...`)
- [x] Full unit test suite passes (`go test ./...`, excluding pre-existing e2e/chart failures requiring a live cluster)
- [x] `go build ./...` clean

### DIFF
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -41,6 +41,7 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/collector/source"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/config"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/constants"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/datastore"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/domain"
 	queueingmodel "github.com/llm-d/llm-d-workload-variant-autoscaler/internal/engines/analyzers/queueingmodel"
 	saturation_v2 "github.com/llm-d/llm-d-workload-variant-autoscaler/internal/engines/analyzers/saturation_v2"
@@ -145,6 +146,10 @@ type Engine struct {
 	vaEventTracker map[string]bool
 
 	Config *config.Config // Unified configuration (injected from main.go)
+
+	// Datastore is used to scope HPA/ScaledObject list calls to tracked namespaces.
+	// When nil, list calls fall back to cluster-wide.
+	Datastore datastore.Datastore
 
 	// ReplicaMetricsCollector is the collector for replica metrics using the source infrastructure
 	ReplicaMetricsCollector *collector.ReplicaMetricsCollector
@@ -450,7 +455,11 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 		logger.Info("Scaling to zero is enabled")
 	}
 
-	activeVAs, _, err := utils.ActiveVariantAutoscaling(ctx, e.client)
+	var trackedNamespaces []string
+	if e.Datastore != nil {
+		trackedNamespaces = e.Datastore.ListTrackedNamespaces()
+	}
+	activeVAs, _, err := utils.ActiveVariantAutoscaling(ctx, e.client, trackedNamespaces)
 	if err != nil {
 		logger.Error(err, "Unable to get active variant autoscalings")
 		return err
--- a/internal/engines/scalefromzero/engine.go
+++ b/internal/engines/scalefromzero/engine.go
@@ -129,7 +129,7 @@ func (e *Engine) optimize(ctx context.Context) error {
 	logger := log.FromContext(ctx)
 
 	// Get all inactive (replicas == 0) VAs
-	inactiveVAs, scaleTargets, err := utils.InactiveVariantAutoscaling(ctx, e.client)
+	inactiveVAs, scaleTargets, err := utils.InactiveVariantAutoscaling(ctx, e.client, e.Datastore.ListTrackedNamespaces())
 	if err != nil {
 		return err
 	}
--- a/internal/utils/variant.go
+++ b/internal/utils/variant.go
@@ -41,8 +41,9 @@ type VariantFilter func(scaletarget.ScaleTargetAccessor) bool
 // ActiveVariantAutoscalingByModel retrieves all VariantAutoscaling resources that are ready for optimization
 // and have at least one target replica.
 // Returns the shallow-copied VAs (not safe for mutation) grouped by ModelID.
-func ActiveVariantAutoscalingByModel(ctx context.Context, client client.Client) (map[string][]wvav1alpha1.VariantAutoscaling, error) {
-	vas, _, err := ActiveVariantAutoscaling(ctx, client)
+// trackedNamespaces scopes HPA/ScaledObject discovery to specific namespaces; nil means cluster-wide.
+func ActiveVariantAutoscalingByModel(ctx context.Context, client client.Client, trackedNamespaces []string) (map[string][]wvav1alpha1.VariantAutoscaling, error) {
+	vas, _, err := ActiveVariantAutoscaling(ctx, client, trackedNamespaces)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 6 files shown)


lionelvillard:
@Goutham-Annem can you please fix the linter error?

internal/utils/variant.go
```diff
@@ -194,53 +199,64 @@ func readyVariantAutoscalings(ctx context.Context, k8sClient client.Client) []wv
...
-		return nil, fmt.Errorf("listing HPAs: %w", err)
-	}
-	for i := range hpaList.Items {
-		hpa := &hpaList.Items[i]
-		if !annotations.IsManaged(hpa) || !hpa.DeletionTimestamp.IsZero() {
-			continue
+	for _, opts := range nsCalls {
+		var hpaList autoscalingv2.HorizontalPodAutoscalerList
+		if err := k8sClient.List(ctx, &hpaList, opts...); err != nil {
+			return nil, fmt.Errorf("listing HPAs: %w", err)
```
ev-shindin, line 230: Inside the per-namespace HPA loop, a `List` error on *any* namespace does `return nil, err`, discarding every variant already collected from earlier namespaces this call. The ScaledObject loop below (`L253–267`) instead keeps partial results on error. With one cluster-wide call this asymmetry didn't exist (no prior data to lose); the per-namespace loop introduces a new partial-failure mode where one namespace's transient error wipes the whole tick's HPA discovery (it's non-fatal at the caller, so the tick just sees zero HPAs). Suggest matching the ScaledObject pattern — record the error, continue, and return partial results.

internal/engines/saturation/engine.go
```diff
@@ -359,7 +364,11 @@ func (e *Engine) optimize(ctx context.Context) (retErr error) {
 		logger.Info("Scaling to zero is enabled")
 	}
 
-	activeVAs, _, err := utils.ActiveVariantAutoscaling(ctx, e.client)
+	var trackedNamespaces []string
+	if e.Datastore != nil {
```
ev-shindin, line 459: The new `Datastore` scoping wiring (`engine.Datastore = ds` in `cmd/main.go`, the `e.Datastore != nil` branch here) has no saturation-package test - nothing sets `Engine.Datastore` or checks the nil-fallback. A dropped assignment or inverted nil-check would pass the whole suite. Add a test seeding a `datastore.Datastore` with tracked namespaces + managed scalers in tracked and untracked namespaces, asserting only tracked ones are optimized (plus a `Datastore == nil` -> cluster-wide case).

internal/engines/scalefromzero/engine.go
```diff
@@ -127,7 +127,7 @@ func (e *Engine) optimize(ctx context.Context) error {
 	logger := log.FromContext(ctx)
 
 	// Get all inactive (replicas == 0) VAs
-	inactiveVAs, scaleTargets, err := utils.InactiveVariantAutoscaling(ctx, e.client)
+	inactiveVAs, scaleTargets, err := utils.InactiveVariantAutoscaling(ctx, e.client, e.Datastore.ListTrackedNamespaces())
```
ev-shindin, line 132: Similar to `saturation/engine.go` : the only touched test just updates the call signature with `nil`; no engine-level test asserts scoping to tracked namespaces.

internal/utils/variant.go
ev-shindin, line 230: Two untested new paths: `namespacedListCalls` isn't unit-tested directly for the empty/single/multi cases (only covered indirectly), and the multi-namespace list-error path (succeed on ns1, fail on ns2) -which would surface the HPA data-loss above - has no test. Add an `interceptor.Funcs` case that fails one namespace's List.

ev-shindin (commented):
LGTM. Thanks @Goutham-Annem ! Please address my comments

asm582:
FYI, this PR looks like a duplicate of https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1189

Issue #1364, opened 2026-06-30
Provide KEDA configuration for Replica Rebalancing guide


lionelvillard:

xuhui-lu:
Hi @lionelvillard, I took a quick look at the current KEDA path for replica rebalancing.

My understanding is that WVA already discovers ScaledObjects, but gpu-rebalance currently only handles HPAs. Since KEDA-owned HPAs are intentionally not managed directly, would it make sense to add ScaledObject support in gpu-rebalance first, likely by patching spec.maxReplicaCount, and then update the docs with the KEDA setup?

If that sounds right, I’m happy to work on it.

asm582:
@xuhui-lu can you please update status?

xuhui-lu:
The runtime implementation is in #1384. It now supports rebalancing KEDA ScaledObjects by updating spec.maxReplicaCount, including mixed HPA/ScaledObject allocation, configured minimum replicas, infeasible-minimum handling, and the requested quota-invariant coverage.

The only failed check is e2e-tests-full, which did not reach the test phase: `Connection reset by peer`. I do not have permission to rerun Actions in the upstream repository, so this needs a maintainer to rerun the failed job. Once CI is green, #1384 should be ready for final approval and merge.

I have also prepared the KEDA configuration and Replica Rebalancing guide changes locally as a separate follow-up, and will rebase/open that draft after #1384 merges.

lionelvillard:
looks like this is done. Closing.

Issue #1347, opened 2026-06-26
feat(guides): Provide KEDA configuration for all existing autoscaling guides


lionelvillard:
**What would you like to be added**:

KEDA configurations for:

**Why is this needed**:

prometheus-adapter is deprecated.

lionelvillard:
all done.

PR #1384, opened 2026-07-07
feat(coordinator): rebalance KEDA ScaledObjects

xuhui-lu:
This is a small implementation pass for the KEDA side of replica rebalancing.

The coordinator already discovers managed `ScaledObject`s, and it also avoids patching KEDA-owned generated HPAs directly. The missing piece was that `gpu-rebalance` only used the HPA path, so a managed ScaledObject selected by the coordinator was effectively ignored by the plugin.

This PR teaches `gpu-rebalance` to treat HPAs and ScaledObjects as the same kind of quota entry:

- direct HPAs still get `spec.maxReplicas` patched
- KEDA ScaledObjects get `spec.maxReplicaCount` patched instead
- HPAs and ScaledObjects in the same namespace share the same ResourceQuota calculation
- RBAC now allows patch/update on `keda.sh/scaledobjects`

I kept this as a draft because I want to confirm this is the expected implementation direction for #1364 before adding the guide updates.

Tested with:

```text
go test ./internal/coordinator/...
git diff --check
make manifests
```

### DIFF
--- a/config/base/rbac/manager-clusterrole.yaml
+++ b/config/base/rbac/manager-clusterrole.yaml
@@ -84,6 +84,8 @@ rules:
   verbs:
   - get
   - list
+  - patch
+  - update
   - watch
 - apiGroups:
   - leaderworkerset.x-k8s.io
--- a/internal/coordinator/plugins/gpurebalance/plugin.go
+++ b/internal/coordinator/plugins/gpurebalance/plugin.go
@@ -7,10 +7,12 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
 	promv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/common/model"
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -19,6 +21,8 @@ const (
 	AnnotationInferencePool = "llm-d.ai/epp-inference-pool"
 	gpuQuotaResource        = "requests.nvidia.com/gpu"
 	eppQueueMetric          = `sum(inference_extension_flow_control_queue_size{inference_pool=%q})`
+	displayKindHPA          = "HorizontalPodAutoscaler"
+	displayKindScaledObject = "ScaledObject"
 )
 
 // Plugin implements the Coordinator Plugin interface for GPU rebalancing.
@@ -29,6 +33,7 @@ type Plugin struct {
 
 // +kubebuilder:rbac:groups="",resources=resourcequotas,verbs=get;list;watch
 // +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs=get;list;watch;patch;update
+// +kubebuilder:rbac:groups=keda.sh,resources=scaledobjects,verbs=get;list;watch;patch;update
 
 // New constructs the gpu-rebalance Plugin.
 func New(c client.Client, promAPI promv1.API) *Plugin {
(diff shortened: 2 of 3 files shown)


nourey:
Hi @xuhui-lu, I took a closer look at this PR because it overlaps with the KEDA migration work I’ve been following across [#1347](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1347) and [#1364](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1364). I recently worked on the related HPA+EPP guide migration in [llm-d/llm-d#1981](https://github.com/llm-d/llm-d/pull/1981) and I have also been looking at how these guide changes can eventually be backed by guide specific behavior tests, as discussed in [llm-d/llm-d#1517](https://github.com/llm-d/llm-d/issues/1517). Since this PR is part of the same transition, I wanted to understand how the runtime path behaves, not just whether the KEDA configuration is valid.

The overall direction here looks right to me: for a KEDA managed workload, `gpu-rebalance` should update the parent `ScaledObject.spec.maxReplicaCount` rather than patching the KEDA generated HPA directly. While following that path, I noticed one min max edge case that may be worth handling as part of this change. The shared [`scalerEntry`](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/88e3be6a4e8d5093d610e75483844c4c236f6322/internal/coordinator/plugins/gpurebalance/plugin.go#L46-L52) currently carries `currentMax` but it does not appear to carry the scaler’s effective minimum. The target calculation still uses a lower bound of **1** and the ScaledObject path then writes that target directly to `spec.maxReplicaCount`.

For example considering:

```yaml
minReplicaCount: 5
maxReplicaCount: 10
````

Under the plugin’s current quota arithmetic, with namespace quota `2`, two ScaledObjects and equal queue depths, each target is calculated as:

```text
floor(2 * 0.5) = 1
```

The resulting ScaledObject can therefore become:

```yaml
minReplicaCount: 5
maxReplicaCount: 1
```

In KEDA **v2.18.0**, this update may reach the API but the KEDA controller checks the replica bounds during reconciliation and returns an error before reconciling the generated HPA. Would it make sense for each entry to carry an `effectiveMin`, derived from `HPA.spec.minReplicas` or `ScaledObject.spec.minReplicaCount` and guarantee that the plugin never patches a maximum below that value?

One possible direction would be:

```text
1. determine the effective minimum for every scaler
2. check whether the aggregate minima are feasible under the available quota
3. reserve the feasible minima
4. distribute the remaining capacity using the existing queue weights
```

a plain clamp such as:

```text
target = max(target, effectiveMin)
```

would prevent invalid min/max objects but it could still silently over allocate when the configured minima alone exceed the available quota. For example two scalers with minimum `5` and quota `2` would produce a total minimum allocation of `10`. In that infeasible case, leaving the namespace unchanged and logging the condition may be safer than applying a partial allocation, although I may be missing an intended policy here.

The same underlying minimum replica issue appears to predate this PR on the direct HPA path, so I do not mean to attribute that existing behavior to the ScaledObject work. I only noticed it while tracing the new shared path and it looks like the shared abstraction may provide a good place to make the behavior consistent for both resource types.

Focused unit coverage could verify at least:

* a ScaledObject with `minReplicaCount > 1`
* a direct HPA with `minReplicas > 1`
* mixed HPA and ScaledObject entries with different minima
* aggregate minima exceeding the available quota
* the intended behavior for nil minima and `minReplicaCount: 0`

Separately before the user facing Replica Rebalancing guide claims KEDA support, I think it would be useful for the guide level validation to cover the complete path:

```text
gpu-rebalance -> ScaledObject.spec.maxReplicaCount -> KEDA-generated HPA.spec.maxReplicas -> target workload replica behavior
```

That integration coverage does not necessarily need to block this implementation PR but it would confirm that KEDA propagates the new ceiling, remains healthy during reconciliation and produces the expected workload behavior.

Thanks for taking the time to read this. I appreciate the work and thought that has gone into the PR and I hope this observation is useful. Please let me know if I’ve missed an intended assumption in the current allocation policy.

xuhui-lu:
> Hi @xuhui-lu, I took a closer look at this PR because it overlaps with the KEDA migration work I’ve been following across [#1347](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1347) and [#1364](https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1364). I recently worked on the related HPA+EPP guide migration in [llm-d/llm-d#1981](https://github.com/llm-d/llm-d/pull/1981) and I have also been looking at how these guide changes can eventually be backed by guide specific behavior tests, as discussed in [llm-d/llm-d#1517](https://github.com/llm-d/llm-d/issues/1517). Since this PR is part of the same transition, I wanted to understand how the runtime path behaves, not just whether the KEDA configuration is valid.
> 
> The overall direction here looks right to me: for a KEDA managed workload, `gpu-rebalance` should update the parent `ScaledObject.spec.maxReplicaCount` rather than patching the KEDA generated HPA directly. While following that path, I noticed one min max edge case that may be worth handling as part of this change. The shared [`scalerEntry`](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/88e3be6a4e8d5093d610e75483844c4c236f6322/internal/coordinator/plugins/gpurebalance/plugin.go#L46-L52) currently carries `currentMax` but it does not appear to carry the scaler’s effective minimum. The target calculation still uses a lower bound of **1** and the ScaledObject path then writes that target directly to `spec.maxReplicaCount`.
> 
> For example considering:
> 
> ```yaml
> minReplicaCount: 5
> maxReplicaCount: 10
> ```
> 
> Under the plugin’s current quota arithmetic, with namespace quota `2`, two ScaledObjects and equal queue depths, each target is calculated as:
> 
> ```
> floor(2 * 0.5) = 1
> ```
> 
> The resulting ScaledObject can therefore become:
> 
> ```yaml
> minReplicaCount: 5
> maxReplicaCount: 1
> ```
> 
> In KEDA **v2.18.0**, this update may reach the API but the KEDA controller checks the replica bounds during reconciliation and returns an error before reconciling the generated HPA. Would it make sense for each entry to carry an `effectiveMin`, derived from `HPA.spec.minReplicas` or `ScaledObject.spec.minReplicaCount` and guarantee that the plugin never patches a maximum below that value?
> 
> One possible direction would be:
> 
> ```
> 1. determine the effective minimum for every scaler
> 2. check whether the aggregate minima are feasible under the available quota
> 3. reserve the feasible minima
> 4. distribute the remaining capacity using the existing queue weights
> ```
> 
> a plain clamp such as:
> 
> ```
> target = max(target, effectiveMin)
> ```
> 
> would prevent invalid min/max objects but it could still silently over allocate when the configured minima alone exceed the available quota. For example two scalers with minimum `5` and quota `2` would produce a total minimum allocation of `10`. In that infeasible case, leaving the namespace unchanged and logging the condition may be safer than applying a partial allocation, although I may be missing an intended policy here.
> 
> The same underlying minimum replica issue appears to predate this PR on the direct HPA path, so I do not mean to attribute that existing behavior to the ScaledObject work. I only noticed it while tracing the new shared path and it looks like the shared abstraction may provide a good place to make the behavior consistent for both resource types.
> 
> Focused unit coverage could verify at least:
> 
> * a ScaledObject with `minReplicaCount > 1`
> * a direct HPA with `minReplicas > 1`
> * mixed HPA and ScaledObject entries with different minima
> * aggregate minima exceeding the available quota
> * the intended behavior for nil minima and `minReplicaCount: 0`
> 
> Separately before the user facing Replica Rebalancing guide claims KEDA support, I think it would be useful for the guide level validation to cover the complete path:
> 
> ```
> gpu-rebalance -> ScaledObject.spec.maxReplicaCount -> KEDA-generated HPA.spec.maxReplicas -> target workload replica behavior
> ```
> 
> That integration coverage does not necessarily need to block this implementation PR but it would confirm that KEDA propagates the new ceiling, remains healthy during reconciliation and produces the expected workload behavior.
> 
> Thanks for taking the time to read this. I appreciate the work and thought that has gone into the PR and I hope this observation is useful. Please let me know if I’ve missed an intended assumption in the current allocation policy.

It makes sense. I will take it into consideration and update this PR. Thanks!

asm582:
Thanks for this PR; it handles scenarios where minReplicas is greater than 1.

A critical test to be added is TestRebalance_AllocationNeverExceedsQuota. This test verifies the fundamental invariant: sum(all target maxReplicas) == quota

It matters because :

- If allocation ever exceeds quota, only ResourceQuota enforcement prevents pod scheduling issues
- Pods get silently stuck in the Pending state

nourey:
I added the requested TestRebalance_AllocationNeverExceedsQuota coverage in [xuhui-lu/llm-d-workload-variant-autoscaler#1](https://github.com/xuhui-lu/llm-d-workload-variant-autoscaler/pull/1).

The test covers the minimum-replica allocation edge case I raised while reviewing this PR, using a mixed HPA and ScaledObject scenario. It verifies the exact minimum aware targets as well as the aggregate invariant that all resulting max replica ceilings sum to the namespace quota.

I also validated that it fails against both a historical allocation that ignores configured minima and a naive full quota allocation followed by minimum clamping.

asm582:
@nourey @xuhui-lu, can you please sign commits for this PR. Once CI passes, let's merge it.

xuhui-lu:
> @nourey @xuhui-lu, can you please sign commits for this PR. Once CI passes, let's merge it.

done.

PR #1389, opened 2026-07-07
docs: drop CRD-reference generator and dead crd-docs references (VA removal 3/5)

lionelvillard:
Part of Phase 3 of the VariantAutoscaling CRD removal (#1130, step 3.7).

The generated `docs/user-guide/crd-reference.md` no longer exists and the VariantAutoscaling CRD is being removed, so its documentation generator has no source to render.

- Remove the `crd-docs` Makefile target and `hack/crd-doc-gen/config.yaml`.
- Drop the dead `make crd-docs` steps from `CONTRIBUTING.md` and the developer guide.
- Fix the broken CRD Reference link in `docs/design/controller-behavior.md` and point the configuration link at its real location.
- Mark `docs/proposals/deprecate-va-crd.md` as Implemented.

The annotation-based discovery rewrite of `controller-behavior.md` is deferred to a follow-up where the final controller shape can be documented accurately.

### PR stack (VA CRD removal — merge in order)
1. PR1 (#1387) — remove VA RBAC helper roles
2. PR2 (#1388) — remove VA samples
3. PR3 — remove CRD-reference doc generator
4. PR4 — remove the CRD API (types → `internal/variant`, delete reconciler/manifests/scheme)
5. PR5 — update test suites for annotation-based discovery (`make test` green)

> Stacked via the fork; all PRs target `main`, so PRs 2–5 show cumulative diffs until the earlier ones merge. Review/merge in order.

PR #1388, opened 2026-07-07
chore(samples): remove VariantAutoscaling CRD samples (VA removal 2/5)

lionelvillard:
Part of Phase 3 of the VariantAutoscaling CRD removal (#1130, step 3.5).

Deletes the VariantAutoscaling sample manifests now that annotation-based discovery is the only supported path:
- `config/samples/variantautoscaling-{integration,with-cost,with-lws}.yaml`
- `config/samples/legacy/` (archived reference VA samples)
- the legacy VA-dependent `config/samples/keda/scaledobject.yaml` and `config/samples/hpa/hpa.yaml`

Repoints `config/samples/keda` and `config/samples/hpa` at their `annotations/` overlays so the sample entry points keep working without a CRD. The deprecated Helm chart is left untouched.

### PR stack (VA CRD removal — merge in order)
1. PR1 (#1387) — remove VA RBAC helper roles
2. PR2 — remove VA samples
3. PR3 — remove CRD-reference doc generator
4. PR4 — remove the CRD API (types → `internal/variant`, delete reconciler/manifests/scheme)
5. PR5 — update test suites for annotation-based discovery (`make test` green)

> Stacked via the fork; all PRs target `main`, so PRs 2–5 show cumulative diffs until the earlier ones merge. Review/merge in order.

PR #1387, opened 2026-07-07
chore(rbac): remove VariantAutoscaling admin/editor/viewer roles (VA removal 1/5)

lionelvillard:
Part of Phase 3 of the VariantAutoscaling CRD removal (#1130, step 3.4).

Deletes the standalone CRD helper ClusterRoles (admin/editor/viewer) under `config/base/rbac`. These roles only existed to help cluster admins delegate permissions over the VariantAutoscaling CRD, which is being removed. The controller itself never used them.

The deprecated Helm chart (see AGENTS.md) is left untouched.

### PR stack (VA CRD removal — merge in order)
1. **PR1 (this)** — remove VA RBAC helper roles
2. PR2 — remove VA samples
3. PR3 — remove CRD-reference doc generator
4. PR4 — remove the CRD API (types → `internal/variant`, delete reconciler/manifests/scheme)
5. PR5 — update test suites for annotation-based discovery (`make test` green)

> Stacked via the fork; all PRs target `main`, so PRs 2–5 show cumulative diffs until the earlier ones merge. Review/merge in order.

PR #1390, opened 2026-07-07
refactor: remove the VariantAutoscaling CRD API (VA removal 4/4)

lionelvillard:
Final PR of Phase 3 of the VariantAutoscaling CRD removal (#1130, steps 3.1–3.3 and 3.6).

The VariantAutoscaling CRD is no longer a user-facing API. Variants are discovered by synthesizing in-memory objects from annotated HPAs and KEDA ScaledObjects (Phase 1 annotation-based discovery), so the CRD, its reconciler, and its scheme/index wiring are removed.

**Production**
- Move the VariantAutoscaling types to `internal/variant` as plain in-memory structs (hand-written DeepCopy; still implements `runtime.Object` for the shared EventRecorder, but never scheme-registered or persisted). Drop kubebuilder markers and `groupversion_info.go`.
- Delete `api/v1alpha1`, the VA reconciler, the VA field indexer, and the now-dead predicates.
- Repoint every caller from `api/v1alpha1` to `internal/variant`.
- `readyVariantAutoscalings` sources solely from annotation discovery; the saturation and scale-from-zero writeback paths drop the CRD status patch / DecisionCache / DecisionTrigger. `wva_desired_replicas` emission is the sole output, which KEDA/HPA read directly.
- `cmd/main.go`: drop VA scheme registration, CRD detection, reconciler wiring, and the index gate.
- Delete `config/base/crd` (VA was the only CRD); drop CRD generation and the VA rule blocks from the manager ClusterRole (markers relocated to `internal/controller/rbac.go`).

**Tests** (folded in so `make test` is green on this PR)
- Add `internal/variant/scheme.go` with `AddToScheme`/`GroupVersion` for fake-client unit tests only — the production manager scheme does not register the type.
- Point the envtest suites away from the deleted `config/base/crd`; rewrite the saturation / scale-from-zero engine tests to seed annotated HPAs and drop the obsolete CRD dual-mode test cases.
 

### PR stack (VA CRD removal)
1. #1387 — remove VA RBAC helper roles (merged)
2. #1388 — remove VA samples (merged)
3. #1389 — remove CRD-reference doc generator (merged)
4. PR4 (this) — remove the CRD API + update tests

PR5 (#1391) has been folded into this PR and closed — the test changes are meaningless without the API removal, and CI runs `make test` per-PR.


mamy-CS (approved):
/lgtm

PR #1391, opened 2026-07-07
test: update suites for annotation-based variant discovery (VA removal 5/5)

lionelvillard:
Part of Phase 3 of the VariantAutoscaling CRD removal (#1130, step 3.6) — follow-up to the CRD API removal so `make test` is green again.

- Add `internal/variant/scheme.go` providing `AddToScheme`/`GroupVersion` so fake-client unit tests can register the in-memory VariantAutoscaling type. **The production manager scheme does not register it.**
- Point the controller/actuator/saturation/indexers envtest suites away from the deleted `config/base/crd`; they now rely on built-in types (Deployments, HPAs) and KEDA only where needed.
- Rewrite the saturation and scale-from-zero engine tests to seed annotated HPAs (the new discovery surface) instead of creating VA objects, and assert on the loop running rather than on patched VA status (there is none).
- Drop obsolete CRD-based dual-mode cases from the collector `buildInstanceKey` tests, the indexer VA-lookup tests, and the `readyVariantAutoscalings` merge tests; keep the annotation-only cases.
- Remove VA Create/Delete calls from the actuator tests.

**Verification:** `make test` scope = 37 packages passing. Pre-existing unrelated failure `TestThroughput/FitITLModel` also fails on the base commit (darwin/arm64 slope-fit edge case) and is untouched here.

**Deferred follow-ups** (compile, not blocking `make test`): rewriting the e2e VA fixtures (`test/e2e/fixtures/va_builder.go`) to HPA/ScaledObject builders, and the annotation-path rewrite of `docs/design/controller-behavior.md`.

### PR stack (VA CRD removal — merge in order)
1. PR1 (#1387) — remove VA RBAC helper roles
2. PR2 (#1388) — remove VA samples
3. PR3 (#1389) — remove CRD-reference doc generator
4. PR4 (#1390) — remove the CRD API
5. PR5 (this) — update test suites for annotation-based discovery

> Stacked via the fork; all PRs target `main`, so this PR shows the cumulative diff of the whole stack until the earlier ones merge. Review/merge in order.


lionelvillard:
Folded into #1390 — the test-suite updates are meaningless without the CRD API removal, and CI runs `make test` per-PR, so #1390 now carries both as one self-contained change.

PR #1392, opened 2026-07-07
Bug fix: Calculate missing V1 saturation utilization metric

shuynh2017:
Fixes #
- Before the fix, in a scale test, there's scaling action as shown in `Scaling Decision Rate` panel, but strangely the `Saturation Utilization`  panel shows `0` utilization!
saturation-utilization-before

- The problem is that the metric wasn't calculated. The fix is to calculate the metric as original intended. Here's the original intention:

  ```
	  // Utilization is the variant-level utilization ratio (0.0-1.0) reported for
	  // observability. The exact formula differs by analyzer because V1 and V2
	  // reason about saturation differently:
	  //   V1: mean of per-replica KvCacheUsage fractions (matches what V1's
	  //       per-replica threshold check operates on).
	  //   V2: ...
	  Utilization float64
  ```

- After the fix:

image

## Proposed Changes

- :bug: Fix bug

### Pre-review Checklist

- [N/A] **E2E tests** for any new behavior

**Release Note**

```release-note
N/A
```

**Docs**

N/A

### DIFF
--- a/internal/engines/saturation/engine.go
+++ b/internal/engines/saturation/engine.go
@@ -1074,6 +1074,8 @@ func (e *Engine) convertSaturationTargetsToDecisions(
 			decision.Cost = va.Cost
 			// Use average spare KV capacity as the SpareCapacity indicator for limiter prioritization
 			decision.SpareCapacity = va.AvgSpareKvCapacity
+			// V1 Utilization: mean of per-replica KvCacheUsage fractions
+			decision.Utilization = va.AvgKvCacheUsage
 		} else {
 			logger.Info("No variant analysis found for decision (metrics may be unavailable)",
 				"variant", variantName)
--- a/internal/engines/saturation/engine_v1_utilization_test.go
+++ b/internal/engines/saturation/engine_v1_utilization_test.go
@@ -1,0 +1,91 @@
+package saturation
+
+import (
+	"context"
+	"testing"
+
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/interfaces"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
+)
+
+func init() {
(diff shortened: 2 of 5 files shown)


internal/engines/saturation/engine.go
```diff
@@ -1074,6 +1074,8 @@ func (e *Engine) convertSaturationTargetsToDecisions(
 			decision.Cost = va.Cost
 			// Use average spare KV capacity as the SpareCapacity indicator for limiter prioritization
 			decision.SpareCapacity = va.AvgSpareKvCapacity
+			// V1 Utilization: mean of per-replica KvCacheUsage fractions
+			decision.Utilization = va.AvgKvCacheUsage
```
shuynh2017, line 1078: decision.Utilization must be set in order for emitted metric to have a value.

shuynh2017:
@deanlorenz pls help to review. Thanks.

deanlorenz:
Nice fix — matches the formula already documented on `VariantDecision.Utilization`. Two non-blocking notes:

**1. (non-blocking) Doc example now shows the exact bug this fixes**

`docs/developer-guide/prometheus.md`'s example for `wva_saturation_utilization` shows `"value": [..., "0"]` — looks captured while hitting this bug. Worth swapping in a non-zero example now that it's fixed.

**2. (non-blocking) New tests use exact float equality**

`TestAnalyzeVariant_AvgKvCacheUsage` and `engine_v1_utilization_test.go` compare a computed average with `!=` against a literal. Passes here by luck of float rounding — an epsilon check (`math.Abs(got-want) > 1e-9`) would be safer for future additions to this test.

deanlorenz (approved):
LGTM — fix matches the formula already documented on `VariantDecision.Utilization`. Left two non-blocking notes above (doc example, test float-equality).

PR #1393, opened 2026-07-08
Bug fix: correctly record the metric showing number of pods with metrics 

shuynh2017:
Fixes #

## The Problem:
- In my cluster, there's only `1` model pod running, but the dashboard shows `7` pods with metrics!

## Proposed Changes

- :bug: Fix bug

The Fix:
- After some tracing, it turns out that the query for metrics return 7 pods, and 6 of them no longer run in the namespace. They are the pods that were scaled up then scaled down yesterday.
- There's code to filter out pods w/o metrics, and these 6 pods are filtered out in the metric collector. The fix is to move the code recoding the metric to after the filter. The dashboard now shows correctly:

image

- Addition, there's also a change to move where we determine metric freshness (trackMetricFreshness) in the same function but just a bit later - just before building the replicaMetric. This makes sure we only track metric freshness for the metrics that will be returned from the function.

- ### Pre-review Checklist

- [ N/A] **E2E tests** for any new behavior
- [ N/A] **Docs PR** for any user-facing impact
- [ N/A] **Proposal PR** for any new enhancement or change to existing behavior

**Release Note**

```release-note
N/A
```

**Docs**

N/A

### DIFF
--- a/internal/collector/replica_metrics.go
+++ b/internal/collector/replica_metrics.go
@@ -879,9 +879,6 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
 		// This replaces the previous ownership traversal approach
 		vaName := data.vaName
 
-		// Track freshness for metrics in this pod for this variant right away
-		trackMetricFreshness(vaName, data, collectedAt, vaMetricsFreshnessStatus)
-
 		// Skip pods that have no metrics at all. This can happen when the query returns pods that
 		// were scaled up then scaled down, i.e. no longer running in the namespace.
 		if !data.hasKv && !data.hasQueue {
@@ -920,6 +917,7 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
 				"scale targets", getScaleTargetNames(scaleTargets))
 			continue
 		}
+
 		variantKey := utils.GetNamespacedKey(namespace, vaName)
 		// Get accelerator name from Deployment/LWS nodeSelector/nodeAffinity or VA label
 		acceleratorName := ""
@@ -976,6 +974,8 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
 			logger.Info("Pod has engine metrics but no dispatch rate — possible pod/pod_name label mismatch", "pod", podName, "model", modelID, "namespace", namespace)
 		}
 
+		// Track freshness for metrics in this pod
+		trackMetricFreshness(vaName, data, collectedAt, vaMetricsFreshnessStatus)
 		metric := interfaces.ReplicaMetrics{
 			PodName:               podName,
 			ModelID:               modelID,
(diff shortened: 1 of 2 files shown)


internal/collector/replica_metrics.go
```diff
@@ -1015,6 +1015,9 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
 		}
 	}
 
+	// Only set this after all pods have been processed, making sure not to include pods without metrics (which are skipped above).
+	// This ensures that the discovered pod count reflects only those pods for which metrics were actually collected.
```
mamy-CS: Tiny wording nit: `len(replicaMetrics)` also excludes pods that do have metrics but get skipped by the `vaName == ""` check above. That's probably the right semantics, but "pods for which metrics were actually collected" slightly overstates it, maybe "pods that produced replica metrics"?

internal/collector/replica_metrics.go
```diff
@@ -1015,6 +1015,9 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
...
+	metrics.SetMetricsPodsDiscovered(namespace, len(replicaMetrics))
```
mamy-CS, line 1020: Could you add a unit test pinning this? The existing check in replica_metrics_test.go only asserts the metric exists, so it will pass both before and after this fix. Maybe a case with one pod that has no KV/queue metrics, asserting the gauge value, would catch a regression

mamy-CS (commented):
some nits, and a general comment pre-existing, not caused by this pr but the gauge is keyed by namespace just noting it in case it's worth an issue. And looks like there are some places the old semantics are mentioned doc wise maybe do a sweep through. Thanks.

internal/collector/replica_metrics.go
shuynh2017: updated to be more precise. Thanks.

internal/collector/replica_metrics.go
shuynh2017: added test to catch regression. Thanks.

shuynh2017:
> some nits, and a general comment pre-existing, not caused by this pr but the gauge is keyed by namespace just noting it in case it's worth an issue. And looks like there are some places the old semantics are mentioned doc wise maybe do a sweep through. Thanks.

Yes, it's intended to be keyed by namespace, and as documented in prometheus.md.

internal/collector/replica_metrics.go
```diff
@@ -976,6 +974,8 @@ func (c *ReplicaMetricsCollector) collectReplicaMetrics(
 			logger.Info("Pod has engine metrics but no dispatch rate — possible pod/pod_name label mismatch", "pod", podName, "model", modelID, "namespace", namespace)
 		}
 
+		// Track freshness for metrics in this pod
```
mamy-CS, line 977: Was moving this here intentional? if that's the intent, could you mention it in the PR description? It changes a second metric the description doesn't cover.

mamy-CS (commented):
One more questiion, btw wording reads right now, and the test pins it thanks!

internal/collector/replica_metrics.go
shuynh2017: yep. Added to description.

mamy-CS (approved):
/lgtm

PR #1414, opened 2026-07-13
fix: postprocess EPP queue depth metric extraction

kahilam:
## Summary

- The postprocessing script (`hack/benchmark/postprocess.py`) was unable to extract the `Avg queue depth (EPP)` metric because:
  1. The metric was renamed from `inference_extension_flow_control_queue_size` to `llm_d_epp_average_queue_size` (with `inference_pool_average_queue_size` as a deprecated alias), so the old name no longer matched.
  2. A logic bug in the per-file iteration caused only the first EPP metrics snapshot to be read — subsequent log files were skipped due to an incorrect `break` condition that checked the global values list instead of a per-file flag.
- This resulted in `Avg queue depth (EPP): ?` (unrecognised metric name) or `0.0` (only the pre-load snapshot was captured).
- **Fix**: try all known metric names in priority order and use a per-file `found` flag so every snapshot is included in the average.

### DIFF
--- a/hack/benchmark/postprocess.py
+++ b/hack/benchmark/postprocess.py
@@ -209,18 +209,26 @@ def _extract_queue_depth_avg(results_dir):
     if not os.path.isdir(raw_dir):
         return None
 
+    metric_names = [
+        "llm_d_epp_average_queue_size",
+    ]
+
     values = []
     for fname in sorted(os.listdir(raw_dir)):
         if not fname.endswith(".log") or "router-epp" not in fname:
             continue
         fpath = os.path.join(raw_dir, fname)
+        found = False
         with open(fpath) as f:
             for line in f:
-                val = _parse_prometheus_value(
-                    line.strip(),
-                    "inference_extension_flow_control_queue_size")
-                if val is not None:
-                    values.append(val)
+                stripped = line.strip()
+                for metric_name in metric_names:
+                    val = _parse_prometheus_value(stripped, metric_name)
+                    if val is not None:
+                        values.append(val)
+                        found = True
+                        break
+                if found:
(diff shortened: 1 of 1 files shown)


asm582:
@kahilam, let's keep using the current latest metric names. Also, can you please share the command output post changes?

kahilam:
@kahilam, let's keep using the current latest metric names. Also, can you please share the command output post changes?

```
2026-07-13 16:08:25,023 - INFO    - ✅ Run complete (mode=full, harness=guidellm).
2026-07-13 16:08:26,857 - INFO    - Run parameters stored in configmap/llm-d-benchmark-run-parameters (key=run-20260713T230825Z) in ns/kahila

=========================================
  Generating benchmark report...
=========================================
Results directory: /Users/kahila/llmd/inferno-autoscaler/kahila-20260713-155504-297/results/guidellm-1783983343-zfigo2_1

Processing: /Users/kahila/llmd/inferno-autoscaler/kahila-20260713-155504-297/results/guidellm-1783983343-zfigo2_1

| Metric | Run 1 |
|--------|------|
| P99 TTFT (ms) | 448,772 |
| P99 ITL (ms/token) | 108.32 |
| Avg primary replicas | 7.32 |
| Max primary replicas | 10 |
| Avg secondary replicas | 0.00 |
| Max secondary replicas | 0 |
| Avg KV cache utilization | 33.3% |
| Avg queue depth (EPP) | 123.2 |
| Error count | 0 |
| Avg pod startup (s) | 101 |
| Cost (weighted avg replicas × GPU/hr) | 7.32 |
```

asm582 (approved):
/lgtm

PR #1415, opened 2026-07-13
docs: add EPP+KEDA saturation benchmark results

kahilam:
## Summary

- Adds direct EPP+KEDA saturation autoscaling benchmark results for all three workload scenarios:
  - **Prefill heavy**
  - **Decode heavy**
  - **Symmetrical**

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -48,6 +48,25 @@ Summary of WVA benchmark runs with configuration details.
 | Scale-down policy | 10 Pods / 150s |
 | Metric source | External (`wva_desired_replicas`) |
 
+## EPP+KEDA Saturation Configuration
+
+| Parameter | Value |
+|-----------|-------|
+| **EPP** | |
+| Scorer weights | Default (queue=2, kv-cache=2, prefix-cache=3) |
+| Feature gates | Default (no flowControl) |
+| **KEDA ScaledObject** | |
+| KV cache target | 700m |
+| Queue depth target | 2 |
+| Min replicas | 1 |
+| Max replicas | 10 |
+| Polling interval | 15s |
+| Scale-up policy | 1 Pod / 180s, stabilization 0s |
+| Scale-down policy | 1 Pod / 300s, stabilization 300s |
+| **Benchmark** | |
+| Model | Qwen/Qwen3-32B |
+| Duration | 600s |
+
 
 ## Prefill Heavy Scenario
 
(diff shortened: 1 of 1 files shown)


asm582:
Thanks, @kahilam for the PR. The setup information is duplicated; we should move it to the top and add a new section for KEDA.

kahilam:
> Thanks, @kahilam for the PR. The setup information is duplicated; we should move it to the top and add a new section for KEDA.

Done.

asm582:
Hmm, Why do we have code changes in the doc PR?

kahilam:
> Hmm, Why do we have code changes in the doc PR?

you are right, we already merged this here:
https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1414/changes
I will remove it now.

docs/benchmark.md
```diff
@@ -48,6 +48,19 @@ Summary of WVA benchmark runs with configuration details.
 | Scale-down policy | 10 Pods / 150s |
 | Metric source | External (`wva_desired_replicas`) |
 
+## EPP+KEDA Saturation Configuration
```
asm582, line 51: Please add the EPP configuration.

asm582 (changes requested):

docs/benchmark.md
```diff
@@ -48,6 +48,25 @@ Summary of WVA benchmark runs with configuration details.
...
+
+| Parameter | Value |
+|-----------|-------|
+| **EPP** | |
+| Scorer weights | Default (queue=2, kv-cache=2, prefix-cache=3) |
+| Feature gates | Default (no flowControl) |
```
asm582, line 57: Why was there no flow control enabled?

asm582 (dismissed):

docs/benchmark.md
kahilam: The EPP+KEDA scenario uses router: {} (default EPP config, no flowControl). From [epp-keda-saturation.yaml](https://github.com/llm-d/llm-d-benchmark/blob/main/config/scenarios/guides/epp-keda-saturation.yaml). I thought flowControl is only needed for WVA's request-level flow control/queuing.

asm582: WVA is an autoscaler. Let's keep these results for now. We need to run another experiment with optimized baseline plugins. Thank you.

kahilam:
@asm582 Thanks for the /lgtm/approve. Could you also dismiss your earlier "requested changes"

PR #1416, opened 2026-07-13
RBAC: reduce WVA manager Secret privs

shuynh2017:
Fixes #

This issue was flagged by a security scan in downstream to reduce privileges on secret to minimum required by WVA manager.

- Currently WVA manager has **clusterwide** `get,list,watch` on `secrets` as defined in `manager-role` `ClusterRole`. It can `get,list,watch` any secrets in any namespace ! 
- But actually, WVA manager only need to reads 1 secret in 1 namespce: `wva-epp-metrics-token` in `workload-variant-autoscaler-system`.
- To address this security concern, we will reduce it to be able to read  1 secret in 1 namespce: `wva-epp-metrics-token` in `workload-variant-autoscaler-system`. This is done by creating a new `Role` and bind the new role to WVA manager service account:
```
 kind: Role
metadata:
  name: wva-epp-metrics-role
  labels:
    app.kubernetes.io/name: workload-variant-autoscaler
rules:
- apiGroups:
  - ""
  resources:
  - secrets              <=== secrets
  resourceNames:
  - epp-metrics-token  <=== exact secret name
```
```
kind: RoleBinding
metadata:
  annotations:
   
  labels:
    app.kubernetes.io/name: workload-variant-autoscaler
  name: wva-epp-metrics-rolebinding
  namespace: workload-variant-autoscaler-system
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: wva-epp-metrics-role
subjects:
- kind: ServiceAccount
  name: wva-controller-manager
  namespace: workload-variant-autoscaler-system
```
- As a consequence, WVA manager can only get 1 secret in 1 namespace.
- Implementation Notes:
  - `list,watch` are removed since we can't have these with `resourceNames`.
  - With `list,watch` are removed, we also need to change API call from `k8sClient.Get` to `apiReader.Get`. The rest of the changes are due to this.
- Manual Testing:
  - can get `wva-epp-metrics-token` in `workload-variant-autoscaler-system`.
  - can't get an existing secret  `wva-epp-metrics-token2` in `workload-variant-autoscaler-system`.
  - can't get an existing secret `llm-d-hf-token`  in `llm-d-sim`

## Proposed Changes

- 🧹 

### Pre-review Checklist

- [Y] **E2E tests** for any new behavior

**Release Note**

```release-note
N/A
```

**Docs**

N/A

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -333,7 +333,6 @@ func main() {
 	}
 
 	watchNS := cfg.WatchNamespace()
-	secretNS := config.SystemNamespace()
 	if watchNS != "" {
 		setupLog.Info("Watching single namespace", "namespace", watchNS)
 		mgrOptions.Cache = cache.Options{
@@ -348,12 +347,10 @@ func main() {
 			"app.kubernetes.io/name": "workload-variant-autoscaler",
 		})
 
-		setupLog.Info("Configuring cache with label selector for ConfigMaps and namespace-scoped Secrets",
-			"labelSelector", wvaConfigSelector.String(),
-			"secretNamespace", secretNS)
+		setupLog.Info("Configuring cache with label selector for ConfigMaps",
+			"labelSelector", wvaConfigSelector.String())
 
-		// Configure cache to only watch configmaps with the WVA labels and
-		// secrets in the controller namespace (for epp-metrics-token).
+		// Configure cache to only watch configmaps with the WVA labels.
 		// Other resource types are cached normally without filtering.
 		mgrOptions.Cache = cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -333,6 +333,7 @@ func main() {
 	}
 
 	watchNS := cfg.WatchNamespace()
+	secretNS := config.SystemNamespace()
 	if watchNS != "" {
 		setupLog.Info("Watching single namespace", "namespace", watchNS)
 		mgrOptions.Cache = cache.Options{
@@ -347,17 +348,24 @@ func main() {
 			"app.kubernetes.io/name": "workload-variant-autoscaler",
 		})
 
-		setupLog.Info("Configuring cache with label selector for ConfigMaps",
-			"labelSelector", wvaConfigSelector.String())
+		setupLog.Info("Configuring cache with label selector for ConfigMaps and namespace-scoped Secrets",
+			"labelSelector", wvaConfigSelector.String(),
+			"secretNamespace", secretNS)
 
-		// Configure cache to only watch configmaps with the WVA labels
-		// Other resource types are cached normally without filtering
+		// Configure cache to only watch configmaps with the WVA labels and
+		// secrets in the controller namespace (for epp-metrics-token).
+		// Other resource types are cached normally without filtering.
 		mgrOptions.Cache = cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
 				&corev1.ConfigMap{}: {
 					// Empty map means cache all namespaces, but filter by label
 					Namespaces: map[string]cache.Config{},
 					Label:      wvaConfigSelector,
+				},
+				&corev1.Secret{}: {
+					Namespaces: map[string]cache.Config{
+						secretNS: {}, // tells the informer to only sync secrets from the one namespace
+					},
 				},
 			},
 		}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -362,11 +362,6 @@ func main() {
 					Namespaces: map[string]cache.Config{},
 					Label:      wvaConfigSelector,
 				},
-				&corev1.Secret{}: {
-					Namespaces: map[string]cache.Config{
-						secretNS: {}, // tells the informer to only sync secrets from the one namespace
-					},
-				},
 			},
 		}
 	}
@@ -540,6 +535,7 @@ func main() {
 	inferencePoolReconciler := &controller.InferencePoolReconciler{
 		Datastore: ds,
 		Client:    mgr.GetClient(),
+		APIReader: mgr.GetAPIReader(),
 		PoolGKNN:  poolGKNN,
 	}
 
--- a/config/base/rbac/epp-metrics-role.yaml
+++ b/config/base/rbac/epp-metrics-role.yaml
@@ -10,7 +10,7 @@ rules:
   - ""
   resources:
   - secrets
+  resourceNames:
+  - epp-metrics-token
   verbs:
   - get
-  - list
-  - watch
--- a/config/base/rbac/epp-metrics-role.yaml
+++ b/config/base/rbac/epp-metrics-role.yaml
@@ -1,0 +1,16 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: epp-metrics-role
+  labels:
+    app.kubernetes.io/name: workload-variant-autoscaler
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
--- a/config/base/kustomizeconfig.yaml
+++ b/config/base/kustomizeconfig.yaml
@@ -5,6 +5,12 @@ nameReference:
   - kind: Secret
     path: metadata/annotations/kubernetes.io\/service-account.name
     create: false
+- kind: Secret
+  version: v1
+  fieldSpecs:
+  - kind: Role
+    path: rules/resourceNames
+    create: false
 # Track Secret renames inside ServiceMonitor's bearerTokenSecret.name so
 # overlay JSON patches can use the unprefixed Secret name and have kustomize
 # apply namePrefix/nameSuffix automatically.
--- a/config/base/rbac/epp-metrics-rolebinding.yaml
+++ b/config/base/rbac/epp-metrics-rolebinding.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: epp-metrics-rolebinding
+  labels:
+    app.kubernetes.io/name: workload-variant-autoscaler
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: epp-metrics-role
+subjects:
+- kind: ServiceAccount
+  name: controller-manager
--- a/config/base/rbac/kustomization.yaml
+++ b/config/base/rbac/kustomization.yaml
@@ -16,3 +16,6 @@ resources:
 - epp-metrics-reader-clusterrole.yaml
 - epp-metrics-reader-clusterrolebinding.yaml
 - epp-metrics-token-secret.yaml
+# EPP metrics token: namespaced Role for epp-metrics-token Secret access.
+- epp-metrics-role.yaml
+- epp-metrics-rolebinding.yaml
--- a/config/base/rbac/manager-clusterrole.yaml
+++ b/config/base/rbac/manager-clusterrole.yaml
@@ -13,7 +13,6 @@ rules:
   - nodes/status
   - pods
   - resourcequotas
-  - secrets
   - services
   verbs:
   - get
--- a/internal/collector/source/pod/pod_scraping_source.go
+++ b/internal/collector/source/pod/pod_scraping_source.go
@@ -32,6 +32,7 @@ import (
 type PodScrapingSource struct {
 	config     PodScrapingSourceConfig
 	k8sClient  client.Client
+	apiReader  client.Reader
 	httpClient *http.Client
 	registry   *source.QueryList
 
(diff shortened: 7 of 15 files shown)


config/base/rbac/epp-metrics-role.yaml
lionelvillard, line 1: how is this related to EPP? This gives read access to all secrets.

cmd/main.go
```diff
@@ -333,6 +333,7 @@ func main() {
 	}
 
 	watchNS := cfg.WatchNamespace()
+	secretNS := config.SystemNamespace()
 	if watchNS != "" {
```
mamy-CS, line 336: the namespaced role and epp-metrics-token live in `SystemNamespace`, if `WATCH_NAMESPACE` is ever set to a namespace other than the controller's the token secret won't be readable. What's the plan here? Is single-namespace mode expected to always watch the system namespace?

cmd/main.go
```diff
@@ -347,18 +348,25 @@ func main() {
...
+		// secrets in the controller namespace (for epp-metrics-token).
+		// Other resource types are cached normally without filtering.
 		mgrOptions.Cache = cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
 				&corev1.ConfigMap{}: {
 					// Empty map means cache all namespaces, but filter by label
 					Namespaces: map[string]cache.Config{},
 					Label:      wvaConfigSelector,
 				},
+				&corev1.Secret{}: {
```
mamy-CS: since you already know you only need epp-metrics-token, can further shrink what the informer caches. RBAC still needs namespace-wide list/watch, but the cache doesn't, I think.

mamy-CS (commented):
e2e-openshift is skipping here, since this touches base RBAC that the OpenShift overlay inherits, can we confirm the nightly OCP run validates the manager still reads epp-metrics-token under the namespaced Role? Just want eyes on it there before we rely on it.

config/base/rbac/epp-metrics-role.yaml
shuynh2017: this should only gives WVA manager access to secrets in its namespace, as opposed to  secrets in all namespaces originally. We can narrow down and specify the exact name of the secret but then we need to drop `list` and `watch`. May be we should specify the exact secret name?

cmd/main.go
shuynh2017: This is single namespace for now (i.e. no change). Until we start to implement multiple namespaces, then we can consider. Incidentally, there's another security concern that would be addressed by supporting multiple namespaces, but I think we need to take that separately and whether there's a simpler solution for this concern.

cmd/main.go
shuynh2017: mentioned above ^^, if we specify secret name then we need to drop list/watch. I can provide another iteration and see what's the trade-offs.

shuynh2017:
> e2e-openshift is skipping here, since this touches base RBAC that the OpenShift overlay inherits, can we confirm the nightly OCP run validates the manager still reads epp-metrics-token under the namespaced Role? Just want eyes on it there before we rely on it.

by doing `ok-to-test` ?

shuynh2017:
@lionelvillard @mamy-CS , I've updated the description with a different implementation https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1416#issue-4877468196. Pls review. Thanks.

lionelvillard:
> But actually, WVA manager only need to reads 1 secret in 1 namespce: wva-epp-metrics-token in workload-variant-autoscaler-system.

what happens when multiple EPPs are deployed? Isn't 1 secret per EPP?

shuynh2017:
> > But actually, WVA manager only need to reads 1 secret in 1 namespce: wva-epp-metrics-token in workload-variant-autoscaler-system.
> 
> what happens when multiple EPPs are deployed? Isn't 1 secret per EPP?

@lionelvillard, I think you have a point.  Let's look more carefully:
- Currently, wva manager pod mounts secret `wva-epp-metrics-token` in `workload-variant-autoscaler-system` to `/var/run/secret/epp-metrics`. It then reads this token. If this token exists, then it is used to scrape pods. This is currently working on kind cluster. If this token doesn't exist, then it will call API to get secret which can be configured from different namespaces (EPP's), and the secret name is also configurable.
- I think may be this is the case where the scan thinks there is one secret in one namespace where the design is to be able to get secrets from namespaces - in which case the original RBAC is needed and no changes is needed. Pls comment.

shuynh2017:
@lionelvillard if ^^ align with your thinking, then I will close this PR. Thanks.

lionelvillard:
@shuynh2017 this PR is going to the right direction. However the provided Role is not correct. There should be no need to allow WVA to read secret. The WVA SA token is automatically mounted (and not related to any EPPs). When an EPP deployment is created, it should include a ClusterRoleBinding to allow WVA to scrape metrics. The documentation is here: https://book.kubebuilder.io/reference/metrics.html#metrics (a bit hard to read).

shuynh2017:
> There should be no need to allow WVA to read secret. When an EPP deployment is created, it should include a ClusterRoleBinding to allow WVA to scrape metrics. The documentation is here: https://book.kubebuilder.io/reference/metrics.html#metrics (a bit hard to read).

Thanks @lionelvillard. It looks like there are couple of things here. Let's take this first:
-  Having RBAC reading all secrets from all namespaces is the main security concern. If `There should be no need to allow WVA to read secret. ` then we can drop the RBAC. This means code like this should be removed, @dumb0002 @mamy-CS, I think you are familiar with the following code, do you have any concerns for removing this in `pod_scraping_source.go`:
```
	// Try to read from secret
	secret := &corev1.Secret{}
	// Use the specified secret namespace, or fall back to service namespace
	secretNamespace := p.config.MetricsReaderSecretNamespace
	if secretNamespace == "" {
		secretNamespace = p.config.ServiceNamespace
	}
	secretKey := types.NamespacedName{
		Name:      p.config.MetricsReaderSecretName,
		Namespace: secretNamespace,
	}

	if err := p.k8sClient.Get(ctx, secretKey, secret); err != nil {
		// Secret doesn't exist - treat authentication as optional
		// Return no error, just indicate auth should not be used
		return "", false, nil
	}
```

dumb0002:
@shuynh2017, a clarification question: why do we need the proposed `wva-epp-metrics-role` and  `wva-epp-metrics-rolebinding`? The `secret wva-epp-metrics-token` is mounted as a volume in the WVA controller. Also this secret is created during deployment of WVA.

dumb0002:
> > There should be no need to allow WVA to read secret. When an EPP deployment is created, it should include a ClusterRoleBinding to allow WVA to scrape metrics. The documentation is here: https://book.kubebuilder.io/reference/metrics.html#metrics (a bit hard to read).
> 
> Thanks @lionelvillard. It looks like there are couple of things here. Let's take this first:
> 
> * Having RBAC reading all secrets from all namespaces is the main security concern. If `There should be no need to allow WVA to read secret. ` then we can drop the RBAC. This means code like this should be removed, @dumb0002 @mamy-CS, I think you are familiar with the following code, do you have any concerns for removing this in `pod_scraping_source.go`:
> 
> ```
> 	// Try to read from secret
> 	secret := &corev1.Secret{}
> 	// Use the specified secret namespace, or fall back to service namespace
> 	secretNamespace := p.config.MetricsReaderSecretNamespace
> 	if secretNamespace == "" {
> 		secretNamespace = p.config.ServiceNamespace
> 	}
> 	secretKey := types.NamespacedName{
> 		Name:      p.config.MetricsReaderSecretName,
> 		Namespace: secretNamespace,
> 	}
> 
> 	if err := p.k8sClient.Get(ctx, secretKey, secret); err != nil {
> 		// Secret doesn't exist - treat authentication as optional
> 		// Return no error, just indicate auth should not be used
> 		return "", false, nil
> 	}
> ```

@shuynh2017 (cc: @mamy-CS), IMK `config/base/rbac/manager-clusterrole.yaml:7-21` grants the manager `get/list/watch` on secrets (cluster-wide). This is not for mounting the EPP token. It's there because the Go code path in `PodScrapingSource.getAuthToken()` has an alternative token-fetch mechanism. Hence, my vote is to drop the RBAC.

shuynh2017:
> @shuynh2017, a clarification question: why do we need the proposed `wva-epp-metrics-role` and `wva-epp-metrics-rolebinding`? The `secret wva-epp-metrics-token` is mounted as a volume in the WVA controller. Also this secret is created during deployment of WVA.

Pls ignore all current changes. I will update all the changes once we are done with the discussion. Thanks.

shuynh2017:
> > > There should be no need to allow WVA to read secret. When an EPP deployment is created, it should include a ClusterRoleBinding to allow WVA to scrape metrics. The documentation is here: https://book.kubebuilder.io/reference/metrics.html#metrics (a bit hard to read).
> > 
> > 
> > Thanks @lionelvillard. It looks like there are couple of things here. Let's take this first:
> > 
> > * Having RBAC reading all secrets from all namespaces is the main security concern. If `There should be no need to allow WVA to read secret. ` then we can drop the RBAC. This means code like this should be removed, @dumb0002 @mamy-CS, I think you are familiar with the following code, do you have any concerns for removing this in `pod_scraping_source.go`:
> > 
> > ```
> > 	// Try to read from secret
> > 	secret := &corev1.Secret{}
> > 	// Use the specified secret namespace, or fall back to service namespace
> > 	secretNamespace := p.config.MetricsReaderSecretNamespace
> > 	if secretNamespace == "" {
> > 		secretNamespace = p.config.ServiceNamespace
> > 	}
> > 	secretKey := types.NamespacedName{
> > 		Name:      p.config.MetricsReaderSecretName,
> > 		Namespace: secretNamespace,
> > 	}
> > 
> > 	if err := p.k8sClient.Get(ctx, secretKey, secret); err != nil {
> > 		// Secret doesn't exist - treat authentication as optional
> > 		// Return no error, just indicate auth should not be used
> > 		return "", false, nil
> > 	}
> > ```
> 
> @shuynh2017 (cc: @mamy-CS), IMK `config/base/rbac/manager-clusterrole.yaml:7-21` grants the manager `get/list/watch` on secrets (cluster-wide). This is not for mounting the EPP token. It's there because the Go code path in `PodScrapingSource.getAuthToken()` has an alternative token-fetch mechanism. Hence, my vote is to drop the RBAC.

@dumb0002, the above snip code is the alternative token-fetch mechanism which currently coded as needing to be able to read all secrets in any namespace. If this is the intent then we can't drop the RBAC.

lionelvillard:
RBAC is way better than allowing WVA to read secrets. The ClusterRoleBinding is deployed alongside EPP, not WVA controller.

dumb0002:
> RBAC is way better than allowing WVA to read secrets. The ClusterRoleBinding is deployed alongside EPP, not WVA controller.

@lionelvillard, which ClusterRoleBinding object are you referring to here?

dumb0002:
> > > > There should be no need to allow WVA to read secret. When an EPP deployment is created, it should include a ClusterRoleBinding to allow WVA to scrape metrics. The documentation is here: https://book.kubebuilder.io/reference/metrics.html#metrics (a bit hard to read).
> > > 
> > > 
> > > Thanks @lionelvillard. It looks like there are couple of things here. Let's take this first:
> > > 
> > > * Having RBAC reading all secrets from all namespaces is the main security concern. If `There should be no need to allow WVA to read secret. ` then we can drop the RBAC. This means code like this should be removed, @dumb0002 @mamy-CS, I think you are familiar with the following code, do you have any concerns for removing this in `pod_scraping_source.go`:
> > > 
> > > ```
> > > 	// Try to read from secret
> > > 	secret := &corev1.Secret{}
> > > 	// Use the specified secret namespace, or fall back to service namespace
> > > 	secretNamespace := p.config.MetricsReaderSecretNamespace
> > > 	if secretNamespace == "" {
> > > 		secretNamespace = p.config.ServiceNamespace
> > > 	}
> > > 	secretKey := types.NamespacedName{
> > > 		Name:      p.config.MetricsReaderSecretName,
> > > 		Namespace: secretNamespace,
> > > 	}
> > > 
> > > 	if err := p.k8sClient.Get(ctx, secretKey, secret); err != nil {
> > > 		// Secret doesn't exist - treat authentication as optional
> > > 		// Return no error, just indicate auth should not be used
> > > 		return "", false, nil
> > > 	}
> > > ```
> > 
> > 
> > @shuynh2017 (cc: @mamy-CS), IMK `config/base/rbac/manager-clusterrole.yaml:7-21` grants the manager `get/list/watch` on secrets (cluster-wide). This is not for mounting the EPP token. It's there because the Go code path in `PodScrapingSource.getAuthToken()` has an alternative token-fetch mechanism. Hence, my vote is to drop the RBAC.
> 
> @dumb0002, the above snip code is the alternative token-fetch mechanism which currently coded as needing to be able to read all secrets in any namespace. If this is the intent then we can't drop the RBAC.

@mamy-CS , can you comment here? IMK the above code snipe is one of the available options to retrieves the authentication token in `pod_scraping_source`. We already support using a `beartoken` which is used for the EPP metrics. Is it a strong requirement to also support an authentication via reading a secret??

shuynh2017:
Close this PR. Will refer to the discussion here in a future PR.

mamy-CS:
I agree. No, the secret reading path is not a strong requirement. And I agree we should drop it along with the RBAC. 
One clarification here so we don't conflate two things, @lionelvillard's the clusterrolebinding point is about the reverse direction authorizing WVA's SA to be allowed to scrape an EPP's protected metrics endpoint (the epp side), deployed alongside each EPP. That's independent of WVA reading a secret via the API, which is what we're removing here. Removing the secret read path shouldn't affect that.

PR #1419, opened 2026-07-14
update: harden TLS for openshift etc

zdtsw:
These requirements are from security scan in product:
- Flip PROMETHEUS_TLS_INSECURE_SKIP_VERIFY default to false in config/manager/configmap.yaml and Helm values.yaml.
- Pin the ci-signed-commits.yaml reusable-workflow ref to a SHA.
- Set ServiceMonitor insecureSkipVerify: false (or wire CA/serving-cert).
- Digest-pin the upstream Dockerfile builder stage.

## Proposed Changes
- pin image and GHAction with SHA1
- disable insecureSkipVerify by inject service-ca bundle

- :broom: Update or clean up current behavior

### Pre-review Checklist

**Release Note**

```release-note

```

**Docs**

cc @shuynh2017

### DIFF
--- a/config/components/openshift/deployment-patch.yaml
+++ b/config/components/openshift/deployment-patch.yaml
@@ -18,3 +18,12 @@ spec:
           value: ""
         - name: PROMETHEUS_SERVER_NAME
           value: "thanos-querier.openshift-monitoring.svc"
+        volumeMounts:
+        - name: metrics-serving-cert
+          mountPath: /tmp/k8s-metrics-server/serving-certs
+          readOnly: true
+      volumes:
+      - name: metrics-serving-cert
+        secret:
+          secretName: metrics-server-cert
+          defaultMode: 420
--- a/config/components/openshift/deployment-patch.yaml
+++ b/config/components/openshift/deployment-patch.yaml
@@ -20,7 +20,7 @@ spec:
           value: "thanos-querier.openshift-monitoring.svc"
         volumeMounts:
         - name: metrics-serving-cert
-          mountPath: /tmp/k8s-metrics-server/serving-certs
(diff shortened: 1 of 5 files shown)


shuynh2017:
@zdtsw, can you point to the fix for "Flip PROMETHEUS_TLS_INSECURE_SKIP_VERIFY default to false in config/manager/configmap.yaml and Helm values.yaml."

zdtsw:
> @zdtsw, can you point to the fix for "Flip PROMETHEUS_TLS_INSECURE_SKIP_VERIFY default to false in config/manager/configmap.yaml and Helm values.yaml."

Helm part i think we have dropped it , right? 
we do not use Helm anymore here
as for the PROMETHEUS_TLS_INSECURE_SKIP_VERIFY
https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/main/config/components/openshift/configmap-patch.yaml#L12 is for openshift

shuynh2017:
/lgtm
@lionelvillard for approval.

config/components/openshift/deployment-patch.yaml
```diff
@@ -18,3 +18,12 @@ spec:
           value: ""
         - name: PROMETHEUS_SERVER_NAME
           value: "thanos-querier.openshift-monitoring.svc"
+        volumeMounts:
+        - name: metrics-serving-cert
+          mountPath: /tmp/k8s-metrics-server/serving-certs
```
lionelvillard: `/tmp` is not the natural place for the cert. `/etc` is better

zdtsw: updated

lionelvillard (approved):

PR #1434, opened 2026-07-18
fix(coordinator): respect HPA spec.minReplicas in GPU rebalance floor

### LINKED ISSUE #1406: fix(coordinator): respect scaler minimum replicas during rebalancing
### Description

The `gpu-rebalance` plugin currently uses a hardcoded minimum target of `1`. It can therefore patch `HPA.spec.maxReplicas` or `ScaledObject.spec.maxReplicaCount` below the scaler's configured minimum.

A simple clamp is not sufficient because the aggregate minima may already exceed the namespace GPU quota.

The rebalancer should:

- derive the effective minimum from `HPA.spec.minReplicas` or `ScaledObject.spec.minReplicaCount`;
- skip the namespace without patching when aggregate minima exceed quota;
- otherwise reserve the minima and distribute the remaining quota using the existing queue weights.

### Acceptance criteria

- Support HPA, ScaledObject, and mixed entries.
- Never patch a maximum below its effective minimum.
- Do not partially patch an infeasible namespace.
- Cover minima greater than `1`, nil/zero minima, and aggregate minima exceeding quota.

Related to #1348. Discovered while reviewing #1384.

Goutham-Annem:
Fixes #1406

## Proposed Changes

The GPU rebalance plugin's proportional allocation algorithm clamped every computed `maxReplicas` target to a minimum of `1`, ignoring the HPA's `spec.minReplicas`. Setting `maxReplicas` below `minReplicas` produces an invalid HPA spec that Kubernetes rejects with a `422 Unprocessable Entity` error. A concrete example: an HPA with `spec.minReplicas = 3` could receive `maxReplicas = 2` from the rebalancer when its queue share was less than 20% on a quota-10 pool.

- :bug: Replace the hardcoded floor of `1` with `max(1, hpa.Spec.MinReplicas)` so the patched `maxReplicas` is always a value Kubernetes will accept
- :broom: Add `makeHPAWithMin` test helper alongside the existing `makeHPA`
- :broom: Add `TestRebalance_RespectsHPAMinReplicas` — verifies `maxReplicas` is clamped up when proportional allocation would go below `minReplicas`
- :broom: Add `TestRebalance_NoMinReplicasDefaultsToOne` — verifies the nil `minReplicas` (Kubernetes default of 1) behaviour is preserved

### Pre-review Checklist

- [x] **E2E tests** for any new behavior — unit tests added for both branches
- [ ] **Docs PR** for any user-facing impact — n/a
- [ ] **Proposal PR** for any new enhancement or change to existing behavior — n/a (bug fix)

**Release Note**
```release-note
[BUG] GPU rebalance coordinator now uses max(1, hpa.Spec.MinReplicas) as the minimum target, preventing invalid HPA patches when spec.minReplicas is greater than 1.
```

### DIFF
--- a/internal/coordinator/plugins/gpurebalance/plugin.go
+++ b/internal/coordinator/plugins/gpurebalance/plugin.go
@@ -117,34 +117,37 @@ func (p *Plugin) rebalanceNamespace(ctx context.Context, log logr.Logger, ns str
 		}
 	}
 
-	// TODO: the minimum target is hardcoded to 1 but an HPA may have spec.minReplicas
-	// set higher (e.g. 2 or 3). Setting maxReplicas below minReplicas produces an
-	// invalid HPA and Kubernetes will reject the patch. The floor should be
-	// max(1, hpa.Spec.MinReplicas) so the computed target is always a valid value.
-
-	// TODO: when len(entries) > quota the minimum-1 clamp causes every HPA to
-	// receive maxReplicas=1 and allocated exceeds quota (e.g. 100 HPAs, quota=10
-	// → allocated=100). The ResourceQuota becomes the only enforcement and the
-	// first 10 pods to be scheduled win while the other 90 are stuck pending.
-	// Fix: rank HPAs by queue depth, assign maxReplicas=1 only to the top-quota
-	// pools, and park the rest at minReplicas so the budget is not over-committed.
+	// TODO: when len(entries) > quota the minimum-minReplicas clamp can still
+	// cause allocated to exceed quota (e.g. 100 HPAs each with minReplicas=1,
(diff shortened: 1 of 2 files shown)


asm582:
@Goutham-Annem Thanks for this PR. We have migrated to KEDA. @xuhui-lu has already sent a PR for this. Can we close this PR?

Goutham-Annem:
Closing — the minReplicas concern was addressed upstream in #1384 (`feat(coordinator): rebalance KEDA ScaledObjects`) which introduced the `effectiveMin` approach that reserves each scaler's minimum before distributing remaining quota proportionally. That implementation is more complete (handles both HPAs and ScaledObjects, and warns when minimumTotal > quota).

PR #1308, opened 2026-06-22
feat(benchmark): add two-variant WVA efficiency-aware scaling scenario

biranofer:
## Summary

Adds a **two-variant WVA efficiency-aware scaling** benchmark scenario: a single model deployed as two variants of differing \`variantCost\` under one shared \`InferencePool\` / EPP, exercising the WVA saturation-V2 cost-aware optimizer. The optimizer scales up the **most efficient** variant first (best serving-capacity per unit cost), with the cheaper variant absorbing spillover.

## What's included

- \`hack/benchmark/scenarios/guides/two-variant-wva.yaml\` — primary-variant standup spec (cost 10, TP=2), pins WVA chart/controller \`v0.8.0-rc5\`.
- \`hack/benchmark/scenarios/guides/variants/v2-tp1-cheaper.yaml\` — secondary variant (cost 5, **TP=1 / 1 GPU**), a genuinely smaller shape, not just cheaper-same-shape.
- \`hack/benchmark/add_variant.py\` — layers the secondary \`Deployment\`/\`VariantAutoscaling\`/\`HPA\` onto the running primary; resolves primary TP from the \`VLLM_TENSOR_PARALLELISM\` env / \`sh -c\` wrapping so the TP override actually takes effect.
- \`hack/benchmark/scenarios/wva_threshold/wva_saturation_v2_config.yaml\` — ConfigMap selecting saturation V2.
- \`hack/benchmark/post_run_analyze.sh\` — wraps the 5 post-run dump+plot steps that must run promptly after \`benchmark-run\` (WVA controller log buffer rotates; runs \`dump_wva_target_timeseries.py\`, \`dump_capacity_demand_estimate.py\`, \`dump_epp_throughput.py\`, \`dump_wva_full_timeseries.py\`, \`plot_two_variant_pipeline.py\`).
- \`hack/benchmark/dump_wva_target_timeseries.py\` — extracts WVA controller decisions and V2 saturation analysis numbers from pod logs.
- \`hack/benchmark/dump_capacity_demand_estimate.py\` — computes per-variant capacity/demand estimate from raw vLLM/EPP scrapes.
- \`hack/benchmark/dump_epp_throughput.py\` — derives request rate from EPP counters.
- \`hack/benchmark/dump_wva_full_timeseries.py\` — extracts WVA Prometheus metrics timeseries.
[...]

PR #1435, opened 2026-07-18
feat(benchmark): migrate two-variant benchmark from VA+HPA to KEDA ScaledObjects

biranofer:
## Summary

Replaces the deprecated `VariantAutoscaling` CRD + direct HPA path in the
two-variant benchmark with KEDA ScaledObjects using annotation-based WVA
discovery (`llm-d.ai/managed: "true"`). This aligns the benchmark with the
current recommended WVA deployment path and removes the dependency on the
deprecated VA CRD (scheduled for removal in v0.10.0).

Supersedes the benchmark scenario introduced in #1308, updating it to the
KEDA ScaledObject path that is now the recommended autoscaling approach.

## What changed

**`hack/benchmark/add_variant.py`** — core rewrite:
- Creates a KEDA `TriggerAuthentication` (auto-detects the WVA SA token secret
  name: `wva-controller-manager-token` on fresh Kustomize installs)
- Creates a `ScaledObject` for the primary variant on first run, converting any
  legacy direct HPA left by the benchmark harness (deletes HPA first to satisfy
  KEDA's admission webhook)
- Creates a `ScaledObject` for the added variant
- ScaledObjects carry required labels: `wva.llmd.ai/controller-instance`,
  `inference.optimization/acceleratorName`
- Supports N variants (naming is generic, not primary/secondary-specific)
- Fix: `find_primary_deployment` now selects by `llm-d.ai/role=decode` + absence of
  `wva.llmd.ai/variant` label (prior version checked a non-existent kebab-case label,
  causing a fallback to the first Deployment in the namespace)
- **Fix (added after initial review): correct `llm-d.ai/variant` pod-label
  attribution.** WVA's collector trusts this label as the pod's
  VariantAutoscaling name whenever it's *present* — it only falls back to
  the correct owner-chain lookup when the label is *missing*, not when
  it's wrong. Two places set it to a value that matched no tracked
  variant, silently dropping that pod's queue-backlog demand contribution
  (in-use-KV demand is attributed via a different, unaffected path, which
  is why runs still scaled *something* while quietly undercounting total
  demand by ~2×):
  - the added variant's own label was set to the bare Deployment name
    instead of its ScaledObject name (`<deployment>-scaler`) — now fixed
    in `make_variant_deployment()`. This also restores a selector/label
    pop on `llm-d.ai/inference-serving` that a prior commit in this
    branch removed as believed-dead-code — it isn't: leaving it makes
    the variant's selector a superset of the primary's, which trips the
    native HPA controller's `AmbiguousSelector` safety check and blocks
    scaling entirely on **both** ScaledObjects.
  - primary's label is set unconditionally by the `llm-d-benchmark`
    modelservice chart template regardless of any scenario-yaml
    override, and is baked into the Deployment's selector too (immutable
    once created). Rather than fight a chart default we can't durably
    patch, `add_variant.py` now reads primary's live label and names its
    ScaledObject to match, instead of assuming a fixed
    `<deployment>-scaler` pattern.

  Verified via an independent raw-scrape demand estimate (no dependency
  on WVA's own attribution): `totalDemand` recovered from a consistent
  ~0.5× ratio (undercounted) to ~0.9–1.3× (correct, within sampling
  noise). See **Validation** below — this fix changes the validation
  numbers materially, since the original Run 1/Run 2 table was captured
  on code with this undetected bug.

**`hack/benchmark/scenarios/guides/two-variant-wva.yaml`**:
- removes `wva.variantAutoscaling` and `wva.hpa` blocks; autoscaling
  objects are now created by `benchmark-add-variant`.
- removes a hardcoded `modelArtifacts.extraLabels["llm-d.ai/variant"]`
  value that was previously believed "metadata only" — it isn't; a wrong
  value here is exactly the present-but-wrong case that bypasses the
  collector's safe fallback described above. Leaving the label unset lets
  `add_variant.py` set it correctly post-standup instead.

**`Makefile`**:
- Adds `PROMETHEUS_URL` variable
- `benchmark-restart-controller` and `benchmark-enable-v2-saturation`
  auto-detect the controller deployment and saturation ConfigMap names
  (differ between Kustomize and legacy Helm installs)
- `benchmark-run` now copies the local scenario YAML to the benchmark repo
  before invoking the harness, matching what `benchmark-standup` already did
  (previously the run step used the upstream FMA-based scenario instead)
- `BENCHMARK_MODEL_ID` is now empty by default; the scenario YAML drives the
  model name (previously inherited `MODEL_ID=e2ewva/dummy-model` from the e2e
  test default, causing a model mismatch error)

**`docs/developer-guide/two-variant-wva-benchmark.md`** — updated topology
diagram, step instructions, cluster prerequisites, and failure modes for the
KEDA path.

## Prerequisites (not in this PR)

- WVA image must include **PR #1341** ("fix: exclude KEDA-generated HPAs from
  managed-HPA index"). Not in v0.8.0; use a nightly build from main.
- On OpenShift: the WVA controller `ServiceMonitor` must have a `tlsConfig`
  pointing to `wva-metrics-server-ca`. The OCP kustomize patch can fail
  silently if the base ServiceMonitor lacks the field (tracked separately).

## Validation

**The table below supersedes the original Run 1/Run 2 validation** — those
runs were captured on code that (per a commit already in this branch,
`c7d1908f`, "remove incorrect llm-d.ai/inference-serving label check")
had removed the selector fix this PR restores, and predate the
label-attribution fix described above. Neither issue produced visibly
broken output — the controller's own decision log looked healthy in
both cases — so the original numbers cannot be trusted as evidence the
cost-aware prioritization logic itself is correct. The new run below was
independently cross-checked against a raw-scrape demand estimate for
this reason.

End-to-end validation on `pokprod001` (OpenShift, NVIDIA H100 80 GB), fresh
Kustomize WVA deployment, saturation-V2 cost-aware optimizer, both variants
starting at 1/1 ready.

**Configuration**: prefill-heavy workload (4K/1K tokens, Poisson arrival,
5 min @ 2 RPS then 12 min @ 10 RPS — longer and heavier than the original
single-stage 15 RPS/900s config), `unsloth/Meta-Llama-3.1-8B-Instruct`,
primary TP=2 (cost 10), variant TP=1 (cost 5), max replicas 10 each.

| Metric | Value |
|--------|-------|
| Requests | 7800 |
| Successful | 7696 |
| Errors | 104 (1.33%) |
| Achieved RPS | 7.14 |
| Avg primary replicas (TP=2) | **1.68** |
| Max primary replicas (TP=2) | **5** |
| Avg variant replicas (TP=1) | **1.00** |
| Max variant replicas (TP=1) | **1** |
| Avg KV cache utilization | 21.3% |
| Cost (weighted avg replicas × GPU/hr) | 4.36 |
| TTFT p50 / p95 / p99 (ms) | 211 / 23,620 / 35,968 |
| TPOT (ITL) p50 / p95 / p99 (ms/token) | 24.1 / 40.0 / 73.8 |
| Demand cross-check ratio (self-reported vs. independent raw-scrape estimate) | ~0.9–1.3× throughout |

**Efficiency-aware prioritization confirmed, and for the first time
verifiably on correct demand**: primary (TP=2, the more GPU-efficient
pick per unit cost — cost ratio 10/5 = GPU ratio 2/1) is the one that
scales, up to 5 replicas; the variant (TP=1) never leaves 1. Real
replica counts cross-checked against `replica_status_timeseries.json`,
not just the controller's own decision log.

**Known remaining issue, not fixed in this PR**: the replica-count
timeseries shows a scale-up/scale-down oscillation under sustained
saturation (WVA's desired-target line swings 1→3→4→5→2→1→3→4 rather than
settling), and the 104 errors cluster at 31–45s request latency —
consistent with in-flight streams being severed by pod termination
during a scale-down, not client-side timeouts. This is a
scaleDownBoundary/hysteresis tuning question, orthogonal to the
correctness fixes in this PR.

Pipeline plot (replicas solid=KEDA actual / dashed=WVA desired, KV cache
demand vs capacity, requests running/waiting, EPP queue depth):

Two-variant KEDA pipeline plot — corrected run

## Open items

- [ ] `inference.optimization/acceleratorName: NVIDIA-H100-80GB-HBM3` is
  hardcoded in `add_variant.py`; should be auto-detected from Deployment nodeSelector
- [ ] Existing `vllm-*` PodMonitors with `llm_d_ai_variant` relabeling (deployment
  name) conflict with the KEDA path (ScaledObject names); `add_variant.py` should
  remove the relabeling automatically on OCP installs
- [ ] The WVA controller's `ServiceMonitor` has a hardcoded TLS `serverName`
  (`wva-controller-manager-metrics-service.wva-system.svc`) that doesn't
  match the real tenant namespace on namespace-scoped installs, silently
  breaking the HTTPS scrape (`wva_desired_replicas` never reaches
  Prometheus). Resets on every fresh `benchmark-standup`; not yet fixed at
  the template level, currently requires a manual patch:
  `oc patch servicemonitor wva-controller-manager-metrics-monitor -n <ns> --type=json -p '[{"op":"replace","path":"/spec/endpoints/0/tlsConfig/serverName","value":"wva-controller-manager-metrics-service.<ns>.svc"}]'`
- [ ] Scale-up/scale-down oscillation under sustained saturation (see
  Validation above) — causes real request errors from severed in-flight
  streams. Needs a tuning follow-up (raise `scaleDownBoundary`, add
  scale-up hysteresis), not addressed in this PR.


asm582:
@biranofer, can we convert this to a regular PR instead of a draft?

biranofer:
> @biranofer, can we convert this to a regular PR instead of a draft?

Still validating the labeling issues, soon.

kahilam (commented):
Hi @biranofer,  LGTM overall. Will do end-to-end testing once this is out of draft.

biranofer:
@asm582 @kahilam  note the PR was updated - a fix for labeling issue that caused the WVA Sat-v2 to not getting the queue metrics and thus not assess correctly the entire demand.

hack/benchmark/scenarios/guides/wva-sat2-tp1.yaml
```diff
@@ -0,0 +1,248 @@
...
+  # --------------------------------------------------------------------------
+  images:
+    vllmOpenai:
+      repository: docker.io/vllm/vllm-openai
+      tag: v0.14.0
+
+  # --------------------------------------------------------------------------
+  # EPP PLUGIN CONFIG — flowControl gate required for KEDA trigger
+  # --------------------------------------------------------------------------
+  inferenceExtension:
```
kahilam: Same bug fixed in two-variant-wva.yaml (commit 3ea55cf): this should be router: epp: not inferenceExtension:. The inferenceExtension key silently no-ops — flowControl won't actually be enabled.

kahilam (commented):
I added a comment.

hack/benchmark/scenarios/guides/wva-sat2-tp1.yaml
biranofer: God catch was not pushed tx ! fixed.

Issue #1441, opened 2026-07-21
Track: deprecate & remove the V1 (percentage-based) saturation analyzer


ev-shindin:
## Summary

WVA has two saturation analysis paths: **V1** (percentage/utilization-based, `optimizeV1`) and **V2** (token/capacity-based, `optimizeV2`, `internal/engines/analyzers/saturation_v2`). V2 supersedes V1. This issue tracks V1's staged deprecation and removal across releases so the default flip and the removal don't surprise operators.

The code already carries the intent — `internal/engines/saturation/engine.go:457`: *"V1 will be deprecated once V2 is fully validated."*

## How the path is selected today (v0.8.0)

- The analyzer is chosen per cycle in `engine.go` (`analyzerName` switch, ~L459–465): `"saturation"` → **V2** (`optimizeV2`), `"queueing-model"` → queueing model, **default / unset → V1** (`optimizeV1`).
- `analyzerName` comes from `SaturationScalingConfig.GetAnalyzerName()` (`internal/config/saturation_scaling.go:97`): returns `"saturation"` when the `analyzers:` list is non-empty, otherwise the `analyzerName` string field; empty → V1.
- **So V1 is the current default; V2 is opt-in** via `analyzers: [...]` or `analyzerName: saturation`.
- The queueing-model analyzer is orthogonal (presence-based, selected by its own ConfigMap) and is **not** affected by this deprecation.

### v0.8.0 — V2 recommended, V1 default *(current)*
- [x] V2 shipped and runs in parallel with V1; selectable via config.
- [x] Docs clearly **recommend V2** and show how to opt in (`analyzers:` list / `analyzerName: saturation`).
- [x] `wva_config_info` / a startup log surfaces the active analyzer path so operators can confirm which one is running.

### v0.9.0 — V2 default, V1 still selectable ("possible to return to V1")
- [ ] **Flip the default:** unset/empty analyzer selection resolves to **V2**, not V1 (update `GetAnalyzerName`/the engine's default switch case).
- [ ] Keep V1 reachable by **explicit opt-out** (e.g. `analyzerName: v1`); document the escape hatch.
- [ ] **Release note / migration note** — this is a behavior change for any cluster relying on the implicit default; call it out, with the one-line opt-out for anyone who needs V1 back.
- [ ] Confirm no test/e2e implicitly depends on V1 being the default.

### v0.10.0 — V1 marked deprecated
[...]

PR #1442, opened 2026-07-21
feat(saturation)!: default to V2 (token/capacity-based) saturation analyzer

ev-shindin:
## Fixes

Implements the **v0.9.0 milestone** of the V1-analyzer deprecation timeline (see https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1441)
*V2 default, V1 still selectable*. (Part of the V1-deprecation tracking issue;
link the issue number here.)

## Proposed Changes

Makes the **V2 (token/capacity-based)** saturation analyzer the default. Selection
uses the existing mechanism — the shipped saturation ConfigMap `default` entry now
carries an `analyzers:` section, which `SaturationScalingConfig.GetAnalyzerName()`
resolves to the V2 path. Operators opt back out to the legacy **V1
(percentage-based)** analyzer by removing the `analyzers:` section.

**Config / docs**

- **ConfigMaps** (`config/base/manager/saturation-scaling-configmap.yaml`,
  `deploy/configmap-saturation-scaling.yaml`): add `analyzers:` +
  `scaleUpThreshold`/`scaleDownBoundary` to the `default` entry. Comments label
  each threshold's ownership:
  - Shared (V1 + V2): `kvCacheThreshold`, `queueLengthThreshold`
  - V1-only (ignored by V2): `kvSpareTrigger`, `queueSpareTrigger`
  - V2-only: `scaleUpThreshold` (0.85), `scaleDownBoundary` (0.70)
- **README**: new "Upgrading to v0.9.0" section — behavioral-change notice + V1
  opt-out steps (breaking-change requirement per `release-process.md`).
- **docs/developer-guide/saturation-scaling-config.md**: new "Analyzer Selection
  (V1 vs V2)" section with a threshold-ownership table; every example `default`
  entry carries `analyzers:` so a copy-paste keeps V2; corrected the stale
  "V1 is the default" framing.
- **deploy/ci-e2e-openshift/deploy-infrastructure.sh**: its threshold patch pins
  the run to V1; marked `TODO(v2)`.

**Per-model / namespace threshold calibration** (`internal/config/saturation_scaling.go`,
`internal/engines/saturation/engine.go`)

Analyzer selection is resolved once from the **global** `default` entry, but
thresholds are resolved per-model/namespace. A per-model override or namespace-local
entry written V1-style (no `analyzers:`) can therefore run on the V2 path. Because
entries are `ApplyDefaults()`'d individually at parse time, the V2 thresholds are
defaulted on the **final resolved config** via the new `ApplyV2ThresholdDefaults()`
(post-merge) rather than in `ApplyDefaults()` — this calibrates such entries without
letting a V1-style override clobber a tuned global during `Merge`.
`resolveSaturationConfig` also resets an inverted threshold pair produced by a
cross-entry merge back to the defaults. `Validate()` now range/consistency-checks the
V2 thresholds whenever they are set, regardless of `IsV2()`, so an invalid explicit
value on a V1-style entry can no longer reach the V2 path unchecked.

**E2E tests** — V1-specific tests guarded and marked for V2 reconfiguration:

- `sglang_backend_test.go`: pins to V1 by snapshotting/restoring the saturation
  ConfigMap in `BeforeAll`/`AfterAll` (its scale-up assertion is calibrated to V1
  thresholds; the fixture's `token_usage=0.85` sits exactly at V2's
  `scaleUpThreshold=0.85`, non-deterministic under V2). Marked `TODO(v2)`.
- `saturation_analyzer_path_test.go`: already self-guards both paths — added a note
  plus `TODO(v1-removal)`.
- `saturation_v2_test.go` / `throughput_analyzer_test.go` already write their own
  analyzer config — unaffected.

**Unit tests**

- `internal/config/shipped_configmap_test.go` (new): asserts both shipped ConfigMaps'
  `default` resolves to V2 — guards against a YAML typo silently reverting to V1.
- `resolveSaturationConfig` tests model production parse order and cover tuned-global
  inheritance, fully-V1-style calibration, and inverted-pair reset;
  `ApplyV2ThresholdDefaults` and the new `Validate` rules have direct unit tests.
  (Guard strength verified empirically: the fix-dependent specs fail when the
  post-merge calibration is removed.)

## Pre-review Checklist

- [x] Commit message follows [Conventional Commits](https://www.conventionalcommits.org/) (`feat(saturation)!:` + `BREAKING CHANGE:` footer)
- [x] Docs updated (developer guide + README upgrade note)
- [x] Unit tests added/updated; `go vet` clean; changed Go files `gofmt`-clean
- [x] `internal/config`, `internal/engines/saturation` (envtest), and `internal/controller` (envtest) suites pass
- [x] E2E smoke run on the V2 default (reviewer / CI)

## Release Note

```release-note
BREAKING CHANGE: The default saturation analyzer is now V2 (token/capacity-based)
instead of V1 (percentage-based). V2 may produce different scaling decisions for
the same workload; review dashboards and alert thresholds after upgrading. To stay
on V1, remove the `analyzers:` section (and the V2-only scaleUpThreshold/
scaleDownBoundary fields) from the `default` entry of the saturation ConfigMap.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,40 @@ spec:
 
 More examples in [config/samples/hpa/](config/samples/hpa/) and [config/samples/keda/](config/samples/keda/).
 
+## Upgrading
+
+### Upgrading to v0.9.0 — V2 saturation analyzer is now the default
+
+**Behavioral change.** The default saturation analyzer changes from **V1**
+(percentage/spare-capacity-based) to **V2** (token/capacity-based). The shipped
+`default` entry in the saturation ConfigMap now includes an `analyzers:` section,
+which selects V2. No code change or image rebuild is involved — analyzer selection
+is driven entirely by config.
+
+V2 may produce different scaling decisions than V1 for the same workload. Review
+your dashboards and alert thresholds after upgrading.
+
+**Staying on V1 (opt-out).** Remove the `analyzers:` section (and the V2-only
+`scaleUpThreshold` / `scaleDownBoundary` fields) from the `default` entry of your
+saturation ConfigMap. The remaining `kvCacheThreshold`, `queueLengthThreshold`,
+`kvSpareTrigger`, and `queueSpareTrigger` fields drive V1:
+
+```yaml
+data:
+  default: |
+    kvCacheThreshold: 0.80
+    queueLengthThreshold: 5
+    kvSpareTrigger: 0.1
+    queueSpareTrigger: 3
+```
+
+Apply with `kubectl apply -f deploy/configmap-saturation-scaling.yaml`; the change
+takes effect immediately (the controller watches the ConfigMap).
+
+> V1 is deprecated and scheduled for removal in a future release. See the
+> [saturation scaling configuration guide](docs/developer-guide/saturation-scaling-config.md#analyzer-selection-v1-vs-v2)
+> for threshold ownership (which fields each analyzer reads) and migration details.
+
 ## Contributing
 
 We welcome contributions! See [CONTRIBUTING.md](CONTRIBUTING.md) for guidelines.
--- a/config/base/manager/saturation-scaling-configmap.yaml
+++ b/config/base/manager/saturation-scaling-configmap.yaml
@@ -15,10 +15,23 @@ metadata:
     # are never picked up at runtime.
     app.kubernetes.io/name: workload-variant-autoscaler
 data:
-  # Global defaults applied to all variants unless overridden
+  # Global defaults applied to all variants unless overridden.
+  #
+  # The `analyzers` section selects the V2 (token/capacity-based) saturation
+  # analyzer — the default since v0.9.0. Remove the `analyzers` section (and the
+  # V2-only thresholds) to opt out to the legacy V1 (percentage-based) analyzer.
+  # See docs/developer-guide/saturation-scaling-config.md.
   default: |
+    analyzers:
+      - name: saturation
+        score: 1.0
+    # V2-only thresholds:
+    scaleUpThreshold: 0.85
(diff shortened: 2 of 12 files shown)


deanlorenz (approved):
LGTM — clean, well-tested flip. I traced the selection/threshold flow and the "selection is global, thresholds are per-model/namespace" reasoning holds up, so calibrating the V2 thresholds post-merge (rather than in `ApplyDefaults`) is genuinely necessary to keep a V1-style override from clobbering a tuned global. The production-parse-order specs and the shipped-ConfigMap `IsV2()` guard are the right tests for it. Approving.

Two **non-blocking** items — take or leave, no need to hold the merge:

1. **Inverted-pair reset in `resolveSaturationConfig`.** When a cross-entry merge yields `scaleUp <= scaleDown`, resetting *both* thresholds to the hardcoded defaults silently discards a tuned value — e.g. a global `scaleUpThreshold: 0.95` is thrown away if a namespace entry sets `scaleDownBoundary: 0.97`. Since this runs in the reconcile loop, erroring out would drop the model from autoscaling (worse than a fallback). Could instead **keep the base entry's already-validated pair and drop only the override's threshold contribution**, plus emit a K8s Event / warning log so the reset isn't silent — preserves the tuned global and makes the conflict visible.

2. **README upgrade note.** It explains the global default flip and the opt-out well, but an operator with existing per-model/namespace overrides written V1-style (no `analyzers:`) might assume those stay on V1 — since selection is global, they flip to V2 too. The developer guide's "selection is global; thresholds are per-model" note covers this; one sentence in the README upgrade section would close the gap for anyone who only reads the upgrade note.

PR #1443, opened 2026-07-21
feat: benchmark-run copies local workload profiles to harness

asm582:
## Summary
- Allow `benchmark-run` to use local `.yaml` workload files from `test/benchmark/scenarios/`
- Automatically copies local workloads to the llm-d-benchmark harness directory during run phase
- Enables running workloads without requiring download from the inference-perf catalog

## Test plan
- [x] Run benchmark with local workload file: `make benchmark-run BENCHMARK_WORKLOAD=interactive-chat`
- [x] Verify local file is copied to harness instead of attempting download
- [x] Verify benchmark run completes successfully with local workload

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -428,6 +428,10 @@ benchmark-standup: ## Stand up the benchmark environment (set BENCHMARK_NAMESPAC
 	   $(BENCHMARK_REPO_DIR)/config/scenarios/$(BENCHMARK_SPEC).yaml; \
 	exit $$rc
 
+# Note: For KEDA autoscaling to work, the benchmark namespace must have the label:
+#   oc label namespace <namespace> openshift.io/user-monitoring=true
+# This enables Prometheus to scrape metrics from ServiceMonitors in that namespace.
+
 .PHONY: benchmark-run
 benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<namespace>, MODEL_ID=<model>, BENCHMARK_HARNESS=guidellm|inference-perf)
 	@if [ -z "$(BENCHMARK_NAMESPACE)" ]; then \
--- a/Makefile
+++ b/Makefile
@@ -482,7 +482,8 @@ benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<name
 		-l $(BENCHMARK_HARNESS) \
 		-w $(BENCHMARK_WORKLOAD) \
 		$(if $(BENCHMARK_MODEL_ID),-m $(BENCHMARK_MODEL_ID),) \
-		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,)
+		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,) \
+		--wait-timeout $(BENCHMARK_WAIT_TIMEOUT)
 	@echo ""
 	@echo "========================================="
 	@echo "  Generating benchmark report..."
--- a/Makefile
+++ b/Makefile
@@ -477,7 +477,7 @@ benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<name
 	$(LLMDBENCHMARK) $(BENCHMARK_CLI_FLAGS) run \
 		-p $(BENCHMARK_NAMESPACE) \
 		-l $(BENCHMARK_HARNESS) \
-		-w $(BENCHMARK_WORKLOAD) \
+		-w $(BENCHMARK_WORKLOAD).yaml \
 		$(if $(BENCHMARK_MODEL_ID),-m $(BENCHMARK_MODEL_ID),) \
 		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,) \
 		--wait-timeout $(BENCHMARK_WAIT_TIMEOUT)
--- a/Makefile
+++ b/Makefile
@@ -425,12 +425,13 @@ benchmark-standup: ## Stand up the benchmark environment (set BENCHMARK_NAMESPAC
 		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,); \
 	rc=$$?; \
 	mv $(BENCHMARK_REPO_DIR)/config/scenarios/$(BENCHMARK_SPEC).yaml.bak \
-	   $(BENCHMARK_REPO_DIR)/config/scenarios/$(BENCHMARK_SPEC).yaml; \
+	   $(BENCHMARK_REPO_DIR)/config/scenarios/$(BENCHMARK_SPEC).yaml
+	@if [ $$rc -eq 0 ] && [ "$(BENCHMARK_MONITORING)" = "true" ]; then \
+		echo "Enabling user-workload monitoring for namespace $(BENCHMARK_NAMESPACE)..."; \
+		oc label namespace $(BENCHMARK_NAMESPACE) openshift.io/user-workload-monitoring=enabled --overwrite 2>/dev/null && \
+		echo "✅ Monitoring label applied. Prometheus will begin scraping ServiceMonitors in this namespace."; \
+	fi
 	exit $$rc
-
-# Note: For KEDA autoscaling to work, the benchmark namespace must have the label:
-#   oc label namespace <namespace> openshift.io/user-monitoring=true
-# This enables Prometheus to scrape metrics from ServiceMonitors in that namespace.
 
 .PHONY: benchmark-run
 benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<namespace>, MODEL_ID=<model>, BENCHMARK_HARNESS=guidellm|inference-perf)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -428,13 +428,29 @@ benchmark-standup: ## Stand up the benchmark environment (set BENCHMARK_NAMESPAC
...
 .PHONY: benchmark-run
 benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<namespace>, MODEL_ID=<model>, BENCHMARK_HARNESS=guidellm|inference-perf)
 	@if [ -z "$(BENCHMARK_NAMESPACE)" ]; then \
 		echo "ERROR: BENCHMARK_NAMESPACE is required. Usage: make benchmark-run BENCHMARK_NAMESPACE=<namespace>"; \
 		exit 1; \
 	fi
 	@mkdir -p "$(BENCHMARK_SCENARIOS_DIR)"
+	@if [ -n "$(BENCHMARK_MODEL_ID)" ] && [ -f "$(BENCHMARK_SCENARIOS_DIR)/$(BENCHMARK_WORKLOAD)" ]; then \
+		echo "Injecting MODEL_ID=$(BENCHMARK_MODEL_ID) into local workload file before rendering plan..."; \
+		sed -i.bak 's|model_name: .*|model_name: $(BENCHMARK_MODEL_ID)|' \
```
kahilam: sed -i mutates source files in git — The sed commands at linesmodify the original file in $(BENCHMARK_SCENARIOS_DIR) (i.e., test/benchmark/scenarios/). This permanently changes the tracked source file. Running with different models will dirty the git tree.

Makefile
```diff
@@ -428,13 +428,29 @@ benchmark-standup: ## Stand up the benchmark environment (set BENCHMARK_NAMESPAC
...
+			"$(BENCHMARK_SCENARIOS_DIR)/$(BENCHMARK_WORKLOAD)"; \
+		rm -f "$(BENCHMARK_SCENARIOS_DIR)/$(BENCHMARK_WORKLOAD).bak"; \
+	fi
+	@if [ "$(BENCHMARK_DIRECT_KEDA)" = "true" ] && [ -f "$(BENCHMARK_SCENARIOS_DIR)/$(BENCHMARK_WORKLOAD)" ]; then \
+		echo "Injecting external model endpoint for direct-KEDA mode..."; \
+		sed -i.bak 's|base_url: .*|base_url: http://infra-llmdbench-inference-gateway.$(BENCHMARK_NAMESPACE).svc.cluster.local:80|' \
```
kahilam, line 445: same issue here also

Makefile
```diff
@@ -452,13 +468,26 @@ benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<name
...
 		cp "$(BENCHMARK_SCENARIOS_DIR)/$(BENCHMARK_WORKLOAD).in" \
 		   "$(BENCHMARK_REPO_DIR)/workload/profiles/$(BENCHMARK_HARNESS)/$(BENCHMARK_WORKLOAD)"; \
+	elif [ -f "$(BENCHMARK_SCENARIOS_DIR)/$(BENCHMARK_WORKLOAD)" ]; then \
+		echo "Copying local workload from $(BENCHMARK_SCENARIOS_DIR)/$(BENCHMARK_WORKLOAD) to harness..."; \
+		cp "$(BENCHMARK_SCENARIOS_DIR)/$(BENCHMARK_WORKLOAD)" \
+		   "$(BENCHMARK_REPO_DIR)/workload/profiles/$(BENCHMARK_HARNESS)/$(BENCHMARK_WORKLOAD)"; \
+		if [ -n "$(BENCHMARK_MODEL_ID)" ]; then \
+			echo "Injecting MODEL_ID=$(BENCHMARK_MODEL_ID) into workload profile..."; \
+			sed -i.bak 's|model_name: .*|model_name: $(BENCHMARK_MODEL_ID)|' \
+				"$(BENCHMARK_REPO_DIR)/workload/profiles/$(BENCHMARK_HARNESS)/$(BENCHMARK_WORKLOAD)"; \
```
kahilam: model_name is already injected in the source file at line 442. The first injection should be removed and only this one kept (after the copy).

Makefile
```diff
@@ -452,13 +468,26 @@ benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<name
...
+			rm -f "$(BENCHMARK_REPO_DIR)/workload/profiles/$(BENCHMARK_HARNESS)/$(BENCHMARK_WORKLOAD).bak"; \
+		fi; \
+		cp "$(BENCHMARK_REPO_DIR)/workload/profiles/$(BENCHMARK_HARNESS)/$(BENCHMARK_WORKLOAD)" \
+		   "$(BENCHMARK_REPO_DIR)/workload/profiles/$(BENCHMARK_HARNESS)/$(BENCHMARK_WORKLOAD).yaml"; \
```
kahilam: Having both the non-extension and .yaml copies in the profiles dir can cause the configmap to include the unrendered version. I hit this exact bug — the harness picked up the wrong file from the configmap.

Makefile
```diff
@@ -452,13 +468,26 @@ benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<name
...
 	fi
 	$(LLMDBENCHMARK) $(BENCHMARK_CLI_FLAGS) run \
 		-p $(BENCHMARK_NAMESPACE) \
 		-l $(BENCHMARK_HARNESS) \
 		-w $(BENCHMARK_WORKLOAD) \
 		$(if $(BENCHMARK_MODEL_ID),-m $(BENCHMARK_MODEL_ID),) \
-		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,)
+		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,) \
+		--wait-timeout $(BENCHMARK_WAIT_TIMEOUT)
```
kahilam, line 483: Missing default for BENCHMARK_WAIT_TIMEOUT. If this variable is not set, it will pass an empty value to the CLI and may break the command.

kahilam (commented):
@asm582,I ran into this exact issue yesterday while running symmetrical workload benchmarks. The local profile wasn't being found because the Makefile's copy logic expected $(BENCHMARK_WORKLOAD).in but the actual file was symmetrical.yaml.in. The harness also expects the .yaml extension. Will review the PR shortly.

Makefile
asm582: benchmark already runs with default, upto user to set this.

Makefile
asm582: This is the goal. This is cloned locally inside the repo root. This should not be used for code changes.

Makefile
asm582: Replied above

Makefile
```diff
@@ -425,7 +425,12 @@ benchmark-standup: ## Stand up the benchmark environment (set BENCHMARK_NAMESPAC
 		$(if $(filter true,$(BENCHMARK_MONITORING)),--monitoring,); \
 	rc=$$?; \
 	mv $(BENCHMARK_REPO_DIR)/config/scenarios/$(BENCHMARK_SPEC).yaml.bak \
-	   $(BENCHMARK_REPO_DIR)/config/scenarios/$(BENCHMARK_SPEC).yaml; \
+	   $(BENCHMARK_REPO_DIR)/config/scenarios/$(BENCHMARK_SPEC).yaml
```
kahilam: The \ was removed at line 428. So everything after it runs in a new shell that has no idea what $rc is. If standup fails, exit $rc does nothing and returns success.

Makefile
```diff
@@ -452,13 +463,24 @@ benchmark-run: ## Run a single benchmark workload (set BENCHMARK_NAMESPACE=<name
...
+			sed -i.bak 's|model_name: .*|model_name: $(BENCHMARK_MODEL_ID)|' \
+				"$(BENCHMARK_REPO_DIR)/workload/profiles/$(BENCHMARK_HARNESS)/$(BENCHMARK_WORKLOAD).yaml"; \
+			rm -f "$(BENCHMARK_REPO_DIR)/workload/profiles/$(BENCHMARK_HARNESS)/$(BENCHMARK_WORKLOAD).yaml.bak"; \
+		fi; \
 	fi
 	$(LLMDBENCHMARK) $(BENCHMARK_CLI_FLAGS) run \
 		-p $(BENCHMARK_NAMESPACE) \
 		-l $(BENCHMARK_HARNESS) \
-		-w $(BENCHMARK_WORKLOAD) \
+		-w $(BENCHMARK_WORKLOAD).yaml \
```
kahilam, line 480: Appending .yaml here breaks the .yaml.in path — that path copies the file without .yaml, so the CLI won't find it. This only works for the new elif (local workload) path.

kahilam (commented):
Two comments are added.

Makefile
asm582: Without appending .yaml, the inference-perf breaks.

Makefile
asm582: Resolved.

asm582:
> Two comments are added.

both comments are resolved.

kahilam (approved):

PR #1444, opened 2026-07-21
docs: propose metric-based analyzer interface

deanlorenz:
## Summary

Proposes a **metric-based analyzer interface** for WVA. Every analyzer produces, per finest-grain item, exactly two numbers — a **demand** `D` and a **target** `P` (per-replica capacity) — in a unit of the analyzer's own choosing, such that `D / P` is a replica count. The shape is deliberately symmetric with KEDA/HPA.

Two capabilities fall out of the same contract:

1. **WVA's own analyzer results become Prometheus metrics** that KEDA/HPA or dashboards can consume.
2. **External analyzers can be defined as plain PromQL** — WVA is extensible with no code change.

The optimizer's coordination logic and the actuation path are unchanged, and each ScaledObject remains owned by exactly one of {KEDA, WVA}.

## What's here

- New proposal doc: `docs/proposals/analyzer-metric-interface.md` (Status: Draft). No code changes.

This is a **proposal for discussion** (the "why" and "what"). Feedback is especially welcome on the `(demand, target)` contract shape, the analyzer-centric external-analyzer definition, and the KEDA-shaped framing.

### DIFF
--- a/docs/proposals/analyzer-metric-interface.md
+++ b/docs/proposals/analyzer-metric-interface.md
@@ -85,13 +85,16 @@ agree, because each analyzer's contribution is reduced to replicas before anythi
 
 ### Provenance label $E$
 
-An optional **error/provenance hint** $E$ records *how* a value was obtained (e.g. `direct`,
-`fallback`, `stale`, `fetch-failed`). $E$ is **observability only** — the engine and optimizer never
-branch on it. It is **set inside WVA** — by an internal analyzer (in Go) or, for an external analyzer,
-by the wrapper from collection state — and is **never read from an external query result** (a PromQL
-result cannot carry it). It surfaces in the engine's per-cycle logs and, if emitted as a metric, on a
-**separate** provenance series — not as a label on the `wva_analyzer_*` value series, whose label set
-stays stable (a flipping `e` would churn series).
+An optional **error/provenance hint** $E$ records *how* a value was obtained. It is a **free-form
+token chosen by the value's producer**: an internal analyzer's own reason code (WVA already carries
+such a per-result reason today), or, for an external analyzer, the **`e:` field of whichever target
+fallback succeeded** (e.g. `direct`, `fallback`, or a source name like `epp` / `vllm-rps`). WVA
+additionally reserves a few **health** tokens that the wrapper sets itself (`stale`, `fetch-failed`,
+`too-few-pods`). $E$ is **observability only** — the engine and optimizer never branch on it — and is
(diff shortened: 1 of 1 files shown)


deanlorenz:
Thanks @ev-shindin — the correctness fixes all look right and I've taken them as-is (the `last()` → `sum(max_over_time)` / `avg(avg_over_time)` reductions, bare-selector + `{{scope}}` for expressions, configurable `modelLabel`/`namespaceLabel` + escaping, `P ≤ 0` → missing, and moving provenance/description off the value series).

I pushed a small follow-up (`ff3e168b`) with a few refinements from an internal review:

- **`perRole` → a `match:` selector.** Each `demand`/`target` rule carries a `match` over ScaledObject attributes, applied uniformly to both, instead of keying by role name. Same intent as your `perRole` (still one query per role, since no metric carries `role`), just more general and symmetric — this is the one spot I diverged from your commit, so flagging it explicitly.
- **Grounded `role`** in the `llm-d.ai/role` pod-template label (`getRoleFromScaleTarget`), and noted it's *not* a metric label — which is why per-role queries are needed.
- **Provenance `E`** defined as a free-form producer token (internal = existing reason code; external = the winning fallback's `e:`), on the separate series you introduced; fixed a stale "records an `e` label" line to match.
- **Per-role demand** reconciled as utilization-space coordination (min over roles) — no cross-role demand normalization needed.
- **`orZero`** promoted from a comment to an explained option.

Happy to keep `perRole` if you'd rather — let me know.

ev-shindin (approved):
Thanks @deanlorenz ! As a proposal it is good enough, Please create issue/issues to achieve the goal.

docs/proposals/analyzer-metric-interface.md
```diff
@@ -0,0 +1,342 @@
...
+optimizer's job — per-analyzer weights live in its config, not in the contract.
+
+### Metric emission
+
+WVA emits the results of **every** analyzer it knows — internal and external alike — as Prometheus
+signals with a small, common label set. Metric names cannot contain dots, so the analyzer label is a
+**label**, not part of the name:
+
+```
+wva_analyzer_demand{analyzer, namespace, model, role?}          # per model instance (role only if role-specific)
```
lionelvillard, line 168: Can we have metric with meaningful names?

lionelvillard:
I think the scope is too broad for a single proposal. There are at least three distinct concerns here—the analyzer contract, extensibility, and exposing WVA's reasoning—and each deserves its own proposal.

Issue #1447, opened 2026-07-22
Priority-weighted rescale: redistribute the GPU budget under contention


ev-shindin:
### Enhancement Description

- **One-line enhancement description** (can be used as a release note): Priority-weighted **rescale** —
  under GPU contention, redistribute the whole budget by `priority × demand`, reclaiming capacity from
  lower-priority models (even if still hungry) so higher-priority work can run, instead of handing out
  only the free GPUs. Opt-in; unchanged when uncontended; reclaim never crosses a namespace quota or an
  accelerator type.
- **Enhancement Proposal**: llm-d/llm-d-workload-variant-autoscaler#1238 — `docs/proposals/design-rescale.md`
  (answers the *why* and *what*: today's additive fair-share freezes allocation by arrival order once
  the budget is full, so `priority` stops mattering and a saturated low-priority model is never
  reclaimed — a spiking production model can't scale up while an internal model keeps its GPUs).
- **PRs by stage and milestone**:
  - [x] Alpha - v0.xx — reclaim/fill to target behind the `rescale` flag; single budget; **V2 greedy
        path only**; water-filling targets (reserve `minReplicas` floors, split the rest by
        `priority × demand`); free-capacity-gated fill so the group is never transiently over-budget;
        reclaim bypasses scale-down damping. Requires an HPA `scaleDown.stabilizationWindowSeconds = 0`
        passthrough or direct actuation.
    - [x] Code (`llm-d/llm-d-workload-variant-autoscaler`) update PR(s):
    - [ ] Docs (`llm-d/llm-d`) update PR(s):
    - [ ] Guides (`llm-d/llm-d`) update PR(s):

  - [ ] Beta - v0.xx — observability (`Rescaled` status condition; per-model targets and the
        priorities/shares behind them; a signal for "grantee Pending pending an in-flight reclaim") and
        rescale-owned **hysteresis** (minimum share-gap and/or cool-down) against threshold oscillation.
    - [ ] Code (`llm-d/llm-d-workload-variant-autoscaler`) update PR(s):
    - [ ] Docs (`llm-d/llm-d`) update PR(s):
    - [ ] Guides (`llm-d/llm-d`) update PR(s):
  - [ ] Stable - v0.xx — **multiple budgets**: partition by `(accelerator type, namespace)` via
        `availableByNS` under the cluster ceiling; physical∧quota composition depends on #1003.
[...]

PR #1452, opened 2026-07-22
feat(rescale): priority-weighted GPU rescale under contention (Alpha)

ev-shindin:
Implements the **Alpha** stage of priority-weighted rescale for the V2 GPU-constrained optimizer.

Today the V2 optimizer only *adds* GPUs when they are free; under contention it cannot take capacity back from a low-priority model that is holding more than its fair share. This PR adds an opt-in pass that, when GPUs are contended, redistributes a competition group's whole budget by `priority × demand` — reclaiming from over-share (lower-priority) models and giving that capacity to starved higher-priority ones.

## Proposed Changes

- :gift: **Opt-in, off-by-default `enableRescale` flag** on the saturation-scaling config. Rescale runs only on the GPU-limited (`greedy-by-score`) path; with the flag off (or without the GPU limiter) behavior is unchanged.
- :gift: **Priority-weighted water-filling** (`computeRescaleTargets`): `minReplicas` floors reserved first, `share = floor + (budget − Σfloor)·weight/Σweight`, iteratively capped at `min(demand, maxReplicas)` with the freed excess re-split, largest-remainder integer rounding, and whole-replica quantization at each variant's `gpusPerReplica`.
- :gift: **Paced, never-over-budget convergence.** Reclaims free nothing the same cycle (usage is `CurrentReplicas`-based); fills are gated by the GPUs *physically* free this cycle, so a group is never over-budget. A namespace fill is bounded by `min(namespace quota free, cluster physical free)` and debits **both** budgets, so it can never over-subscribe the shared physical pool.
- :gift: **P/D (disaggregated) models**: the model's GPU target is split across prefill/decode by role demand with each role's floor reserved first; reclaim/fill run per role sharing one free-this-cycle counter.
- :gift: **Scope-coupled enablement**: rescale runs at scope *X* iff a budget exists at *X* **and** the flag is set at *X* — cluster flag from the global `default`; per-namespace flag from that namespace's **own** `default` only (never the global fallback), so the cluster flag cannot leak onto a namespace quota.
- :broom: Reclaim scale-down decisions are tagged `DecisionReasonRescale` so downstream stabilization does not damp them as noise. Bookkeeping is keyed by `(namespace, ModelID)`; group and intra-group processing order are deterministic.

**Deferred to Beta / follow-ups:** observability (a `Rescaled` status condition and surfacing a reclaim *stall* when a role can't shed a whole replica) and hysteresis (min share-gap / cool-down); the physical∧quota namespace budget partition (#1003); multi-accelerator models (variants spanning GPU types).

### Pre-review Checklist

- [ ] **E2E tests** for any new behavior — *Alpha scope:* covered by extensive unit + envtest specs (water-filling worked examples, reclaim/fill pacing, cluster-vs-namespace budget bounding and double-debit, `(namespace, ModelID)` isolation, scope-coupling matrix, P/D role split, `min`/`maxReplicas` and `gpusPerReplica>1` quantization, deterministic ordering, Conflict path). The engine-level wiring is a 2-line type-assert whose both sides are unit-tested; a full `engine.optimize` e2e (requires VA + metrics + inventory scaffolding) is deferred.
- [ ] **Docs PR** for any user-facing impact — feature is **off by default**; the `enableRescale` setting is documented inline in both saturation configmaps and in `docs/plans/engine/rescale-alpha.md`. User-facing guide/docs to land when the feature graduates from Alpha.
- [x] **Proposal PR** — design proposal `docs/proposals/design-rescale.md` (PR #1238).

**Release Note**

```release-note
Adds an opt-in, off-by-default `enableRescale` saturation-scaling setting. When enabled (alongside the GPU limiter), the autoscaler redistributes a competition group's GPU budget by model priority and demand under contention — reclaiming capacity from lower-priority models to serve higher-priority ones, without ever exceeding physical or namespace-quota capacity. No behavior change unless explicitly enabled.
```

### DIFF
--- a/config/base/manager/saturation-scaling-configmap.yaml
+++ b/config/base/manager/saturation-scaling-configmap.yaml
@@ -37,3 +37,11 @@ data:
     # Enable GPU limiter to constrain scaling based on available cluster resources
     # When true, scale-up decisions are limited by available GPU capacity
     enableLimiter: false
+    # Enable priority-weighted rescale: under GPU contention the V2 optimizer
+    # redistributes the whole budget by priority x demand, reclaiming from
+    # lower-priority models so higher-priority work can run. Off by default.
+    # Budget-scope flag (read only from this `default` entry): the value here
+    # governs the cluster budget; a namespace-local config's `default` governs
+    # that namespace's quota budget. Has no effect without a same-scope GPU budget
+    # (requires enableLimiter and a matching cluster/namespace budget).
+    enableRescale: false
--- a/deploy/configmap-saturation-scaling.yaml
+++ b/deploy/configmap-saturation-scaling.yaml
@@ -43,6 +43,13 @@ data:
     # Enable GPU limiter to constrain scaling based on available cluster resources
     # When true, scale-up decisions are limited by available GPU capacity
     enableLimiter: false
+    # Enable priority-weighted rescale: under GPU contention the V2 optimizer
+    # redistributes the whole budget by priority x demand, reclaiming from
+    # lower-priority models so higher-priority work can run. Off by default.
+    # Budget-scope flag (read only from this `default` entry): the value here
+    # governs the cluster budget; a namespace-local config's `default` governs
+    # that namespace's quota budget. Has no effect without a same-scope GPU budget.
+    enableRescale: false
 
   # Per-model overrides inherit the default's analyzer selection (V2) via
   # field-level merge, so they only need the thresholds they change. Under V2,
--- a/docs/plans/engine/rescale-alpha.md
+++ b/docs/plans/engine/rescale-alpha.md
@@ -1,0 +1,119 @@
+# Plan: Priority-Weighted Rescale — Alpha stage
+
+Implements the **Alpha** milestone of the rescale proposal (`docs/proposals/design-rescale.md`, PR #1238):
+a redistributive, priority-weighted allocation in the V2 GPU-constrained optimizer.
+
+## Goal
+
+Under contention, reallocate the **whole group budget** by each model's `priority × demand`, reclaiming
+GPUs from models holding more than their share (lower-priority, even if still hungry) so higher-priority
+work can run — instead of today's additive fair-share that hands out only *free* GPUs and freezes
+allocation by arrival order once the budget is full.
+
+## Scope
+
+**In Alpha**
+- Opt-in flag; behavior **unchanged when off, or when a group is uncontended**.
+- V2 `GreedyByScoreOptimizer` only (cost-aware/unlimited path untouched).
+- **All models, including P/D** — role-aware reclaim/fill (dependency #1237 *role-aware scale-down* is
+  merged; the machinery — `scaleDownRoleIterated`, `safeRemovalReplicasForRole`,
+  `allocateForModelPaired` — is in `main`).
+- Competition group = **`(accelerator type, budget scope)`** where budget scope is `cluster`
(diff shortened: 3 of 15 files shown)


deanlorenz (commented):
Nice mechanism — `computeRescaleTargets` reproduces the proposal's worked examples exactly, and the pacing/never-over-budget guarantees hold up (traced the double-debit and free-capacity-gating by hand). Test coverage is unusually thorough for an Alpha, and I confirmed by inspection that the off-flag and no-limiter paths are structurally unchanged.

Opt-in and off-by-default, so no behavior changes today, and nothing below is strictly blocking for an Alpha — but a few questions before it's on anywhere:

1. **`DecisionReasonRescale` — stabilization consumer coming later?** The comment says downstream stabilization must not damp a reclaim as noise, but nothing in-tree special-cases this reason yet (`gpurebalance`'s damping is a TODO and doesn't read `DecisionReason`). Today it only feeds the metrics label. Forward-looking, or is there a consumer I'm missing?

2. **Reclaim bypasses the multi-analyzer scale-down gate — intentional?** `reclaimRole` sheds replicas from a bound computed purely off the saturation water-fill delta — it skips `needsScaleDownForRole`/`safeRemovalReplicasForRole`, which `scaleDownRoleIterated` uses to require every analyzer's agreement before touching a role. So if a co-enabled analyzer (e.g. TA) reports a role fully loaded (`RoleSpare ≤ 0`) while saturation sees spare, the regular path leaves it alone but rescale reclaims anyway. Example: model at 6 GPUs, saturation-only demand implies a target of 2, TA reports the role fully loaded — rescale still sheds 4 GPUs from it. Is saturation-only reclaim deliberate for Alpha, or should it reuse the existing gate?

3. **Cross-scope fill order vs. the regular path's priority ordering.** `fairShareScaleUp` interleaves cluster- and namespace-scoped models in one priority-ordered loop; rescale processes cluster groups entirely before namespace groups, priority only breaking ties within a group. So under a same-type physical-pool oversubscription across scopes, a lower-priority cluster model could win scarce GPUs over a higher-priority namespace one. Looks like it only bites in the physical∧quota case already deferred to #1003 — just confirming that's the intended scope.

4. **P/D fill has no joint per-role throttle.** The regular path (`allocateForModelPaired`) throttles both roles to the same `deltaUtil` each iteration and never commits one without the other. Rescale's P/D fill instead processes roles sequentially off one shared `freeThisCycle` counter, so a scarce cycle can fill one role to 100% of target while its pair gets almost nothing. Example: prefill:decode demand 6000:2000 (true ratio 3:1), only 3 GPUs free → decode fills fully (2/2) before prefill's turn, landing at 1:2 — inverted for that cycle. Self-corrects over later cycles, but worth confirming the transient's acceptable for Alpha.

Minor: branch's a bit behind current `main` — worth a rebase before merge, no urgency since CI's green.

ev-shindin:
@deanlorenz  thanks. All points you are raised are valid and will be addressed in the beta version. Please look to https://github.com/llm-d/llm-d-workload-variant-autoscaler/issues/1447

deanlorenz (approved):
Thanks for the follow-up and the pointer to #1447 — good enough for Alpha given it's off-by-default. #1447 already tracks RC-1 (damping bypass, intentional for now) and RC-3 (physical∧quota partition, deferred to #1003). I'll follow up with you directly to make sure RC-2 (reclaim bypassing the multi-analyzer scale-down gate) and RC-4 (P/D fill's missing joint per-role throttle) don't get lost heading into Beta — they're captured in the PR thread either way. Approving.

PR #1453, opened 2026-07-23
docs(proposal): SLO-Driven Autoscaling via the llm-d Latency Predictor

ev-shindin:
## Proposed Changes

- :page_facing_up: Add a **Draft** design proposal — *SLO-Driven Autoscaling via the llm-d Latency Predictor* (`docs/proposals/slo-latency-predictor-autoscaling.md`).

**What it proposes:** a new opt-in `PredictiveSLOAnalyzer` that derives **per-variant** capacity from the `llm-d-latency-predictor` (predicted TTFT/TPOT at a percentile) instead of static, hand-tuned thresholds. It runs **behind the existing multi-analyzer combiner** with the static `saturation` analyzer as the floor (max-for-scale-up / unanimous-for-scale-down), so it is **non-regression by construction** — it can only add replicas or block a removal relative to today, never under-provision below static. Off by default. WVA reuses the predictor's models (trained online for the scheduler) and **owns no training** — it pulls each variant's pre-trained artifact and evaluates it locally.

**Framing for reviewers:** the proposal is deliberately conservative about the predictor's limits — it is least accurate near saturation (optimistically biased), so boundary safety is **non-regression vs. static, not absolute SLO safety**, and its value is **contingent on offline load-test seeding**. The recommended first step is a **narrow, non-actuating Phase-0 spike** — surface `num_request_running`, then measure predicted-vs-actual latency near the SLO knee (magnitude *and* sign of the optimism bias) **head-to-head against simpler per-variant baselines fit from the same data** — before any registry / multi-model / analyzer machinery is built. Open questions (predictor access topology, coverage-with-artifact export, per-variant model provenance, the R↔W coupling) are listed in the doc.

cc: @kaushik

### DIFF
--- a/docs/proposals/slo-latency-predictor-autoscaling.md
+++ b/docs/proposals/slo-latency-predictor-autoscaling.md
@@ -1,0 +1,109 @@
+# Proposal: SLO-Driven Autoscaling via the llm-d Latency Predictor
+
+**Authors:** [TBD]
+**Status:** Draft
+**Created:** 2026-07-23
+**Last Updated:** 2026-07-23
+
+---
+
+## Problem Statement
+
+WVA sizes a model by asking an **analyzer** for each variant's capacity. The default `saturation` (V2) analyzer models a per-replica token capacity `min(k1, k2)` from **static, hand-tuned, workload-/hardware-agnostic** thresholds (`KvCacheThreshold`, `QueueLengthThreshold`) that are only *indirectly* tied to any latency SLO — there is no principled way to set them per workload or per TTFT/TPOT percentile.
+
+Meanwhile the inference scheduler (EPP) already runs the **`llm-d-latency-predictor`**: a served ML model, trained online from live traffic, mapping a `(load, request)` feature vector to **p90 TTFT/TPOT**, used to route requests. This proposal reuses it to derive autoscaling capacity anchored to *measured, per-workload, percentile* latency — including prefix-cache and learned batch/memory effects a static model cannot express.
+
+> **Honest framing (load-bearing).** The predictor is trained on *healthy* traffic, so it is **least accurate near saturation** — exactly where the autoscaler operates — and its errors there are **optimistic** (under-predict latency → under-provision → miss SLO). The design therefore: (a) trusts the predictor **only inside its data support**; (b) always runs it **behind the multi-analyzer safety combiner** with static `saturation` as the floor; (c) adds a runtime guardrail. **Boundary safety is *non-regression* (never worse than static), not absolute.** And **pre-seeding, this analyzer ≈ "static `saturation` + a sanity check"** — it earns value only once offline load-test seeding pushes support out to the knee.
+
+## Goals
+
+- Derive **per-variant** capacity (roles via `pod_type`) anchored to predicted TTFT/TPOT (mean or percentile), **inside the predictor's data support**.
+- Emit the standard `interfaces.AnalyzerResult`; optimizer/limiter/rescale/scale-from-zero paths unchanged.
+- Fail safe to static `saturation` when the predictor is cold, unavailable, out-of-support, or confidently wrong.
+
+## Non-Goals
+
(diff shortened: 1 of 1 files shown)


kaushikmitr:
Thanks @ev-shindin, reviewed the proposal. Overall I think the core approach will work: using the predictor in reverse (sweeping load to find where predicted latency crosses the SLO, and sizing capacity from that point) is sound, and the design is safe by construction since the static analyzer always acts as a floor via the combiner. I verified the predictor-side claims against the `llm-d-latency-predictor` code and the combiner claims against this repo; they hold up. A few suggestions below.

## Main suggestion: make remote prediction the Phase-1 default, not local artifact eval

The doc treats "query a per-variant prediction Service over HTTP" as the fallback alternative. I would flip that. Querying the prediction server directly is simpler and removes three problems the local-eval path has:

1. **No artifact pipeline needed.** For what it's worth, the export path already exists if we do want it later: the training server has `/model/export` (merged in llm-d-latency-predictor #34), which bundles all model artifacts plus a `metadata.json` in one tarball, so model and metadata are atomic by construction. The missing piece would only be per-bucket coverage in the metadata (today it carries aggregate sample counts). There are also `/model/{name}/download`, `/models/list`, and `/model/download/info`. So Requirements (i)/(ii) are a small delta on existing endpoints, not new machinery. But with remote prediction none of this is needed for Phase 1.

2. **No ensemble mismatch.** The predictor has a gated ensemble mode (`QueueGatedModel`, surfaced as `ensemble_active` in `/status`). When active, served predictions route through gated wrappers around noqueue/queued sub-models, and `/model/export` deliberately excludes the gate. A WVA evaluating exported base models locally would see different predictions than the scheduler does, and they would disagree exactly under queueing, where autoscaling matters most. This could also spuriously trip the divergence-disable guardrail. Remote prediction evaluates whatever the server actually serves, so WVA and routing always agree.

3. **Schema drift is handled by the API contract.** The predictor's feature set is evolving (e.g. new flag-gated multimodal encoder features). With local eval, WVA builds feature vectors itself and any drift between the artifact's training schema and WVA's featurization mispredicts silently; the artifact would need a feature-schema version and WVA should abstain on mismatch. With remote prediction, new features have server-side defaults and inputs are validated, so an older client keeps working.

The cost side is manageable:

- **Sweep cost is a non-issue.** The server already has `/predict/bulk` (up to 10k requests per call), so an entire grid sweep per variant is one HTTP round trip, at a control-loop cadence of minutes.
- **The network dependency fails safe.** Predictor unreachable means abstain, and static governs. That is already the designed behavior.
- **Topology work is plumbing.** Today the predictor is a pod-local EPP sidecar, so it needs a per-variant Micro Service (which is something we will do anyway once we go to active-active epp model). The prediction server is a stateless serving layer that syncs models from the training server, so it can be scaled out independently without touching the scheduler's sidecars.


If local eval stays in the doc as a later optimization, points 1 to 3 above become its prerequisites (coverage in export metadata, ensemble in the artifact contract or `ensemble_active == false` required, and a feature-schema version with abstain-on-mismatch), and I would put them in Phase 0 next to the coverage blocker.

## Implementation-stage tuning (not blockers)

These can be tightened during Phase 0/1, but worth calling out in the doc:

- **Sweep cadence and stability.** Capacity R* will shift with every workload change and model refresh. Some smoothing or hysteresis is needed to avoid replica churn, and abstain-boundary jitter can make the analyzer blink in and out of the combiner, which toggles scale-down gating.
- **Request shape for the sweep.** A fixed reference request (stable, well-defined SLO contract, matches what WVA can measure today) versus one that tracks the live traffic mix (adapts to drift, but noisier and needs token histograms WVA does not have yet). A fixed shape plus a drift alarm is probably the right start, with mix-tracking in Phase 2 alongside histograms.
- **The feedback loop.** Scaling decisions change per-replica load, which changes what the predictor trains on, which changes the next sweep (the R-W coupling already listed in the open questions). Needs empirical watching; not solvable on paper.

PR #1464, opened 2026-07-24
fix(metrics): evict stale replica-gauge series on variant removal

### LINKED ISSUE #1459: fix(metrics): replica gauges and accel-tracking map never evicted on variant removal — stale series report dead variants
## Summary

When a variant is removed (its managed scaler / annotation deleted), nothing evicts its metric state. Two leaks result, and one produces stale, incorrect observability:

- **File:** `internal/metrics/metrics.go`
- **Exported gauges:** `wva_current_replicas`, `wva_desired_replicas`, `wva_desired_ratio`

Discovery is annotation-driven (the HPA / ScaledObject reconcilers); there is no VA delete handler and no metrics-cleanup hook. When a variant leaves the discovered set, `EmitReplicaMetrics` simply stops being called for it and its state is orphaned.

## What leaks

1. **Exported gauge series.** `currentReplicas` / `desiredReplicas` / `desiredRatio` (`metrics.go:138-158`) accumulate one series per `(variant_name, namespace, accelerator_type, controller_instance)`. The only `.Delete()` calls are for accelerator-label *transitions* inside `EmitReplicaMetrics` (`metrics.go:602-607`). On variant removal there is no `.Delete()`, so the series are orphaned.
2. **Accelerator-tracking map.** `replicaSeriesAccel` (`metrics.go:31-34`) keeps one entry per `(variant, namespace)` ever seen and is never pruned (already flagged by the TODO at `metrics.go:26`).

## Impact

- **Stale / incorrect observability (primary).** A Prometheus `GaugeVec` keeps exporting a series on every `/metrics` scrape until `.Delete()` is called. A removed variant therefore keeps reporting its **last** `wva_current_replicas` / `wva_desired_replicas` value indefinitely, so dashboards and alerts see a dead variant as if it were still live.
- **The in-code comment is inaccurate.** The comment at `metrics.go:28-30` claims the orphaned series "expire via Prometheus staleness." They do not: Prometheus only inserts a staleness marker when a series *disappears* from a scrape, which never happens while WVA keeps exporting it. The only cleanup today is a controller process restart (which rebuilds the registry).
- **Unbounded growth (secondary).** Exported series and map entries grow with every distinct `(variant, namespace, accelerator)` ever seen. Churny clusters accumulate stale series, increasing the `/metrics` payload size and scrape cost over the controller's lifetime.

## Suggested fix

- Add a removal hook: when a variant leaves the discovered set (or its scaler is deleted), `.Delete()` its `currentReplicas` / `desiredReplicas` / `desiredRatio` series for every label combination and drop its `replicaSeriesAccel` entry (under `replicaSeriesMu`). The HPA / ScaledObject reconcilers' delete events are the natural trigger.
- Correct the misleading "expire via Prometheus staleness" comment.

## Acceptance criteria

- [ ] Removing a managed variant / scaler evicts its `wva_current_replicas`, `wva_desired_replicas`, and `wva_desired_ratio` series from the `/metrics` endpoint (verified: series absent on the next scrape after deletion).
- [ ] The corresponding `replicaSeriesAccel` entry is pruned on removal (map does not grow across create/delete churn).
- [ ] The inaccurate staleness comment at `metrics.go:28-30` is corrected.
- [ ] Unit coverage for the eviction path (delete → series gone, map entry gone).

## Notes

Related to #1072 (associate metrics to variants via Kubernetes labels).

Goutham-Annem:
Fixes #1459

## Summary

When a managed HPA or ScaledObject was deleted or de-annotated, the
three replica scaling gauges (`wva_current_replicas`, `wva_desired_replicas`,
`wva_desired_ratio`) and the internal accelerator-tracking map entry were
never cleaned up. The controller kept exporting the last-seen values for
the removed variant on every `/metrics` scrape — not via Prometheus
staleness (which does not apply to live GaugeVec series), but as
permanently orphaned series until the controller process restarted.

## Changes

- **`internal/metrics/metrics.go`**
  - Add `MetricsEmitter.DeleteReplicaMetrics(variantName, namespace string)`:
    deletes all three replica-gauge series for the variant and prunes its
    `replicaSeriesAccel` map entry under the mutex.
  - Correct the misleading comment that claimed series "expire via Prometheus
    staleness" — they do not while the controller is alive.

- **`internal/controller/hpa_reconciler.go`**
  - Add `MetricsEmitter *metrics.MetricsEmitter` field.
  - Call `DeleteReplicaMetrics` on the `IsNotFound` path (HPA gone) and on
    the de-annotation / deletion-timestamp path.

- **`internal/controller/scaledobject_reconciler.go`**
  - Same changes as HPAReconciler for the ScaledObject code path.

- **`cmd/main.go`**
  - Retain the `MetricsEmitter` instance and pass it to both reconcilers.

- **`internal/metrics/replica_metrics_test.go`**
  - 5 new Ginkgo specs under `Describe("DeleteReplicaMetrics")`:
    - evicts all three gauge series for a removed variant
    - prunes the accelerator-tracking map entry
    - does not evict a different VA's series in the same namespace
    - is a no-op when `EmitReplicaMetrics` was never called
    - evicts series when the accelerator label was unresolved at emit time

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/...` — all 26 Ginkgo specs in the metrics suite pass (including 5 new `DeleteReplicaMetrics` specs)
- [x] `go test ./...` (unit) — all packages green

### DIFF
--- a/internal/controller/hpa_reconciler.go
+++ b/internal/controller/hpa_reconciler.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/annotations"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/datastore"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/metrics"
 )
 
 // HPAReconciler tracks namespaces for annotation-based WVA discovery via HPA objects.
@@ -34,7 +35,8 @@ import (
 // loop can scope its List calls to namespaces that contain managed scalers.
 type HPAReconciler struct {
 	client.Client
-	Datastore datastore.Datastore
+	Datastore      datastore.Datastore
+	MetricsEmitter *metrics.MetricsEmitter
 }
 
 // +kubebuilder:rbac:groups=autoscaling,resources=horizontalpodautoscalers,verbs=get;list;watch
@@ -48,6 +50,7 @@ func (r *HPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	if err := r.Get(ctx, req.NamespacedName, hpa); err != nil {
 		if apierrors.IsNotFound(err) {
 			r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+			r.MetricsEmitter.DeleteReplicaMetrics(req.Name, req.Namespace)
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, err
@@ -55,6 +58,7 @@ func (r *HPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 
 	if !hpa.DeletionTimestamp.IsZero() || !annotations.IsManaged(hpa) {
 		r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+		r.MetricsEmitter.DeleteReplicaMetrics(req.Name, req.Namespace)
 	} else {
 		r.Datastore.NamespaceTrack("AnnotatedScaler", req.Name, req.Namespace)
 	}
--- a/internal/controller/hpa_reconciler.go
+++ b/internal/controller/hpa_reconciler.go
@@ -50,7 +50,7 @@ func (r *HPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	if err := r.Get(ctx, req.NamespacedName, hpa); err != nil {
 		if apierrors.IsNotFound(err) {
 			r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
-			r.MetricsEmitter.DeleteReplicaMetrics(req.Name, req.Namespace)
+			r.MetricsEmitter.DeleteReplicaMetrics(ctx, req.Name, req.Namespace)
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, err
@@ -58,7 +58,7 @@ func (r *HPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 
 	if !hpa.DeletionTimestamp.IsZero() || !annotations.IsManaged(hpa) {
 		r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
-		r.MetricsEmitter.DeleteReplicaMetrics(req.Name, req.Namespace)
+		r.MetricsEmitter.DeleteReplicaMetrics(ctx, req.Name, req.Namespace)
 	} else {
 		r.Datastore.NamespaceTrack("AnnotatedScaler", req.Name, req.Namespace)
 	}
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -579,8 +579,14 @@ func (m *MetricsEmitter) EmitReplicaMetrics(ctx context.Context, va *llmdOptv1al
 		return l
 	}
 
-	// Set the current series first so a concurrent scrape never sees a gap.
+	// Update the tracking map and Set the Prometheus series under the same lock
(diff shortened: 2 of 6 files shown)


internal/controller/hpa_reconciler.go
```diff
@@ -48,13 +50,15 @@ func (r *HPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	if err := r.Get(ctx, req.NamespacedName, hpa); err != nil {
 		if apierrors.IsNotFound(err) {
 			r.Datastore.NamespaceUntrack("AnnotatedScaler", req.Name, req.Namespace)
+			r.MetricsEmitter.DeleteReplicaMetrics(req.Name, req.Namespace)
```
ev-shindin: Emit and delete now live in two independent goroutines with no coordination, which opens a
re-leak window. The engine's `optimize()` snapshots `activeVAs`
(`internal/engines/saturation/engine.go:376`) and only emits much later
(`engine.go:1609`), after Prometheus queries + optimization. If a scaler is deleted during that
window - listed before the cache delete, emitted after this reconciler ran - `EmitReplicaMetrics`
re-creates the series. Since the object is gone, no further reconcile event fires, so nothing
ever deletes it again -> the stale series persists until process restart, which is exactly the
bug #1459 targets. The window is ~one optimize cycle wide, so it's reachable in practice.

Suggest a belt-and-suspenders sweep in the optimize loop (diff the discovered `vaMap`
against the previous cycle and `DeleteReplicaMetrics` the vanished keys — self-heals within one
cycle)

internal/metrics/metrics.go
```diff
@@ -608,6 +607,41 @@ func (m *MetricsEmitter) EmitReplicaMetrics(ctx context.Context, va *llmdOptv1al
...
 
+// DeleteReplicaMetrics removes all replica-gauge series for the named variant
+// from the /metrics endpoint and prunes its entry from the accelerator-tracking
+// map. Call this when a variant's managed scaler (HPA or ScaledObject) is
+// deleted or de-annotated so that stale series are evicted immediately rather
+// than persisting until a controller restart.
+//
+// If EmitReplicaMetrics was never called for this variant (no entry in the
+// tracking map), this is a no-op.
+func (m *MetricsEmitter) DeleteReplicaMetrics(variantName, namespace string) {
```
ev-shindin: `DeleteReplicaMetrics` doesn't take a `ctx`, unlike the sibling emitter methods
(`EmitReplicaMetrics`, `RecordSaturationMetrics`). Not a correctness issue, but for consistency
(and future logging of what was evicted) consider `DeleteReplicaMetrics(ctx, variantName,
namespace)`. The nil-guard on the package gauges and the nil-safe receiver (method never derefs
`m`) are both good — no panic risk from an unset `MetricsEmitter`.

internal/controller/hpa_reconciler.go
ev-shindin: The five new unit tests exercise `DeleteReplicaMetrics` in isolation thoroughly, but nothing
tests the reconciler wiring — that a delete/de-annotate actually calls it with identifiers that
match the emitted series. Given the `va.Name == hpa.Name` coupling this fix silently depends on,
an envtest (annotate an HPA -> let a series emit -> delete the HPA -> assert the series is gone)
would guard the part most likely to regress under a future refactor of VA-name synthesis.

internal/metrics/replica_metrics_test.go
```diff
@@ -142,3 +142,93 @@ var _ = Describe("EmitReplicaMetrics", func() {
...
+		ctx      context.Context
+	)
+
+	BeforeEach(func() {
+		resetMetrics()
+		registry = prometheus.NewRegistry()
+		Expect(InitMetrics(registry)).To(Succeed())
+		emitter = NewMetricsEmitter()
+		ctx = context.Background()
+		_ = ctx // used by EmitReplicaMetrics
```
ev-shindin: nit: This blank-assignment/comment is unnecessary; can be dropped.

ev-shindin (changes requested):
Thanks @Goutham-Annem ! Please address review comments.

PR #1469, opened 2026-07-26
fix(saturation_v2): count DP-rank instances, not pods, in aggregateByVariant

### LINKED ISSUE #1467: Saturation V2: DP>1 capacity is undercounted because aggregation uses pod count, not instance count
## Summary

In `aggregateByVariant()` (`internal/engines/analyzers/saturation_v2/analyzer.go`),
`TotalCapacity` and `ReplicaCount` are computed from the **pod count**
(`readyCount`), but `perReplicaCapacity` is a **per-instance** value.

For a data-parallel (DP>1) vLLM deployment, one pod runs N engine instances,
each with its own port and KV-cache pool. So multiplying a per-instance capacity
by the pod count makes `TotalCapacity` N times too small.

```go
// origin/main
readyCount := vs.CurrentReplicas - vs.PendingReplicas        // pod count
perReplicaCapacity = float64(median(capacities))             // per-instance
totalCapacity := float64(readyCount) * perReplicaCapacity    // wrong unit
ReplicaCount:   readyCount,                                  // wrong unit
```

## Impact

`Utilization = TotalDemand / TotalCapacity` becomes N times too high. This
triggers scale-up that is not needed.

Example (production DP=8 decode deployment): 7 ready pods = 56 instances.

- Before: `TotalCapacity = 7 × perReplica` → `Utilization ≈ 5.89` (impossible; utilization is at most 1.0).
- After: `TotalCapacity = 56 × perReplica` → `Utilization ≈ 0.74` (below the 0.85 scale-up threshold, so no scale-up — which matches reality).

## Notes

- This is the aggregation-layer follow-up to #1121, which added per-instance
  metric keying at the collector layer but did not update this multiplier.
- The fix depends on Prometheus scraping every DP-rank port (one `instance`
  series per port), which needs a PodMonitor, not a single-port ServiceMonitor.

go-bai:
## Summary

In `aggregateByVariant()` (`internal/engines/analyzers/saturation_v2/analyzer.go`), `TotalCapacity` and `ReplicaCount` were computed from the **pod count** (`readyCount`), but `perReplicaCapacity` is a **per-instance** value.

For a data-parallel (DP>1) vLLM deployment, one pod runs N engine instances, each with its own port and KV-cache pool. Multiplying a per-instance capacity by the pod count makes `TotalCapacity` N times too small.

## Impact

`Utilization = TotalDemand / TotalCapacity` becomes N times too high, triggering scale-up that is not needed.

Example (DP=8 decode deployment): 7 ready pods = 56 instances.

- Before: `TotalCapacity = 7 × perReplica` → `Utilization ≈ 5.89` (impossible; utilization is at most 1.0).
- After: `TotalCapacity = 56 × perReplica` → `Utilization ≈ 0.74` (below the 0.85 scale-up threshold, so no scale-up — which matches reality).

## Fix

Scale `TotalCapacity` and `ReplicaCount` by `len(replicas)` (live per-instance records) instead of `readyCount`, mirroring the throughput analyzer's `computeVariantSupply`/`nKV` pattern.

Both fields must change: `aggregation.go` ignores the `TotalCapacity` field and recomputes totals as `ReplicaCount × PerReplicaCapacity` (in `SumTotalSupply`, `SumTotalAnticipatedSupply`, `AggregateByRole`), so fixing `TotalCapacity` alone would still leave the wrong value downstream.

This is the aggregation-layer follow-up to #1121, which added per-instance metric keying at the collector layer but did not update this multiplier.

Fixes #1467

### DIFF
--- a/internal/engines/analyzers/saturation_v2/analyzer.go
+++ b/internal/engines/analyzers/saturation_v2/analyzer.go
@@ -368,6 +368,9 @@ func (a *SaturationAnalyzer) aggregateByVariant(
 		// supply totals — so both must stay in sync. Defaults to readyCount (pod
 		// count) for the fallback branches, which have no per-instance data.
 		replicaCount := readyCount
+		// pendingCount must match replicaCount's unit (SumTotalAnticipatedSupply
+		// adds them); converted to instances below when replicaCount switches.
+		pendingCount := vs.PendingReplicas
 
 		var capacityLabel string
 		if len(replicas) > 0 {
@@ -381,6 +384,22 @@ func (a *SaturationAnalyzer) aggregateByVariant(
 			}
 			perReplicaCapacity = float64(median(capacities))
 			replicaCount = len(replicas)
+			// Convert pending to instance units. readyCount and vs.PendingReplicas
+			// share the scale-target unit (pods for Deployment, groups for LWS), so
+			// len(replicas)/readyCount approximates the instances-per-unit (DP)
+			// factor. This is exact only in steady state: len(replicas) is live
+			// metrics while readyCount is (lagging) scale-target status, so during a
+			// scale-up — when new instances report metrics before their unit is
+			// counted ready — the ratio is inflated and pendingCount overshoots
+			// (understating RequiredCapacity). A metrics-derived instances-per-unit
+			// (grouping by pod/LWS group) would remove this skew; tracked as a
+			// follow-up. TODO: scale-from-zero (readyCount == 0) can't infer DP at
+			// all and is left unconverted here.
+			if readyCount > 0 {
+				if instancesPerUnit := len(replicas) / readyCount; instancesPerUnit > 1 {
+					pendingCount = vs.PendingReplicas * instancesPerUnit
+				}
+			}
 			if accelerator == "" {
 				accelerator = replicas[0].AcceleratorName
 			}
@@ -411,7 +430,7 @@ func (a *SaturationAnalyzer) aggregateByVariant(
 			Cost:               cost,
 			Role:               vs.Role,
 			ReplicaCount:       replicaCount,
-			PendingReplicas:    vs.PendingReplicas,
+			PendingReplicas:    pendingCount,
 			PerReplicaCapacity: perReplicaCapacity,
 			TotalCapacity:      totalCapacity,
 			TotalDemand:        totalDemand,
--- a/internal/engines/analyzers/saturation_v2/analyzer.go
+++ b/internal/engines/analyzers/saturation_v2/analyzer.go
@@ -363,15 +363,24 @@ func (a *SaturationAnalyzer) aggregateByVariant(
 			readyCount = 0
 		}
 
+		// replicaCount multiplies perReplicaCapacity to form TotalCapacity. It also
+		// sets VariantCapacity.ReplicaCount, which aggregation.go uses to recompute
+		// supply totals — so both must stay in sync. Defaults to readyCount (pod
+		// count) for the fallback branches, which have no per-instance data.
+		replicaCount := readyCount
+
 		var capacityLabel string
 		if len(replicas) > 0 {
-			// Use median effective capacity from ready pods
+			// len(replicas) counts vLLM engine instances (DP ranks), not pods: a DP=8
+			// pod hosts 8 independently-capacitied instances. Using the pod count would
+			// undercount TotalCapacity by the DP factor.
 			capacities := make([]int64, 0, len(replicas))
 			for _, rc := range replicas {
 				capacities = append(capacities, rc.EffectiveCapacity)
 				totalDemand += float64(rc.ReplicaDemand)
 			}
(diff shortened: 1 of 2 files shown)


ev-shindin (changes requested):
Thanks @go-bai! Nice catch - the `len(replicas)` switch for `ReplicaCount`/`TotalCapacity` is correct and fixes the utilization inflation (the impossible `5.89 → 0.74` case). Demand is already summed over `len(replicas)` scraped instances, so this brings capacity into the same unit. 👍

One follow-up before this fully delivers "count DP instances, not pods": the **anticipated-supply path is still in pod units for pending replicas**.

`SumTotalAnticipatedSupply` computes:

```go
total += float64(vc.ReplicaCount + vc.PendingReplicas) * vc.PerReplicaCapacity
```

This PR fixes `ReplicaCount` → instances, but leaves `PendingReplicas = vs.PendingReplicas`, which is a **pod** count (`domain/saturation_analyzer.go`: `CurrentReplicas − ReadyReplicas`). So the formula now adds **instances + pods**, which is the same unit mismatch the PR set out to remove — just moved to the pending term.

Concrete effect for DP=8 with 7 ready + 2 pending pods:

- Anticipated supply = `(56 + 2) × PRC` instead of the correct `(56 + 16) × PRC`.
- Since the engine post-step does `requiredCapacity = max(0, demand/scaleUpThreshold − anticipatedSupply)`, undercounting the 16 in-flight instances as 2 makes WVA think far less capacity is arriving than really is → **redundant scale-ups during an already-in-flight scale-up**, and weaker scale-down flap suppression.

Less severe than the bug being fixed (transient, errs toward over-capacity), but it's the same class of issue and directly undercuts the PR's stated goal.

Since `readyCount` is ready **pods** and `len(replicas)` is ready **instances**, we can convert pending into the same unit using the observed ratio:

```go
replicaCount := readyCount
pendingCount := vs.PendingReplicas
if len(replicas) > 0 {
    replicaCount = len(replicas)
    // Convert pending pods to DP-instance units so anticipated supply
    // stays consistent with ReplicaCount. readyCount is ready pods;
    // len(replicas) is ready instances (DP ranks).
    if readyCount > 0 {
        if ipp := len(replicas) / readyCount; ipp > 1 {
            pendingCount = vs.PendingReplicas * ipp
        }
    }
    // ... existing median/demand logic ...
}
// ...
ReplicaCount:    replicaCount,
PendingReplicas: pendingCount,
```

Integer division is exact in steady state and conservatively rounds down under partial scraping. The `readyCount == 0` scale-from-zero case can't infer DP from live metrics — worth a `// TODO` to source instances-per-pod from the workload spec (LWS group size / vLLM `--data-parallel-size`) rather than inferring. That same explicit-DP signal would also let us fix the identical gap in the throughput analyzer (`ReplicaCount = nKV`, `PendingReplicas` = pods), which shares this limitation today.

Could you also add a **DP>1 test with pending replicas** asserting `TotalAnticipatedSupply`? The current coverage looks like it only exercises the ready-instance path.

go-bai:
Thanks @ev-shindin for the careful review. You're right — switching `ReplicaCount` to instances while leaving `PendingReplicas` in pod units just relocates the same unit mismatch to the anticipated-supply term.

Fixed in `b6d360e`: `PendingReplicas` is now converted to instance units using `len(replicas) / readyCount`. Both `readyCount` and `PendingReplicas` come from the same scale-target status, so the ratio is instances-per-unit — this stays correct for LWS as well, where the unit is a group rather than a pod. Added a DP>1 test with a pending replica asserting `TotalAnticipatedSupply` (confirmed it fails against the pre-fix `vs.PendingReplicas`).

One caveat I documented inline: the ratio is exact only in steady state. `len(replicas)` is live metrics while `readyCount` is lagging status, so **during a scale-up** — when new instances report metrics before their unit is counted ready — the ratio inflates and `pendingCount` overshoots, understating `RequiredCapacity`. (LWS example, per-group DP=8, scaling 1→2 groups: mid-transition `len(replicas)=16`, `readyCount=1` → ratio 16 instead of 8.) So the note about rounding down conservatively holds for partial scraping, but not for this transient.

On the follow-ups you raised — plus that transient — I filed #1475: derive instances-per-unit by grouping live metrics per pod / LWS group (numerator and denominator from the same snapshot, which removes the skew and also covers "reporting but not ready" at cold start), with a spec-derived fallback only for fully cold pending units. That issue also covers the identical `nKV` / pod-unit `PendingReplicas` gap in the throughput analyzer, and notes the DP parsing that doesn't exist yet (`EngineParams` has TP but no DP; needs vLLM `--data-parallel-size` and SGLang `--dp-size`).

Would it be OK to keep this PR scoped to saturation_v2 and handle those in #1475, since they share the same signal? Happy to fold the throughput fix in here instead if you'd prefer.

ev-shindin (approved):
LGTM Thanks @go-bai ! Please work on #1475. I will aslo fold another related issue.

PR #1470, opened 2026-07-26
fix(saturation-v2): charge waiting requests by P/D role

### LINKED ISSUE #1456: Saturation V2: per-replica demand ignores output tokens for waiting requests on decode/"both" replicas
## Summary

In the Saturation V2 analyzer, the per-replica demand estimate charges **only input
(prompt) tokens** for waiting (queued) requests, regardless of the replica's P/D role:

```go
// internal/engines/analyzers/saturation_v2/analyzer.go:158-161 (main path)
replicaDemand := rm.TokensInUse
if rm.AvgInputTokens > 0 {
    replicaDemand += int64(rm.QueueLength) * int64(rm.AvgInputTokens)
}
```

The same input-only term appears in the fallback path
(`computeReplicaCapacityFallback`, `analyzer.go:249-251`).

For **decode** and **"both"** replicas, a waiting request's KV footprint is dominated by
the output tokens it will generate, not just its prompt. Charging input-only under-counts
decode demand and can delay scale-up for decode-heavy / long-generation workloads.

## Current behavior

The waiting-queue demand term is `QueueLength × AvgInputTokens` in both the main and
fallback per-replica paths, applied uniformly to every replica irrespective of role.

Note this is **inconsistent with the sibling estimator** in the same file. The
model-level scheduler-queue demand is already role-aware and *does* count output tokens for
decode/both:

```go
// internal/engines/analyzers/saturation_v2/analyzer.go:634-642 (estimateSchedulerQueueDemand)
switch role {
case "prefill":
    byRole["prefill"] = inputTokens
case "decode":
    byRole["decode"] = inputTokens + outputTokens
default: // "both" or unknown
    byRole[role] = total // inputTokens + outputTokens
}
```

So the codebase already establishes the intended semantics — prefill charges input only;
decode/both charge input **and** output — but the per-replica local waiting-queue term does
not follow it.

## Why this is wrong

- **KV demand on a decode replica is output-driven.** Once a queued request is admitted, its
  KV grows toward `prefill + OL`. Modeling it as `AvgInputTokens` alone ignores the decode
  KV growth that is the entire point of a decode replica.
- **Under-provisioning for long-output workloads.** A deep queue of long-generation requests
  looks far cheaper than it will actually be, so V2 can under-count demand exactly when a
  decode backlog is building.
- **Internal inconsistency.** For the same "queued request KV cost" the repo currently has
  three different answers:

  | Estimator | per-queued-request KV charged |
  |---|---|
  | Per-replica local queue (`analyzer.go:160`) | `IL` |
  | Model-level scheduler queue (`estimateSchedulerQueueDemand`) | `IL + OL` (decode/both) |
  | Throughput analyzer `KVreq` | `IL + OL/2` |

  The per-replica term is the outlier and the least physically accurate.

Note: output tokens of **running** requests are already accounted for — `TokensInUse`
(`= KvCacheUsage × TotalKvCapacityTokens`) is measured occupancy and includes generated
tokens' KV. The gap is only in the **waiting** term.

## Proposed change

Make the per-replica waiting-queue demand role-aware, mirroring
`estimateSchedulerQueueDemand`:

- **prefill** role: `QueueLength × AvgInputTokens` (unchanged)
- **decode / both** role: include output tokens

```go
// role-aware waiting-queue demand
queuedInput  := float64(rm.QueueLength) * rm.AvgInputTokens
queuedOutput := float64(rm.QueueLength) * rm.AvgOutputTokens

switch role {
case interfaces.RolePrefill:
    replicaDemand += int64(queuedInput)
default: // RoleDecode, RoleBoth, "" (non-disaggregated)
    replicaDemand += int64(queuedInput + queuedOutput)
}
```

Apply the same treatment in `computeReplicaCapacityFallback`.

### Open decision: full `OL` vs time-averaged `OL/2`

- `IL + OL` matches the sibling `estimateSchedulerQueueDemand` and is the more conservative
  (OOM-safe) choice for a saturation/capacity analyzer. **Recommended for consistency.**
- `IL + OL/2` matches the throughput analyzer's `KVreq` (time-averaged footprint) and is the
  physically principled steady-state value, but less conservative.

Recommendation: use `IL + OL` for decode/both to keep the two waiting-queue estimators in
this file consistent; revisit unifying all three on `IL + OL/2` as a separate cleanup.

## Implementation notes

- **Role is not on `ReplicaMetrics`.** It lives on `interfaces.VariantReplicaState.Role`
  (`"prefill"`, `"decode"`, `"both"`, or `""`). Plumb it into the per-replica computation the
  same way `gpuCount` is: build a `rolesByVariant` map from `input.VariantStates` in
  `Analyze()` (alongside the existing `gpusByVariant` at `analyzer.go:69-72`) and pass the
  role into `computeReplicaCapacity` / `computeReplicaCapacityFallback`.
- **Add a `RoleDecode = "decode"` constant.** Only `RolePrefill = "prefill"` and
  `RoleBoth = "both"` exist today (`internal/interfaces/saturation_analyzer.go:33,36`);
  `"decode"` is currently a bare string literal in `estimateSchedulerQueueDemand`. Add the
  constant and use it in both places.
- Treat empty role `""` (non-disaggregated) the same as `both` (include output).
- Whether a decode-only replica should still be charged the *input* tokens of its waiting
  requests (the prefill KV is transferred from a separate prefill replica but is still
  resident) is a modeling nuance; `estimateSchedulerQueueDemand` charges `input + output` for
  decode, so mirror that for now and track any refinement separately.

## Acceptance criteria

- [ ] Per-replica waiting-queue demand includes output tokens for decode and "both" replicas
      in both the main and fallback paths.
- [ ] Prefill replicas remain input-only.
- [ ] `RoleDecode` constant added and used in both `computeReplicaCapacity*` and
      `estimateSchedulerQueueDemand`.
- [ ] Unit tests cover: prefill (input-only), decode (input+output), both (input+output),
      and empty-role (input+output) waiting-queue demand.
- [ ] No change for non-disaggregated deployments with an empty queue (`QueueLength == 0`).

ev-shindin:
Fixes #1456

## Proposed Changes

- :bug: Charge requests waiting in a replica's **local engine queue** by P/D role. Previously every replica was charged `QueueLength × AvgInputTokens` regardless of role. Decode and `both` replicas grow KV per generated token, so long-generation workloads under-reported demand — by 81× for an illustrative 100-input/8000-output request — making them look cheaper than they are and delaying scale-up.

  | Role | Per-request charge |
  |------|--------------------|
  | `prefill` | `AvgInputTokens` |
  | `decode` / `both` | `AvgInputTokens + AvgOutputTokens` |

  This mirrors the role attribution `estimateSchedulerQueueDemand` already applied model-wide. The role is plumbed from `VariantStates` through **both** the main and the fallback (no `vllm:cache_config_info`) demand paths, with a new `waitingQueueDemand` helper as the single charge site.

- :broom: Harden the charge against bad inputs. The gate now validates in `float64` before converting to `int64`, so one check covers NaN, infinities, and finite-but-out-of-range values. The previous `AvgInputTokens > 0` gate rejected NaN but a `<= 0` gate does not (every NaN comparison is false), and an out-of-range float→int conversion is implementation-defined in Go. The collector filters NaN/Inf on the token averages, so this is a guard rather than a live bug. Multiplying before truncating also drops the per-request rounding the old truncate-then-multiply form accumulated (queue 3 at avg 100.9 tokens now yields 302 rather than 300).

- :broom: Gate on the *combined* per-request cost rather than input alone, so a decode replica reporting only generation tokens is no longer charged zero for its queue.

- :gift: Add `domain.RoleDecode`, and switch `estimateSchedulerQueueDemand` off its bare `"prefill"` / `"decode"` string literals.

- :broom: Emit the resolved role per variant on the `analyzer-result` log line. Demand is no longer interpretable without it — a missing `llm-d.ai/role` label reads as `both` and changes the charge. Canonicalized so an analyzer that leaves the role unset still logs a meaningful value.

- :broom: Docs: document the per-role formula and its operator-visible impact in `saturation-scaling-config.md`; add `variants[].role` to the `analyzer-result` schema in `cycle-log.md` and replace its "total observed token demand" description, which named neither the local-queue projection nor the scheduler-queue term; note in `prometheus.md` that V2 utilization is unbounded above.

### Why `I + O` and not `I + O/2`

`I + O` is this analyzer's **demand-side** unit: a request's footprint at its final decode step, chosen as a peak / no-preemption planning charge, since KV grows monotonically once generation starts and cannot be shed without preemption and recompute. Two distinctions are documented in the code so they aren't mistaken for inconsistencies:

- The throughput analyzer's `ILeff + O/2` is time-averaged, but it is a **separate analyzer** with its own demand/supply pairing.
- Internal to saturation V2, `estimateCapacityFromParams` prices a *concurrent* request at `I + O/2` when deriving k2. This term is therefore deliberately more conservative than the capacity it is compared against, biasing toward scale-up — accepted, because under-provisioning decode capacity causes preemption and recompute thrash. Revisiting that means moving the demand/capacity pair together, not this function alone.

### Scope note

Because an absent role canonicalizes to `both`, **non-disaggregated deployments also include output tokens now**. That is the physically correct charge for a replica handling the full request lifecycle, but it means the common single-variant case sees higher computed demand wherever replicas have a non-empty local queue. See the release note.

### Reviewer notes

- **A pre-existing unit mismatch on the fallback path is documented, not changed.** For a deployment-derived record, `EffectiveCapacity` is `EffectiveMaxBatchedTokens` — a *per-step* budget the store itself calls "a safe lower bound" — while this addend is absolute KV tokens. A deep queue can therefore report saturation the replica's actual KV occupancy doesn't support, and raising the per-request charge lowers the queue depth at which that happens. Fixing it means pairing that path's demand and capacity units, which is out of scope here. A test pins the safe end (idle replica, shallow queue → not saturated). Happy to file this and the other pre-existing items as separate issues.
- Residual risk worth watching after merge is **not** the coefficient but its interaction with P1-observed k2: when `queueLen >= queueLengthThreshold`, k2 is pinned to `tokensInUse`, so utilization becomes `1 + q·(I+O)/tokensInUse`. For long-generation workloads that factor lands on the *magnitude* of the first scale-up step, not just its trigger.

### Pre-review Checklist

- [ ] **E2E tests** for any new behavior — **not added; see below.** No V2 e2e fixture can exercise this. All V2 suites drive the analyzer with `--fake-metrics`, which sets only `kv-cache-usage` / `running-requests` / `waiting-requests`; `AvgInputTokens` and `AvgOutputTokens` come from 5m-rate PromQL over token histograms that static fakes never populate, so both are `0` and the queue term is inert. The one fixture emitting real token metrics is the SGLang emulator, which is explicitly pinned to V1 by a guard carrying `TODO(v2): recalibrate the SGLang fixture metrics + assertion for the V2 analyzer and drop this guard`. Notably this change may unblock that TODO — the fixture sits exactly at V2's `0.85` threshold (which V2 must *exceed*), and the queue term would rise from `3×500` to `3×750`, plausibly making the V2 arc deterministic. That needs a real kind-cluster run to confirm and is left as follow-up. Unit coverage is thorough instead: all role variants, empty/negative queue, missing metrics, output-only decode, non-finite and out-of-int64-range values, both demand paths, and end-to-end `Analyze` assertions including a two-variant P/D model.
- [x] **Docs PR** for any user-facing impact — included in this PR (`saturation-scaling-config.md`, `cycle-log.md`, `prometheus.md`), since the affected docs live in this repo.
- [ ] **Proposal PR** — not applicable. This corrects an existing demand term to match the role attribution the codebase already applied at model level; it does not introduce a new mechanism.

**Release Note**

```release-note
Saturation V2 now accounts for output tokens when estimating demand from requests waiting in a replica's local engine queue. Decode and non-disaggregated ("both") replicas are charged the average input plus output tokens per waiting request; prefill replicas remain input-only. This corrects a significant under-estimate of demand for long-generation workloads, which could delay scale-up.

Because a variant with no `llm-d.ai/role` label is treated as "both", this affects non-disaggregated deployments as well. Where replicas have a non-empty local queue, expect `wva_saturation_utilization` and `wva_required_capacity{unit="continuous"}` to step up, `wva_spare_capacity` to step down, and `wva_desired_replicas` to take a one-time step. `wva_kv_cache_tokens_used` is unaffected. Alerts thresholded on saturation utilization may need re-baselining.
```

**Docs**

Documentation for this change lives in this repo and is included in this PR. No `llm-d/llm-d` docs change is needed — the per-role charge is an internal demand-model detail, and the user-facing knobs (`scaleUpThreshold`, `scaleDownBoundary`) are unchanged.

### DIFF
--- a/docs/developer-guide/cycle-log.md
+++ b/docs/developer-guide/cycle-log.md
@@ -30,8 +30,8 @@ optimizer actually receives.
   "scaleUpThreshold": 1.1,
   "scaleDownBoundary": 0.7,
   "variants": [
-    {"name": "primary", "prc": 1152000, "reason": "P3-k2"},
-    {"name": "v2",      "prc":  403391, "reason": "P1-obs"}
+    {"name": "primary", "prc": 1152000, "role": "both", "reason": "P3-k2"},
+    {"name": "v2",      "prc":  403391, "role": "both", "reason": "P1-obs"}
   ]
 }
 ```
@@ -42,7 +42,7 @@ optimizer actually receives.
 | `namespace` | Kubernetes namespace |
 | `analyzer` | Analyzer name, e.g. `"saturation"`, `"throughput"` |
 | `supply` | Total token supply across ready replicas (readyCount × perReplicaCapacity) |
-| `demand` | Total observed token demand |
+| `demand` | Total token demand. Not purely observed: for saturation V2 it is the sum of three terms — resident KV tokens, a role-aware projection of requests waiting in each replica's local engine queue, and a model-level, prefix-cache-discounted projection of requests still queued upstream in llm-d flow control (`SchedulerQueue`, not attributed to any variant). See [saturation-scaling-config.md](saturation-scaling-config.md) |
 | `util` | `demand / supply`; > 1.0 means the model is over capacity |
 | `rc` | Required capacity signal (post-threshold): > 0 triggers scale-up |
 | `sc` | Spare capacity signal (post-threshold): > 0 permits scale-down |
@@ -50,6 +50,7 @@ optimizer actually receives.
 | `scaleDownBoundary` | Scale-down boundary resolved for this analyzer (from config) |
 | `variants[].name` | Variant name |
 | `variants[].prc` | Per-replica capacity in analyzer units (tokens for saturation) |
+| `variants[].role` | Resolved P/D role: `prefill`, `decode`, or `both`. Renders as `both` both when the scale target has no `llm-d.ai/role` label and when the analyzer does not populate the role at all. Saturation V2 charges waiting requests by this role, so it is needed to interpret `demand` |
 | `variants[].reason` | How the variant's capacity was computed (see below) |
 
 If an analyzer does not compute per-variant capacity, `variants` is an empty
--- a/docs/developer-guide/prometheus.md
+++ b/docs/developer-guide/prometheus.md
@@ -510,7 +510,7 @@ With WVA metrics, the value for the label `namespace` is the WVA controller name
 
 ### `wva_saturation_utilization`
 - **Type**: Gauge
-- **Description**: Per-variant utilization ratio (0.0-1.0) from saturation analysis. V1 path: mean of per-replica KV-cache-usage fractions (matches the per-replica threshold V1 checks). V2 path: TotalDemand / TotalCapacity from the analyzer result. Numerically equivalent for uniform-capacity replicas; V2 is capacity-weighted for mixed-capacity cases.
+- **Description**: Per-variant utilization ratio from saturation analysis. V1 path: mean of per-replica KV-cache-usage fractions, bounded 0.0-1.0 (matches the per-replica threshold V1 checks). V2 path: TotalDemand / TotalCapacity from the analyzer result — **unbounded above**, where > 1.0 means demand exceeds supply. Numerically equivalent for uniform-capacity replicas; V2 is capacity-weighted for mixed-capacity cases.
 - **Labels**:
   - `variant_name`: Name of the variant
   - `namespace`: Kubernetes namespace
--- a/docs/developer-guide/saturation-scaling-config.md
+++ b/docs/developer-guide/saturation-scaling-config.md
@@ -263,6 +263,69 @@ kv-tokens; throughput: tokens/sec). Each analyzer also decides how to
 attribute that demand across roles or variants — sat_v2 splits it among
 active roles.
 
+#### Per-replica demand and the role-aware waiting-queue charge
+
+Distinct from `SchedulerQueue` above, each replica also reports its own **local
+engine queue** (`vllm:num_requests_waiting` / `sglang:num_queue_reqs`) — requests
+already admitted to that pod but not yet generating. sat_v2 charges them to the
+replica, so per-replica demand has two terms:
+
+```
+replicaDemand = <resident KV tokens> + QueueLength × <per-request charge>
+```
+
+The resident term is `TokensInUse` on the main path, or
+`KvCacheUsage × effectiveCapacity` on the fallback path (no
+`vllm:cache_config_info`). The per-request charge depends on the variant's P/D
+role, because a replica only pays for KV it actually materializes:
+
+| Role | Per-request charge | Rationale |
+|------|--------------------|-----------|
+| `prefill` | `AvgInputTokens` | Prompt KV only; output is generated on a decode pod |
+| `decode` | `AvgInputTokens + AvgOutputTokens` | Holds prompt KV and grows it per generated token |
+| `both` (default) | `AvgInputTokens + AvgOutputTokens` | Handles the full request lifecycle |
+
+An absent or empty role canonicalizes to `both`, so **non-disaggregated
+deployments take the decode-style charge** — correct for a replica serving the
+whole request, but it means the majority of deployments include output tokens.
+
+`AvgInputTokens + AvgOutputTokens` is the footprint at a request's *final* decode
+step, chosen as a peak / no-preemption planning charge: KV grows monotonically
+once generation starts and cannot be shed without preemption and recompute.
+
+**`I + O` is the saturation analyzer's demand-side unit; `ILeff + O/2` is the
+throughput analyzer's.** The two are separate analyzers with separate
+demand/supply models, so the difference is by design, not an inconsistency:
+
+| Analyzer | Per-request KV unit | Meaning |
+|----------|--------------------|---------|
+| Saturation V2 (demand) | `I + O` | Peak footprint — what a replica must be able to hold |
+| Throughput (`WorkloadShape.KVreq`) | `ILeff + O/2` | Time-averaged residency (`ILeff = I × (1 − PrefixHitRate)`) |
+
+One asymmetry *is* internal to saturation V2 and should not be confused with the
+above: its own k2 derivation (`estimateCapacityFromParams`) prices a *concurrent*
+request at `I + O/2`. So a queued request is charged more than it occupies on
+average, biasing this term toward scale-up — accepted deliberately, since
+under-provisioning decode capacity causes preemption thrash. Changing it means
+moving the demand/capacity pair together. See the `waitingQueueDemand` doc
+comment for the full trade-off.
+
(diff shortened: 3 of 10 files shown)


deanlorenz (approved):
LGTM. Thorough fix, and the PR body does most of a reviewer's work for them — thank you for the depth on the `I+O` rationale and the two reviewer notes.

I traced the role plumbing and the numeric guards and am satisfied it's correct:

- **Role source is consistent across every consumer.** The charge in `waitingQueueDemand`, the `activeRoles`/log attribution via `VariantCapacity.Role`, and the scheduler-queue split all resolve from the same `VariantStates[].Role`, so there's no path where the per-request charge and the role bookkeeping can disagree for a variant.
- **The per-role decision path stays clean.** `aggregateByRole` → `applyUniversalThreshold` → `initRoleState` feeds the optimizer per-role `RoleCapacities[role]` RC/SC for disaggregated models (the pooled model-level scalars are only used in the all-`both` case, where they're single-unit anyway). So charging decode `I+O` and prefill `I` never get compared against a shared supply on the path that produces a scaling decision.
- **The `float64`-before-`int64` guard is sound**, including the `>= float64(math.MaxInt64)` boundary for the `2^63` round-up — nice catch, and good that it's pinned by the `math.Pow(2,62)×2` case.

I confirmed both of your reviewer notes against the source and agree with your scoping:

- **Fallback-path unit mismatch** (per-step `EffectiveMaxBatchedTokens` budget vs. absolute KV tokens): genuinely pre-existing, and raising the per-request charge does lower the queue depth at which it can invent saturation. Out of scope here; the idle-replica/shallow-queue test pins the safe end. Worth a tracking issue.
- **Interaction with P1-observed k2**: confirmed — when `queueLen >= queueLengthThreshold`, k2 pins to `tokensInUse`, so utilization becomes `1 + q·(I+O)/tokensInUse` and the coefficient lands on the *magnitude* of the first scale-up step, not just its trigger. Agree this is the residual most worth watching on a real cluster; a follow-up issue (or a note on an existing V2-calibration issue) would be good so it isn't lost.

The `both`-default scope change (non-disaggregated deployments now include output tokens) is the highest operator-facing impact, and the release note + `saturation-scaling-config.md` operational note cover the expected metric moves and alert re-baselining well.

+1 to filing the two pre-existing items above as separate issues. Approving.

PR #1472, opened 2026-07-26
test(e2e): validate the llm-d KEDA EPP guide with the WVA Kind lifecycle

nourey:
## Motivation

This PR adds the WVA-owned Kind qualification path for the llm-d KEDA+EPP queue guide.

The intended ownership split is:

* WVA owns the deterministic Kind setup, lifecycle, conformance spec, diagnostics, and CI lane
* llm-d remains the source of truth for the user-facing guide and owns real-model / real-GPU nightly qualification

The Kind path is intentionally CPU-only. It uses a deterministic inference simulator with the canonical optimized-baseline decode Deployment identity so that the Prometheus → KEDA → HPA control path can be exercised without requiring real GPUs or KV-cache behavior.

### Disposable Kind setup

Add:

```bash
make test-e2e-keda-epp-guide-with-setup
```

This target owns the complete disposable lifecycle:

1. creates a fresh single-node CPU-only Kind cluster;
2. resolves the current official llm-d `main` once to an immutable commit SHA;
3. materializes that exact revision for the complete run;
4. installs low-resource Prometheus, KEDA, and EPP infrastructure;
5. deploys the deterministic simulator under the canonical decode Deployment identity;
6. applies the canonical KEDA `TriggerAuthentication` and `ScaledObject`;
7. runs only the `keda-epp-guide` Ginkgo spec;
8. tears down the isolated Kind cluster on exit.

`LLMD_SOURCE_SHA=<full SHA>` remains available as an explicit opt-in pin for reproducibility, debugging, compatibility investigation, or temporary stabilization. `LLMD_SOURCE` may provide a local checkout for materialization, but its checked-out branch or `HEAD` never selects the canonical revision.

### Canonical guide composition

The EPP install uses the current llm-d inputs in this exact order:

1. canonical router base;
2. canonical optimized-baseline values;
3. canonical monitoring values;
4. canonical KEDA+EPP queue-guide values;
5. one narrow WVA-owned Kind override.

The WVA override provides only the deterministic / disposable-Kind differences needed for this environment, including bounded resources and `maxConcurrency: 1`.

The canonical Kustomize resources are materialized from the selected llm-d revision rather than maintained as an independent WVA copy. A temporary overlay changes only the environment-specific namespace and Prometheus endpoint; normalized comparison rejects other canonical drift.

### Secure KEDA / Prometheus path

The Kind setup keeps Prometheus HTTPS verification enabled.

A generated Prometheus CA is:

* used to issue the Prometheus serving certificate;
* exposed to the workload namespace through the canonical `keda-prometheus-auth` Secret;
* mounted into the KEDA operator as a public-CA-only trust directory.

The Prometheus private key remains only in the monitoring namespace. No `unsafeSsl`, `insecureSkipVerify`, or equivalent TLS bypass is used.

### Deterministic contract

The isolated spec verifies the canonical:

* `ScaledObject` target, replica bounds, generated HPA name, and HPA behavior;
* `TriggerAuthentication` relationship;
* queue trigger:
  `llm_d_epp_flow_control_queue_size`;
* running-request trigger:
  `llm_d_epp_request_running`;
* exact PromQL selectors, thresholds, and `AverageValue` metric type;
* Prometheus target health and exact source-query liveness;
* KEDA external metric freshness and phase-specific values;
* generated HPA ownership, target, selectors, metrics, and behavior;
* independent HPA decision, Deployment actuation, and ReadyReplicas convergence.

Because the EPP per-model metric series are lazy, the spec first sends one bounded warmup request, proves both source series exist through Prometheus, removes the warmup, and requires the downstream KEDA running metric to reset to zero before the deterministic scaling phases.

The deterministic scaling phase then proves:

1. one running request / one queued request remains stable at exactly one replica;
2. adding one more queued request produces the exact running-request and queue stimulus required for scale-up;
3. the HPA desires two replicas and the Deployment is actuated to two replicas;
4. ReadyReplicas converges stably to two;
5. no sampled HPA or Deployment replica value exceeds two.

## CI

Add a dedicated `keda-epp-guide` job to `.github/workflows/ci-pr-checks.yaml`.

The job invokes the same:

```bash
make test-e2e-keda-epp-guide-with-setup
```

entry point used for local execution.

This gives WVA its own deterministic Kind drift signal while real-GPU qualification remains upstream in llm-d.

## Relationship to llm-d GPU qualification

The metric contract and lazy-series ordering are aligned with the real-GPU KEDA+EPP queue path in llm-d #2192 / #2208.

This PR does **not** duplicate that qualification.

Out of scope here:

* real models or GPUs;
* KV-event generation or KV-routing quality;
* sustained/performance load;
* scale-down qualification;
* post-scale inference;
* OpenShift / Thanos qualification;
* stable-guide promotion;
* generic cross-repository synchronization machinery.

Those remain upstream or follow-up concerns.

## Validation

Local/static validation includes:

* `git diff --check`;
* Bash syntax and ShellCheck;
* Go formatting, compile-only E2E validation, vet, and lint;
* workflow/YAML validation;
* canonical render and normalized-equivalence validation;
* exact five-layer Helm ordering;
* source-policy tests for current-main resolution, explicit SHA pins, local-source safety, and fail-closed missing revisions.

A clean fresh-Kind semantic execution passed:

```text
Ran 1 of 73 Specs
SUCCESS! -- 1 Passed | 0 Failed | 72 Skipped
```

The run verified:

* secure Prometheus/KEDA CA trust with no TLS verification bypass;
* lazy-series warmup through Prometheus and downstream KEDA running-metric reset;
* Prometheus source-query liveness and KEDA external-metric liveness;
* exact canonical ScaledObject / TriggerAuthentication contract;
* generated HPA identity, ownership, metric translation, and behavior;
* stable one-running / one-queued Phase A;
* exact running-request and queue scale stimulus in Phase B;
* HPA desired replicas reaching `2`;
* Deployment spec reaching `2`;
* stable ReadyReplicas at `2`;
* no sampled HPA or Deployment replica value above `2`;
* request and fresh-cluster cleanup.

The repository-default Kind image currently uses Kubernetes v1.32. KEDA 2.20 does not list Kubernetes v1.32 in its tested compatibility matrix, although KEDA's deployment prerequisites allow Kubernetes 1.30 and newer. The local pass therefore demonstrates this contract on the repository-default environment but is not presented as evidence of official KEDA compatibility for that pairing.

## Scope

This PR completes the WVA side of the agreed split: the **Kind setup + deterministic spec + CI lane**.

Real-GPU/nightly qualification remains in llm-d.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -264,6 +264,27 @@ test-e2e-smoke: ## Run smoke e2e tests
 	echo "=========================================="; \
 	exit $$TEST_EXIT_CODE
 
+# Runs only the direct KEDA+EPP guide contract against resources already applied
+# from the llm-d repository. The WVA controller is intentionally not required.
+# The guide's bounded inner waits total less than 30m; 40m reserves time for diagnostics and cleanup.
+.PHONY: test-e2e-keda-epp-guide
+test-e2e-keda-epp-guide: ## Run the llm-d KEDA+EPP guide contract
+	@echo "Running direct KEDA+EPP guide contract..."
+	KUBECONFIG=$(KUBECONFIG) \
+	ENVIRONMENT=kind-emulator \
+	DEPLOY_WVA=false \
+	WVA_NAMESPACE=workload-variant-autoscaler-system \
+	LLMD_NAMESPACE=llm-d-optimized-baseline \
+	MONITORING_NAMESPACE=llm-d-monitoring \
+	KEDA_NAMESPACE=keda \
+	EPP_SERVICE_NAME=optimized-baseline-epp \
+	USE_SIMULATOR=true \
+	SCALE_TO_ZERO_ENABLED=false \
+	SCALER_BACKEND=keda \
+	MODEL_ID=Qwen/Qwen3-32B \
+	go test ./test/e2e/ -timeout 40m -v -ginkgo.v \
+		-ginkgo.label-filter="keda-epp-guide"
+
 # Runs the complete e2e test suite (KEDA backend, excluding smoke and flaky tests).
 .PHONY: test-e2e-full
 test-e2e-full: ## Run full e2e test suite
--- a/Makefile
+++ b/Makefile
@@ -316,8 +316,8 @@ test-e2e-smoke: ## Run smoke e2e tests
 
 # Runs only the direct KEDA+EPP guide contract against resources already applied
 # from the llm-d repository. The WVA controller is intentionally not required.
-# Its successful-spec observation bound is 55m30s; Ginkgo gets 65m and Go gets
-# 70m so suite preflight, diagnostics, and cleanup remain outside that bound.
+# Ginkgo gets 65m and Go gets 70m so suite preflight, diagnostics, and cleanup
+# remain inside the outer test-process boundary.
 .PHONY: test-e2e-keda-epp-guide
 test-e2e-keda-epp-guide: ## Run the llm-d KEDA+EPP guide contract
 	@echo "Running direct KEDA+EPP guide contract..."
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ IMAGE_TAG_BASE ?= ghcr.io/llm-d
 IMG_TAG ?= latest
 IMG ?= $(IMAGE_TAG_BASE)/llm-d-workload-variant-autoscaler:$(IMG_TAG)
 KIND_ARGS ?= -t mix -n 3 -g 2   # Default: 3 nodes, 2 GPUs per node, mixed vendors
+CLUSTER_NAME ?= kind-wva-gpu-cluster
 CLUSTER_GPU_TYPE ?= nvidia-mix
 CLUSTER_NODES ?= 3
 CLUSTER_GPUS ?= 4
@@ -33,6 +34,8 @@ E2E_MONITORING_NAMESPACE    ?= workload-variant-autoscaler-monitoring
 E2E_EMULATED_LLMD_NAMESPACE ?= llm-d-sim
 E2E_KEDA_NAMESPACE          ?= keda-system
 E2E_WVA_SECONDARY_OVERLAY_PATH ?= $(CURDIR)/test/e2e/testdata/secondary-controller
+LLMD_SOURCE                 ?=
+LLMD_SOURCE_SHA             ?=
 # llm-d-benchmark CLI configuration
 # Ensure brew-installed tools (helm >=3.19) take precedence over Rancher Desktop
 export PATH := /opt/homebrew/bin:$(PATH)
@@ -132,12 +135,12 @@ test: manifests generate fmt vet setup-envtest helm ## Run tests.
 .PHONY: create-kind-cluster
 create-kind-cluster:
 	export KIND=$(KIND) KUBECTL=$(KUBECTL) && \
-		deploy/kind-emulator/setup.sh -t $(CLUSTER_GPU_TYPE) -n $(CLUSTER_NODES) -g $(CLUSTER_GPUS)
+		deploy/kind-emulator/setup.sh -c $(CLUSTER_NAME) -t $(CLUSTER_GPU_TYPE) -n $(CLUSTER_NODES) -g $(CLUSTER_GPUS)
 
 # Destroys the Kind cluster created by `create-kind-cluster`
 .PHONY: destroy-kind-cluster
 destroy-kind-cluster:
-	export KIND=$(KIND) KUBECTL=$(KUBECTL) && \
+	export KIND=$(KIND) KUBECTL=$(KUBECTL) KIND_NAME=$(CLUSTER_NAME) && \
         deploy/kind-emulator/teardown.sh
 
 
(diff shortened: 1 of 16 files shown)


docs/developer-guide/testing.md
```diff
@@ -187,6 +187,58 @@ make test-e2e-full
...
 See the [E2E Test Suite README](../../test/e2e/README.md) for full configuration options and examples.
 
+### Direct KEDA+EPP guide contract
+
+`test-e2e-keda-epp-guide` is a narrow, controller-free contract for the
+canonical KEDA+EPP guide in the `llm-d/llm-d` repository. It observes resources
+that the caller already installed; it does not create a replacement
+`ScaledObject`, deploy the WVA controller, or provide a generic guide runner.
+
+Before invoking the target, the caller must create a fresh disposable Kind
```
lionelvillard: there is a make target for that: `create-kind-cluster`

docs/developer-guide/testing.md
```diff
@@ -187,6 +187,58 @@ make test-e2e-full
...
+cluster and install the following from the matching llm-d checkout:
```
lionelvillard: In addition, `SKIP_BUILD=true DEPLOY_WVA=false make deploy-e2e-infra` can be used to install llm-d.

Makefile
```diff
@@ -264,6 +264,27 @@ test-e2e-smoke: ## Run smoke e2e tests
...
+# The guide's bounded inner waits total less than 30m; 40m reserves time for diagnostics and cleanup.
+.PHONY: test-e2e-keda-epp-guide
+test-e2e-keda-epp-guide: ## Run the llm-d KEDA+EPP guide contract
+	@echo "Running direct KEDA+EPP guide contract..."
+	KUBECONFIG=$(KUBECONFIG) \
+	ENVIRONMENT=kind-emulator \
+	DEPLOY_WVA=false \
+	WVA_NAMESPACE=workload-variant-autoscaler-system \
+	LLMD_NAMESPACE=llm-d-optimized-baseline \
+	MONITORING_NAMESPACE=llm-d-monitoring \
```
lionelvillard: by default, `make deploy-e2e-infra` install prometheus in `workload-variant-autoscaler-monitoring`.  Either change the default or remove `MONITORING_NAMESPACE`

lionelvillard:
thanks @nourey! Overall this PR LGTM, only a few things are missing: 
- eventually we want to run the KEDA+EPP guide e2e tests from the WVA repo, with kind. AFAICT, the only missing bit is installing the kind-compatible `optimized-baseline-nvidia-gpu-vllm-decode` deployment (and corresponding EPP).
- in order to run the guide locally, using make target, namespaces need to be aligned. See inlined comments. 

As for https://github.com/llm-d/llm-d/pull/2098, the nightly tests are meant to run on real GPUs. Better to keep kind tests in this repo.

FYI @mamy-CS

mamy-CS:
Thanks @lionelvillard, aligned on all of this. Direction: keep the duplication in WVA and let nightly-vs-e2e drift be the sync signal for now. WVA owns the kind KEDA+EPP path, GPU nightlies stay upstream.

To close the gaps this needs (a) the kind-compatible `optimized-baseline` config vendored here, decode deployment, EPP alignment, the ScaledObject+TriggerAuthentication, and the kind-only KEDA CA values. Behind a `test-e2e-keda-epp-guide-with-setup` target mirroring the smoke/full ones, and (b) a CI job to actually run it. 

One alignment gotcha: the guide's ScaledObject points at release `llmd-kube-prometheus-stack` but the new `PROMETHEUS_RELEASE_NAME` defaults to `kube-prometheus-stack` needs setting or KEDA can't scrape based on findings run locally. I'll also drop a few inline nits on flake-prone poll loops and the operator-log scan. Thanks for the work @nourey

nourey:
Thanks, this direction makes sense to me.

Keeping the Kind compatible configuration vendored in WVA will increase maintenance and synchronization costs, since the WVA and llm-d implementations may evolve independently. However, if we treat this separation as an explicit architecture rather than simple duplication, I think it can provide several valuable advantages.

  1. It allows autoscaling logic to be developed independently from real model and GPU behavior. Deterministic simulators can exercise metrics, KEDA, HPA, quota and replica transition logic without waiting for expensive model startup or GPU capacity
  
  2. It provides a safe environment for negative and failure scenarios. Missing metrics, unavailable Prometheus endpoints, KEDA restarts, invalid authentication, blocked replica readiness and similar failures can be introduced deliberately without destabilizing an upstream GPU environment
  
  3. It gives us a place to define and test autoscaling invariants. Rules such as respecting minimum and maximum replicas, never exceeding namespace quota, avoiding unsafe scaling when metrics disappear and preserving bounded transitions can be verified directly
  
  4. It makes different autoscaling approaches comparable under the same controlled workload. KEDA, HPA, WVA policies and future proactive scaling mechanisms could be evaluated against the same load patterns and observable outcomes.
  
  5. It makes broader test matrices affordable. Thresholds, replica limits, load shapes, failure modes and scaling policies can be exercised in Kind, while only the most meaningful configurations need to progress to real GPU qualification
  
  6. It should make debugging easier. If the deterministic WVA test is green while the upstream GPU nightly fails, the investigation can focus on real model, GPU, deployment or platform integration rather than the basic Prometheus to KEDA to HPA control path
  
  7. It also enables a contract first guide development flow. The expected metrics, labels, scaling target, replica bounds and observable behavior can be defined and validated in WVA before the corresponding llm-d guide implementation is qualified with a real model


The additional maintenance cost is real, so I think this works only if the relationship is documented explicitly. Each vendored path should identify its upstream origin, list the intentional Kind specific differences, state exactly which behaviors it proves and does not prove and map those assertions to the corresponding llm-d GPU nightly coverage. Stable promotion should then be based on both layers: deterministic conformance in WVA and real environment qualification in llm-d.

With the upcoming inline review fixes, this PR can establish the first WVA owned Kind KEDA+EPP path and its CI entry point. I do not think this PR needs to solve the complete long term bridge between the repositories but it may be useful to confirm the intended destination: WVA as the deterministic autoscaling conformance and experimentation layer and llm-d as the owner of the user facing guide and real GPU qualification.

The long term CI layout, shared metadata format and any automated synchronization mechanism can be designed separately. This PR only needs to preserve a path toward that architecture rather than implementing the entire bridge now.

This also seems relevant to the existing KEDA+EPP guide promotion path. The guide is already available as experimental documentation, so after the WVA Kind path and the upstream GPU qualification are established, I think we should make the stable promotion criteria and ownership explicit rather than leaving the two test layers disconnected. That broader decision does not need to block this PR but this PR should provide the first concrete foundation for it.

If that is the intended direction, later work can formalize the shared contract, provenance, intentional differences and promotion evidence without blocking this initial Kind setup. I really appreciate the kind words and thoughtful guidance from both of you

asm582:
@mamy-CS @lionelvillard what does this mean: `the kind-compatible optimized-baseline config` ? Without any real KV events, does it matter whether we have an optimized baseline?

nourey:
I pushed the follow up that completes the WVA side as the Kind setup + deterministic spec + CI lane: `test-e2e-keda-epp-guide-with-setup` now owns a fresh CPU-only Kind lifecycle, resolves the current official llm-d `main` once per run and materializes that immutable revision, composes the canonical router/optimized-baseline/monitoring/queue-guide inputs with a narrow Kind override, installs the corresponding EPP and deterministic simulator under the canonical decode Deployment identity, keeps the Prometheus→KEDA path TLS verified with the Kind only public CA trust setup, and verifies the exact queue/running metric contract through Prometheus, KEDA, the generated HPA, and a bounded `1 -> 2` transition. To clarify the “kind-compatible optimized-baseline” wording: the goal is not to reproduce full optimized baseline routing behavior in Kind; it is to preserve the canonical decode identity, EPP/Flow Control topology, labels and autoscaling contract needed for this path. Since there are no real KV events here, KV-sensitive routing behavior is intentionally not qualified by this test. That remains with the real-model/GPU layer upstream, consistent with #2192/#2208 and the agreed WVA-Kind / llm-d-GPU split. The new CI lane is now wired; the current run only surfaced a small `prealloc` lint finding in the test slice allocation, which I’ll fix next.

mamy-CS:
@nourey really nice work getting the whole path green on a kind. Some comments to simplify a bit. 
The contract is verified at three overlapping layers (in shell and Go), and collapsing each to a single source of truth would cut a few hundred lines. Maybe cut the require_line_count greps and the awk extractor? And the same metric is confirmed through multiple channels, we can simplify there as well. The HPA behavior is asserted three times as well, asserting once on the live HPA should be enough. The timeout-budget helpers, could probably fold down to a small handful of constants. I think one authoritative check per property reads better and won't flake on upstream formatting. Thanks.

mamy-CS:
> @mamy-CS @lionelvillard what does this mean: `the kind-compatible optimized-baseline config` ? Without any real KV events, does it matter whether we have an optimized baseline?

"kind-compatible optimized-baseline config" means the canonical guide artifacts. But I agree maybe we could probably trim the routing value layer to just what carries the decode identity + Flow Control queue metric. Maybe as a follow-up.

nourey:
I pushed a focused simplification follow-up for the Kind KEDA+EPP contract. The update removes overlapping rendered YAML and duplicate runtime assertions while keeping one authoritative check for each meaningful boundary across Prometheus → KEDA → HPA → Deployment, strengthens the live ScaledObject, TriggerAuthentication, HPA ownership and TLS checks, simplifies the timeout model and updates the canonical llm-d guide path after the upstream relocation. The fresh Kind replay still passes end to end. Overall this reduces the review follow-up by **213 net lines** across the four touched files, with 228 additions and 441 deletions

mamy-CS:
@nourey dco and e2es are failing.

nourey:
Thanks @mamy-CS. The E2E failures were caused by the Go 1.26 builder digest pinned on `main` no longer being available in Quay, so the image build failed before Kind/Ginkgo ran. I opened #1540 as a separate minimal fix; its smoke/full E2Es are green now.

mamy-CS:
Thanks, merged. Please rebase @nourey

mamy-CS (approved):
Thank you @nourey for the great work /lgtm

PR #1449, opened 2026-07-22
chore: rename internal/interfaces to internal/domain

mamy-CS:
Relates to #1402 (Section 3 of 7: rename the misnamed internal/interfaces hub; follow-up to #1445)

## Proposed Changes

- :broom: Rename package `internal/interfaces` → `internal/domain` (9 files moved, byte-identical except the `package` declaration)
- :broom: Update all importers: import paths, ~1,265 `interfaces.X` → `domain.X` qualifier references across 71 files, and remove the 4 redundant `interfaces "..."` import aliases
- :broom: Fix stale path/name references in `docs/developer-guide/{multi-analyzer-pipeline,slo-queuemodel}.md` and a package comment in `internal/collector/collector.go`

### Why

`internal/interfaces` was not an interfaces package; it held mostly concrete structs, constants, and methods (`Allocation`, `ReplicaMetrics`, `VariantDecision`, `AnalyzerResult`, `ServiceClass`, ...), with only four small interfaces (`Analyzer`, `AnalyzerConfig`, `Actuator`, `SaturationAnalyzer`). `internal/domain` reflects its real role. This is the lowest-risk option from the issue (a mechanical rename rather than splitting types from interfaces).

### Pre-review Checklist

- [ ] **E2E tests** for any new behavior — N/A, pure rename, no behavior change
- [ ] **Docs PR** for any user-facing impact — N/A, internal-only; developer-doc references updated in this PR
- [ ] **Proposal PR** for any new enhancement or change to existing behavior — N/A, no behavior change

**Release Note**

```release-note
NONE
```

**Docs**

```Docs
Developer-doc path references updated
```
---

Verification for reviewers: git tracks all 9 as renames (history preserved); every moved file is byte-identical to its original except the `package` line (confirmed by diff); no string literal in the original code ever contained `interfaces.`, so nothing embedded was corrupted; `go build`, `go vet`, `gofmt -l` (repo-wide clean), `make lint` (0 issues), and the full non-e2e `go test` suite all pass.

PR #1477, opened 2026-07-27
fix(saturation): update stale internal/interfaces refs

DolevAdas:
Fixes stale internal/interfaces references in the saturation engine left behind by the interfaces → domain rename (#1449), which currently breaks main's build.


ev-shindin:
Thanks @DolevAdas ! Another PR solving same issue already

PR #1479, opened 2026-07-27
fix(saturation): make analyzer effectiveEnabled opt-in and log non-registration

deanlorenz:
Make analyzer participation opt-in and add a startup signal when the ThroughputAnalyzer is not registered.

- effectiveEnabled is now opt-in: an absent config entry resolves to false (previously defaulted to enabled), so an analyzer only participates when explicitly configured.
- Log at startup when the ThroughputAnalyzer is not registered, so a misconfiguration is visible rather than silent.
- Document the opt-in participation semantics in the developer guide.

```release-note
Analyzer participation is now opt-in: an analyzer runs only when explicitly enabled in its configuration. An absent config entry no longer defaults to enabled. A startup log line now reports when the ThroughputAnalyzer is not registered.
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -97,11 +97,12 @@ func init() {
 
 // throughputAnalyzerEnabled reports whether any saturation config entry lists
 // the throughput analyzer with enabled != false. Startup-time gate: when no
-// entry enables throughput the analyzer is never registered, so it cannot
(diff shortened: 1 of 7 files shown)


ev-shindin (approved):
Thanks @deanlorenz ! Please add release note to PR body.

PR #1480, opened 2026-07-27
feat(throughput): compute decode demand from model-level arrival rate

deanlorenz:
Compute the throughput analyzer's decode demand from a model-level request arrival rate rather than a per-instance summation.

- Collector exposes a model-level request arrival rate for the analyzer.
- Decode demand is derived from the model-level arrival sum (Λ_req × avgOL) plus queue-drain, matching the per-model demand model.
- avgOL is weighted by replica (KV) count across non-prefill variants.
- Document the ArrivalRate input in the multi-analyzer pipeline guide.

```release-note
The throughput analyzer now derives decode demand from the model-level request arrival rate (arrival rate × average output length) rather than summing per-instance arrival rates, matching the model-level demand design.
```

### DIFF
--- a/docs/developer-guide/multi-analyzer-pipeline.md
+++ b/docs/developer-guide/multi-analyzer-pipeline.md
@@ -189,6 +189,7 @@ Key `AnalyzerInput` fields:
 | `VariantStates` | `[]VariantReplicaState` | Current/desired/pending replica counts per variant |
 | `Config` | `AnalyzerConfig` | Resolved config (cast to your config type as needed) |
 | `SchedulerQueue` | `*SchedulerQueueMetrics` | Scheduler queue metrics; nil when flow control is off |
+| `ArrivalRate` | float64 | Model-level request arrival rate (req/s), no per-pod labels; zero when EPP absent or no traffic yet |
 
 ### Output invariants
 
--- a/docs/developer-guide/throughput-analyzer.md
+++ b/docs/developer-guide/throughput-analyzer.md
@@ -368,7 +368,7 @@ On leader failover the incoming leader starts with an empty analyzer. During war
 │    demand: EPP primary → engine fallback → k*-local      │
 │                                                          │
 │  model-level:                                            │
-│    + queue demand from QueueSize / (factor×ITL)          │
+│    demand: ArrivalRate×avgOL + queue demand              │
 │    TotalSupply, TotalDemand, TotalAnticipatedSupply       │
 │    RequiredCapacity = 0, SpareCapacity = 0 (engine fills) │
 └──────┬───────────────────────────────────────────────────┘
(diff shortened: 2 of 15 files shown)


ev-shindin (commented):
Overall this looks good — the model-level arrival approach is a clean fix for the fragile per-pod metric merge, and the tests are thorough. Two things worth calling out before merge:

**1. When EPP is missing, a busy model can now look empty and get scaled down.**

With this change, `TotalDemand` comes entirely from the model-level `ArrivalRate` (plus queue). If that arrival metric isn't there — no EPP, or the query doesn't match — `ArrivalRate` is 0, so `TotalDemand` drops to 0 even if the replicas are actually busy. The engine then sees the whole model as spare capacity and can scale it down.

Before this PR, the k\*-local fallback (derived from KV usage) kept demand above zero in the no-EPP case, so this used to be safe. I see it's intentional and deferred to the `k_knee` work, which is fine — but it means **throughput must not be enabled without EPP arrival metrics present**, or it'll push spurious scale-downs. Can we make that a hard requirement in the user-facing enablement docs (not just the developer guide)?

**2. If the arrival query silently returns 0, we can't tell "no traffic" from "broken query."**

`ArrivalRate` now depends on one query filtered by `target_model_name={modelID}`, with no fallback. If a future EPP version labels the metric differently, or `modelID` doesn't match `target_model_name` (aliasing/LoRA), the query just returns 0 — and that only shows up at DEBUG level. Combined with point 1, a broken query looks exactly like an idle model and leads to scale-down.

Suggestion: log a WARN (or emit a metric) when model-level `ArrivalRate == 0` but replicas still report `RequestRate > 0` — i.e. the engine is clearly completing work but the arrival query says nothing arrived. That's a cheap, reliable signal that the query is misconfigured rather than the model being idle. Could be this PR or a fast-follow.

deanlorenz:
Thanks for the review, @ev-shindin. Both points addressed — pointers to the changes:

**1. No-EPP / arrival = 0 → possible scale-down.** Confirmed intentional and safe within this PR's scope: zero demand only *permits* scale-down — it can't force a scale action or drive scale-up — and the all-live-agree gate still governs whether a scale-down actually happens. I documented that reasoning in-code at the demand-assembly site in `94accd09`.

On the operator requirement — agreed it should be a hard, user-facing statement, not just the developer guide. Operator enablement docs live in llm-d/llm-d, tracked here as #1498 (TA is opt-in + restart-required); I've extended it with a third requirement: **do not enable the ThroughputAnalyzer without EPP arrival metrics present**, since without them a busy model reads as idle. The doc change lands as a PR against llm-d/llm-d.

**2. Broken query vs. idle model.** I considered the `ArrivalRate == 0 && ΣRequestRate > 0` cross-check but didn't use it: `ReplicaMetrics.RequestRate` is a completion/processed rate, not an arrival rate — a draining engine keeps `RequestRate > 0` after arrivals have legitimately dropped to zero, so that condition is a normal ramp-down state and would false-positive constantly. I documented why RequestRate isn't an arrival proxy in `94accd09`.

The sound signal is temporal, not instantaneous — "supply has been live but demand has never been observed for a full staleness window." That's implemented as an observability-only WARN in the liveness path in the companion PR #1481 (`c32235be`); it never sets liveness and never affects a scale decision.

ev-shindin (dismissed):
Approve. One accuracy note for the record: the merged #1481 `detectDemandLiveness` path *diagnoses* a stuck-at-zero-demand condition (live supply, no observed arrivals) but is observability-only -  it does not gate scale-down. So the real protection against a wrong scale-down when arrivals read zero is the multi-analyzer live-consensus veto (the saturation analyzer disagreeing), not the liveness warning. Worth phrasing the safety section that way rather than "#1481 addresses it."

ev-shindin (approved):

PR #1481, opened 2026-07-27
feat(saturation): gate the scale-down veto on per-analyzer liveness

deanlorenz:
Gate the scale-down veto on per-analyzer liveness so an uninformative analyzer (never had metrics, error state, or stale) cannot block scale-down.

- Track per-analyzer liveness via the last informative analysis time.
- Exclude non-live analyzers from the scale-down veto and the safe-removal minimum; this applies uniformly, with no name-based exemption.
- Keep the queueing-model path always-live under the new gate.
- Safety floor: if no live analyzer remains, do not scale down.
- Document the liveness gate in the multi-analyzer pipeline guide.

```release-note
Scale-down decisions now ignore analyzers with no recent informative result (never reported, errored, or stale): an uninformative analyzer can no longer veto a scale-down. If no analyzer is live for a model, scale-down is withheld as a safety measure.
```

### DIFF
--- a/internal/engines/saturation/engine_v2.go
+++ b/internal/engines/saturation/engine_v2.go
@@ -11,6 +11,7 @@ import (
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/accelerator"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/config"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/domain"
+	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/engines/analyzers/throughput"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/engines/pipeline"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/logging"
 	"github.com/llm-d/llm-d-workload-variant-autoscaler/internal/utils"
@@ -164,7 +165,7 @@ func (e *Engine) runAnalyzersAndScore(
 			ScaleDownBoundary: down,
 		})
 	}
-	e.updateLivenessAndSetLive(namespace, modelID, namedResults)
+	e.updateLivenessAndSetLive(ctx, namespace, modelID, namedResults)
 
 	for _, nr := range namedResults {
 		logAnalyzerResult(ctx, modelID, namespace, nr)
@@ -176,8 +177,10 @@ func (e *Engine) runAnalyzersAndScore(
 // every informative result's AnalyzedAt, then sets nr.Live on each entry in
 // place based on whether its last good analysis (if any) is within the
 // staleness window. Applies uniformly to every analyzer, including
-// saturation — no name-based exemption.
+// saturation — no name-based exemption. After the liveness pass it runs the
+// warn-only demand-liveness detector (see detectDemandLiveness).
 func (e *Engine) updateLivenessAndSetLive(
+	ctx context.Context,
 	namespace, modelID string,
 	namedResults []pipeline.NamedAnalyzerResult,
 ) {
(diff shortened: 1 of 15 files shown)


ev-shindin (commented):
Nice change - the liveness gate is applied uniformly (disaggregated and non-disaggregated via the synthesized `both` role), every result path sets `Live`, and the test coverage is genuinely thorough (recovery, staleness boundary, per-model scoping, real-veto-preserved). Two minor follow-ups, neither blocking:

**1. The `no-data`/`error` sentinels are duplicated string literals, not shared constants.**

`ResultIsInformative` in `analyzer_helpers.go` hardcodes `"no-data"` and `"error"`, which have to stay in sync with the saturation analyzer's own `satReasonNoData = "no-data"` (`saturation_v2/types.go`) and its `"error"` label fallback. If saturation ever renames one of those reasons, the pipeline sentinel silently drifts — an informative result gets misclassified as uninformative (or vice-versa) with no compile error, and the liveness gate quietly misbehaves.

Suggestion: reference shared constants across the packages, or add a small cross-package test that pins the string values so a rename fails loudly.

**2. `lastGoodAnalysis` is never evicted.**

The map is keyed per model but entries are never removed when a model/VA is deleted, so on a long-lived controller with model churn it grows unbounded (small, but monotonic). If this matches the existing `vaEventTracker` convention that's fine to defer — otherwise a cheap prune of keys not seen in the current cycle (or cleanup on VA delete) would bound it.

(Minor aside: liveness keys off `Result.AnalyzedAt`; an informative result that left `AnalyzedAt` unset would read as stale. Saturation and throughput both set it today, so this is just an implicit contract worth being aware of.)

deanlorenz:
Thanks, @ev-shindin — both follow-ups folded into this PR:

**1. Shared `no-data`/`error` sentinels.** Done in `61060530` — exported `pipeline.ReasonNoData` / `pipeline.ReasonError` as the single source of truth; `ResultIsInformative` and the saturation analyzer now reference the same constants, so a rename can't silently drift the liveness classification.

**2. `lastGoodAnalysis` eviction.** Done in `c32235be` — a selective prune at the per-cycle boundary evicts model keys no longer in the active-model set while keeping timestamps for still-active models. It's targeted eviction, not a per-cycle reset (the latch is deliberately cross-cycle: a transient `no-data` with a recent good result must still count as live). The map is in-memory, so it also clears on a controller restart.

The same commit adds an observability-only demand-liveness warning (supply live but no demand for a full staleness window — the temporal broken-arrival signal referenced from #1480 point 2); it's log-only and never gates a decision.

On the `AnalyzedAt` aside — good to flag. Both saturation and throughput set it today; the liveness code keys off it deliberately, and that implicit contract is called out in the liveness comments (`33226cd4`/`c32235be`).

internal/engines/saturation/engine_v2.go
```diff
@@ -156,12 +165,173 @@ func (e *Engine) runAnalyzersAndScore(
...
 	return namedResults, nil
 }
 
+// updateLivenessAndSetLive refreshes e.lastGoodAnalysis for this model with
+// every informative result's AnalyzedAt, then sets nr.Live on each entry in
+// place based on whether its last good analysis (if any) is within the
+// staleness window. Applies uniformly to every analyzer, including
+// saturation — no name-based exemption. After the liveness pass it runs the
+// warn-only demand-liveness detector (see detectDemandLiveness).
+func (e *Engine) updateLivenessAndSetLive(
```
ev-shindin, line 182: Liveness silently depends on every informative analyzer setting `AnalyzedAt`. If a future
analyzer returns an informative result but leaves `AnalyzedAt` zero-valued, `now.Sub(lastGood)`
is enormous → it's computed as instantly stale → excluded from the veto → it can *enable* a
scale-down it should have blocked.

The `AnalyzedAt` contract is documented, but I'd make it defensive so a forgotten field can't
disarm the veto — e.g. treat a zero `AnalyzedAt` on an informative result as `now` (or as
non-informative). Non-blocking.

internal/engines/saturation/engine_v2.go
```diff
@@ -156,12 +165,173 @@ func (e *Engine) runAnalyzersAndScore(
...
+// utils.GetNamespacedKey(namespace, modelID)); any outer key in
+// e.lastGoodAnalysis absent from activeKeys belongs to a departed model (its
+// VariantAutoscalings are gone) and is deleted, bounding the map to live models
+// rather than letting it grow unboundedly across the controller's lifetime.
+//
+// Guards against an empty active set: a cycle that enumerates no models (e.g.
+// saturation config not loaded yet) must not wipe accumulated state, so pruning
+// is skipped when activeKeys is empty.
+func (e *Engine) pruneLastGoodAnalysis(activeKeys map[string]bool) {
+	if len(activeKeys) == 0 || e.lastGoodAnalysis == nil {
```
ev-shindin, line 325: The `len(activeKeys) == 0` early return means a genuine "all models removed" state never
reclaims — entries leak until a model reappears. I assume this is intentional (don't wipe
state on a transient empty/list-failure cycle), but the two cases are indistinguishable here.
A one-line comment on *why* empty ≠ prune would settle it. Bounded and low-severity.

internal/engines/saturation/engine_v2.go
```diff
@@ -156,12 +165,173 @@ func (e *Engine) runAnalyzersAndScore(
...
+	perAnalyzer := e.lastGoodAnalysis[modelKey]
+
+	// Config is nil in unit tests that construct a minimal Engine directly
+	// (bypassing NewEngine, which panics on a nil Config in production).
+	interval := 30 * time.Second
+	if e.Config != nil {
+		interval = e.Config.OptimizationInterval()
+	}
+	now := time.Now()
+	threshold := analyzerLivenessStaleCycles * interval
```
ev-shindin, line 203: If `OptimizationInterval()` ever returns 0, `threshold` is 0 → every analyzer is non-live
every cycle → the safety floor makes scale-down impossible. A `interval > 0` guard (fall back
to the 30s default) would harden this. Minor.

ev-shindin (approved):
Thanks @deanlorenz ! Please address my comments in follow-up PR.

PR #1487, opened 2026-07-29
fix(engine): wire GLOBAL_OPT_INTERVAL to the actual optimize loop

biranofer:
## Summary

The saturation engine's optimize loop interval is hardcoded to `30 * time.Second` at construction (`internal/engines/saturation/engine.go`), completely disconnected from `Config.OptimizationInterval()` (sourced from the `GLOBAL_OPT_INTERVAL` ConfigMap key / env var).

This isn't just a missing nice-to-have — it's a silently inert, already-documented config knob:
- `docs/design/controller-behavior.md` documents "Default Interval: 60 seconds... Configurable via `GLOBAL_OPT_INTERVAL`"
- `config/base/manager/manager-configmap.yaml` ships `GLOBAL_OPT_INTERVAL: "15s"`
- the OpenShift overlay (`config/components/openshift/configmap-patch.yaml`) patches it to `"60s"`
- `deploy/kubernetes/README.md` has a runbook for patching `GLOBAL_OPT_INTERVAL` live and restarting to apply it

None of these ever had any effect — the controller always ran the optimize loop at the hardcoded 30s regardless of what the ConfigMap said.

`optimize()` also re-read `Config.OptimizationInterval()` every cycle specifically to "update the executor interval if changed," but the update itself was a no-op stub with a comment admitting the gap:
```go
// TODO: Support dynamic interval in Executor if needed. For now, we log and proceed.
if interval > 0 {
    // e.executor.SetInterval(interval) // If supported
    _ = interval
}
```

## What changed

- `Interval: 30 * time.Second` → `Interval: cfg.OptimizationInterval()` at executor construction.
- Removed the dead re-read-and-discard block in `optimize()`. `PollingExecutor` only supports a fixed interval set at construction (no `SetInterval`), so this is a **static** fix — a controller restart is required to pick up a `GLOBAL_OPT_INTERVAL` change, consistent with how other infrastructure-level settings in `internal/config` behave. Live hot-reload (matching how the saturation-thresholds ConfigMap already reloads without a restart) would need `PollingExecutor` to gain a `SetInterval` and `Start()` to stop locking the period in via `wait.UntilWithContext` — left as a possible follow-up, out of scope here.

## Behavior impact on upgrade

The loop now actually honors whatever `GLOBAL_OPT_INTERVAL` resolves to instead of always running at 30s:
- Base Kustomize install (`GLOBAL_OPT_INTERVAL: "15s"`): faster than before.
- OpenShift overlay (`GLOBAL_OPT_INTERVAL: "60s"`): slower than before.
- No ConfigMap/env override at all: falls back to the viper default of 60s.

Operators following `deploy/kubernetes/README.md`'s existing "patch `GLOBAL_OPT_INTERVAL` and restart" runbook will see it actually take effect for the first time.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/engines/saturation/... ./internal/engines/executor/... ./internal/config/...`
- [x] `go test ./internal/engines/saturation/...` — 86 specs pass unchanged
- [x] `go test ./internal/config/...` — pass unchanged (existing tests already assert `OptimizationInterval()` default/override behavior; this PR makes the runtime loop actually consume that value)
- [x] `gofmt -l` clean

### DIFF
--- a/config/base/manager/manager-configmap.yaml
+++ b/config/base/manager/manager-configmap.yaml
@@ -26,6 +26,9 @@ data:
 
     # Optimization
     # Global optimization loop interval for autoscaling decisions.
+    # Requires a duration unit and must be at least 1s; anything else is logged
+    # and falls back to 15s. Applied at startup — changing it requires a
+    # controller restart.
     GLOBAL_OPT_INTERVAL: "15s"
 
     # Feature Flags
--- a/deploy/kubernetes/README.md
+++ b/deploy/kubernetes/README.md
@@ -657,7 +657,7 @@ make deploy-wva-on-k8s   # WVA + monitoring + scaler + LWS; llm-d is managed sep
 ### Optimization Interval
 
 ```bash
-# Change how often WVA runs optimization (default: 60s)
+# Change how often WVA runs optimization (default: 15s, minimum 1s, unit required)
 kubectl patch configmap wva-manager-config \
   -n workload-variant-autoscaler-system \
   --type merge \
--- a/docs/design/controller-behavior.md
+++ b/docs/design/controller-behavior.md
@@ -178,8 +178,12 @@ func ServiceMonitorPredicate() predicate.Predicate {
 
 In addition to event-driven reconciliation, the controller performs **periodic reconciliation** of all VariantAutoscaling resources:
 
-- **Default Interval**: 60 seconds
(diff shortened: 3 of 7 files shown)


ev-shindin (approved):
Thanks @biranofer . I added a bit of code to your fix.

PR #1486, opened 2026-07-28
feat(config): ScalingPolicy schema Phase 1 in the saturation ConfigMap

ev-shindin:
Implements **Phase 1** of the ScalingPolicy proposal (#1245): the pluggable
scaling-policy envelope in the admin-owned saturation-scaling ConfigMap. Purely
additive to the schema — no CRD, no migration, no new install steps — plus a
follow-on that makes the GPU limiter ConfigMap-driven and applied live.

## Proposed Changes

- :gift: **Analyzer plugin envelope.** `analyzers[]` entries accept `{type, parameters}`.
  `type` falls back to `name` (so `- name: saturation` is unchanged); well-known
  parameters (`scaleUpThreshold`, `scaleDownBoundary`, `score`, `enabled`) fold into
  the typed fields via `Normalize()`, so downstream code is untouched. Analyzer
  selection matches on `EffectiveType()`. Unknown analyzer types are accepted and
  ignored at runtime, preserving the `RegisterAnalyzer` extension path and
  forward-compat.
- :gift: **Inline `scaleToZero: {enabled}`.** On an entry it overrides the separate
  `wva-model-scale-to-zero-config` ConfigMap for that model/namespace; omitted, that
  ConfigMap still governs (`ResolveScaleToZeroEnabled`).
- :gift: **ConfigMap-driven, live GPU limiter.** `limiters:` on the cluster `default`
  entry is the sole source that selects the limiter (a `quota` entry wins over
  `gpu-inventory`; inventory is the default when none is declared). Inline `type: quota`
  entries reuse `QuotaLimiterConfig`, so they share the quota schema, validation, and
  consumer. The engine rebuilds the limiter when the ConfigMap changes
  (`SetLimiterBuilder`/`refreshLimiter`, keyed by a config signature) — **no restart**.
- :wastebasket: **Remove the unreleased limiter flags.** `--limiter-type` / `LIMITER_TYPE`,
  `--quota-config-file` / `QUOTA_CONFIG_FILE`, `Config.LimiterMode()/QuotaConfigFile()/QuotaEntries()`,
  the `limiterConfig` struct, the loader's quota-file loading, and the startup limiter
  validation are removed in favor of the single ConfigMap source.
- :broom: Opting out to the V1 engine is unchanged (delete `analyzers:`). Refreshed the
  shipped ConfigMap examples (correct `{modelID}#{namespace}` keys, single-mode limiters)
  and the `saturation-scaling-config`, `quota-limiter`, and `configuration` developer guides.

[...]

PR #1501, opened 2026-07-30
Feat/rate anchored k2

### LINKED ISSUE #1500: V2 saturation: the compute bound (k2) measures a KV stock, but the constraint it stands for is a rate
## Summary

The V2 saturation analyzer models a replica's capacity as `min(k1, k2)` in KV tokens:

- `k1 = TotalKvCapacityTokens × kvCacheThreshold` — the memory bound.
- `k2` — the compute bound, recorded as `tokensInUse` at the moment the replica was seen
  queueing, then averaged into a rolling history keyed
  `model|accelerator|gpuCount|outputBucket`.

`k2` therefore measures a **KV stock**, while the constraint it is supposed to represent is
a **rate**. On prefill-heavy traffic the two are unrelated: the engine exhausts prompt-token
throughput and begins queueing while KV occupancy is still low. In that regime the analyzer
reports abundant headroom on a replica that is already dropping requests, and the
scale-down path releases replicas it should be holding.

## Evidence

Sustained 1000-input/250-output run, 10 replicas × 16 concurrency. The WVA leg was queueing,
dropping requests and cycling replicas at **16.2% average KV utilization** — at 16%
occupancy `demand/supply` reads as abundant headroom, so the engine cannot express what is
happening.

In the comparison run at 100-input/1000-output the same mechanism shows up as its other
face: mid-experiment WVA drains to **one replica** while load is unchanged, then has to
climb back, and the tail latency for that window is a direct consequence.

Six threshold/window/policy legs were run against the 1000/250 workload — varying
`scaleUpThreshold`, `scaleDownBoundary`, the stabilization window and the KEDA `scaleDown`
policy — and none of them broke the cycling. This is not a tuning problem.

## Root cause

`internal/engines/analyzers/saturation_v2/analyzer.go` L302–L320 (`computeK2`, Priorities 1
and 2). Three distinct mechanisms, all observed in the runs above:

1. **The compute bound is silently discarded.** `tokensInUse` derives from
   `kv_cache_usage_perc` collected as `max_over_time(...[1m])` — a peak. Whenever that peak
   exceeds `kvCacheThreshold` (0.80), `k2 ≥ k1`, so `min(k1, k2)` returns the memory bound
   and the compute signal never binds at all.
2. **Supply is over-stated exactly when load is highest**, so utilization reads low and
   `spareCapacity = totalSupply − totalDemand/scaleDownBoundary` grows.
3. **Demand deflates faster than supply.** The queue term
   `QueueLength × (avgInput + avgOutput)` dominates demand at peak and collapses to zero the
   moment new replicas absorb the backlog, while `k2` stays inflated through the historical
   average (Priority 2). The controller sheds, the backlog returns, and the cycle repeats.

Mechanism 3 is the shed-to-one: after a drain, demand has lost its dominant term while
supply still reflects the peak.

## What a fix has to separate

Two questions are conflated today. Suggested direction, and the shape of the change I have
prototyped:

```
detector:    rates decide WHEN a replica is at its limit
measurement: tokens record WHAT that limit is
```

A replica is at its limit when it has a backlog at least `QueueLengthThreshold` deep, or
when its arrival rate has reached the service rate measured while it *was* backlogged. At
that moment its resident token count is a measurement of the limit — including at 16%
occupancy, which is the whole point. That measurement is stored **per workload bucket**
(model, accelerator, role, GPU count, request shape) as a running minimum, and every replica
of the bucket reads the same value.

Two constraints any implementation has to respect, both of which cost me a rewrite:

- **The value must be identical across replicas of a variant.** `aggregateByVariant` takes
  the MEDIAN of per-replica capacities. A figure that varies with each replica's own load is
  not commensurable across siblings — an idle replica's value blends with a backlogged one's
  and can lift variant capacity enough to turn a scale-up into a scale-down, reintroducing
  shed-to-one by a new route. A bucket ceiling makes the median a no-op.
- **The value must not move with the current cycle's load.** A capacity recomputed from this
  cycle's arrival rate changes every cycle, which is an oscillation waiting to happen. A
  stored ceiling moves only as a new measurement lowers the running minimum or age relaxes
  it upward — both slow by construction.

Also worth recording: λ and μ do not sit on the same time base. A completion happens one
residence time after the arrival that caused it, so an instantaneous λ against a
completion-derived μ reads as saturation on a replica that is coping during a ramp. λ needs
smoothing over `AvgTTFT + AvgOutputTokens × AvgITL`.

**No new metrics are required.** λ (`inference_extension_scheduler_attempts_total`, EPP),
μ (`vllm:request_generation_tokens_count`), occupancy, queue depth and KV capacity are all
collected today. Two of them are registered only when the throughput analyzer is enabled,
which a fix must not depend on.

An ITL model (`ITL(k) = A·k + B`, already fitted by the throughput analyzer) does **not**
cover this: on prefill-heavy traffic ITL can stay flat while TTFT and the queue explode. It
remains the better model for decode-bound workloads.

## What would close this

- A capacity estimate that binds below `k1` when the replica is compute-bound at low
  occupancy, rather than being discarded by `min(k1, k2)`.
- No shed while arrival rate is unchanged and only the queue has drained.

ev-shindin:
Fixes #1500 

The V2 saturation analyzer models a replica's capacity as `min(k1, k2)` in KV tokens,
where `k2` — the compute bound — is recorded as `tokensInUse` at the moment the replica
was seen queueing, then averaged into a rolling history. That measures a **KV stock**
while the constraint it stands for is a **rate**, and on prefill-heavy traffic the two
are unrelated: the engine exhausts prompt-token throughput and starts queueing while KV
occupancy is still low.

In the sustained 1000/250 comparison run the WVA leg was queueing, dropping requests and
cycling replicas at **16.2% average KV utilization** — a regime the current model cannot
express, because at 16% occupancy `demand/supply` reads as abundant headroom. Three
consequences, all observed in that run:

1. `tokensInUse` comes from `max_over_time(kv_cache_usage[1m])`, a peak — whenever the
   peak exceeds `kvCacheThreshold` (0.80), `k2 >= k1` and `min(k1, k2)` returns the memory
   bound, so the compute signal never binds at all.
2. Supply is over-stated exactly when load is highest, so utilization reads low and
   `spareCapacity` grows.
3. Demand deflates faster than supply: the queue term dominates demand at peak and
   collapses the moment new replicas absorb the backlog, while `k2` stays inflated via the
   historical average. The controller sheds to one replica and the cycle repeats.

Threshold tuning does not reach this — six threshold/window/policy legs were run against
that workload and none of them broke the cycling.

This PR adds a second, rate-anchored estimator behind a build-time constant that is
**false**, so nothing changes for any user until the validation in
`docs/plans/engine/rate-anchored-k2.md` is done. The design separates two questions the
occupancy estimator conflates: **rates decide when a replica is at its limit, tokens
record what that limit is.**

## Proposed Changes

- :gift: `internal/engines/analyzers/saturation_v2/rate_capacity.go` — rate-anchored `k2`.
  A replica is at its limit when it has a backlog at least `QueueLengthThreshold` deep, or
  when its arrival rate has reached the service rate measured while it *was* backlogged; at
  that moment its resident token count is a measurement of the limit. The measurement is
  stored **per workload bucket** (model, accelerator, role, GPU count, request shape) as a
  running minimum, and every replica of the bucket reads the same value. Two properties this
  buys, both learned from earlier iterations of this code: the value is identical across
  siblings, so `aggregateByVariant`'s MEDIAN is a no-op rather than a way for an idle
  replica to lift variant capacity and turn a scale-up into a scale-down; and it does not
  move with this cycle's load, so it cannot oscillate — the stored ceiling changes only as a
  new measurement lowers the running minimum or age relaxes it upward, both slow by
  construction.
- :gift: Selected by `EnableRateAnchoredK2`, a build-time constant — deliberately not a
  ConfigMap setting, since this is under evaluation against the incumbent and is not
  something to toggle in a running cluster. With it false the stores are never allocated and
  every path in the new file returns immediately; the existing estimator is untouched.
- :gift: Two new `k2Source` values (`RATE-now`, `RATE-learned`) so the offline replay can
  tell a limit measured this cycle from one carried over.
- :broom: `internal/collector/registration/rate_capacity.go` — single definition of
  `QueryRequestRate` and `QueryKvUsageInstant` (vLLM + SGLang), registered unconditionally
  by the saturation engine. Both were previously registered only when the throughput
  analyzer was enabled; this estimator must not depend on that analyzer. Both registrars now
  use register-if-absent helpers, so either order works and neither panics on the shared
  queries. **No new PromQL** — every metric this needs is already collected.
- :broom: λ is smoothed per replica over a residence time
  (`AvgTTFT + AvgOutputTokens × AvgITL`, both already collected), because a completion
  happens one residence time after the arrival that caused it — comparing an instantaneous λ
  against a completion-derived μ would read as saturation on a replica that is coping during
  a ramp. The weight comes from the actual sample gap, so an irregular optimize interval
  does not distort it; without latency data the rate passes through unsmoothed rather than
  being averaged by a made-up constant.
- :broom: Both new stores prune themselves on insert past `BucketPruneThreshold`, since
  nothing in the engine drives eviction for the analyzer's existing stores either.
- :gift: `docs/plans/engine/rate-anchored-k2.md` — problem, design, the damping argument,
  where the two estimators actually differ (entirely in the post-drain state), guards, known
  limitations, and the validation gate that must pass before the constant is flipped.

### Pre-review Checklist

- [ ] **E2E tests** for any new behavior — N/A while the constant is false: there is no
  reachable behavior change to assert. Unit and envtest coverage is in
  `rate_capacity_test.go` (store rate/ceiling semantics, arrival smoothing, residence time,
  role isolation, an oscillation-stability band, bounded growth) and
  `rate_capacity_integration_test.go` (what the store learns, flag-off equivalence with the
  current path, the post-drain divergence, and equal capacity for a backlogged replica and
  an idle sibling). The e2e gate belongs on the PR that flips the constant, alongside the
  cluster legs in the plan.
- [ ] **Docs PR** for any user-facing impact — N/A. A build-time constant, off; no API,
  ConfigMap field, metric or default changes.
- [ ] **Proposal PR** for any new enhancement or change to existing behavior — the design
  is in `docs/plans/engine/rate-anchored-k2.md` in this PR, and no existing behavior changes
  while the constant is false. Happy to split it into a proposal PR if maintainers would
  rather review the design separately from the code.

**Release Note**

```release-note
```

**Docs**

N/A — no user-visible impact in this PR.

### DIFF
--- a/docs/plans/engine/rate-anchored-k2.md
+++ b/docs/plans/engine/rate-anchored-k2.md
@@ -329,6 +329,45 @@ as replicas drift across it. Input and output thresholds are also separate now: 
 run an order of magnitude longer than generations, and bucketing input at the output
 thresholds put almost every real workload in one bucket.
 
+### The scale-down floor
+
+Capacity estimation alone cannot make a scale-down safe, because the decision is a
+counterfactual — would `N−1` replicas still cope? — and the token-space figures cannot
+answer it. Demand in resident tokens is measured at the current replica count and does
+not survive the removal being evaluated: residence rises, occupancy per replica rises
+with it, and past a point a queue appears from nothing. Arrivals are the one quantity
+the decision does not move, so the same question asked in rate space has an answer that
+holds.
+
+The constraint is on the aggregate service rate within a role, not on replicas within a
+variant:
+
+```
+sum over v in role of (N_v x mu_v)  >=  lambda_role / scaleDownBoundary
+```
+
+Per-variant would be wrong: shedding from one variant re-routes its traffic to a
+sibling, so a variant's own arrival rate is itself a function of the decision. Only the
+role's total stream is invariant. Heterogeneous accelerators fall out naturally, since
+`mu_v` is per bucket and the bucket key already carries accelerator and GPU count.
+
+The optimizer enforces it in `scaleDownVariantSet`, which is already role-scoped, and
+only bounds *how many* replicas may go — which variant gives one up is still the cost
+ordering's decision. One consequence worth expecting: when the expensive variant's
+replicas are individually large in service-rate terms, the floor can hold them and shed
+a cheap one instead. That is correct. Capacity you need cannot be removed merely
+because it is the capacity you would rather not pay for.
+
+If any variant holding replicas in a role has no calibrated `mu`, the role's available
+side is understated while its arrivals still count, which would hold replicas that are
+not needed. Partial knowledge is therefore treated as no knowledge and the constraint is
+skipped.
+
+For P/D, both roles see the same request stream and each gets its own constraint. A
+prefill replica completes few or no generations, so its service rate comes from prompts
+processed (`vllm:request_prompt_tokens_count`, `sglang:prompt_tokens_histogram_count`)
+rather than from the generation-tokens counter.
+
 ### Residual risks
 
 - **Mixed time bases.** Demand still uses `TokensInUse`, a 1-minute peak, while capacity uses
--- a/docs/plans/engine/rate-anchored-k2.md
+++ b/docs/plans/engine/rate-anchored-k2.md
@@ -1,0 +1,187 @@
+# Rate-anchored compute capacity (k2) — plan
+
+## Problem
+
+The V2 saturation analyzer models a replica's capacity as `min(k1, k2)` in KV tokens:
+
+- `k1 = TotalKvCapacityTokens × kvCacheThreshold` — the memory bound.
+- `k2` — the compute bound, currently recorded as `tokensInUse` at the moment the
+  replica was seen queueing (`computeK2` priority 1), then averaged into a rolling
+  history keyed by `model|accelerator|gpuCount|outputBucket`.
+
+`k2` measures a **KV stock** while the constraint it stands for is a **rate**. On a
+prefill-heavy workload the two are unrelated: the engine exhausts prompt-token
+throughput and begins queueing while KV occupancy is still low. In the sustained
+1000/250 comparison run the WVA leg was queueing, dropping requests, and cycling
+replicas at **16.2% average KV utilization** — a regime the current model cannot
+express, because at 16% occupancy `demand/supply` reads as abundant headroom.
+
+Three consequences follow, all observed in that run:
+
+1. **The compute bound is silently discarded.** `tokensInUse` derives from
+   `kv_cache_usage` which is `max_over_time(...[1m])` — a peak. Whenever the peak
+   exceeds `kvCacheThreshold` (0.80), `k2 ≥ k1` and `min(k1, k2)` returns the
+   memory bound, so the compute signal never binds.
+2. **Supply is over-stated exactly when load is highest**, so utilization reads low
+   and `spareCapacity = totalSupply − totalDemand/scaleDownBoundary` grows.
+3. **Demand deflates faster than supply.** The queue term
+   (`QueueLength × (avgInput + avgOutput)`) is the dominant part of demand at peak
+   and collapses to zero the moment new replicas absorb the backlog, while `k2`
+   stays inflated via the historical average. The controller then sheds to one
+   replica and the cycle repeats.
+
+Tuning thresholds cannot fix this: six threshold/window/policy legs were run
+against that workload and none of them broke the cycling.
+
+## Approach
+
+Add a second, **rate-anchored** estimator for `k2`, selected by an internal flag,
+leaving the existing estimator in place and untouched when the flag is off.
+
+When a replica has a backlog, its completion rate *is* its service rate — this
+holds whichever resource binds (prefill compute, decode bandwidth, memory), which
+is what makes it the right anchor for a workload-agnostic capacity model:
+
+```
+μ  = max completion rate observed for this workload bucket while QueueLength > 0
+λ  = this replica's arrival rate (EPP dispatch rate)
+k2 = KvUsageInstant × TotalKvCapacityTokens × (μ / λ)
+```
+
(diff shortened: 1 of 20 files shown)


deanlorenz (commented):
Nice fix — the detector (rates decide *when* we're at the limit) / measurement (resident
tokens record *what* it's worth) split is clean, and the per-bucket running-min ceiling
read identically by every replica is the key move: it collapses the variant median to one
value so an idle sibling can't lift capacity and re-trigger the shed-to-one loop.
Estimator is genuinely inert with the switch off, tests are thorough. Two asks before merge:

**1. Gate the query registration on the switch too.** `RegisterRateCapacityQueries` runs
unconditionally in `NewEngine`, independent of `EnableRateAnchoredK2`. Since
`QueryRequestRate`/`QueryKvUsageInstant` are already in the always-executed
`engineSpecificReplicaQueries` list, this changes execution, not just registration:
pre-PR with TA off, `Build()` returned "query not found" and no Prometheus call was made;
post-PR both run every replica-collection cycle. That's the default config (TA off), so
we ship a per-cycle Prometheus load increase even though nothing reads those fields until
the switch flips — contradicting the doc's "inert until the constant is flipped" framing.
No correctness impact (I checked — no always-on reader). Suggest gating
`RegisterRateCapacityQueries` on `EnableRateAnchoredK2`; or, if you mean to always collect
μ, note that in the doc.

**2. Please rebase onto current `main`.** A few commits behind, and #1486 touches the same
`NewEngine` (limiter-builder machinery). Should be a clean adjacency, but worth doing
explicitly and re-running gates.

Minor: `QueryKvUsageInstant` is collateral in the shared registrar (estimator never reads
it) — a one-line doc note would help; dissolves if you take the gating suggestion.

PR #1502, opened 2026-07-30
feat(controller): warn operator when a live ConfigMap edit can't change ThroughputAnalyzer registration

### LINKED ISSUE #1497: engine: warn when runtime ConfigMap analyzer-enable requires a controller restart
Part of #1492 (Epic A — Multi-analyzer pipeline & analyzer lifecycle).

Enabling an analyzer by editing the saturation ConfigMap at runtime is silently ignored until the controller restarts. Emit a warning (log + Kubernetes Event) so the operator knows a restart is required to pick up the change.

Fixed by #1502 (merged, 1d5553ee).

deanlorenz:
## What

Second half of the ThroughputAnalyzer (TA) gate-observability work. Analyzer registration is **frozen at controller startup** — `RegisterAnalyzer(throughput, …)` runs only if TA is enabled at boot, and the registry is never mutated afterward. The `ConfigMapReconciler` hot-updates the in-memory config on every ConfigMap edit, but that only affects *already-registered* analyzers. So **enabling TA by editing the ConfigMap at runtime silently does nothing** until the controller restarts, with no signal that the edit was inert.

This PR adds the signal (0.9 stopgap — a restart is an acceptable operator action for 0.9):
- a **Kubernetes Warning event** on the ConfigMap, reason `ThroughputAnalyzerRestartRequired`, and
- a **log line** telling the operator to restart the controller,

emitted when the live config's TA-enabled state diverges from the registration decision frozen at startup.

Commit 1 extracts `ThroughputAnalyzerEnabled()` onto `config.Config` (one shared predicate for the startup gate and the reconciler). It is otherwise a pure refactor with one deliberate change: rebasing onto #1486 (which added `AnalyzerScoreConfig.Type` / `EffectiveType()`), the extracted gate now matches on `aw.EffectiveType()` like every other analyzer-selection site, so `- type: throughput` enables TA as well as `- name: throughput` (covered by two new test cases). Commit 2 adds the detection + event/log.

## Scope

- "Disabled by default" needs no code (verified at all three layers) — unchanged.
- Out of scope: namespace-local ConfigMaps; the proper fix (removing the frozen-registration gate for true hot enable/disable) is tracked separately as post-0.9 lifecycle work.

## Testing

New `internal/config` + `configmap_reconciler` tests; envtest suite green; lint/gofmt/build clean.

```release-note
The controller now emits a Kubernetes Warning event and a log line advising a restart when a ConfigMap edit enables or disables the ThroughputAnalyzer at runtime, since analyzer registration is fixed at startup.
```


deanlorenz:
Heads-up for review: this branch was rebased past #1486 (ScalingPolicy Phase 1). Two interactions worth knowing:

- **`cmd/main.go`** had a real conflict (both #1486 and this branch's Commit 1 edit it) — resolved keeping both.
- **`internal/config/config.go`**: the extracted gate now uses `aw.EffectiveType() == "throughput"` to match #1486's sibling change in `engine_v2.go`, so it accepts `- type: throughput` as well as `- name: throughput`. The three-way merge couldn't surface this (it's new code #1486 never touched), so it's a deliberate hand-applied consistency fix, with two new test cases. Full rationale is in the `5614afb4` commit message.

ev-shindin (approved):
@deanlorenz thanks for this - nice, well-tested change. I reviewed it and found one real bug plus a few smaller items.

I pushed the fix straight to this branch (`552fda63`) via maintainer edit - no action needed on your side. Please take a look and let me know if you'd rather I back it out and open it as a separate PR.

deanlorenz:
Thanks for the careful review — the bootstrap-window bug is a real one, and gating on `ConfigMapsBootstrapComplete()` is the right fix (glad that mechanism was already there to hang it on). Happy to keep `552fda63` as-is on this branch rather than splitting it out — no need to back it out. Appreciate the structured-logging cleanup and the constant-drift guard test too.

PR #1503, opened 2026-07-30
fix(throughput,saturation): correctness guards for ThroughputAnalyzer and the liveness engine

deanlorenz:
## What

A batch of defensive correctness guards for the ThroughputAnalyzer / saturation pipeline (no behavior change on the happy path):

- **reject NaN `k`** in `ObservationWindow.Add` (a bad fit can't poison the observation window);
- **wire real per-replica metric freshness** in the collector (the staleness gate was previously inert);
- **share one ITL-model validator** across both fit tiers (removes divergent validation);
- **reject NaN / out-of-range KV usage** in `computeLocalDemand`;
- **harden the liveness engine** against a zero optimization interval and a zero `AnalyzedAt` timestamp, and extend the all-removed-vs-transient prune comment (the three follow-ups raised on the veto-liveness review).

## Testing

Unit tests per guard; `make test` full-repo green; lint 0; gofmt/build clean.

```release-note
Hardened the ThroughputAnalyzer and saturation liveness engine against NaN/out-of-range inputs, a zero optimization interval, and a zero analysis timestamp; per-replica metric freshness is now honored.
```

### DIFF
--- a/docs/developer-guide/multi-analyzer-pipeline.md
+++ b/docs/developer-guide/multi-analyzer-pipeline.md
@@ -251,7 +251,13 @@ same live-only filter.
 **Liveness.** An analyzer is live for the current cycle iff it produced a
 non-error, capacity-bearing result within the staleness window (a fixed
 multiple of the optimization interval, `analyzerLivenessStaleCycles` in
-`internal/engines/saturation/engine_v2.go`). A non-live analyzer — one that
+`internal/engines/saturation/engine_v2.go`). The resolved interval falls back
+to a 30s default whenever `Config` is absent **or** reports a non-positive
+value, so a misconfigured interval can never zero the staleness window and
+latch every analyzer non-live. An informative result with a zero-valued
+`AnalyzedAt` is treated as current (recorded as "now") rather than
+instantly-stale, so a forgotten timestamp on a future analyzer cannot
+silently disarm the veto. A non-live analyzer — one that
 has never produced a usable result, is currently erroring, or whose last
 usable result has aged past the staleness window — is excluded from the
 scale-down vote entirely: it neither vetoes nor constrains the safe-removal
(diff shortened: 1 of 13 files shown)


deanlorenz:
Thanks — good catch. Confirmed the bug: `worstFreshnessStatus` gated the *age* update on `hasTimestamp` but not the *severity* comparison, so an absent-by-design metric (no EPP, prefix caching off) always won the max via its `"missing"` classification and masked a genuinely stale driving metric from the `sanity.go` stale-metrics gate. `27523605`'s fix (skip absent timestamps entirely, fall back to `"missing"` only when none are present) is correct and minimal, and the two new regression tests exercise exactly the fixed branch. Happy to keep it on this branch.

ev-shindin (approved):
@deanlorenz Thanks for the thorough hardening pass — pushed one small follow-up commit
(`fix(collector): skip absent-by-design metrics in per-replica freshness`) to the branch.

**Why:** `worstFreshnessStatus` folded all nine timestamps into the rollup, so a zero
(never-scraped) one counted as `"missing"`. Several metrics are absent by design — EPP
arrival rate with no EPP, prefix-cache / cache-config when caching is off. Since `"missing"`
outranks `"stale"`, such a replica reported `"missing"` with near-zero `Age`, and a genuinely
stale driving metric got masked: `CheckModelMetrics` keys on `== "stale"`, so the sanity issue would never fire when any optional metric was absent — the common non-EPP case.

**Change:** worst status over *present* timestamps only, falling back to `"missing"` when none are present. Two regression tests added (absent stays `"fresh"`; stale isn't masked) — both fail pre-fix. `go vet` + full `internal/collector` pass, gofmt clean.

Happy to switch to capability-based exclusion if you'd prefer. Minor untouched test gaps if you want them here: Tier-2 `validITLModel` rejection, `validITLModel` with `Inf B`, and
`computeLocalDemand`'s non-positive-capacity / negative-ITL skips.

PR #1509, opened 2026-07-31
fix(crd): restart when KEDA or LWS CRDs are installed after startup

### LINKED ISSUE #1461: fix: KEDA/LeaderWorkerSet CRD presence checked only at startup — installing either after WVA starts silently requires a restart
## Summary

`CheckKEDACRD` and `CheckLeaderWorkerSetCRD` (`internal/utils/crd/crd.go:70-80`) probe CRD availability **once at startup**. Their results gate manager wiring that is only built during setup, so installing KEDA or LWS *after* WVA has started has no effect until the controller restarts. Both problems are flagged by in-code TODOs (`crd.go:71`, `crd.go:77`).

## What's gated (all at startup only)

- **`kedaEnabled`** (`cmd/main.go:212`) → `ScaledObjectReconciler` registration (`cmd/main.go:573`, *only if enabled*), field indexers (`cmd/main.go:384`), pod-locator ScaledObject lookups (`cmd/main.go:220`), engine `KEDAEnabled` (`cmd/main.go:620`).
- **`lwsEnabled`** (`cmd/main.go:200`) → LeaderWorkerSet scheme registration (`cmd/main.go:202`).

## Impact

Install-order dependency: deploying WVA before KEDA/LWS silently leaves the entire ScaledObject reconciliation path (or LWS support) disabled. Startup logs the disabled state, but a *later* CRD install produces **no new signal** — the operator has no runtime indication that a restart is required. This is common in GitOps / bootstrap flows where component ordering is not guaranteed.

## Suggested fix (increasing effort)

1. **Surface it better:** on the disabled path, log that a controller restart is required after installing the CRD, and/or emit an Event / condition.
2. **Periodic re-check + graceful self-restart:** re-run the discovery probe on an interval; when the CRD appears, exit non-zero so the Deployment restarts and re-runs the startup wiring. Preferred — avoids unsupported dynamic scheme / informer mutation after `mgr.Start()`.
3. *(Stretch)* Watch `CustomResourceDefinition` and trigger the restart on create.

## Acceptance criteria

- [ ] Installing KEDA / LWS after WVA starts results in the corresponding support becoming active without a manual restart, **or** a clear operator-facing signal (log / Event) that a restart is required.
- [ ] The chosen mechanism does not attempt unsupported post-`Start()` scheme / informer mutation.

## Related

Touches the annotation-based ScaledObject discovery path; see the discovery skip at `internal/utils/variant.go:235`.

cne27:
Fixes #1461

## Summary

This change enables KEDA and LeaderWorkerSet support to become active when those CRDs are installed after the controller has already started. Today both are probed once at startup, so installing either one later leaves that support disabled until someone manually restarts the pod. This is easy to hit in GitOps or bootstrap flows, where component ordering is not guaranteed.

## Changes Made

This change adds `internal/utils/crd/watcher.go`. The `Watcher` type is a manager `Runnable` that probes discovery once a minute for the group-versions that were absent at boot, using the same `DetectCRDInstalled` function the startup path already calls. Targets present at boot are filtered out, and probe errors are retried rather than read as absence. On detection it returns an error wrapping `ErrRestartRequired`, which shuts the manager down so the Deployment restarts the controller with correct wiring. `NeedLeaderElection()` returns false so every replica restarts rather than leaving standbys stale.

In `cmd/main.go`, the change builds the targets from the existing `lwsEnabled` and `kedaEnabled` results, registers the watcher, matches `ErrRestartRequired` in the `mgr.Start` error path so the restart is not logged as `problem running manager`, and rewords the two "CRD not found" startup lines. It adds no dependencies and no RBAC.

## Test evidence

```
$ make test
ok   .../internal/utils/crd   0.471s   coverage: 77.4% of statements

$ make lint
0 issues.
```

Seven new unit tests in `internal/utils/crd`. `make test-e2e-smoke` on kind passed 15 of 15. The feature also fired on its own during `make deploy-e2e-infra`, which installs WVA before KEDA: the controller detected the CRD one interval after boot, restarted, and came back with ScaledObject discovery enabled.

## Notes

- **Exit code.** Non-zero, as suggested fix 2 specifies. Exiting zero would report `Completed` rather than `Error`, and I am happy to switch.
- **Install ordering.** `deploy/install.sh` installs WVA before KEDA, so a fresh `make deploy-e2e-infra` now restarts the controller about a minute in. This is harmless, but it makes the restart routine rather than exceptional on the project's own install path. Fixing the ordering there is the better long-term answer, and it is out of scope here.

### Pre-review Checklist

- [x] **E2E tests** for any new behavior. Verified on kind; no new spec added.
- [ ] **Docs PR** for any user-facing impact. N/A, covered by the release note.
- [ ] **Proposal PR** for any new enhancement or change to existing behavior. N/A, implements #1461.

```release-note
WVA now detects KEDA and LeaderWorkerSet CRDs that are installed after the controller has started. Previously, installing either CRD afterwards left the corresponding support silently disabled until the controller was manually restarted. The controller now restarts itself once the newly installed API is served.
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -177,7 +177,8 @@ func main() {
 		}
 		setupLog.Info("LeaderWorkerSet CRD detected - support enabled")
 	} else {
-		setupLog.Info("LeaderWorkerSet CRD not found - support disabled (Deployment-only mode)")
+		setupLog.Info("LeaderWorkerSet CRD not found - support disabled (Deployment-only mode); " +
+			"support requires a controller restart, which happens automatically if the CRD is installed later")
 	}
 
 	// Detect KEDA for annotation-based ScaledObject discovery
@@ -185,11 +186,27 @@ func main() {
 	if kedaEnabled {
 		setupLog.Info("KEDA ScaledObject CRD detected - annotation-based ScaledObject discovery enabled")
 	} else {
-		setupLog.Info("KEDA ScaledObject CRD not found - annotation-based discovery limited to HPAs")
+		setupLog.Info("KEDA ScaledObject CRD not found - annotation-based discovery limited to HPAs; " +
+			"support requires a controller restart, which happens automatically if the CRD is installed later")
 	}
 	// Gate the pod locator's ScaledObject lookups on KEDA availability. Set before
 	// the saturation engine goroutine constructs its locator.
 	locator.SetKEDAEnabled(kedaEnabled)
+
+	crdTargets := []crd.Target{
+		{
+			CRDName:       "leaderworkersets.leaderworkerset.x-k8s.io",
+			GroupVersion:  "leaderworkerset.x-k8s.io/v1",
+			Kind:          "LeaderWorkerSet",
+			PresentAtBoot: lwsEnabled,
+		},
+		{
+			CRDName:       "scaledobjects.keda.sh",
+			GroupVersion:  "keda.sh/v1alpha1",
+			Kind:          "ScaledObject",
+			PresentAtBoot: kedaEnabled,
+		},
+	}
 
 	tlsOpts := []func(*tls.Config){
 		func(c *tls.Config) {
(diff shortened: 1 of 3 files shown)


lionelvillard:
thanks @cne27! Overall LGTM. To be of the safe side, can you please add an opt-out command line flag disabling auto-restart? 

FYI @mamy-CS

cne27:
Thanks for the review @lionelvillard! Added the `--disable-crd-auto-restart` opt-out flag as requested. Thanks again!

cne27:
@lionelvillard the `--disable-crd-auto-restart` opt-out flag is in as requested, and CI is green. Could you drop a `/lgtm` + `/approve` when you get a chance? @mamy-CS in case you have thoughts too.

lionelvillard (approved):
/lgtm
/approve

thanks @cne27 , sorry for the delay.

PR #1510, opened 2026-08-01
Enforce namespace-scoped GPU inventory on the V2 optimizer path (part of #1441)

jia-gao:
Follow-up to #1222, per the discussion there. Carries namespace-scoped GPU inventory onto the V2 optimizer path.

**Stacked on #1222.** It is the base, so this shows its commits too. Only the last commit, `feat: enforce namespace-scoped GPU inventory on the V2 path`, is new here. Ready for review, but please merge #1222 first.

### Why now

#1441 removes V1 at v0.11.0 and defaults to V2 at v0.9.0. Namespace isolation currently applies to V1 only, so without this it disappears with V1.

### What it does

`NamespaceInventory` implements `NamespaceAwareInventory` and `NamespaceLimiter` implements `ConstraintProvider`, following the contract #1129 established for quotas. An excluded namespace is omitted (open), a namespace with no selector and no default is present with an empty map (deny-all), and a type absent from a namespace's pool is denied rather than falling through to the cluster aggregate. Physical inventory is always finite, so the unlimited sentinel is never emitted. Usage is remapped via `chargeBucket`, so the excluded-namespace accounting carries over to V2.

No engine special-casing was needed. Since the inline-schema migration in #1222 made the selected mode the pipeline's limiter, `gpuConstraintProviders` picks the namespace limiter up directly, and the "not enforced on V2" warning is gone.

### Two things worth review

**Shared default pool.** It is one physical pool, but V2 consumes a static budget per namespace. Handing each namespace the whole pool would let them all allocate it independently and would inflate the cluster aggregate derived from summing the per-namespace pools. So its remaining capacity is split evenly: the shares sum to exactly the pool's remaining capacity, and it can never be overcommitted. This is the one place V2 diverges from V1, whose live shared counter is first-come-first-served and cannot be expressed as a static budget. A named selector avoids the split.

**Cluster aggregate.** `ComputeConstraints` derives it from the active namespace pools, not the static all-buckets `GetResourcePools` sum, so it cannot advertise capacity no active namespace can draw from or feed an over-broad cap into a later intersection.

### Question

@ev-shindin you mentioned having drafted a V2 follow-up around a composable `NamespaceConstraintProvider` in the constraint chain. If that supersedes this, say so and I will close this in favor of it. I went with `ComputeConstraints` on the existing `NamespaceLimiter` since #1129's plumbing was already generic enough, but I have no attachment to the shape.

Composing this with another provider as `min(physical, quota)` is still #1003's job.


jia-gao:
Rebased onto latest main (picks up the Go 1.26 bump), CI is green.

Worth flagging that v0.9.0 shipped, which per #1441 is where V2 becomes the default. Namespace isolation currently only applies to V1, so as of that release the default path has no per-namespace enforcement. That makes this more time-sensitive than when I opened it.

@ev-shindin no rush on the review, but the one answer that would help most is whether the `NamespaceConstraintProvider` you drafted supersedes this. If it does I'll close this and stay out of the way.

PR #1511, opened 2026-08-02
test(throughput): cover ITL-model / demand / supply guard branches

deanlorenz:
## What

Adds unit-test coverage for several already-shipped defensive guards in the ThroughputAnalyzer
that were flagged as untested during the #1503 review. **Test-only — no production code changes.**

During the #1503 review, @ev-shindin noted three untested guard branches (and left them as optional
follow-ups). This PR covers those three plus one adjacent supply-path gap found while adding them:

- **`validITLModel` — `Inf B` rejection.** Symmetric with the already-tested `Inf A` case; the
  `math.IsInf(b, 0)` branch was previously uncovered.
- **Tier-2 fit rejection in `resolveITLModel`.** A single below-baseline replica produces a
  negative slope that `validITLModel` rejects, exercising the `n>0 && sumK2>0` fit-then-reject
  path — distinct from the existing "all replicas idle" (`n==0`) test, which never reaches the fit.
- **`computeLocalDemand` skip guards.** Non-positive `TotalKvCapacityTokens`, and a finite (non-NaN)
  negative predicted ITL — distinct from the existing NaN-coefficient case.
- **`computeVariantSupply` direct coverage.** The same non-positive-capacity guard on the supply
  path, previously only exercised indirectly through an `Analyze`-level test. Adds a direct
  aggregate case plus the skip case.

Each new spec was verified to be non-vacuous (it fails if the guard is removed) and to reach its
intended branch by control-flow + arithmetic, not just a green suite. No behavior changed; no new
test failed against current code, so no latent bug was uncovered.

## Testing

- `make test` — PASS (throughput package coverage 93.2% → 93.4%)
- `make lint` — 0 issues
- `go test ./internal/engines/analyzers/throughput/... -race` — PASS
- `go build ./...` — clean · `gofmt` — clean

## Notes

The matching capacity/ITL guards in `checkVariantGPSMismatch` (a diagnostic-only path) remain a
known coverage gap, intentionally left for a separate follow-up.

```release-note
NONE
```

### DIFF
--- a/internal/engines/analyzers/throughput/analyzer_test.go
+++ b/internal/engines/analyzers/throughput/analyzer_test.go
@@ -2097,4 +2097,20 @@ var _ = Describe("computeLocalDemand", func() {
 		total := computeLocalDemand([]domain.ReplicaMetrics{replicaAt(0.5)}, shape, nanModel)
 		Expect(total).To(Equal(0.0))
 	})
+
+	It("skips a replica with non-positive TotalKvCapacityTokens", func() {
+		noCapacity := replicaAt(0.5)
+		noCapacity.TotalKvCapacityTokens = 0
+		total := computeLocalDemand([]domain.ReplicaMetrics{noCapacity}, shape, model)
+		Expect(total).To(Equal(0.0))
+	})
+
+	It("skips a replica whose model produces a finite non-positive ITL", func() {
+		// B negative enough that A*k+B <= 0 at k=0.5 without A or B being NaN/Inf —
+		// distinct from the existing NaN-ITL case, which uses a NaN model coefficient.
+		negativeITLModel := ITLModel{A: 0.01, B: -0.1}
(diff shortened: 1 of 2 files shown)


ev-shindin (approved):
Nice targeted work. I checked each new spec by deleting the guard it targets and re-running: `rejects Inf B`, the Tier-2 `validITLModel` gate and the non-positive-ITL skip are each the only detector in the suite for their branch.

Two didn't survive that check, so I pushed a fixup (`081d20dd`):

- **`computeLocalDemand` non-positive-capacity was vacuous.** The term is `KvUsageInstant × capacity / KVreq / ITL`, so capacity `0` contributes `0` either way and the spec stayed green with the guard deleted. Now `-65536`.
- **`computeVariantSupply` happy path had no detection power.** `total > 0` on one replica admits anything — dropping `DefaultKSat`, dropping `/itlSat`, and returning `sum` for `sum/n` all survived it. Now asserts the closed form exactly over two replicas of differing capacity plus one with no KV capacity, so `nKV (2) != len(metrics) (3)` pins the denominator too.

PR #1516, opened 2026-08-06
refactor(pipeline): derive the per-model anchor on demand; add analyzer enablement

deanlorenz:
## Summary

Reshapes the multi-analyzer optimizer pipeline so the per-model **anchor** (the topology/identity carrier used for sizing and sorting) is **derived on demand** from the analyzer ballot rather than pre-computed and stored. The engine passes the enabled-analyzer list as the ballot; a getter merges, per variant (keyed by variant name), the identity/metadata from the saturation carrier with the sizing fields from the binding analyzer. This is the structural groundwork for running additional analyzers (e.g. the throughput analyzer) alongside saturation. It is **single-vote only** (0 or 1 enabled analyzers) and changes **zero combine arithmetic** — the multi-analyzer combine lands in a stacked follow-up.

## What changes

- **Ballot enablement tag** — analyzer results carry an `Enabled` flag; the pipeline builds the ballot from the enabled set.
- **On-demand anchor** — derived by a getter via a per-variant merge; no anchor field is persisted on the request.
- **Analyzer enablement** — three configs: saturation-only (default, unchanged), throughput-analyzer-only (saturation as a non-voting topology carrier), both-enabled (two-vote path turned on by the follow-up).
- **Queueing-model path** — the optimizer now refuses the queueing-model optimize path (holds, no-op) instead of dispatching it; folding QM in as a first-class voting analyzer is deferred.
- **Scale-from-zero** — the throughput analyzer emits a per-replica-capacity-only signal for previously-live zero-replica variants.
- **Developer guide** updated for the on-demand anchor + saturation-as-carrier model.

## Behavior / compatibility

Default saturation-only decisions are **byte-for-byte unchanged**, frozen by the characterization goldens in this diff (decision-set identity keyed by variant name). Opt-in: nothing beyond saturation is enabled by default.

## Testing

Characterization golden harness + aggregated (RoleBoth), disaggregated (P/D), and quota-constrained goldens, plus the two-analyzer combine characterization spec — all green after every commit. `make test` / `make lint` / `go build` / `gofmt` clean; DCO signed.

## Stacking / ordering

The bottom 5 commits are the characterization goldens (also proposed standalone as #1513). This branch was rebased onto current `main` before #1513 merged, so those goldens ride in this diff; once #1513 merges, a rebase drops them and this PR reduces to the 5 refactor commits. A follow-up PR (stacked on this branch) adds the multi-analyzer combine, the combine-arithmetic fixes, per-iteration dynamic re-binding, and liveness/notation hardening.

### DIFF
--- a/docs/developer-guide/multi-analyzer-pipeline.md
+++ b/docs/developer-guide/multi-analyzer-pipeline.md
@@ -37,8 +37,8 @@ over it via shared free functions in `internal/engines/pipeline/`.
                            ▼
 ┌──────────────────────────────────────────────────────────┐
 │ Engine: run analyzers, build per-analyzer slice          │
-│ Saturation V2 (always first), then each registered       │
-│ non-saturation analyzer in registration order:           │
+│ Saturation V2 (always run — the (a)/identity carrier),   │
+│ then each registered non-saturation analyzer:            │
 │   • skip if Enabled:false                                │
 │   • Analyze(ctx, input) → *AnalyzerResult                │
 │   • applyUniversalThreshold(result, scaleUp, scaleDown)  │
@@ -161,9 +161,15 @@ no entry at all does not run and is not included in the result slice,
 exactly as if `enabled: false` had been set. This prevents a
 registered-but-unconfigured analyzer from returning `SpareCapacity=0` and
 silently vetoing scale-down, since the per-role scale-down decision requires
-every analyzer in the slice to agree. Saturation is exempt from this gate —
-it is always run, independent of `analyzers` config, because the engine
-identifies it by name before this check applies.
+every voting analyzer in the slice to agree. Saturation is always run
+regardless of `analyzers` config — the engine identifies it by name and
+appends it as the identity/(a) carrier that supplies per-variant metadata
+(`Cost`, `AcceleratorName`, `Role`) for every configured variant. Its *vote*,
+however, is opt-in like any other analyzer's: saturation votes in the combine
+math only in the default single-analyzer config (no explicit `analyzers` list)
+or when its name is explicitly enabled. A `[throughput]`-only config leaves
+saturation present as a non-voting carrier — it is pruned from the voting
+subset (`votingResults`) and neither vetoes nor constrains scale-down.
 
 ---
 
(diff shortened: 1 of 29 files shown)


ev-shindin (approved):
Nice refactor — I checked the characterization goldens by mutation rather than by eye, and they genuinely pin behaviour. The default saturation-only path really is unchanged.

I've pushed a commit to this branch with fixes and regression tests for what I found, all in the newly opt-in throughput path. Happy to split it out if you'd rather review it separately.

**The blocking one — scale-from-zero has to carry `Role`.** The SFZ complement left `Role` unset, on the reasoning that the anchor merge sources identity fields from saturation. That holds for `Cost`/`AcceleratorName`/`ReplicaCount`, but `distributeDemandByRole` and `aggregateRoleCapacities` consume the same slice later in `Analyze`, before any merge, and both canonicalise `""` → `RoleBoth`. On a P/D model that adds a phantom `both` bucket: decode demand gets split across one more role than exists, and `pick("both")` finds no anchor variant, returns `""`, clears `allPicked` and `break`s the *whole* allocation loop. So a disaggregated model with any variant at zero replicas produced no scale-up decisions at all — precisely the case SFZ was added for. Fix is `Role: st.role`.

**Refusal reporting.** The queueing-model hold is correct, but every held VA reported `OptimizationReady=True` / "no scaling change needed" with no event, so a cluster that had stopped autoscaling looked healthy. Now a distinct `OptimizationRefused` reason plus a Warning event, raised from the refusing path so it wins the one-event-per-VA-per-cycle budget over lesser notices. Worth knowing: nothing in the tree reads `TypeOptimizationReady` back, so the event is the whole of the operator-visible signal.

**A scale-to-zero hazard I hit while in there.** For a merely-held variant, `targetReplicas` falls back to `Status.DesiredOptimizedAlloc`, then to `currentAllocations` — which is created but never written anywhere — and only then to the live scale target, and that last rescue sat inside the accelerator-unresolved branch. A VA with a resolved accelerator but no prior replica count published `wva_desired_replicas=0`, which KEDA reads as scale-to-zero. Pre-existing, but the refusal makes it routine: every model now takes the no-decision path each cycle.

Plus three smaller ones: nothing asserted the `Enabled` tag the engine derives from config (hardcoding it `true` broke no test), an explicit analyzer list omitting saturation silently drops its scale-down veto and now logs, and `rescaleModelDecisions` was the one `bindingAnchor` call site without a nil guard.

PR #1517, opened 2026-08-06
Add KEDA autoscaling setup script for vLLM metrics

asm582:
## Summary

Adds a setup script (`hack/setup-keda-vllm-scaling.sh`) that creates KEDA ScaledObjects to autoscale deployments on raw vLLM metrics.

## What this does

- Creates a KEDA `ScaledObject` to autoscale on `vllm:kv_cache_usage_perc` and `vllm:num_requests_waiting`
- Configures sensible defaults: 1-10 replicas, 15s polling interval
- Automatic prerequisite validation (KEDA, TriggerAuthentication, target resource)
- Prevents accidental patching of existing ScaledObjects

## Usage

```bash
# Deployment
./hack/setup-keda-vllm-scaling.sh my-namespace my-decode-deployment

# LeaderWorkerSet
./hack/setup-keda-vllm-scaling.sh my-namespace my-lws \
  --target-kind LeaderWorkerSet \
  --target-api-version leaderworkerset.x-k8s.io/v1alpha1

# Preview
./hack/setup-keda-vllm-scaling.sh my-namespace my-deployment --dry-run
```

## Thresholds

- **KV Cache: 60%** — Router's KV cache scorer (score = 1 - usage/100) degrades to 0.4 at this utilization, indicating meaningful deprioritization
- **Queue Depth: 30 requests** — Calculated from GPU pod startup time (180s) / KEDA polling (15s) × 1.5 buffer, ensuring new replicas arrive before queue saturation

## Testing

- Tested on Deployment resources only
- Does not support scale-to-zero

## Documentation

Full usage and troubleshooting guide available in `hack/SETUP_KEDA_VLLM_SCALING.md`

### DIFF
--- a/hack/SETUP_KEDA_VLLM_SCALING.md
+++ b/hack/SETUP_KEDA_VLLM_SCALING.md
@@ -1,0 +1,118 @@
+# Setup KEDA Autoscaling on vLLM Metrics
+
+This script creates a KEDA `ScaledObject` to autoscale deployments on raw vLLM metrics (`vllm:kv_cache_usage_perc`, `vllm:num_requests_waiting`) instead of pool-averaged EPP metrics.
+
+## Why
+
+Raw vLLM metrics provide finer-grained autoscaling signals than pool averages. This enables proactive scaling based on:
+- **KV Cache Utilization** (60%): Before endpoints become deprioritized by the router
+- **Queue Depth** (30 requests): Before queue exceeds capacity during pod startup
+
+## Prerequisites
+
+- KEDA installed in the cluster
+- `TriggerAuthentication` named `prometheus-auth` in the target namespace (with bearer token for Prometheus)
+- vLLM pods exposing metrics (scraped into Prometheus)
+- Target resource must exist: `Deployment`, `StatefulSet`, `LeaderWorkerSet`, or other scalable resource
+
+## Usage
+
+### Create ScaledObject for a Deployment
+
+```bash
+./scripts/setup-keda-vllm-scaling.sh my-namespace my-decode-deployment
+```
+
+### Create ScaledObject for a LeaderWorkerSet
+
+```bash
+./scripts/setup-keda-vllm-scaling.sh my-namespace my-lws-resource \
+  --target-kind LeaderWorkerSet \
+  --target-api-version leaderworkerset.x-k8s.io/v1alpha1
+```
+
+### Dry-run (preview without creating)
+
+```bash
+./scripts/setup-keda-vllm-scaling.sh my-namespace my-deployment --dry-run
+```
+
+## What it does
+
+1. Validates prerequisites (kubectl, jq, namespace, target resource, TriggerAuthentication)
+2. Checks if a ScaledObject already exists (if so, asks you to delete it first)
+3. Generates a `ScaledObject` manifest with vLLM-based triggers:
+   - **KV Cache Utilization**: 60% threshold (average across pods)
+   - **Queue Depth**: 30 requests threshold (average across pods)
+4. Shows the manifest and asks for confirmation before applying
+5. Verifies creation and shows status
+
+## Cleanup
+
+To remove autoscaling:
+
+```bash
+kubectl delete scaledobject -n <namespace> <deployment-name>-saturation
+```
+
+## Thresholds explained
+
+### KV Cache @ 60%
+
+The router's load-balancing uses a KV cache scorer: `score = 1 - (usage / 100)`. At 60% cache utilization, this score drops to 0.4, indicating meaningful degradation in routing quality. This threshold triggers scale-up before endpoints become strongly deprioritized (70%+), allowing proactive capacity increases.
+
+### Queue Depth @ 30 requests
+
+Calculated from GPU pod startup time (typically 180s) divided by KEDA polling interval (15s), with 1.5× buffer:
+```
+(180s / 15s) × 1.5 = 18 requests ≈ 30 (with margin)
+```
+
+Setting to 30 ensures new pods arrive before the queue explodes, while respecting the router's soft load-balancing.
+
+## Troubleshooting
+
+**"TriggerAuthentication 'prometheus-auth' not found"**
+
+Create it:
+```bash
+kubectl create secret generic prometheus-token -n <namespace> --from-literal=token=<bearer-token>
+kubectl apply -f - <<EOF
+apiVersion: keda.sh/v1alpha1
+kind: TriggerAuthentication
+metadata:
+  name: prometheus-auth
+  namespace: <namespace>
+spec:
+  secretTargetRef:
+    - parameter: bearerToken
+      name: prometheus-token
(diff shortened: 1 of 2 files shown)


asm582:
## Note on Resource Support

While the PR description mentions testing on Deployment, **the script also supports LeaderWorkerSet and other scalable resources** via the `--target-kind` and `--target-api-version` flags.

Example for LeaderWorkerSet:
```bash
./hack/setup-keda-vllm-scaling.sh my-namespace my-lws   --target-kind LeaderWorkerSet   --target-api-version leaderworkerset.x-k8s.io/v1alpha1
```

The limitation is that this has only been tested on Deployment resources; other resource types should work but may need validation in your environment.

hack/setup-keda-vllm-scaling.sh
```diff
@@ -0,0 +1,294 @@
...
+    - type: prometheus
+      name: kv-cache
+      metricType: AverageValue
+      authenticationRef:
+        name: prometheus-auth
+      metadata:
+        serverAddress: "https://thanos-querier.openshift-monitoring.svc.cluster.local:9091"
+        authModes: bearer
+        query: |
+          avg(vllm:kv_cache_usage_perc{namespace="$NAMESPACE",pod=~"$POD_REGEX"})
```
mamy-CS: should it be `sum` here with threshold 0.6 or keep `avg` and switch to `metricType: Value`? Since `vllm:kv_cache_usage_perc` is a 0–1 fraction. So threshold 60 on the next line is ~100× too high. HPA divides the query result by the replica count

hack/setup-keda-vllm-scaling.sh
```diff
@@ -0,0 +1,294 @@
...
+    - type: prometheus
+      name: queue-size
+      metricType: AverageValue
+      authenticationRef:
+        name: prometheus-auth
+      metadata:
+        serverAddress: "https://thanos-querier.openshift-monitoring.svc.cluster.local:9091"
+        authModes: bearer
+        query: |
+          avg(vllm:num_requests_waiting{namespace="$NAMESPACE",pod=~"$POD_REGEX"})
```
mamy-CS, line 232: Same comment as above

hack/setup-keda-vllm-scaling.sh
```diff
@@ -0,0 +1,294 @@
...
+    # At 60% usage, score drops to 0.4, indicating meaningful degradation in routing quality.
+    # Triggers scale-up before endpoints become strongly deprioritized (70%+).
+    # This threshold respects the router's soft load balancing and scales proactively.
+    - type: prometheus
+      name: kv-cache
+      metricType: AverageValue
+      authenticationRef:
+        name: prometheus-auth
+      metadata:
+        serverAddress: "https://thanos-querier.openshift-monitoring.svc.cluster.local:9091"
```
mamy-CS, line 210: Add the service-CA bundle to the TriggerAuthentication (ca param) or set `unsafeSsl: "true"`. Also, worth noting that the address is OpenShift specific too.

hack/setup-keda-vllm-scaling.sh
```diff
@@ -0,0 +1,294 @@
...
+    log_warn "To recreate with new triggers, first delete the existing one:"
+    log_warn "  kubectl delete scaledobject -n $NAMESPACE $SCALEDOBJECT_NAME"
+    exit 1
+fi
+
+log_success "No existing ScaledObject found"
+
+# === BUILD SCALEDOBJECT MANIFEST ===
+
+POD_REGEX="^${TARGET_NAME}-[a-z0-9-]*$"
```
mamy-CS: if both foo and foo-v2 exist, foo-... also matches foo-v2-* pods, so the trigger sums the wrong workload. Selecting on a pod label (app / llm-d selector) would be safer than a name prefix, I believe.

hack/setup-keda-vllm-scaling.sh
```diff
@@ -0,0 +1,294 @@
+#!/usr/bin/env bash
+#
+# Create a KEDA ScaledObject to autoscale on raw vLLM metrics
+# (vllm:kv_cache_usage_perc, vllm:num_requests_waiting) instead of EPP pool aggregates.
+#
+# Usage:
+#   ./scripts/setup-keda-vllm-scaling.sh <namespace> <target-resource-name> [--dry-run] [--target-kind KIND] [--target-api-version VERSION]
```
mamy-CS: nit, script lives in hack/, not scripts/

hack/setup-keda-vllm-scaling.sh
```diff
@@ -0,0 +1,294 @@
...
+fi
+
+log_success "Prerequisites validated"
+
+# === CHECK TARGET RESOURCE EXISTS ===
+
+log_info "Looking for $TARGET_KIND: $TARGET_NAME in namespace/$NAMESPACE..."
+
+RESOURCE_NAME=$(echo "$TARGET_KIND" | tr '[:upper:]' '[:lower:]' | sed 's/set$/sets/')
+[ "$RESOURCE_NAME" = "leaderworkersets.x-k8s.io" ] && RESOURCE_NAME="leaderworkersets"
```
mamy-CS: This looks unreachable. `$RESOURCE_NAME` is derived from the kind (leaderworkersets), so it's never leaderworkersets.x-k8s.io. Can drop it.

hack/SETUP_KEDA_VLLM_SCALING.md
```diff
@@ -0,0 +1,118 @@
...
+- `TriggerAuthentication` named `prometheus-auth` in the target namespace (with bearer token for Prometheus)
+- vLLM pods exposing metrics (scraped into Prometheus)
+- Target resource must exist: `Deployment`, `StatefulSet`, `LeaderWorkerSet`, or other scalable resource
+
+## Usage
+
+### Create ScaledObject for a Deployment
+
+```bash
+./scripts/setup-keda-vllm-scaling.sh my-namespace my-decode-deployment
```
mamy-CS: ./scripts/... -> ./hack/...

hack/setup-keda-vllm-scaling.sh
asm582: This is a temp script.

hack/SETUP_KEDA_VLLM_SCALING.md
asm582: This script is intended to be in hack folder

hack/setup-keda-vllm-scaling.sh
asm582: Thanks, done some fixes. LWS is not tested, so I assume there will be more changes.

hack/setup-keda-vllm-scaling.sh
asm582: Tried to do the best effort fix for this temp script.

mamy-CS:
@asm582 if this is ready, it will be good to merge.

Issue #1458, opened 2026-07-23
test(e2e): add coordinator / gpu-rebalance e2e coverage


mamy-CS:
## Problem

The Coordinator loop (and its only plugin, `gpu-rebalance`) has **no automated end-to-end coverage**. It ships disabled by default (`EXPERIMENTAL_COORDINATOR_ENABLED: "false"` in `config/base/manager/manager-configmap.yaml`), and nothing in `test/e2e/` enables or exercises it — a grep for `coordinator` / `gpu-rebalance` / the flag across the suite returns nothing. Every existing e2e spec covers the saturation/throughput engine → `wva_desired_replicas` → KEDA path or supporting infra, not `internal/coordinator/...`.

Net effect: the coordinator's rebalancing logic is protected only by unit tests plus whatever manual validation a given PR happens to do. This surfaced while reviewing #1457 (optimistic-locking fix for ceiling patches) — the changed code path had unit coverage but had to be validated by hand on a live cluster because CI exercises none of it.

> **Note:** The Coordinator is currently an **experimental, opt-in feature** (gated behind `EXPERIMENTAL_COORDINATOR_ENABLED`, off by default, and it logs an experimental-feature warning on startup). This issue is about not leaving even an experimental feature entirely uncovered by CI; the priority should be weighed accordingly against its experimental status.

## Proposed change

Add an e2e spec (KEDA-based, matching the current suite) that:

1. Deploys two managed KEDA ScaledObjects (`llm-d.ai/managed: "true"` + `llm-d.ai/epp-inference-pool`, no `wva_desired_replicas` trigger so the coordinator adopts them) and a GPU `ResourceQuota` in the test namespace.
2. Enables the coordinator (`EXPERIMENTAL_COORDINATOR_ENABLED=true`, short `COORDINATOR_INTERVAL`).
3. Asserts the coordinator patches `spec.maxReplicaCount` proportionally — e.g. with all queues zero it splits the GPU budget evenly; with asymmetric EPP queue depth the busy pool's ceiling rises and the idle pool's falls.

The manual scaffold in `config/samples/hpa/co-ordinator/` (`poc.mk`, model-a/model-b workloads, gateway, EPP values) is a good starting point, but note it is HPA + prometheus-adapter based and predates the KEDA-only migration — it would need porting to KEDA (ScaledObjects instead of HPAs; drop the prometheus-adapter rules).

### Optional (nice to have)

[...]

PR #1533, opened 2026-08-15
test(e2e): add Coordinator gpu-rebalance e2e coverage

SrikarChittemsetty:
## What this PR does

Adds e2e coverage for the Coordinator's `gpu-rebalance` plugin, per #1458.

Following @mamy-CS's guidance on the issue, the Coordinator is toggled at **deploy time** rather than per-spec. It reads `EXPERIMENTAL_COORDINATOR_ENABLED` once at manager startup, so a runtime toggle would need a ConfigMap patch plus a rollout restart mid-suite and would race the other specs. This mirrors the existing scale-from-zero pattern:

- `deploy/lib/infra_wva.sh` yq-patches `EXPERIMENTAL_COORDINATOR_ENABLED="true"` and a shorter `COORDINATOR_INTERVAL` into `wva-manager-config`, **before** the `kubectl wait --for=condition=Ready`, gated on `ENABLE_COORDINATOR=true`.
- `deploy/install.sh` defaults and exports `ENABLE_COORDINATOR=false` / `COORDINATOR_INTERVAL=5s`.
- `Makefile` gains `COORDINATOR_ENABLED` / `COORDINATOR_INTERVAL`, plus `test-e2e-coordinator` and `test-e2e-coordinator-with-setup`.
- `test/e2e/config.go` surfaces it as `cfg.CoordinatorEnabled`; the specs `Skip()` with an explanatory message when it is false.

The specs carry their own `coordinator` Ginkgo label, deliberately **outside** the `full && !smoke && !flaky` filter, and run from their own target — so the saturation-path specs are never exercised against a cluster with the Coordinator rewriting replica ceilings underneath them.

## Coverage

The acceptance-criteria case: two managed ScaledObjects sharing one namespace GPU `ResourceQuota`.

1. **Selection contract.** Asserts both objects are annotated managed *and* carry no `wva_desired_replicas` trigger. This one is load-bearing: the default `buildScaledObject` fixture installs exactly that trigger, and `IsScaledObjectUnderControl` excludes objects that have it — so without a distinct fixture option the Coordinator would silently skip both and the split assertion would fail for an unrelated reason. Hence the new `WithScaledObjectCoordinatorManaged`.
2. **Even split when every pool queue is zero.** Both ceilings equal, and their sum equals the quota, checked in the same `Eventually` so the invariant holds at one instant rather than each value merely having passed through at some point.
3. **Shrinking quota.** Halving the quota re-derives lower ceilings — the case that catches a plugin that only ever ratchets upward.

The quota is 8 so the expected split is exact rather than rounding-dependent: two scalers at `effectiveMin=1` reserve 2, leaving 6 to split at weight 0.5, giving 1 + 3 = 4 each with no remainder to hand to the highest-weight pool.

Also adds a `ResourceQuota` fixture (none existed), scoped to `requests.nvidia.com/gpu` only — adding pod/CPU limits would make the quota admission controller reject the spec's workloads for unrelated reasons and turn a rebalance assertion into a scheduling failure.

## Deliberately left out

**Asymmetric ceilings under uneven EPP queue depth** is the other half of the acceptance criteria and is committed as a `PIt` rather than asserted. Driving it needs real flow-control queue depth, which means binding these pool annotations to live inference pools and pushing burst load through one of them, instead of the standalone scale targets used here. I did not want to assert it against a fabricated metric and claim coverage the test does not have.

@mamy-CS — two questions on that:
1. Would you rather I wire it to the sglang emulator pools in this PR, or land the deterministic cases first and do the load-driven one as the follow-up?
2. For the CI side, I wired `ENABLE_COORDINATOR` so a job can set it, but did not add a job to `ci-e2e-openshift.yaml` — standing up another run against the real GPU cluster felt like your call on resourcing rather than mine. Happy to add it if you want it here.

## Testing

`go build ./test/...`, `go vet ./test/...`, and `gofmt` are clean; `bash -n` passes on both changed shell scripts.

I have **not** run this suite against a live cluster — I don't have one with GPUs and KEDA to hand, so the deterministic split values are derived by reading `rebalanceNamespace`, not observed. Worth a real run before merge; if the CI job is the easiest way to get that, say the word and I'll add it.

## Disclosure

Prepared with the assistance of an AI coding agent. I verified the analysis against the source and existing suites; the reasoning about the startup-only flag, the trigger-exclusion trap, and the exact split arithmetic was checked against `internal/coordinator/` before writing the specs. This description is my own.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,8 @@ NUM_PROMPTS          ?= 3000
 ENVIRONMENT                 ?= kind-emulator
 USE_SIMULATOR               ?= true
 SCALE_TO_ZERO_ENABLED       ?= false
+COORDINATOR_ENABLED         ?= false
+COORDINATOR_INTERVAL        ?= 5s
 DEPLOY_ALERTING_RULES       ?= false
 SCALER_BACKEND              ?= keda  # keda (ScaledObject) or none (skip, use pre-installed backend)
 LLM_D_ROUTER_VERSION        ?= v0.9.0
@@ -209,6 +211,8 @@ deploy-e2e-infra: ## Deploy e2e test infrastructure (WVA + EPP; no model server 
 		ENVIRONMENT=$(ENVIRONMENT) \
 		SCALER_BACKEND=$(SCALER_BACKEND) \
 		ENABLE_SCALE_TO_ZERO=$(SCALE_TO_ZERO_ENABLED) \
+		ENABLE_COORDINATOR=$(COORDINATOR_ENABLED) \
+		COORDINATOR_INTERVAL=$(COORDINATOR_INTERVAL) \
 		DEPLOY_ALERTING_RULES=$(DEPLOY_ALERTING_RULES) \
 		WVA_IMAGE_REPO=$$IMAGE_REPO \
 		WVA_IMAGE_TAG=$$IMAGE_TAG \
@@ -219,6 +223,8 @@ deploy-e2e-infra: ## Deploy e2e test infrastructure (WVA + EPP; no model server 
 		ENVIRONMENT=$(ENVIRONMENT) \
 		SCALER_BACKEND=$(SCALER_BACKEND) \
 		ENABLE_SCALE_TO_ZERO=$(SCALE_TO_ZERO_ENABLED) \
+		ENABLE_COORDINATOR=$(COORDINATOR_ENABLED) \
+		COORDINATOR_INTERVAL=$(COORDINATOR_INTERVAL) \
 		DEPLOY_ALERTING_RULES=$(DEPLOY_ALERTING_RULES) \
 		./deploy/install.sh; \
 	fi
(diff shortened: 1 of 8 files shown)


SrikarChittemsetty:
Taking this out of draft so it is reviewable rather than sitting.

The honest caveat from the description still stands and I do not want it buried: I have never run this against a live cluster. The split values are derived from reading `rebalanceNamespace`, not observed. It builds, vets, and is gofmt clean, and DCO and signed-commits pass, but that is not the same as green.

It also will not run in CI as things stand. The specs carry their own `coordinator` label, deliberately outside the `full` filter, and there is no job wired to run it. That was the point of the label, so the saturation specs never see the Coordinator ticking, but it does mean nothing exercises this until someone adds the job or runs `make test-e2e-coordinator-with-setup` by hand.

@mamy-CS two things outstanding from your side whenever you have a moment:

1. Whether you want the uneven-EPP-queue-depth case wired to the sglang emulator pools in this PR, or left as the follow-up it currently is. It is a `PIt` rather than a fabricated metric.
2. Whether you want a job added to `ci-e2e-openshift.yaml`. I wired `ENABLE_COORDINATOR` so a job can set it, but did not add one, since another run against the real GPU cluster is a resourcing call rather than mine.

Also worth flagging: #1538 adds damping to gpu-rebalance, which means a lowered ceiling no longer lands within a single tick. I already noted that in the shrink spec here so the timeout margin is not mistaken for slack, but the two want reviewing together.

PR #1539, opened 2026-08-18
cleanup: remove dead code for deprecated VariantAutoscaling CR k8s event

zdtsw:
- VA CRD is no longer included, it becomes a in-mem type from annotated HPA/ScaledObject. k8s event recorder cannot construct ObjectReference and emit error in pod log causing confusion for user.
- Removed recorder related code from saturation engine, scale-from-zero engine and replca metrics collector
- Removed const, test function are also cleaned up
- unattributed-ready-pods log moved from debug to default info level.
## Proposed Changes
- :broom: Update or clean up current behavior

### Pre-review Checklist

**Release Note**

```release-note
Removed error  "Could not construct reference" from WVA pod logs caused by deprecated K8s event recording against synthetic VariantAutoscaling objects.
```


zdtsw:
require fix in https://github.com/llm-d/llm-d-workload-variant-autoscaler/pull/1549

zdtsw:
close we we have moved WVA to legacy

PR #1547, opened 2026-08-30
chore: allow sed branch label "ba" in typos config

aliceco01:
The nightly typo scan reported `ba` -> `by` in `benchmark/config/templates/values/defaults.yaml:564`.

That `ba` is not a typo. The line is:

```
export LD_LIBRARY_PATH=$(find / -name libcuda.so.1 -printf '%h\n' 2>/dev/null | sed ':a; N; $!ba; s/\n/:/g'):${LD_LIBRARY_PATH}
```

`:a` defines a sed label and `$!ba` branches back to it, the standard idiom for joining all input lines. Rewriting `ba` to `by` would break the command.

This adds `ba` to the existing "Short code fragments / false positives" list in `_typos.toml`, alongside `ot` and `vas`. It is the only standalone `ba` in the repository, so the entry does not mask a real typo anywhere else.

Fixes #1529

PR #1550, opened 2026-09-02
fix(scalefromzero): do not treat a failed EPP scrape as an idle pool

aliceco01:
When every EPP pod fails to scrape, a variant with queued requests silently stays scaled to zero, logged identically to a genuinely idle one.

## The problem

Three things line up:

1. `scrapeAllPods` ([pod_scraping_source.go](https://github.com/llm-d/llm-d-workload-variant-autoscaler/blob/main/internal/collector/source/pod/pod_scraping_source.go)) logs a failed pod scrape at VERBOSE, bumps an error counter, and returns from the goroutine, dropping that pod from the results map. No `Error` is set on any `MetricResult`.
2. If *every* pod fails, `aggregateResults` returns `{QueryName: "all_metrics", Values: [], Error: nil}`, and `Refresh` returns it with a nil error. That is byte-identical to a successful scrape of a genuinely idle pool.
3. The scale-from-zero engine reads `results["all_metrics"]` directly, finds no queue-size values, and sets `pendingRequestExist = false`.

So when EPP is unreachable (pod down, auth failure, non-200), the engine concludes there is no queued work and leaves the variant at zero replicas. The only trace is a DEBUG line that reads the same as normal idle logging.

## The fix

**Source side.** When no pod scrape succeeded, set `Error` on the aggregated result so a total collection failure is distinguishable from "nothing queued".

`Refresh` still returns a nil function-level error, so consumers that check only `err` are unaffected. Partial failures are deliberately left alone: partial data is still usable and does not set `Error`.

**Consumer side.** Nil-guard the `all_metrics` lookup and check `HasError()` before interpreting the values, returning the error instead of concluding the pool is idle. Both paths increment `IncMetricsCollectionErrors`, so the failure surfaces through existing metrics rather than only in logs.

The nil guard is defensive. `PodScrapingSource` always sets the `all_metrics` key, so it is not reachable through the only implementation that exists today, but any other `MetricsSource` that omitted the key would panic at this line.

## Tests

- `TestPendingRequestsUnknownIsNotIdle` covers three collection outcomes: errored result, absent result, and a clean empty result that must still be read as idle.
- The two existing total-failure cases in the pod source suite (unreachable pods, auth failure) now also assert that the result carries an error.
- A new case asserts that partial failure does **not** set an error.

I confirmed these are not vacuous: with the production changes reverted, the pod-source assertions fail and the consumer test panics with a nil pointer dereference on the absent-result case.

`make test` passes, 36 packages ok. `go build`, `go vet`, and `gofmt` are clean.

## Notes

I could not run `make lint` locally. The pinned `bin/golangci-lint` is built with go1.26 and panics decoding stdlib export data against a go1.27 toolchain, on `test/testconfig/config.go`, which this PR does not touch. It fails the same way on an unmodified `main`, so it looks like a local toolchain mismatch rather than anything in this change, but it does mean lint is unverified on my end.

There is no existing issue for this; I found it while reading the collector. Happy to open one if you would prefer it tracked separately.

### DIFF
--- a/internal/collector/source/pod/pod_scraping_source.go
+++ b/internal/collector/source/pod/pod_scraping_source.go
@@ -131,6 +131,14 @@ func (p *PodScrapingSource) Refresh(ctx context.Context, spec source.RefreshSpec
 	// Aggregate results
 	aggregated := p.aggregateResults(results)
 
+	// Every ready pod failed to scrape. The aggregated result is otherwise
+	// indistinguishable from a successful scrape of genuinely idle pods (empty
+	// Values, nil Error), which lets consumers read a total collection failure
+	// as "no activity". Surface it as an error so they can tell the two apart.
+	if len(results) == 0 {
+		aggregated.Error = fmt.Errorf("failed to scrape all %d ready pod(s)", len(pods))
+	}
+
 	// Cache the aggregated result under the canonical "all_metrics" key.
 	cacheKey := source.BuildCacheKey("all_metrics", nil)
 	p.cache.Set(cacheKey, *aggregated, p.config.DefaultTTL)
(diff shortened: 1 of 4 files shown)


aliceco01:
@lionelvillard could you approve the workflow runs on this PR? Five workflows have been at `action_required` since Sep 2, so only DCO and signed-commits have run and there is no test signal for a reviewer to look at. It needs the "Approve and run workflows" button in the Actions tab; `/ok-to-test` on its own only reaches the OpenShift E2E job. This is the first-time-contributor gate, so it should stop needing a click once anything of mine merges. Same situation on #1547.

